PR #18, opened 2021-02-18
Updated .gitignore, update README to use DataProfiler, add Loading Data example

lettergram:


examples/LOADINGDATA.md
```diff
@@ -0,0 +1,188 @@
...
+```python
+import pandas as pd
+import data_profiler as dp
+
+data = dp.Data(data=pd.DataFrame([1, 2]), data_type='csv')
+```
+
+*Note: This override can still if fail, if the `data_reader` cannot load the file.*
+
+It is important to note that by default the `Data class` does not utilize file extensions to identify *file type*. Instead, it inspects the file itself to determine which `data_reader` it should use. See [data_readers/csv_data.py](/data_profiler/data_readers/csv_data.py) as an example. The `is_match` function is utilized by the `CSVData` class to determine if it can be used to load the given file.
```
AnhTruong, line 61: not sure the last line is needed

doc/DATA_LOADING.md
```diff
@@ -0,0 +1,188 @@
...
+
+Possible `options`:
+
+* data_format: user selected format in which to return data can only be of specified types
+* samples_per_line - chunks by which to read in the specified dataset
+
+**Example**:
+```python
+from dataprofiler.data_readers.text_data import TextData
+data = TextData("your_file.txt")
```
AnhTruong, line 187: should we add the `options` here to be consistent with other data classes?

doc/DATA_LOADING.md
```diff
@@ -0,0 +1,188 @@
...
+# Auto-Detect & Load a file
+data = dp.Data("your_file.csv")
+
+# returns CSVData
+print(type(data))
+
+# <class 'dataprofiler.data_readers.csv_data.CSVData'>
+```
+
+We provide an interface which all data classes must match, with a template provided in the `base_data class` ([data_readers/base_data.py](/data_profiler/data_readers/base_data.py)). What that means is that the end user doesn't have to worry about functions for the specific class, as the template provides a set of required functions. For general purposes, the various the various data classes can be treated the exact same. However, a given class can be checked and special functions added (if desired), making it more extensible.
```
AnhTruong, line 40: change `For general purposes, the various the various data classes can be treated the exact same` to `For general purposes, the various data classes can be treated the same`

AnhTruong (approved):

gme5078 (approved):

PR #31, opened 2021-02-25
Fix delimiter issue

AnhTruong:
Sometimes, the delimiter is incorrectly detected, causing dp to create an additional column with none value.


dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -68,9 +93,12 @@ def test_auto_file_identification(self):
         Determine if the csv file can be automatically identified
         """
         for input_file in self.input_file_names:
+            print(input_file['path'])
```
lettergram, line 96: Please remove print statement

AnhTruong: removed

dataprofiler/data_readers/csv_data.py
```diff
@@ -138,6 +137,20 @@ def _load_data_from_file(self, input_file_path):
             if not self._header:
                 self._header = self._determine_has_header(data_as_str)
                 self._checked_header = True
+
+        # if the delimiter is always at the end of each row,
+        # without any subsequent pattern, set delimiter to None
+        if self._delimiter:
```
ChrisWallace2020, line 143: Since on line 135 we set self._delimiter if it is None, it will always be not None by this line, so this if statement will always be executed, couldn't we then remove the if statement?

dataprofiler/data_readers/csv_data.py
```diff
@@ -138,6 +137,20 @@ def _load_data_from_file(self, input_file_path):
...
+            with open(input_file_path, encoding=self.file_encoding) as csvfile:
+                num_lines, num_lines_read = 5, 0
+                count_delimiter_last = 0
+                for line in islice(csvfile, 1, num_lines):
+                    if line.strip()[-1] == self._delimiter:
+                        count_delimiter_last += 1
+                    num_lines_read += 1
+                if count_delimiter_last == num_lines_read:
+                    self._delimiter = None
```
ChrisWallace2020, line 152: Hypothetically, say the first 5 lines of data had no value for the last column, but subsequent lines did later, that would mean that the delimiter is indeed correct, but just the first 5 lines have no data in that column, sand since this logic only looks at the first 5 lines this would set the delimiter to None even though there is one. That is what would happen here right? Why do we only care about num_lines = 5?

dataprofiler/data_readers/csv_data.py
AnhTruong: the above delimiter detection may return None, so, I have this if statement

ChrisWallace2020: Oh okay, that makes sense, disregard this comment then

dataprofiler/data_readers/csv_data.py
AnhTruong: I pick up 5 first lines as with other check. The corner case you mentioned is reasonable. I'm not sure how often it happens, at least for sparse files. Increasing number of lines to check definitely helps, but I don't know how many lines are enough. In general, I'd love to read 5 random lines, but it seems we have to scan through the file.

ChrisWallace2020: Also, on this topic, is the right response to detecting delimiters at the end of the line to set delimiter to None? It seems like it should instead delete this last column of Nones/ignore it, since the delimiter may still a proper delimiter if it appears before just the last character, but there seems to be no coverage for this sort of case. (e.g. a line like "data_1, data_2, data_3," would make the delimiter None, even though "," is still a delimiter, just used incorrectly at the end.

gme5078: Anh shouldnt we check for delimiters within the line. Am I reading this code wrong or wouldnt this think the delimiter is None?
X,Y,Z,
1,1,1,
2,2,2,
3,3,3,
4,4,4,
5,5,5,

gme5078 (changes requested):

dataprofiler/data_readers/csv_data.py
gme5078: which I think is what Chris is saying too

AnhTruong: Fixed code to only set delimiter to None if there's one delimiter at the end

lettergram: Honestly, I think it's fair to leave the example as 4 columns (the last one being all nones) given they have the commas. I agree setting the delimiter to none is kinda weird

gme5078 (approved):

ChrisWallace2020 (approved):

PR #33, opened 2021-03-01
Add Tests For BaseOption Class

sagars729:
- Added initial tests for BaseOption Class
- Also included a type check for variable path ~~, but the BaseOption class does not currently check for this (i.e. it is a failing test)~~


dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,54 @@
...
+			options._set_helper("notadictionary", "")
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper(["not", "a", "dictionary"], "")
+
+		#Variable Path Is Not A String
+		expected_error = "The variable path must be a string."
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper({"hello": "world"}, 1)
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper({}, 1)
```
AnhTruong, line 38: can we add a case with a list of strings?

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,54 @@
...
+
+		#Options Is Not A Dictionary
+		expected_error = "The options must be a dictionary."
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options.set("notadictionary")
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options.set(["not", "a", "dictionary"])
+	
+	def test_validate_helper(self, *mocks):
+		pass
```
AnhTruong, line 51: are we gonna implement this test?

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,54 @@
...
+	
+	def test_validate(self, *mocks):
+		pass
```
AnhTruong, line 54: same as above

dataprofiler/tests/profilers/test_base_option.py
sagars729: Currently, the BaseOption class would raise a NotImplementedError for _validate_helper. I left the template here in case that changes in the future.

dataprofiler/tests/profilers/test_base_option.py
sagars729: What would this look like?

dataprofiler/tests/profilers/test_base_option.py
sagars729: (See above comment) Since _validate_helper would raise a NotImplementedError for the BaseOption Class, validate would do the same.

dataprofiler/tests/profilers/test_base_option.py
AnhTruong: just like the other cases, e.g., `options._set_helper(['this', 'is', 'a', 'list'], 1)`

AnhTruong: nvm, please disregard my previous comment. I was not clear about this initially.

sagars729:
@AnhTruong PR has been updated to add variable path type check to the BaseOption Class

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,54 @@
...
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.'
+            'DataLabelerColumn.update', return_value=None)
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.DataLabeler')
+class TestBaseOption(unittest.TestCase):
+	
+	@classmethod
+	def setUpClass(cls):
+		cls.data = Data(data=pd.DataFrame([1, 2]), data_type='csv')
+	
+	def sanity(self, *mocks):
```
JGSweets, line 20: what is the intention of sanity?

dataprofiler/tests/profilers/test_base_option.py
JGSweets: @lettergram thoughts? I would suggest instead, we could validate that since this is abstract that we could validate that running the function would raise the `NotImplementedError`

i.e. 
```python
with self.assertRaises(NotImplementedError):
    options._validate_helper()
```

dataprofiler/tests/profilers/test_base_option.py
JGSweets: as above

dataprofiler/tests/profilers/test_base_option.py
JGSweets: that way, if changes happened in the code, this would the raise an error that something was altered and the test would need to be updated with the code changes.

lettergram: ^ Agreed, recommend doing this.

dataprofiler/tests/profilers/test_base_option.py
lettergram: Would recommend naming it something more clear, assuming this was a "sanity check"

AnhTruong: agree

dataprofiler/tests/profilers/test_base_option.py
sagars729: That makes sense, I will add it in

dataprofiler/tests/profilers/test_base_option.py
sagars729: I usually add these to make sure the object can be created and that it has the right property keys. I should probably rename this to test_default_config or test_init

dataprofiler/tests/profilers/test_base_option.py
sagars729: Included in new commit

dataprofiler/tests/profilers/test_base_option.py
sagars729: Included in new commit

dataprofiler/tests/profilers/test_base_option.py
sagars729: Changed to test_init and also added a check to make sure that the properties are empty

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,61 @@
...
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.DataLabeler')
+class TestBaseOption(unittest.TestCase):
+	
+	@classmethod
+	def setUpClass(cls):
+		cls.data = Data(data=pd.DataFrame([1, 2]), data_type='csv')
+	
+	def test_init(self, *mocks):
+		options = BaseOption()
+		self.assertEqual(options.properties, {})
```
JGSweets, line 22: I would do assertDictEqual, just to be more precise, but the more important change here is that the expected value should be set first followed by the actual value provided by the code:
i.e. `self.assertDictEqual({}, options.properties)`

sagars729: Done! Swapped parameters and used assertDictEqual

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,61 @@
...
+		cls.data = Data(data=pd.DataFrame([1, 2]), data_type='csv')
+	
+	def test_init(self, *mocks):
+		options = BaseOption()
+		self.assertDictEqual({}, options.properties) 
+
+	def test_set_helper(self, *mocks):
+		options = BaseOption()
+
+		#Options Is Not A Dictionary
```
JGSweets, line 27: add a space between # and O, `# Options...`

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,61 @@
...
+		expected_error = "The options must be a dictionary."
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper("notadictionary", "")
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper(["not", "a", "dictionary"], "")
+
+		#Variable Path Is Not A String
```
JGSweets, line 34: same space

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -0,0 +1,61 @@
...
+		expected_error = "The variable path must be a string."
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper({"hello": "world"}, 1)
+		with self.assertRaisesRegex(ValueError, expected_error):
+			options._set_helper({}, 1)
+
+	def test_set(self, *mocks):
+		options = BaseOption()
+
+		#Options Is Not A Dictionary
```
JGSweets, line 44: same space

dataprofiler/tests/profilers/test_base_option.py
sagars729: Fixed all spacing in new commit

JGSweets (approved):

lettergram (approved):

PR #34, opened 2021-03-01
Int Column Profile Merge

gme5078:
This prevents Int Column Profiles from merging if the options are different. Will raise an error instead.


AnhTruong (approved):

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -460,3 +460,27 @@ def test_profile_merge_no_bin_overlap(self):
...
+        profiler1.update(df)
+        profiler1.match_count = 0
+
+        # Creating second profiler with separate options
+        options = IntOptions()
+        options.max.is_enabled = False
+        data2 = [10, 15]
+        df2 = pd.Series(data2).apply(str)
+        profiler2 = IntColumn("Int", options=options)
+        profiler2.update(df2)
```
lettergram, line 478: Is this the behavior we want? It's quite possible we could just use the one that is more inclusive. Would take more logic, but if you take the "max enabled" it's straight forward:

```python
options.max.is_enabled |= options2.max.is_enabled
```

Very easy to do for flags, not so much for others. I suppose the risk would be because the profile without the max has already been ran, we wouldn't be able to actually get the max from said data...

So I see that issue, just want to confirm that's indeed what we've considered?

lettergram: I'd be super angry if it took 24 hours to generate both profiles and couldn't merge them. So I think we have to identify a way to still enable merging as best we can.

Issue #39, opened 2021-03-03
Separate "optional" requirements


lettergram:
**Is your feature request related to a problem? Please describe.**

Many people don't care about the data labels (entity recognition), they should be able to install the library without tensorflow and skip the install. This would also help people attempting to use python 3.9, for instance. 

*I'm not 100% sold this is a great idea*, but I think it's worth a discussion.

**Describe the outcome you'd like:**

I'd like to remove the labeling of the requirements in `requirements.txt` and separate them into `requirements-labeling.txt`. Then add them as `dataprofiler[extras]` or `dataprofiler[labeler]` or something to that effect. This can easily be done in the `setup.py`.

The way you would install the labeler would be:

```bash
$ pip install dataprofiler[labeler] --user
```

Without the labeler would simply be:

```bash
$ pip install dataprofiler --user
```

https://stackoverflow.com/questions/6237946/optional-dependencies-in-distutils-pip

In the output report & when executing we can warn the users to install the labeler, if desired.

**Additional context:**

This was a [recommendation on /r/statistics](https://www.reddit.com/r/statistics/comments/lwdmte/q_what_statistics_are_we_missing_for_our/gphjzby?utm_source=share&utm_medium=web2x&context=3)

PR #41, opened 2021-03-03
Profiler runs without TF & TF only executes when necessary

lettergram:
PR makes it so you can load the profiler without loading tensorflow:

**If TensorFlow is installed**:
```python
>>> import dataprofiler as dp
>>> data = dp.Data('dataprofiler/tests/data/csv/guns.csv')
>>> profile = dp.Profiler(data) # TensorFlow loads here
2021-03-03 02:46:53.359832: I tensorflow/stream_executor/platform/....
```

**If TensorFlow is NOT installed**:
```python
>>> import dataprofiler as dp
>>> data = dp.Data('dataprofiler/tests/data/csv/guns.csv')
>>> profile = dp.Profiler(data)

!!! WARNING Partial Profiler Failure !!!

Profiling Type: data_labeler
Exception: AttributeError
Message: module 'tensorflow' has no attribute '__version__'

For labeler errors, try installing tensorflow and tensorflow-addons; command:
  $ pip install tensorflow tensorflow-addons --user

  self._create_profile(df_series, options)
```

-------------

This is related to issue #39, next steps (if we go that route):

1. Removing the requirement(s)
2. Move labeler requirements to `requirements-labeler.txt`
3. Create an "extras" in setup.py moving it to a special `dataprofiler[labeler]`


ChrisWallace2020 (dismissed):

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -6,17 +6,27 @@
...
-import tensorflow as tf
-import tensorflow_addons as tfa
 import numpy as np
 from sklearn import decomposition
 
 from . import labeler_utils
 from .base_model import BaseModel, BaseTrainableModel
 from .base_model import AutoSubRegistrationMeta
 
+import tensorflow as tf
```
JGSweets, line 16: needs to move back for pep8

dataprofiler/profilers/profile_builder.py
```diff
@@ -329,6 +329,23 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
         elif not isinstance(profiler_options, ProfilerOptions):
             raise ValueError("The profile options must be passed as a "
                              "ProfileOptions object.")
+
+        try:
```
JGSweets, line 333: I'm not sure turning off the data labeler in general is appropriate as a labeler maybe made without TF. Need to think more on this of how best to address.

lettergram: I actually don't know where else to put this, but currently there's no other reasonable model besides tensorflow. 

I agree this is very convoluted, but I don't know where else to put this.

gme5078: I agree with Jeremy, a labeler could be used without tensorflow (like our own regex model). Going to look in the repo for where I would put this.

gme5078 (changes requested):

dataprofiler/profilers/profile_builder.py
```diff
@@ -329,6 +329,23 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
...
+            import tensorflow_addons
+        except Exception:
+            import warnings
+            warning_msg = "\n\nTensorFlow and/or TensorFlow Addons"
+            warning_msg += "are not installed.\n\nUnfortunately, this "
+            warning_msg += "means the library cannot label the data\n\n"
+            warning_msg += "To label the data, install tensorflow & "
+            warning_msg += "tensorflow-addons\n\n"
+            warning_msg += "    $ pip install tensorflow "
+            warning_msg += "tesnorflow-addons --user\n\n"
```
gme5078, line 344: typo

gme5078 (changes requested):

AnhTruong (dismissed):

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -43,20 +43,41 @@ def _create_profile(self, df_series, options=None):
...
+                    # This is considered a major error
+                    if type(e).__name__ == "ValueError":
+                        raise ValueError(e)
+                    
+                    warning_msg += "\n\nFor labeler errors, try installing "
+                    warning_msg += "tensorflow and tensorflow-addons; command:\n"
+                    warning_msg += "  $ pip install tensorflow "
+                    warning_msg += "tensorflow-addons --user\n"
+
+                    warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
```
gme5078, line 79: Does this exception occur for every column of the dataset?
What doess warning_flag = False do?

gme5078 (changes requested):

dataprofiler/profilers/column_profile_compilers.py
lettergram: Yes, if it fails on every potential issue

.gitignore
```diff
@@ -3,6 +3,7 @@
 DataProfiler\.egg-info/
 DataProfiler.egg-info/*
 data_profiler/labelers/embeddings/glove-reduced-64D.txt
```
gme5078, line 5: feels like this one should be removed, but not worth another commit

AnhTruong (approved):

gme5078 (approved):

PR #44, opened 2021-03-03
fix header issue

AnhTruong:


dataprofiler/data_readers/csv_data.py
```diff
@@ -110,7 +110,7 @@ def _determine_has_header(data_as_str):
         try:
             has_header = sniffer.has_header(data_as_str)
         except csv.Error:
-            has_header = 0
+            has_header = 1
```
JGSweets, line 113: I'm concerned about this change has pandas does not do header detection. Pandas makes the assumption that the first row is the header.

AnhTruong: My thought process is this: if the sniffer fails to detect the header, we can either set header to 0 (fix from this PR) or None (currently existed). The advantage of setting to 0 is:
- most of the time, the files have headers, so the error rate is less than setting it to None
- setting to 0 will reduce 1 first line, but won't affect the results of dp, while setting to None will make the potential header combined with the value rows, so the detection results of dp may be affected. E.g., header is 'a b c', and values are '1 2 3\n 4 5 6\n...'

AnhTruong: one other way is to raise an error if the sniffer fails, and ask user to input the header.

ChrisWallace2020: So just so I'm understanding, we aren't going to use sniffer any differently, we're just going to begin implicitly assuming a header is present?

lettergram: Why do we get the exceptions?

JGSweets (changes requested):
Could you please provide an assessment of why the `csv.Error` is tripping so we can better understand the case where this occurs and therefore understand if changing `has_header` is appropriate?

My understanding is that the `csv.Error` occurs whenever the delimiter can't be detected. 
Additionally, pandas does not due header detection so this would force the header to be at line 0 if any error in delimiter detection (from the csv.sniffer lib) occurred.

dataprofiler/data_readers/csv_data.py
AnhTruong: @ChrisWallace2020 : yes, which is helpful from my first point above.
@lettergram : in case that the sniffer fails to detect the header, in which case we need to set header to either 0 or 1

JGSweets: @AnhTruong I believe I found a potential solution. The problem itself is very interesting and understanding why it was occurring helped in assessing the solution, hence why I stressed understanding it further to potentially be able to address the problem at its root. Basically, the problem occurs bc the delimiter cannot be determined (in these cases due to single col which has no delimiters). For whatever reason, as oppose to saying no delimiter found (or giving that option), the program just errors. I actually found a case log on their lib which discusses this same issue.

I instead found a potential solution which avoids this error, and should only be executed when the csv.Error occurs as a backup.
```python
# ERRORS
sniffer = csv.Sniffer()
sniffer.has_header('header\n' + "\n".join(map(str, range(100))))
# NO ERRORS
sniffer = csv.Sniffer()
sniffer._guess_delimiter = lambda *args, **kwargs: (' ', 0) # ONLY EXECUTE THIS IF WE FAILED and got csv.Error
sniffer.has_header('header\n' + "\n".join(map(str, range(100))))
```

I think we should consider this as a potential solution, but not 100% confident it is the correct solution.

AnhTruong: still have some error with csv files of text without spaces, but I think that may be the best solution that we can have now

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -175,5 +184,47 @@ def test_header_check_files(self):
...
+        options = dict()
+        CSVData.is_match(path, options)
+        self.assertEqual(1, options.get("header"))
+
+        # File with header, but the csv sniffer fails to detect
+        # set to 0 for auto determine
+        path = os.path.join(test_dir, 'csv/sparse-interchange-none.txt')
+        options = dict()
+        CSVData.is_match(path, options)
+        self.assertEqual(0, options.get("header"))
```
ChrisWallace2020, line 227: Could we make this a for loop? This seems a bit repetitive. Maybe define a list of tuples like [(path, header), ...] where you store the file path and the expected value for options.get("header")?

ChrisWallace2020: Or could you even just index into the list of dictionaries defined in input_file_names?

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,21 +104,114 @@ def _determine_delimiter_of_str(data_as_str):
...
+            len_increase = False
+            len_not_none = len(header_check_list[i]) - header_check_list[i].count("none")
+            if len_not_none >= prior_len:
+                prior_len = len_not_none
+                len_increase = True
+            # Returns the last row that could reasonably be the header
+            if (rstr > str_thresh and none <= none_thresh and diff >= diff_thresh \
+                    and len_increase and not row_classic_header_ends) \
+                    or (i in skip_rows):
+                row_classic_header_ends = i
```
ChrisWallace2020, line 193: I'm confused, if i is a row that should be skipped, then why would that make this conditional pass and set the return value to i? Could you provide an example to help me better understand this reasoning?

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,21 +104,114 @@ def _determine_delimiter_of_str(data_as_str):
...
+        for i in range(0, len(differences)):
+            # Determine ratio of none in row, must be BELOW threshold
+            none = float(header_check_list[i].count("none")) / float(len(header_check_list[i]))
+            # Determine percent of differences between prior row, must be BELOW threshold
+            diff = float(differences[i].count(True)) / float(len(differences[i]))
+            # Determine percent of string, uppercase string or none in row, must be ABOVE threshold
+            rstr = float((header_check_list[i].count("str")
+                          + header_check_list[i].count("upstr")
+                          + header_check_list[i].count("none"))) / float(len(header_check_list[i]))
+            # Determines if the elements in the row is increasing or decreasing
```
ChrisWallace2020, line 183: Do you mean to say "Determines if the **number of** elements in the row is increasing or decreasing?

dataprofiler/tests/data_readers/test_csv_data.py
AnhTruong: yeah, I can make it as a for loop

dataprofiler/data_readers/csv_data.py
AnhTruong: so, either we meet the row with different len, or the row with the same len but different value types of cells, we make that row as potential candidate for the header

dataprofiler/data_readers/csv_data.py
AnhTruong: yeah, number of not-none elements, to be more specific. I'll change the comment when all code are approved

ChrisWallace2020: This comment is still incorrect

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,21 +104,129 @@ def _determine_delimiter_of_str(data_as_str):
...
+    # def _determine_has_header(data_as_str, suggested_delim=None):
+    #     """Automatically checks if the string has a header."""
+    #     if not data_as_str:
+    #         return None
+    #     sniffer = csv.Sniffer()
+    #     if not suggested_delim:
+    #         suggested_delim = ' '
+    #     sniffer._guess_delimiter = lambda *args, **kwargs: (suggested_delim, 0)
+    #     has_header = sniffer.has_header(data_as_str)
+    #     return 0 if has_header else None
```
ChrisWallace2020, line 222: Should this get cleaned up too?

dataprofiler/tests/data/csv/sparse-interchange-none.txt
```diff
@@ -0,0 +1,10 @@
+CONTAR, EL NOMBRE, EL NUMERO
```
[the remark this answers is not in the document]
AnhTruong: the missing values are in different columns as opposed to the same column in consecutive rows

dataprofiler/tests/data/csv/names-col-without-space.txt
```diff
@@ -0,0 +1,61 @@
+NAMES
```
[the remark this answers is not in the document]
AnhTruong: done

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,21 +104,114 @@ def _determine_delimiter_of_str(data_as_str):
...
     @staticmethod
-    def _determine_has_header(data_as_str):
-        """Automatically checks if the string has a header."""
-        sniffer = csv.Sniffer()
-        try:
-            has_header = sniffer.has_header(data_as_str)
-        except csv.Error:
-            has_header = 0
-        return 0 if has_header else None
+    def _determine_has_header(data_as_str, delimiter, diff_thresh=0.3,
```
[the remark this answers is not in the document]
AnhTruong: it can replace the _determine_has_header, right? as it does all the detection for the header.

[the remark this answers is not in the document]
AnhTruong: it corrects some mistakes made by the csv sniffer as well

dataprofiler/tests/data/csv/two-header.txt
```diff
@@ -0,0 +1,11 @@
+COUNT, NAMES, NUMBERS
```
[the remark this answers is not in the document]
AnhTruong: done

dataprofiler/tests/data/csv/names-col-without-space.txt
```diff
@@ -0,0 +1,61 @@
...
+George
```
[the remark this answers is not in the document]
AnhTruong: added

dataprofiler/data_readers/csv_data.py
```diff
@@ -96,29 +96,150 @@ def _determine_delimiter_of_str(data_as_str):
...
+                header_check_list[-1].append(cell_type)
+            
+        # Flags differences in types between each row (true/false)
+        potential_header = header_check_list[0]
+        differences = []
+        for i in range(0, len(header_check_list)):
+            differences.append([])
+        
+            # Identify if the row has any data
+            len_not_none = len(header_check_list[i]) - header_check_list[i].count("none")
```
lettergram, line 181: added: 

```
len_pot_head = len(potential_header) - potential_header.count("none")
```

dataprofiler/data_readers/csv_data.py
```diff
@@ -96,29 +96,150 @@ def _determine_delimiter_of_str(data_as_str):
...
+        
+            # If row has more elements or has no data, mark as "skip", no difference
+            if len(header_check_list[i]) > len(potential_header) or len_not_none == 0:
```
lettergram, line 184: switched to:

```
if len_not_none > len_pot_head or len_not_none == 0:
```

dataprofiler/data_readers/csv_data.py
```diff
@@ -96,29 +96,150 @@ def _determine_delimiter_of_str(data_as_str):
...
+            delimiter = ','
+    
+        # Ensure no empty last col and only 2 cols
+        empty_last_col_flag = True
+        for row in data_as_str.split('\n'):
+            last_cell = row.split(delimiter)[-1]
+            if len(row) == 2 and last_cell != None \
+            and len(last_cell) > 0:
+                empty_last_col_flag = False
+        if empty_last_col_flag:
```
JGSweets, line 143: I'm concerned that this just defaults to `,` bc the last col. This seems. like a delimiter check and to me doesn't make sense here.

I can see this being temp as delimiter check gets updated. However, defaulting to `,` seems dangerous as that could just be completely wrong.

It shouldn't even hit this function if a delimiter can't be detected or is incorrectly detected.

dataprofiler/data_readers/csv_data.py
AnhTruong: removed

dataprofiler/data_readers/csv_data.py
```diff
@@ -96,29 +96,150 @@ def _determine_delimiter_of_str(data_as_str):
...
+        :return: index for row estimated to be the last valid header
+        :type: int
+        """
+        
+        # Catch base cases
+        if not data_as_str or len(data_as_str) == 0:
+            return None
+    
+        # Ensure no "None" delimiter
+        delimiter = suggested_delimiter
```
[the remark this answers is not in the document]
AnhTruong: just to keep the original suggested delimiter

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,21 +104,148 @@ def _determine_delimiter_of_str(data_as_str):
...
+        
+        :return: index for row estimated to be the last valid header
+        :type: int
+        """
+        
+        # Catch base cases
+        if not data_as_str or len(data_as_str) == 0:
+            return None
+    
+        # Ensure no "None" delimiter
```
lettergram, line 131: # Ensure no "None" delimiter, delimiter is required for evaluating single columns

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,21 +104,148 @@ def _determine_delimiter_of_str(data_as_str):
...
+        # there's two columns, but one is empty (seperated by '.')..
+        # This check switches the delimiter to ensure a "hanging col" is not
+        # generated due to an incorrect separator identification
+        empty_last_col_flag = True
+        for row in data_as_str.split('\n'):
+            last_cell = row.split(delimiter)[-1]
+            if len(row) == 2 and last_cell != None \
+            and len(last_cell) > 0:
+                empty_last_col_flag = False
+        if empty_last_col_flag:
```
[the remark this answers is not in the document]
AnhTruong: removed the block

ChrisWallace2020 (approved):

dataprofiler/data_readers/csv_data.py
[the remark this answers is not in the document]
JGSweets: I don't believe it is mutable, hence removable.

JGSweets (approved):
tech debt for suggested delimiter to be removed but will approve provided it is fixed in future PR

Issue #40, opened 2021-03-03
Add training and extended training with an example dataset


kvrameshreddy:
**Please provide the issue you face regarding the documentation**
When using extended training,
when i pass custom label with custom data the following error appears
![error_while transfer_learning](https://user-images.githubusercontent.com/49742709/109766926-d6d2cd80-7c1c-11eb-8dc5-f038b4c8afa1.JPG)

When i add "BACKGROUND" In labels. Labler is getting trained. when i predict entities. the prediction is as follows
![err1](https://user-images.githubusercontent.com/49742709/109804305-1bc02980-7c48-11eb-9833-396db1bc8185.PNG)

to reproduce the error please use the following data
[new _data_label.zip](https://github.com/capitalone/DataProfiler/files/6074177/new._data_label.zip)

Please Update the Documentation with a example data, So that it will be helpful .

Thankyou

lettergram:
Appreciate the additional feedback, the documentation was updated, see:

https://github.com/capitalone/DataProfiler#extending-a-data-labeler-with-transfer-learning

If you'd like to add new custom labels when transfer learning, you have to do the following:

```python
# NOTE: if you want to extend the data labels as oppose to replace with your own,
#       you will need to extend the labels list passed into the fit function:
#       e.g. labels = data_labeler.labels + ['new_label_1', 'new_label_2', ...]

# this will use transfer learning to retrain the data labeler on your new 
# dataset and labels.
# NOTE: data must be in an acceptable format for the preprocessor to interpret.
#       please refer to the preprocessor/model for the expected data format.
#       Currently, the DataLabeler cannot take in Tabular data, but requires 
#       data to be ingested with two columns [X, y] where X is the samples and 
#       y is the labels.
model_results = data_labeler.fit(x=data['samples'], y=data['labels'], 
    validation_split=0.2, epochs=2, labels=labels)
```

The current labels that exist are listed in the readme, if you'd simply like to extend them.

@JGSweets will be working on examples in the future, which would likely include some sample data.

kvrameshreddy:
Hi @lettergram,

When we add custom labels to the existing labels, i am facing the following issue
![pad_error](https://user-images.githubusercontent.com/49742709/109938186-be82b180-7cf5-11eb-97b3-9fa2146894ee.JPG)

I think the problem is with label_mapping internally. 

Please take a look into this. Thankyou

lettergram:
[...]

PR #48, opened 2021-03-05
refactor label extension for data labeler

JGSweets:
allows the user to do `data_labeler.add_label(...)`
Fixes a bug in fit when labels was a list and the model required 0 as first index for PAD

Attempts to address issues referenced in: https://github.com/capitalone/DataProfiler/issues/40


dataprofiler/tests/labelers/test_base_model.py
```diff
@@ -96,25 +96,99 @@ def test_set_parameters(self):
...
 
+    def test_add_labels(self):
+
+        # setup mock
+        mock_model = mock.Mock(spec=base_model.BaseModel)
+        mock_model._convert_labels_to_label_mapping.side_effect = \
+            base_model.BaseModel._convert_labels_to_label_mapping
+        mock_model._label_mapping = {}
+        type(mock_model).num_labels = mock.PropertyMock(
+            side_effect=lambda: max(mock_model._label_mapping.values()) + 1)
```
ChrisWallace2020, line 107: Shouldn't the +1 be determined by whether or not they place a label at 0? It seems like this property is dependent on the requires_zero_mapping flag

dataprofiler/tests/labelers/test_base_model.py
```diff
@@ -96,25 +96,99 @@ def test_set_parameters(self):
...
+                                    "list."):
+            base_model.BaseModel.set_label_mapping(mock_model, label_mapping=None)
+
+        # base case
+        with self.assertRaisesRegex(TypeError,
+                                    "Labels must either be an encoding dict "
+                                    "which maps labels to index encodings or a "
+                                    "list."):
             base_model.BaseModel.set_label_mapping(
                 mock_model, label_mapping=None)
```
ChrisWallace2020, line 168: Am I missing something or is this the same test twice?

ChrisWallace2020: Also what does "base case" mean in this context?

dataprofiler/tests/labelers/test_base_model.py
```diff
@@ -96,25 +96,99 @@ def test_set_parameters(self):
...
+        # test label list to label_mapping
+        labels = ['a', 'b', 'd', 'c']
+        label_mapping = base_model.BaseModel._convert_labels_to_label_mapping(
+            labels, requires_zero_mapping=True)
+        self.assertDictEqual(dict(a=0, b=1, d=2, c=3), label_mapping)
+
+        # test label dict to label_mapping
+        labels = dict(a=1, b=2, d=3, c=4)
+        label_mapping = base_model.BaseModel._convert_labels_to_label_mapping(
+            labels, requires_zero_mapping=True)
```
ChrisWallace2020, line 193: This may bring up a design question, but should requires_zero_mapping only be used/specified if the labels are given as a list? Should we raise a warning or anything in this case or is it on the user to understand that it only affects the mapping generated from a list.

dataprofiler/tests/labelers/test_base_model.py
JGSweets: updated, good catch

dataprofiler/tests/labelers/test_base_model.py
JGSweets: the `requires_zero_mapping` does not impact the +1 as it is only dependent on the final value of the index. i.e. when we add a new label, it should be +1 the max value in the encodings, which is not relative but not dependent on `requires_zero_mapping`

ChrisWallace2020: Okay, I guess I interpreted the name "num_labels" to be more a sign of the number of labels present, not the next free label number

dataprofiler/tests/labelers/test_base_model.py
JGSweets: the labels are given as a dict, it is assumed the user is intentionally assigning the mapping. if they are provided as a list, the user is leaving the preference up to the model. In this case, the user doesn't really need to know what's happening below IMO since they have given the model autonomy to set it up how it needs it.

I welcome further thoughts.

ChrisWallace2020: I understand what it means in the context of providing a list, but I am saying that requires_zero_mapping doesn't do anything when given a dict, so I just think there should be some reflection that of that in the docstring or a warning raised if a dict is passed and requires_zero_mapping is specified as non-None, since doing so implies the user doesn't understand what/when/how that parameter is used.

dataprofiler/tests/labelers/test_base_model.py
JGSweets: I see what you mean, this is a fair point. I'm going to update the add_label to just directly use label_mapping and ignore num_labels. Thought currently this is how the base model determines num_labels, which might be a bug we want to address.

dataprofiler/tests/labelers/test_base_model.py
JGSweets: Warnings / errors are provided to the user for the case of our character model if the user inputs an improper dict with directions of how to fix.

JGSweets: These errors are also model dependent, so I'm not sure there's a generic error that could be made.

In addition, with these new changes, it might be that `requires_zero_mapping` is tech debt and can be removed as the model themselves will alter the list / dict arrangements rather than a base class or the data labeler

dataprofiler/tests/labelers/test_data_labelers.py
```diff
@@ -313,13 +313,13 @@ def test_set_labels(self, mock_open, mock_load_model):
...
         data_labeler._model.set_label_mapping.assert_called_with(
-            label_mapping={'a': 1, 'b': 2})
+            label_mapping=['a', 'b'])
 
         # test require pad false
         data_labeler._model.requires_zero_mapping = False
         data_labeler.set_labels(['a', 'b'])
         data_labeler._model.set_label_mapping.assert_called_with(
-            label_mapping={'a': 0, 'b': 1})
+            label_mapping=['a', 'b'])
```
lettergram, line 322: Can we still manually set the label mapping, if needed? My understanding is that we could still use a dict if desired (though it wouldn't be recommended).

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -105,6 +105,45 @@ def test_data_labeler_change_labels(self):
...
+
+        # get char-level predictions on default model
+        expected_label_mapping = {
+            "PAD": 0, "CITY": 1, "BACKGROUND": 1, "ADDRESS": 2, "BAN": 3,
+            "CREDIT_CARD": 4, "EMAIL_ADDRESS": 5, "UUID": 6, "HASH_OR_KEY": 7,
+            "IPV4": 8, "IPV6": 9, "MAC_ADDRESS": 10, "NAME": 11, "PERSON": 11,
+            "PHONE_NUMBER": 12, "SSN": 13, "URL": 14, "DATETIME": 15,
+            "INTEGER_BIG": 16, "INTEGER": 16, "FLOAT": 17, "QUANTITY": 18,
+            "ORDINAL": 19, "NEW_LABEL": 20}
+        model_predictions = data_labeler.fit(x=self.df[0], y=self.df[1])
```
lettergram, line 135: I don't think this is an appropriate test as the model itself should change as needed. For instance, I believe an upcoming release will not have the "CITY" label.

dataprofiler/tests/labelers/test_base_model.py
```diff
@@ -96,25 +96,98 @@ def test_set_parameters(self):
...
+
         # assert error for empty label_mapping dict
-        with self.assertRaisesRegex(ValueError,
-                                    "`label_mapping` must be a dict which maps "
-                                    "labels to index encodings."):
+        with self.assertRaisesRegex(TypeError,
+                                    "Labels must either be an encoding dict "
+                                    "which maps labels to index encodings or a "
+                                    "list."):
             base_model.BaseModel.set_label_mapping(mock_model, label_mapping={})
```
AnhTruong, line 174: can we adjust the message to cope with non-empty dict requirement?

AnhTruong (commented):
added some comments

dataprofiler/labelers/base_model.py
```diff
@@ -135,17 +156,51 @@ def set_params(self, **kwargs):
...
+        if same_as and not isinstance(same_as, str):
+            raise TypeError('`same_as` must be a str.')
+        elif same_as and same_as not in self._label_mapping:
+            raise ValueError('`same_as` value: {}, did not exist in the '
+                             'label_mapping.'.format(same_as))
+
+        # add label to label_mapping
+        max_label_ind = max(self._label_mapping.values())
+        self._label_mapping[label] = self._label_mapping.get(same_as,
+                                                             max_label_ind + 1)
```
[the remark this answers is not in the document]
JGSweets: The readme still contains the example of replacing the label_mapping.

dataprofiler/tests/labelers/test_data_labelers.py
JGSweets: Technically, but you could also do it at `fit` using the labels `parameter` as in the readme

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
JGSweets: This was an integration test specifically for the model. I could prob test it by mocking some things out as well if you like.

README.md
```diff
@@ -704,14 +701,43 @@ data_labeler = dp.DataLabeler(labeler_type='structured', trainable=True)
...
+new_labels = ['label1', 'label2', ...]
+data = dp.Data("your_file.csv")  # contains data with new labels
+
+# load default structured Data Labeler w/ trainable set to True
+data_labeler = dp.DataLabeler(labeler_type='structured', trainable=True)
+
+# this will maintain current labels and model weights, but extend the model's 
+# labels
+for label in new_labels:
+    data_labeler.add_label(label)
```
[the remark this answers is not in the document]
JGSweets: no, this was discussed and chosen to be iterative as asked by @lettergram. Would also complicate the `same_as` parameter slightly.

dataprofiler/labelers/base_model.py
```diff
@@ -90,6 +91,26 @@ def labels(self):
...
+
+        :param labels: Labels to convert to an encoding dict
+        :type labels: Union[list, dict]
+        :param requires_zero_mapping: boolean if the label mapping requires the
+            mapping for index 0 reserved.
+        :type requires_zero_mapping: bool
+        :return: label encoding dict
+        """
+        if isinstance(labels, dict):
+            return labels
```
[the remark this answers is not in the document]
JGSweets: that does not occur by base_model, each individual model has its own check. if needed.

dataprofiler/tests/labelers/test_base_model.py
JGSweets: Can you explain further what you mean / provide an example?

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -198,17 +198,23 @@ def set_label_mapping(self, label_mapping):
...
-            raise TypeError("`label_mapping` must be a dict which maps labels "
-                            "to index encodings.")
-        if 'PAD' not in label_mapping and 0 not in label_mapping.values():
-            label_mapping.update({'PAD': 0})
-        if label_mapping.get('PAD', None) != 0:
+        if 'PAD' not in label_mapping:
+            if isinstance(label_mapping, list):  # if list missing PAD
+                label_mapping = ['PAD'] + label_mapping
+            elif 0 not in label_mapping.values():  # if dict missing PAD and 0
+                label_mapping.update({'PAD': 0})
```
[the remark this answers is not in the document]
JGSweets: notice below that it will error out if PAD is not 0
https://github.com/capitalone/DataProfiler/pull/48/files/dcd7a73cc57075ea2866f74146e58a7b99dde343#diff-8ae9227e36e999b076bef5c67ba99f6cf22ecb8c02734691cbf045fb931dfa26R212

README.md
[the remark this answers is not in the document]
gme5078: Can we include an example with same_as?

README.md
```diff
@@ -704,14 +701,44 @@ data_labeler = dp.DataLabeler(labeler_type='structured', trainable=True)
...
+#       Currently, the DataLabeler cannot take in Tabular data, but requires 
+#       data to be ingested with two columns [X, y] where X is the samples and 
+#       y is the labels.
+model_results = data_labeler.fit(x=data['samples'], y=data['labels'], 
+                                 validation_split=0.2, epochs=2)
+
+# final_results, final_confidences are a list of results for each epoch
+epoch_id = 0
+final_results = model_results[epoch_id]["pred"]
+final_confidences = model_results[epoch_id]["conf"]
```
gme5078, line 740: Do we need these last several lines to show adding labels?

dataprofiler/tests/labelers/test_base_model.py
```diff
@@ -96,25 +96,98 @@ def test_set_parameters(self):
...
+        with self.assertRaisesRegex(TypeError, '`label` must be a str.'):
+            base_model.BaseModel.add_label(mock_model, label=None)
+
+        with self.assertRaisesRegex(TypeError, '`label` must be a str.'):
+            base_model.BaseModel.add_label(mock_model, label=1)
+
+        # assert existing label
+        label = 'NEW_LABEL'
+        mock_model._label_mapping = {'NEW_LABEL': 1}
+        with self.assertWarnsRegex(UserWarning,
```
gme5078, line 117: more specific warning than UserWarning?

dataprofiler/tests/labelers/test_base_model.py
```diff
@@ -96,25 +96,98 @@ def test_set_parameters(self):
         with self.assertRaisesRegex(ValueError, 'test'):
             base_model.BaseModel.set_params(mock_model, **params)
 
-    def test_set_label_mapping_parameters(self):
+    def test_add_labels(self):
 
         # setup mock
         mock_model = mock.Mock(spec=base_model.BaseModel)
```
gme5078, line 102: This is a lot to mock. Can we mock the tensorflow model instead of the whole baseModel class?

gme5078 (changes requested):
I liked moving the onus of label mapping from the data labeler to the model

AnhTruong (approved):

dataprofiler/tests/labelers/test_base_model.py
JGSweets: I do not think this should be separate from user as it is directly created by the user's add here.

README.md
JGSweets: I think it's good to show a complete story, but if the desire is to remove, i'm okay with that.

README.md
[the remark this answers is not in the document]
JGSweets: done

gme5078 (approved):

dataprofiler/labelers/base_model.py
```diff
@@ -135,17 +156,51 @@ def set_params(self, **kwargs):
...
-        :type label_mapping: dict
+        :param label_mapping: label mapping of the model or list of labels to be
+            converted into the label mapping
+        :type label_mapping: Union[list, dict]
         :return: None
         """
-        if not label_mapping or not isinstance(label_mapping, dict):
-            raise ValueError("`label_mapping` must be a dict which maps labels "
-                             "to index encodings.")
+        if not isinstance(label_mapping, (list, dict)) or not label_mapping:
```
ChrisWallace2020, line 199: Is there a point to the second half of this if statement? if label_mapping is None then it will already make the isinstance() call fail, which will short circuit and the second half of the if statement would never get reached.

PR #49, opened 2021-03-08
Int Profile Merge

gme5078:
Here is an update to int profile merging. This allows calculations to merge to the lowest common denomination without erroring. Numeric stats needed to be updated to handle options as well.


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -130,39 +130,42 @@ def _add_helper(self, other1, other2):
...
+                self.max = other1.max
+            else:
+                self.max = other2.max
+        if "sum" in self.__calculations.keys():
+            if other1.sum != 0 and other2.sum != 0:
+                self.sum = other1.sum + other2.sum
+            elif other2.sum == 0:
+                self.sum = other1.sum
+            else:
+                self.sum = other2.sum
```
ChrisWallace2020, line 168: Are these 3 separate cases necessary? If you case on whether one or the other is zero, so that you can just use the other sum, then thats the same as the first case when you add them, so you can just always add them like in the first case. Right?

gme5078: Yep, good catch, fixed

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -460,3 +460,35 @@ def test_profile_merge_no_bin_overlap(self):
...
+
+        # Creating second profiler with separate options
+        options = IntOptions()
+        options.min.is_enabled = False
+        data2 = [10, 15]
+        df2 = pd.Series(data2).apply(str)
+        profiler2 = IntColumn("Int", options=options)
+        profiler2.update(df2)
+
+        # Asserting error when adding 2 profilers with different options
```
JGSweets, line 484: assert raises error?

JGSweets: also should with a `with self.assertWarnsRegex(...)`

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -176,6 +176,8 @@ def test_timeit_merge(self):
         other2.min, other2.max, other2.variance, other2.sum = 1, 1, 1, 1
 
         # set auto as only histogram to merge
+        other1.histogram_selection = "auto"
```
JGSweets, line 179: this error'ed without it?

JGSweets (changes requested):

JGSweets:
We should create a task for the user to specify raise error instead of warning via options.

dataprofiler/profilers/base_column_profilers.py
```diff
@@ -129,6 +130,15 @@ def _perform_property_calcs(self, calculations, df_series,
...
                                    subset_properties)
 
+    @staticmethod
+    def _merge_calculations(merged_profile_calcs, profile1_calcs, profile2_calcs):
+        calcs = list(merged_profile_calcs.keys())
+        for calc in calcs:
+            if calc not in profile1_calcs or calc not in profile2_calcs:
+                del merged_profile_calcs[calc]
+                if calc in profile1_calcs or calc in profile2_calcs:
+                    warnings.warn("{} is disabled because it is not enabled in both profiles.".format(calc))
```
JGSweets, line 140: validate proper warning class:
https://docs.python.org/3/library/warnings.html

dataprofiler/tests/profilers/test_int_column_profile.py
gme5078: Fixed

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
gme5078: Yes because histogram selection is used whether to determine if the histogram exists or not (deciding, is it mergeable or not)

lettergram (dismissed):

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
JGSweets: Can you test setting it to None, which is the base? Or do we have a test which evaluates merging two profiles which are trained on empty data? Just want to make sure `None` (default value) doesn't cause a failure.

JGSweets: This test may exist.

gme5078: The following assertions fail, but the code does not have an error with None

lettergram (approved):

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
JGSweets: should note their does exist a test which tests empty profile adds for int

JGSweets (approved):

PR #51, opened 2021-03-10
Add Tests For BooleanOption and BaseColumnOptions Classes

sagars729:
- Added tests for BooleanOption Class
- Added class methods for creating a new Options object and also for getting the variable path for an Options object to allow tests to inherit each other (more useful in future tests)
- Added variable path type check to BooleanOption _validate_helper as method and check were not implemented in BaseOption class
- [x] Tests for BaseColumn Class ~~(IP)~~


dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,90 @@
...
+	def setUpClass(cls):
+		cls.data = Data(data=pd.DataFrame([1, 2]), data_type='csv')
+	
+	@classmethod
+	def getOptions(self, **params):
+		return BooleanOption(**params)
+	
+	@classmethod
+	def getOptionsPath(self, **params):
+		return "BooleanOption"
```
sagars729, line 27: This could be a class variable instead of a method. I wasn't sure if the params of the Option classes may eventually change the variable path.

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -17,12 +17,21 @@ class TestBaseOption(unittest.TestCase):
 	def setUpClass(cls):
 		cls.data = Data(data=pd.DataFrame([1, 2]), data_type='csv')
 	
+
+	@classmethod
+	def getOptions(self, **params):
+		return BaseOption(**params)
```
AnhTruong, line 23: should name it as `getBaseOption`

ChrisWallace2020: But shouldn't that not matter if in the boolean tests you just use super().<test_function>? Because then it'll defer to the super's implementation which can be named something different, and it'll still work in the boolean case? Idk maybe I'm misunderstanding the goal.

sagars729: Consider the super().test_validate_helper() call in test_validate_helper for TestBooleanOption. If I used getBaseOption in this class, then regardless of which class is calling super, it will create a new BaseOption object. By creating a getOptions method that is overriden the getOptions call in the TestBaseOption class will create a BooleanOption (when TestBooleanOption is calling its super) which is the intended behavior.

ChrisWallace2020: It seems I had a lapse in my python understanding, I thought in the call to super() it wouldn't make calls to the Boolean class, I just tested this out in a jupyter notebook and realized I was wrong. Disregard my comment.

lettergram, line 23: Why are the tabs 8 spaces here, but only 4 spaces earlier?

sagars729: I believe GitHub's default tab spacing is 4. The view will change if you use the following link: [https://github.com/capitalone/DataProfiler/pull/51/files?ts=4](https://github.com/capitalone/DataProfiler/pull/51/files?ts=4)

lettergram (dismissed):

dataprofiler/tests/profilers/test_base_option.py
JGSweets: I believe int his case we should use 4 spaces, or set the IDE to do 4 spaces when you hit tab.

JGSweets: `setUpClass` is in camel case because of the design of unittest, but i'm not sure `getOptions` should also be camel case but rather snake case as normalized for python.

JGSweets, line 23: I think it might be better if we use a class variable o define what get options is such that the new tests requires it to be defined. Otherwise, anything that imports this class could forget to alter the getOptions and end up with the wrong unittests. Additionally, it might be best to create it as an abstract test set which then forces the setting.

i.e.
```python

class AbstractTestOptions...(object):
    
    option_class = None
    
    @classmethod
    def get_options(self, *args, **kwargs):
        if not self.option_class:
            raise ....
        return self.option_class(*args, **kwargs)

    .
    .
    . <other abstract tests which are universal>


class TestBaseOptions(AbstractTestOptions, unittest.....):

    option_class = BaseOption
    
    .
    .
    . <Tests etc...>
```

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,90 @@
...
+import pandas as pd
+
+from dataprofiler import Data, ProfilerOptions, Profiler
+from dataprofiler.profilers.profiler_options import BooleanOption
+from dataprofiler.tests.profilers.test_base_option import TestBaseOption
+
+
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.'
+            'DataLabelerColumn.update', return_value=None)
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.DataLabeler')
```
JGSweets, line 14: are these necessary?

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,90 @@
...
+class TestBooleanOption(TestBaseOption):
+	
+	@classmethod
+	def setUpClass(cls):
+		cls.data = Data(data=pd.DataFrame([1, 2]), data_type='csv')
+	
+	@classmethod
+	def getOptions(self, **params):
```
JGSweets, line 22: same get options statement

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,90 @@
...
+		return BooleanOption(**params)
+	
+	@classmethod
+	def getOptionsPath(self, **params):
```
JGSweets, line 26: like getOptions, casing / should be abstract and set as class prop?

JGSweets: should path be inferred by the class name?

JGSweets (changes requested):
Made some initial thoughts.

dataprofiler/tests/profilers/test_base_option.py
sagars729: My own IDE does 4 spaces but GitHub does not - I can replace all "\t" with four spaces if that's preferable.

sagars729: Sure, I can create an abstract class that is implemented by the other TestOptions classes and have it check and enforce that setting.

dataprofiler/tests/profilers/test_boolean_option.py
sagars729: I'm not sure - they were included in the original profiler options tests.

dataprofiler/profilers/profiler_options.py
```diff
@@ -145,6 +148,20 @@ def __init__(self):
...
+        Validates the options do not conflict and cause errors.
+
+        :param variable_path: current path to variable set.
+        :type variable_path: str
+        :return: list of errors (if raise_error is false)
+        :rtype: list(str)
+        """
+        if not isinstance(variable_path, str):
+            raise ValueError("The variable path must be a string.")
+
```
sagars729, line 162: This is probably redundant since the call to super() will raise the same error

dataprofiler/tests/profilers/test_base_option.py
JGSweets: Yes, please use spaces. for the IDE, I believe you can set it to convert your tab to spaces. such that it isn't saved as a \t

dataprofiler/profilers/profiler_options.py
```diff
@@ -129,6 +129,9 @@ def _validate_helper(self, variable_path='BooleanOption'):
         :return: list of errors (if raise_error is false)
         :rtype: list(str)
         """
+        if not isinstance(variable_path, str):
```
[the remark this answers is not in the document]
sagars729: See comment below on repetition in BaseColumnOption. I could add the check to the BaseOption class instead (raise a ValueError before the NotImplementedError) to avoid any repetition.

sagars729:
@JGSweets added new commits that

- Removed redundant type check from BaseColumnOptions class
- Fixed casing of getOptions and getOptionsPath to be get_options and get_options_path
- Replaced all "\t" characters with "    "
- Created a option_class class variable to determine the Class
- Created an abstract test options class that contains get_options and get_options_path methods 
- Inferred option variable path in the get_options_path method rather than explicitly defining it

I am still unsure if the mocks are necessary or not necessary.

dataprofiler/tests/profilers/test_base_column_options.py
```diff
@@ -0,0 +1,58 @@
...
+        super().test_set_helper(*mocks)
+
+    def test_set(self, *mocks):
+        super().test_set(*mocks)
+    
+    def test_validate_helper(self, *mocks):
+        super().test_validate_helper(*mocks)
+    
+    def test_validate(self, *mocks):
+        super().test_validate(*mocks)
```
sagars729, line 42: This is redundant (I am overriding the super method to call the super method). It does make the file easier to read though as it's clear what tests are being run (which is good if these tests need to be modified in the future).

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -6,56 +6,56 @@
...
-	def test_validate(self, *mocks):
-		options = BaseOption()
-
-		with self.assertRaises(NotImplementedError):
-			options.validate()
+class TestBaseOption(AbstractTestOptions, unittest.TestCase):
+    
+    @classmethod
+    def setUpClass(cls):
+        cls.option_class = BaseOption
```
JGSweets, line 18: as above may change depending on how abstractbaseclass is setup

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -6,56 +6,56 @@
...
+        options = self.get_options()
+
+        with self.assertRaises(NotImplementedError):
+            options._validate_helper()
+    
+    def test_validate(self, *mocks):
+        options = self.get_options()
+
+        with self.assertRaises(NotImplementedError):
+            options.validate()
```
JGSweets, line 61: If these are generic, I would recommend moving them to the abstract base class.

We are treating it more like a mixing for testing TBH.

JGSweets: if not generic then fine to keep in place.

dataprofiler/tests/profilers/test_boolean_option.py
JGSweets: I don't think they are as we don't use the datalabeler in these tests, however previous tests did have it, e.g. `test_data_labeler`

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,82 @@
...
+        option = self.get_options(is_enabled="Hello World")
+        expected_error = "{}.is_enabled must be a Boolean.".format(optpth)
+        self.assertEqual([expected_error], option._validate_helper())
+    
+    def test_validate(self, *mocks):
+        option = self.get_options(is_enabled=True)
+        optpth = self.get_options_path()
+    
+        # Default Configuration Is Valid
+        self.assertEqual([], option.validate())
```
JGSweets, line 75: I would like weigh in from others, but I think we might want to have option.validate() return None if no errors as opposed to an empty list. Thoughts?

JGSweets: This of course would require updating the validate logic around https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profiler_options.py#L109

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,82 @@
...
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.DataLabeler')
+class TestBooleanOption(TestBaseOption):
+    
+    @classmethod
+    def setUpClass(cls):
+        cls.option_class = BooleanOption    
+        
+    def test_init(self, *mocks):
+        option = self.get_options()
+        self.assertDictEqual({"is_enabled": True}, option.properties)
```
JGSweets, line 23: we should also assert when setting `is_enabled=True` and `is_enabled=False`

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -0,0 +1,82 @@
...
+    
+        # Default Configuration Is Valid
+        self.assertEqual([], option.validate())
+        
+        # Option is_enabled is not a boolean
+        option = self.get_options(is_enabled="Hello World")
+        expected_error = "{}.is_enabled must be a Boolean.".format(optpth)
+        with self.assertRaisesRegex(ValueError, expected_error):
+            option.validate(raise_error=True)
+        self.assertEqual([expected_error], option.validate(raise_error=False))
```
JGSweets, line 82: nice!

dataprofiler/tests/profilers/test_boolean_option.py
JGSweets: I only reference here as this is where the api meets the user.

JGSweets (changes requested):
Overall pretty good, some thoughts etc.

dataprofiler/profilers/profiler_options.py
```diff
@@ -145,6 +148,17 @@ def __init__(self):
...
+        """
+        Validates the options do not conflict and cause errors.
+
+        :param variable_path: current path to variable set.
+        :type variable_path: str
+        :return: list of errors (if raise_error is false)
+        :rtype: list(str)
+        """
+        return super()._validate_helper(variable_path) 
+
```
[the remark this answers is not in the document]
sagars729: It actually ends up being necessary because of the default argument (variable_path = "BaseColumnOptions") without it, all the variable paths end up being incorrect.

dataprofiler/tests/profilers/abstract_test_options.py
```diff
@@ -0,0 +1,14 @@
+
+class AbstractTestOptions():
+
```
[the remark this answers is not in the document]
sagars729: I want to avoid adding a setUpClass method since this class isn't inheriting unittest. I will add an option_class variable set it to None.

dataprofiler/tests/profilers/abstract_test_options.py
```diff
@@ -0,0 +1,14 @@
...
+class AbstractTestOptions():
+    
+    @classmethod
+    def get_options(self, *args, **kwargs):
+        if self.option_class == None:
```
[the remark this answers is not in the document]
sagars729: For now, since the files are separate, I can add in that check.

dataprofiler/tests/profilers/test_base_option.py
sagars729: They aren't generic, but every test class does need to implement these methods. I will at least have the methods raise a NotImplementedError if they aren't implemented by a test class.

dataprofiler/tests/profilers/abstract_test_options.py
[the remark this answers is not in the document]
sagars729: Added in new commit

dataprofiler/tests/profilers/test_base_option.py
sagars729: added in new commit

dataprofiler/tests/profilers/abstract_test_options.py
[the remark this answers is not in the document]
sagars729: Added in new commit. I first checked if option_class was of type "type" and then checked if an object of that type was an instance of BaseOption.

dataprofiler/profilers/profiler_options.py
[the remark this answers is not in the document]
JGSweets: good point!

dataprofiler/tests/profilers/test_boolean_option.py
sagars729: See lines 25-28. Those could be moved to test_init, but I want classes that inherit TestBooleanOption (like TestBaseColumnOptions) to check those (but not call super().test_init() since that would fail).
The purpose of test_init was to verify that default params (which may be assumptions in tests) haven't changed.

dataprofiler/tests/profilers/test_base_column_options.py
```diff
@@ -0,0 +1,58 @@
...
+from dataprofiler import Data, ProfilerOptions, Profiler
+from dataprofiler.profilers.profiler_options import BaseColumnOptions
+from dataprofiler.tests.profilers.test_boolean_option import TestBooleanOption
+
+
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.'
+            'DataLabelerColumn.update', return_value=None)
+@mock.patch('dataprofiler.profilers.data_labeler_column_profile.DataLabeler')
+class TestBaseColumnOptions(TestBooleanOption):
+
```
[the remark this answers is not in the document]
sagars729: Which option class? I am a little unclear on what is being referenced

dataprofiler/tests/profilers/test_boolean_option.py
JGSweets: I would consider these two separate tests as `_set_helper` should only test the `_set_helper` functionality and not whether the `__init__` properly constructs the option class. I just don't find it as intuitive that init tests would be contained elsewhere.

I think if you want to make the `init` more agnostic, you could do 1 of two things. 

1. only look at `options.is_enabled`
this validates class props are properly being set. We also make the test agnostic to the properties call which likely could be validated elsewhere (perhaps a separate test).
2. setup `test_init` in the abstract test class to take in the expected properties for evaluation

JGSweets: of course other options exist as well, just thinking quickly here.

dataprofiler/tests/profilers/test_boolean_option.py
sagars729: My comment on this would be that it may break later implementations of validate_helper. Currently, validate_helper calls its super class to get an initial list of errors then continues to add to that list. if None is returned, you would need to add some different logic:
```python
if errors == None: errors = []
...
```

dataprofiler/tests/profilers/test_base_column_options.py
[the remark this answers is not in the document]
JGSweets: I guess it isn't necessary to move it, but technically we could get rid of 

```python
@classmethod
def setUpClass(cls):
    cls.option_class = BaseColumnOptions
```

and just do 

```python
class ...:
    option_class = BaseColumnOptions
    .
    .
    .
```

which will be less code in the long run for someone to type

dataprofiler/tests/profilers/test_boolean_option.py
JGSweets: I wouldn't alter `validate_helper`, only `validate` which isn't referencing itself.

dataprofiler/tests/profilers/test_boolean_option.py
sagars729: That does make sense - it isn't intuitive to and init checks in the set_helper method. I modified the code slightly to better fit my intention with the tests. Now, it's still testing enabling and disabling the option with set/set_helper and verifying the init separately in the test_init test.

dataprofiler/tests/profilers/test_base_column_options.py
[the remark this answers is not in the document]
sagars729: I think that might defeat the purpose of the check in the abstract class. TestBaseColumnOptions would just inherit option_class from BooleanOption and use the BooleanOption class for its tests. If we enforce using the setUpClass method (which is generally included because of unittest), then if TestBaseColumnOptions does not set option_class it would default to None.

dataprofiler/profilers/profiler_options.py
```diff
@@ -106,7 +106,8 @@ def validate(self, raise_error=True):
         errors = self._validate_helper()
         if raise_error and errors:
             raise ValueError('\n'.join(errors))
-        return errors
+        elif errors: 
+            return errors
```
sagars729, line 110: @JGSweets modified to return None on validate with tests changed accordingly

JGSweets: ty

JGSweets (approved):

JGSweets:
@sagars729 please update the brach with master

AnhTruong (approved):

AnhTruong:
Will merge once the branch is up-to-date with the master

AnhTruong:
@sagars729

PR #53, opened 2021-03-10
Updated Model

lettergram:


dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -420,6 +420,8 @@ def test_data_labeler_toggle(self):
     def test_null_count(self):
         column = pd.Series([1, float('nan')] * 10)
 
+        print(column)
```
AnhTruong, line 423: remove?

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -123,26 +128,18 @@ def test_data_labeler_extend_labels(self):
...
-            "ORDINAL": 19, "NEW_LABEL": 20}
         model_predictions = data_labeler.fit(x=self.df[0], y=self.df[1])
 
         self.assertEqual(1, len(model_predictions))
         self.assertEqual(3, len(model_predictions[0]))  # history, f1, f1_report
         self.assertIsInstance(model_predictions[0][0], dict)  # history
         self.assertIsInstance(model_predictions[0][1], float)  # f1
         self.assertIsInstance(model_predictions[0][2], dict)  # f1_report
-        self.assertDictEqual(expected_label_mapping, data_labeler.label_mapping)
+        self.assertTrue(new_label in data_labeler.label_mapping) # Ensure new label added
```
JGSweets, line 138: If we want to keep this assert, i would suggest using:
`self.assertIn`. However, I do think it would be better to assert that the label is at the lend of the label set and that the original labels are unchanged.

One option would be to manually set the labels, then add, because then we control the labels.
Another option would be to grab the label mapping before and after the add. validate it is just an add to the end. however, this no longer becomes explicit and could succumb to a relational error.

These current two tests, while unlikely could still have errors if add_label were altered i.e. the label added coverts all labels to something else but maintains that label.

Just my thoughts.

JGSweets (changes requested):
small comment, not necessarily a block but should be considered that the new unit tests could cause relational errors.

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
lettergram: See update

AnhTruong (approved):

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -123,26 +136,19 @@ def test_data_labeler_extend_labels(self):
...
         model_predictions = data_labeler.fit(x=self.df[0], y=self.df[1])
 
         self.assertEqual(1, len(model_predictions))
         self.assertEqual(3, len(model_predictions[0]))  # history, f1, f1_report
         self.assertIsInstance(model_predictions[0][0], dict)  # history
         self.assertIsInstance(model_predictions[0][1], float)  # f1
         self.assertIsInstance(model_predictions[0][2], dict)  # f1_report
-        self.assertDictEqual(expected_label_mapping, data_labeler.label_mapping)
+        self.assertIn(new_label, data_labeler.label_mapping) # Ensure new label added
+        self.assertEqual(original_max_label+1, new_max_label) # Ensure new label iterated
```
JGSweets, line 147: redundant with `self.assertDictEqual(expected_label_mapping, data_labeler.label_mapping)`
but will approve

JGSweets (approved):

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -113,8 +113,21 @@ def test_data_labeler_extend_labels(self):
             dp.labelers.StructuredDataLabeler._default_model_loc)
         data_labeler = dp.labelers.TrainableDataLabeler(dirpath=dirpath)
 
-        data_labeler.add_label('NEW_LABEL')
-
+        original_label_mapping = data_labeler.label_mapping.copy()
```
JGSweets, line 116: still concerned that this test could create relative errors as we aren't being explicit

JGSweets: but i'm not blocking approval

JGSweets (approved):
Person i believe is negatively impacted in this model, however there was improvement on other labels and the addition of new labels.

I think improvement on person needs to be made in the next iteration.

AnhTruong (dismissed):

JGSweets (changes requested):
As @AnhTruong noted, drivers license is missing

AnhTruong:
Yeah, please add DRIVERS_LICENSE

JGSweets (approved):
Previous comments about person and other labels still apply.

gme5078 (approved):

AnhTruong (approved):

PR #55, opened 2021-03-12
Update _check_and_return_valid_data_format and preprocessors to operate on np arrays

ChrisWallace2020:
Validated that _check_and_return_valid_data_format was compatible with Data reader objects so they could be passed to fit/predict for labelling. 

Made np arrays the canonical return from _check_and_return_valid_data_format and updated adjacent code accordingly to expect such output.

Modified code to do away with convention of storing unstructured labels in a dictionary with "entities" key, instead just stored as list of entities. 

Updated docstrings and .help() functions accordingly.


dataprofiler/labelers/base_data_labeler.py
```diff
@@ -145,32 +146,31 @@ def _check_and_return_valid_data_format(data, fit_or_predict='fit'):
...
-        else:
+            data = data.data
+
+        if isinstance(data, list):
+            data = np.array(data, dtype="object")
+        elif isinstance(data, pd.Series) or isinstance(data, pd.DataFrame):
+            data = data.values
+        elif not isinstance(data, np.ndarray):
             raise TypeError(
                 "Data must either be imported using the data_readers or "
```
JGSweets, line 166: likely need to update this

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -738,14 +734,11 @@ def fit(self, x, y, validation_split=0.2, labels=None, reset_weights=False,
...
-                train_data_x = x.loc[shuffle_inds].reset_index(drop=True)
-                train_data_y = y.loc[shuffle_inds].reset_index(drop=True)
-
-                # free memory
-                del shuffle_inds
+                # shuffle input data
+                train_data_x = copy.deepcopy(x)
+                train_data_y = copy.deepcopy(y)
+                np.random.shuffle(train_data_x)
+                np.random.shuffle(train_data_y)
```
JGSweets, line 741: this change will now train on the whole dataset as oppose to the subset of training only (removing cv)
I think we should maintain the shuffle_inds for now

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -716,13 +716,9 @@ def fit(self, x, y, validation_split=0.2, labels=None, reset_weights=False,
...
-        # shuffle input data, these are pandas.DataFrame or pandas.Series
-        shuffle_inds = np.random.permutation(num_samples)
-        x = x.loc[shuffle_inds].reset_index(drop=True)
-        y = y.loc[shuffle_inds].reset_index(drop=True)
-
-        # free memory
-        del shuffle_inds
+        # shuffle input data
+        np.random.shuffle(x)
+        np.random.shuffle(y)
```
lettergram, line 721: X and Y wont be lined up

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -738,14 +734,11 @@ def fit(self, x, y, validation_split=0.2, labels=None, reset_weights=False,
...
-                # pandas.Series
-                shuffle_inds = np.random.permutation(cv_split_index)
-                train_data_x = x.loc[shuffle_inds].reset_index(drop=True)
-                train_data_y = y.loc[shuffle_inds].reset_index(drop=True)
-
-                # free memory
-                del shuffle_inds
+                # shuffle input data
+                train_data_x = copy.deepcopy(x)
+                train_data_y = copy.deepcopy(y)
```
lettergram, line 739: No deep copies plz

dataprofiler/labelers/base_data_labeler.py
JGSweets: Also shuffle_inds maintains x,y permutation matching

dataprofiler/labelers/data_processing.py
```diff
@@ -1081,7 +1086,7 @@ def help(cls):
             + param_docs[param_start_ind:param_end_ind]
             + "\nProcess Input Format:\n"
             "    data = List of strings ['1st string', 'second string', ...]\n"
-            "    labels = [<INT>, <INT>, ...(num_samples in data)]")
+            "    labels = [\"<LABEL>\", \"<LABEL>\", ...(num_samples in data)]")
```
JGSweets, line 1089: nice

dataprofiler/labelers/base_data_labeler.py
ChrisWallace2020: resolved

dataprofiler/labelers/base_data_labeler.py
ChrisWallace2020: resolved

dataprofiler/labelers/base_data_labeler.py
ChrisWallace2020: resolved

dataprofiler/labelers/data_processing.py
```diff
@@ -1102,12 +1107,12 @@ def convert_to_unstructured_format(self, data, labels):
         required input data format.
 
         :param data: list of strings
-        :type data: Union[numpy.ndarray, pandas.DataFrame]
+        :type data: numpy.ndarray
         :param labels: labels for each input character
         :type labels: list
         :return: data in the following format
-                 [ dict(text="<SAMPLE><SEPARATOR><SAMPLE>...",
```
JGSweets, line 1109: Hmm we should look into proper way to explain return when multiple outs

dataprofiler/tests/labelers/test_data_labelers.py
```diff
@@ -110,6 +110,69 @@ def test_load_data_labeler(self, *mocks):
         data_labeler = dp.DataLabeler(labeler_type='unstructured')
         self.assertIsInstance(data_labeler, BaseDataLabeler)
 
+    def test_structured_data_labeler_fit_predict_take_data_obj(self):
```
JGSweets, line 113: I'm concerned these are running the actual data labeler rather than validating that the function returns the proper formats and accepts the proper formats.

ChrisWallace2020: This is a good point, I mostly had this at first because I didn't know if the labeler could work with several formats, and if I would return several, but since its canonical now I suppose I could write this without calls to fit if you think that would be more appropriate.

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -716,10 +715,10 @@ def fit(self, x, y, validation_split=0.2, labels=None, reset_weights=False,
         if reset_weights:
             self._model.reset_weights()
 
-        # shuffle input data, these are pandas.DataFrame or pandas.Series
+        # shuffle input data
         shuffle_inds = np.random.permutation(num_samples)
-        x = x.loc[shuffle_inds].reset_index(drop=True)
-        y = y.loc[shuffle_inds].reset_index(drop=True)
+        x = x[shuffle_inds]
+        y = y[shuffle_inds]
```
lettergram, line 721: Why are we changing this?

dataprofiler/labelers/data_processing.py
```diff
@@ -614,7 +615,11 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
 
         # Import tensorflow
         import tensorflow as tf
-        
+
+        if data.ndim > 1:
+            data = data.reshape(-1)
+        data = data.astype(str)
```
lettergram, line 621: Could this cause conflicts of some kind?

dataprofiler/labelers/data_processing.py
```diff
@@ -1151,6 +1156,11 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
             raise ValueError('If `labels` are specified, `label_mapping` must '
                              'also be specified.')
 
+        if data.ndim > 1:
+            data = data.reshape(-1)
+        if labels is not None and labels.ndim > 1:
+            labels = labels.reshape(-1)
```
lettergram, line 1162: Reshaping one but not the other seems like a recipe for a mismatch, also a very expensive operation `O(n)`.

dataprofiler/labelers/data_processing.py
AnhTruong: do we really need it, as the function below this does have the processing for this already, i.e., `sample_buffer = str(sample_buffer)`

AnhTruong (commented):
one comment

dataprofiler/labelers/base_data_labeler.py
ChrisWallace2020: They are np arrays now, not dataframes, so we need to update use of .loc

dataprofiler/labelers/data_processing.py
ChrisWallace2020: I do this since the input data needs to be a list, and the labels are a list of entities, which can be lists themselves. So if the data is given as a 2D array, I want to flatten it, but I don't want to always do this with the labels since they are supposed to be 2D in some sense.

lettergram: At least there should be a test for this or an error.

dataprofiler/labelers/data_processing.py
```diff
@@ -598,23 +600,38 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
...
         :rtype batch_data: dicts
         """
-        if labels is not None and not label_mapping:
-            raise ValueError('If `labels` are specified, `label_mapping` must '
-                             'also be specified.')
+        if len(data.shape) > 1:
+            # Ignore case of 1 col DataFrame
+            if not (len(data.shape) == 2 and data.shape[1] == 1):
+                raise ValueError("Multidimensional data given to "
+                                 "CharPreprocessor")
```
lettergram, line 617: Should provide what they should do as well

dataprofiler/labelers/data_processing.py
```diff
@@ -1129,27 +1146,40 @@ def convert_to_unstructured_format(self, data, labels):
...
-                             'also be specified.')
+        if labels is not None:
+            if not label_mapping:
+                raise ValueError('If `labels` are specified, `label_mapping` '
+                                 'must also be specified.')
+            if data.shape != labels.shape:
+                raise ValueError("Data and labels given to "
+                                 "StructCharPreprocessor have different shapes")
+
+        if len(data.shape) > 1:
```
lettergram, line 1175: Should this be `>=1`?

AnhTruong (approved):

dataprofiler/labelers/data_processing.py
```diff
@@ -598,23 +600,38 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
...
         :type batch_size: int
         :return batch_data: A dict containing  samples of size batch_size
         :rtype batch_data: dicts
         """
-        if labels is not None and not label_mapping:
-            raise ValueError('If `labels` are specified, `label_mapping` must '
-                             'also be specified.')
+        if len(data.shape) > 1:
+            # Ignore case of 1 col DataFrame
+            if not (len(data.shape) == 2 and data.shape[1] == 1):
```
JGSweets, line 615: it is possible this could be more than 2 dimensions.

I would suggest validating that only 1 value in shape is > 1 to account for all cases

dataprofiler/labelers/data_processing.py
```diff
@@ -598,23 +600,38 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
...
+            # Flatten 1 col DataFrame
+            data = data.reshape(-1)
+
+        if labels is not None:
+            if not label_mapping:
+                raise ValueError('If `labels` are specified, `label_mapping` '
+                                 'must also be specified.')
+            # Need to cast if input is coming from structured process
+            if isinstance(labels, list):
+                labels = np.array(labels, dtype="object")
```
JGSweets, line 627: we should assume these labels are coming in as expected. the other processor should input the data into this processor as the correct type so that we don't have to make modifications for every processor that might input data into this one.

dataprofiler/labelers/data_processing.py
```diff
@@ -598,23 +600,38 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
...
+            if len(data) != len(labels):
+                raise ValueError("Data and labels given to "
+                                 "CharPreprocessor have different shapes")
```
JGSweets, line 630: I might add the shapes that are different via a format string
i.e. something like:
```python
raise ValueError("Data and labels given to CharPreprocessor have different shapes, {} != {}".format(data.shape, labels.shape))
```

dataprofiler/labelers/data_processing.py
```diff
@@ -1129,27 +1146,40 @@ def convert_to_unstructured_format(self, data, labels):
...
-        if labels is not None and not label_mapping:
-            raise ValueError('If `labels` are specified, `label_mapping` must '
-                             'also be specified.')
+        if labels is not None:
+            if not label_mapping:
+                raise ValueError('If `labels` are specified, `label_mapping` '
+                                 'must also be specified.')
+            if data.shape != labels.shape:
+                raise ValueError("Data and labels given to "
+                                 "StructCharPreprocessor have different shapes")
```
JGSweets, line 1173: since this is becoming repetitive, i wonder if there's a better way to do this. Might be a refactor in a future PR

dataprofiler/labelers/data_processing.py
```diff
@@ -1129,27 +1146,40 @@ def convert_to_unstructured_format(self, data, labels):
...
+
+        if len(data.shape) > 1:
+            if not (len(data.shape) == 2 and data.shape[1] == 1):
```
JGSweets, line 1176: as before, could be more than 2 dimensions we are accounting for.

dataprofiler/labelers/data_processing.py
```diff
@@ -1129,27 +1146,40 @@ def convert_to_unstructured_format(self, data, labels):
...
+                warnings.warn("Data given to StructCharPreprocessor was "
+                              "multidimensional, it will be flattened")
```
JGSweets, line 1178: maybe needs to mention this could affect results, because doesn't say why it needs to be warned, just warning the user. Also maybe add a little more description as in `flattened for model processing`

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
```diff
@@ -55,16 +53,14 @@ def test_fit_with_default_model(self):
...
+              (20, 22, 'ADDRESS'),
+              (22, 24, 'INTEGER_BIG')]],
             ['How nice.',
-             {'entities': [
-                 (0, 2, 'ADDRESS'),
-                 (4, 5, 'INTEGER_BIG'),
-                 (6, 8, 'INTEGER_BIG')]}]
+             [(0, 2, 'ADDRESS'),
+              (4, 5, 'INTEGER_BIG'),
+              (6, 8, 'INTEGER_BIG')]]
```
JGSweets, line 63: why is this changing?

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
```diff
@@ -12,16 +12,14 @@ class TestUnstructuredDataLabeler(unittest.TestCase):
...
+              (20, 22, 'ADDRESS'),
+              (22, 24, 'INTEGER_BIG')]],
             ['How nice.',
-                {'entities': [
-                 (0, 2, 'ADDRESS'),
-                 (4, 5, 'INTEGER_BIG'),
-                 (6, 8, 'INTEGER_BIG')]}]
+             [(0, 2, 'ADDRESS'),
+              (4, 5, 'INTEGER_BIG'),
+              (6, 8, 'INTEGER_BIG')]]
```
JGSweets, line 22: why is this changing?

JGSweets: nvm

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
JGSweets: nvm

dataprofiler/tests/labelers/test_data_labelers.py
```diff
@@ -110,25 +107,115 @@ def test_load_data_labeler(self, *mocks):
...
+        'ADDRESS': 2,
+        'PERSON': 3,
+    },
+    "decoder": {
+        0: "PAD",
+        1: "BACKGROUND",
+        2: "ADDRESS",
+        3: "PERSON"
+    },
+    "count": 4
```
JGSweets, line 216: no need to change this

dataprofiler/tests/labelers/test_data_labelers.py
```diff
@@ -110,25 +107,115 @@ def test_load_data_labeler(self, *mocks):
...
 }
 
-
```
JGSweets, line 131: need to validate pep8 such that this space should exist, below too

dataprofiler/tests/labelers/test_data_labelers.py
```diff
@@ -16,10 +16,8 @@
 from dataprofiler.labelers.data_labelers import BaseDataLabeler, \
     TrainableDataLabeler
 from dataprofiler.labelers import data_processing
-from dataprofiler.labelers import StructCharPreprocessor
 from dataprofiler.labelers.base_model import BaseModel, BaseTrainableModel
 
-
```
JGSweets, line 22: there should be two space between imports and next set of data

dataprofiler/labelers/data_processing.py
ChrisWallace2020: Good idea

dataprofiler/labelers/data_processing.py
```diff
@@ -1129,27 +1146,40 @@ def convert_to_unstructured_format(self, data, labels):
...
         """
-        if labels is not None and not label_mapping:
-            raise ValueError('If `labels` are specified, `label_mapping` must '
-                             'also be specified.')
+        if labels is not None:
+            if not label_mapping:
+                raise ValueError('If `labels` are specified, `label_mapping` '
+                                 'must also be specified.')
+            if data.shape != labels.shape:
+                raise ValueError("Data and labels given to "
```
gme5078, line 1172: I think this needs to be tested. Same for multiple parts I think

gme5078 (changes requested):

dataprofiler/tests/labelers/test_data_labelers.py
ChrisWallace2020: I actually don't know how this happened, I'll revert it though

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
lettergram, line 63: We should drop the "INTEGER_BIG" as it's not in the model any more

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
```diff
@@ -55,16 +53,14 @@ def test_fit_with_default_model(self):
...
             ['this is my test sentence.',
-             {'entities': [
-                 (5, 7, 'ADDRESS'),
-                 (11, 20, 'INTEGER_BIG'),
-                 (20, 22, 'ADDRESS'),
-                 (22, 24, 'INTEGER_BIG')]}],
+             [(5, 7, 'ADDRESS'),
+              (11, 20, 'INTEGER_BIG'),
+              (20, 22, 'ADDRESS'),
+              (22, 24, 'INTEGER_BIG')]],
```
lettergram, line 59: We should drop the "INTEGER_BIG" as it's not in the model any more

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
```diff
@@ -12,16 +12,14 @@ class TestUnstructuredDataLabeler(unittest.TestCase):
...
             ['this is my test sentence.',
-             {'entities': [
-                 (5 ,  7, 'ADDRESS'),
-                 (11, 20, 'INTEGER_BIG'),
-                 (20, 22, 'ADDRESS'),
-                 (22, 24, 'INTEGER_BIG')]}],
+             [(5,  7, 'ADDRESS'),
+              (11, 20, 'INTEGER_BIG'),
+              (20, 22, 'ADDRESS'),
+              (22, 24, 'INTEGER_BIG')]],
```
lettergram, line 18: We should drop the "INTEGER_BIG" as it's not in the model any more

dataprofiler/tests/labelers/test_integration_unstructured_data_labeler.py
lettergram, line 22: We should drop the "INTEGER_BIG" as it's not in the model any more

lettergram (changes requested):

dataprofiler/tests/labelers/test_data_processing.py
```diff
@@ -671,6 +673,22 @@ def test_process(self):
             self.assertTrue((expected[0] == output[0]).all())
             self.assertTrue((expected[1] == output[1]).all())
 
+    def test_process_input_checks(self):
+        prep = CharPreprocessor()
+        multi_dim_msg = re.escape("Multidimensional data given to "
```
gme5078, line 678: I dont think you need re.escape(), I think you should remove it and remove the re import

gme5078 (changes requested):

dataprofiler/tests/labelers/test_data_processing.py
ChrisWallace2020: I use it because of my use of parenthesis and characters like !, =

gme5078 (approved):

lettergram (approved):

PR #56, opened 2021-03-15
Added ability to likely identify all string delimited (CSV) data

lettergram:
* Fixes bug in header identification (int columns could never be identified, would always be float or date)
* Fixes readme instruction issue
* Adds multiple tests for all-string documents
* Utilizes word counts to suggest header ini all string delimited files; worked on many test files


dataprofiler/data_readers/csv_data.py
```diff
@@ -148,26 +149,28 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
-                        
-                    try:
-                        float(cell)
-                        cell_type = 'float'
-                    except ValueError:
-                        pass
-
                     try: 
-                        if dateutil.parser.parse(cell):
+                        if dateutil.parser.parse(cell, fuzzy=False):
```
lettergram, line 153: Had to move this up, integers will trigger this function

dataprofiler/data_readers/csv_data.py
```diff
@@ -148,26 +149,28 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
                     except ValueError:
                         pass
                     except OverflowError:
                         pass
+
+                    try:
+                        float(cell)
+                        cell_type = 'float'
+                    except ValueError:
+                        pass
```
lettergram, line 164: Float has to come next, as integers can still trigger this

dataprofiler/data_readers/csv_data.py
```diff
@@ -148,26 +149,28 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
+
+                    if cell.isdigit():
+                        cell_type = 'int'
```
lettergram, line 167: integers have to come last as integers can trigger prior try/except statements.

dataprofiler/data_readers/csv_data.py
```diff
@@ -148,26 +149,28 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
                     
                     if cell.isupper():
                         cell_type = 'upstr'
 
+                if cell_type not in ['str', 'none']:
+                    only_string_flag = False
```
lettergram, line 173: Checks if a document contains only strings. If so the standard functionality wont work

dataprofiler/data_readers/csv_data.py
```diff
@@ -220,7 +225,7 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
                 len_increase = True
                 
             # Returns the last row that could reasonably be the header
-            if (rstr > str_thresh and none <= none_thresh and diff <= diff_thresh):
+            if (rstr > str_thresh and none < none_thresh and diff < diff_thresh):
```
lettergram, line 228: Set to non-inclusive, just matches what I'd expect as a "threshold". For instance if there were 4 columns, we'd expect >2 columns to have header if we are to say a header exists at all.

dataprofiler/data_readers/csv_data.py
AnhTruong: I'd suggest using if cell_type=='float', then check if cell == int(float(cell)), cause otherwise, this doesn't work `'-123.isdigit()'`

dataprofiler/data_readers/csv_data.py
AnhTruong: i think keeping <= makes more sense, but we can always change the thresholds

dataprofiler/data_readers/csv_data.py
```diff
@@ -231,7 +236,61 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
         # If change in statistics never occurs, return no header
         if not change_flag:
             row_classic_header_ends = None
+
+
+        # Attempt to resolve case where only strings in every cell
+        if only_string_flag:
```
AnhTruong, line 242: not sure I understand this block. Why the existing code doesn't work for the files with only strings?

AnhTruong (commented):
some comments

dataprofiler/data_readers/csv_data.py
lettergram: I guess I view thresholds as max limits, i.e. you cannot go beyond that. Setting to be inclusive doesn't seem appropriate as it's "on the threshold". If we want to change it, that's probably fine, but then we should change the input constraints. 

Personally, I think this is opinion more than fact, I just wanted to hold it consistent across the board (non-inclusive).

dataprofiler/tests/data/csv/all-strings-skip-header-integers.csv
```diff
@@ -0,0 +1,6 @@
+Author: Austin Walters,,,
+Blog Post,Date,,
+1,2,3,4
```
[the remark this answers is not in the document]
lettergram: I think it's unclear personally 🤷‍♂️ 1-4 could be the header.

dataprofiler/tests/data/csv/all-strings-skip-partial-header.csv
```diff
@@ -0,0 +1,7 @@
+
+Blog Post,Date,,
```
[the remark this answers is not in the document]
lettergram: It is?

dataprofiler/data_readers/csv_data.py
lettergram: We discussed in chat, the logic currently used checks to see if there are changes in things such as: data type, data format, none counts, etc. Unfortunately, if you have a file of all strings none of the above can be applied. Instead, I utilized a method which checked the word counts per column and the difference down the column that is being evaluated. It appears effective at identifying the files I tested and also doesn't negatively impact the non-all-string files I tested.

dataprofiler/tests/data/csv/all-strings-skip-header-integers.csv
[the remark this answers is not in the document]
lettergram: @JGSweets please add an this as a github issue. Currently, integer based headers can't be detected; particularly if they're below an "author" or "description" row

dataprofiler/data_readers/csv_data.py
```diff
@@ -231,7 +235,61 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
+            col_stats = {}
+            rows = data_as_str.split('\n')
+            for i in range(0, len(rows)):
+                cells = rows[i].split(delimiter)
+                for j in range(0, len(cells)):
+
+                    # Determine number of words in cell
+                    words = cells[j].split(' ')
+                    word_count = len(words)
+                    if len(words[0])==0: word_count = 0
```
ChrisWallace2020, line 251: Make this more straightforward

dataprofiler/data_readers/csv_data.py
```diff
@@ -231,7 +235,61 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
+        if only_string_flag:
+            col_stats = {}
+            rows = data_as_str.split('\n')
+            for i in range(0, len(rows)):
+                cells = rows[i].split(delimiter)
+                for j in range(0, len(cells)):
+
+                    # Determine number of words in cell
+                    word_count = 0
+                    if len(cells[j]) > 0:
```
lettergram, line 250: add strip here

dataprofiler/data_readers/csv_data.py
```diff
@@ -231,7 +235,62 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
         if not change_flag:
             row_classic_header_ends = None
+
+
+        # Attempt to resolve case where only strings in every cell
+        if only_string_flag:
+            col_stats = {}
+            rows = data_as_str.split('\n')
+            for i in range(0, len(rows)):
+                cells = rows[i].split(delimiter)
```
JGSweets, line 245: this needs to account for delimiters in quotes

dataprofiler/data_readers/csv_data.py
```diff
@@ -136,6 +136,7 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
         
         # Determine type for every cell
         header_check_list = []
+        only_string_flag = True # Requires additional checks
```
JGSweets, line 139: the line below split(delimiter) this needs to account for delimiters in quotes

JGSweets (changes requested):
Need tests validating these work for cases where a delimiter exists in quotes for both the header and values 
Potentially needs 3 tests:
delimiter in quotes in header
delimiter in quotes in values
delimiter in quotes in values and headers (multi locations)

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -103,6 +106,18 @@ def setUpClass(cls):
...
+                 count=11, delimiter=',', has_header=[0],
+                 num_columns=4, encoding='utf-8'),
+            dict(path=os.path.join(test_dir, 'csv/all-strings-skip-header.csv'),
+                 count=11, delimiter=',', has_header=[1],
+                 num_columns=4, encoding='utf-8'),
+            dict(path=os.path.join(test_dir, 'csv/all-strings-skip-header-author.csv'),
+                count=7, delimiter=',', has_header=[1],
+                 num_columns=4, encoding='utf-8'),
+            dict(path=os.path.join(test_dir, 'csv/all-strings-skip-partial-header.csv'),
+                 count=7, delimiter=',', has_header=[None],
```
AnhTruong, line 119: can you add a comment that we don't detect the header with none values

dataprofiler/data_readers/csv_data.py
```diff
@@ -231,7 +235,64 @@ def _guess_header_row(data_as_str, suggested_delimiter=None,
...
+
+                    # Identify min / max for a column
+                    if word_count > col_stats[j]['max']:
+                        col_stats[j]['max'] = word_count
+                    if word_count < col_stats[j]['min']:
+                        col_stats[j]['min'] = word_count
+
+                    # First index with value
+                    if 'first_index_with_value' not in col_stats[j] and word_count > 0:
+                        col_stats[j]['first_index_with_value'] = (i, word_count)
```
ChrisWallace2020, line 268: You never actually index into [1] for these entries to access the word count, is it still necessary to store it then?

dataprofiler/data_readers/csv_data.py
lettergram: Done

JGSweets (approved):

AnhTruong (approved):

PR #60, opened 2021-03-15
add header test and fix to allow user to set even if 0

JGSweets:
Previously, a user could not set the header if 0 and no valid header options were checked. 

This PR adds both tests and functionality to allow the user to set the header at any index of the file.


dataprofiler/data_readers/csv_data.py
```diff
@@ -86,6 +86,22 @@ def delimiter(self):
...
+        :param options: dictionary of options for the csv reader to validate
+        :type options: dict
+        :return: None
+        """
+        options = super()._check_and_return_options(options)
+        for key, value in options.items():
+            if key == 'header' and value != 'auto' and value is not None \
+                    and not (isinstance(value, int) and value > -1):
+                raise ValueError('`header` must either be \'auto\' or an '
+                                 'integer > -1')
```
AnhTruong, line 102: should we change to non-negative integer?

lettergram (approved):

ChrisWallace2020:
This seems like a substantial backwards incompatible change (before header=None in options meant to infer one, now it means there isn't one, if I'm understanding correctly), so is there some place in docstrings or README where the new convention should be described?

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -172,6 +172,55 @@ def test_data_formats(self):
...
+    def test_set_header(self):
+        test_dir = os.path.join(test_root_path, 'data')
+        filename = 'csv/sparse-first-and-last-column-two-headers.txt'
+        filename = os.path.join(test_dir, filename)
+
+        # set bad header setting
+        options = dict(header=-2)
+        with self.assertRaisesRegex(ValueError,
+                                    '`header` must either be \'auto\' or an '
+                                    'integer > -1'):
```
AnhTruong, line 184: can you update the error message here to have the None (no header) value as well?

AnhTruong (commented):
some comments

dataprofiler/data_readers/csv_data.py
ChrisWallace2020: +1, describing the header relative to the value -1 is strange

lettergram: Personally, I think this description is okay... but should probably give a full explanation. i.e.

"`auto`, `none` for no header, or a non-negative integer for the row that represents the header" 

Something like that.

dataprofiler/data_readers/csv_data.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, input_file_path=None, data=None, options=None):
         self._selected_data_format = options.get("data_format", "dataframe")
         self._delimiter = options.get("delimiter", None)
         self._selected_columns = options.get("selected_columns", list())
-        self._header = options.get("header", 0)  # 0 for pandas to auto get
+        self._header = options.get("header", 'auto')
```
lettergram, line 68: This needs to be documented somewhere in the API. It should also be added to the `CSVData class` README section

JGSweets:
> This seems like a substantial backwards incompatible change (before header=None in options meant to infer one, now it means there isn't one, if I'm understanding correctly), so is there some place in docstrings or README where the new convention should be described?

0 technically was used as infer where as None meant no header. The problem is the user couldn't then set it to 0.

Since these are becoming more front options, should they be described yes, but not sure the readme is the best place bc it is becoming overwhelming in documentation

lettergram (changes requested):
I believe you should update the minor version number as this change actually changes the API.

I also think you need to add a comment explaining the options.

dataprofiler/tests/data_readers/test_csv_data.py
JGSweets: Good catch

dataprofiler/data_readers/csv_data.py
JGSweets: Sure, I find the visual values more direct but that's personal preference. Sometimes I feel blobs of text are less interpretable

lettergram: I think it just needs to be clear what the "header" number is. IMO saying `> -1` is fine, but it assumes they know.

dataprofiler/data_readers/csv_data.py
JGSweets: done

lettergram (approved):

AnhTruong (approved):

Issue #65, opened 2021-03-15
base_column_profilers.py has a variable, property, which shadows a built-in


JGSweets:
**General Information:**
 - Python version: 3.8.7
 - Library version: 0.3.4

**Describe the bug:**
function `_filter_properties_w_options` uses a variable called `property` which shadows the built-in python `property`

JGSweets:
also in method: `_perform_property_calcs`

JGSweets:
being resolved in https://github.com/capitalone/DataProfiler/pull/67

PR #67, opened 2021-03-16
Merge (Int, Float, and Text) Profiles with Different Options

gme5078:
Now a warning is thrown if the profiles have different options.

### DIFF
--- a/dataprofiler/profilers/base_column_profilers.py
+++ b/dataprofiler/profilers/base_column_profilers.py
@@ -124,11 +124,11 @@ class BaseColumnProfiler(with_metaclass(abc.ABCMeta, object)):
         :type subset_properties: dict
         :return: None
         """
-        for property in calculations:
-            calculations[property](self,
-                                   df_series,
-                                   prev_dependent_properties,
-                                   subset_properties)
+        for prop in calculations:
+            calculations[prop](self,
+                               df_series,
+                               prev_dependent_properties,
+                               subset_properties)
 
     @staticmethod
     def _merge_calculations(merged_profile_calcs, profile1_calcs, profile2_calcs):
--- a/dataprofiler/profilers/base_column_profilers.py
+++ b/dataprofiler/profilers/base_column_profilers.py
@@ -245,7 +245,9 @@ class BaseColumnPrimitiveTypeProfiler(with_metaclass(abc.ABCMeta,
         :type name: String
         """
         BaseColumnProfiler.__init__(self, name)
-        self.match_count = 0
+        # Number of values the match the column type. eg. how many floats match
+        # in the float column
+        self.match_count = 0 
 
     def _update_column_base_properties(self, profile):
         """
--- a/dataprofiler/profilers/base_column_profilers.py
+++ b/dataprofiler/profilers/base_column_profilers.py
@@ -103,9 +103,9 @@ class BaseColumnProfiler(with_metaclass(abc.ABCMeta, object)):
         :param options: Contains all the options.
         :type options: BaseColumnOptions
         """
-        for property in list(calculations):
-            if options and not options.is_prop_enabled(property):
-                del calculations[property]
+        for prop in list(calculations):
+            if options and not options.is_prop_enabled(prop):
+                del calculations[prop]
 
     def _perform_property_calcs(self, calculations, df_series,
                                 prev_dependent_properties, subset_properties):
--- a/dataprofiler/profilers/base_column_profilers.py
+++ b/dataprofiler/profilers/base_column_profilers.py
@@ -245,7 +245,7 @@ class BaseColumnPrimitiveTypeProfiler(with_metaclass(abc.ABCMeta,
         :type name: String
         """
         BaseColumnProfiler.__init__(self, name)
-        # Number of values the match the column type. eg. how many floats match
+        # Number of values that match the column type. eg. how many floats match
         # in the float column
         self.match_count = 0 
 
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -23,7 +23,8 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :type options: FloatOptions
         """
         if options and not isinstance(options, FloatOptions):
-            raise ValueError("options must be of type FloatOptions.")
+            raise ValueError("FloatColumn parameter 'options' must be of type"
+                             " FloatOptions.")
         NumericStatsMixin.__init__(self, options)
         BaseColumnPrimitiveTypeProfiler.__init__(self, name)
         self.precision = 0
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -22,9 +22,8 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :param options: Options for the float column
         :type options: FloatOptions
         """
-        if options:
-            if not isinstance(options, FloatOptions):
-                raise ValueError("options must be of type FloatOptions.")
+        if options and not isinstance(options, FloatOptions):
+            raise ValueError("options must be of type FloatOptions.")
         NumericStatsMixin.__init__(self, options)
         BaseColumnPrimitiveTypeProfiler.__init__(self, name)
         self.precision = 0
@@ -50,6 +49,14 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         merged_profile = FloatColumn(None)
         BaseColumnPrimitiveTypeProfiler._add_helper(merged_profile, self, other)
         NumericStatsMixin._add_helper(merged_profile, self, other)
+        
+        self._merge_calculations(merged_profile.__calculations,
+                                 self.__calculations,
+                                 other.__calculations)
+
+        if "precision" in merged_profile.__calculations:
+            merged_profile.precision = max(self.precision, other.precision)
+            
         return merged_profile
 
     @property
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -55,7 +55,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                                  other.__calculations)
 
         if "precision" in merged_profile.__calculations:
-            merged_profile.precision = max(self.precision, other.precision)
+            merged_profile.precision = min(self.precision, other.precision)
             
         return merged_profile
 
@@ -98,13 +98,14 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
     def _get_float_precision(cls, df_series):
         """
         Determines the precision of the numeric value
+        
         :param df_series: a given column
         :type df_series: pandas.core.series.Series
         :return: string representing its precision print format
         :rtype: int
         """
         integer_decimal_loc = -1
-        float_precision = 0
+        float_precision = None
         for value in df_series:
             decimal_loc = value.rfind('.')
 
@@ -117,7 +118,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             value_len = len(value)
             value_precision = value_len - decimal_loc - 1
 
-            if float_precision < value_precision:
+            if float_precision is None or float_precision > value_precision:
                 float_precision = value_precision
         return float_precision
 
@@ -157,9 +158,11 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :type df_series: pandas.DataFrame
         :return: None
         """
-        self.precision = max(
-            self.precision, self._get_float_precision(df_series)
-        )
+        subset_precision = self._get_float_precision(df_series)
+        if self.precision and subset_precision:
+            self.precision = min(self.precision, subset_precision)
+        elif subset_precision:
+            self.precision = subset_precision
 
     def _update_helper(self, df_series_clean, profile):
         """
--- a/dataprofiler/profilers/text_column_profile.py
+++ b/dataprofiler/profilers/text_column_profile.py
@@ -51,7 +51,7 @@ class TextColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                                  other.__calculations)
         if "vocab" in merged_profile.__calculations:
             merged_profile.vocab = self.vocab.copy()
-            merged_profile._update_vocab([other.vocab])
+            merged_profile._update_vocab(other.vocab)
         return merged_profile
 
     @property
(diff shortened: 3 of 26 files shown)


gme5078:
Changes I've Made:
Lowest common denominator of options is selected. 
Warning thrown if options differ.
Throw error if options are incorrect.
Removed options from being saved.
Removed use of "property" as it is a python key word.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -22,9 +22,8 @@ def __init__(self, name, options=None):
         :param options: Options for the float column
         :type options: FloatOptions
         """
-        if options:
-            if not isinstance(options, FloatOptions):
-                raise ValueError("options must be of type FloatOptions.")
+        if options and not isinstance(options, FloatOptions):
```
JGSweets, line 25: nice

dataprofiler/profilers/float_column_profile.py
```diff
@@ -50,6 +49,14 @@ def __add__(self, other):
         merged_profile = FloatColumn(None)
         BaseColumnPrimitiveTypeProfiler._add_helper(merged_profile, self, other)
         NumericStatsMixin._add_helper(merged_profile, self, other)
+        
+        self._merge_calculations(merged_profile.__calculations,
+                                 self.__calculations,
+                                 other.__calculations)
+
+        if "precision" in merged_profile.__calculations:
+            merged_profile.precision = max(self.precision, other.precision)
```
JGSweets, line 58: Do we want consistency with merging or should in this case we always assume that max is what we want for precision? Actually, Now that I think about it, max likely is not correct. As usually precision is set to the lowest precision visible. We should have to correct this  in `_update_precision`

Now since we are using this functionality in two places, should we actually have a _merge_precision function? and then both calls use it?

dataprofiler/profilers/text_column_profile.py
```diff
@@ -45,10 +44,14 @@ def __add__(self, other):
...
-        merged_profile.vocab = self.vocab.copy()
-        merged_profile._update_vocab([other.vocab])
         NumericStatsMixin._add_helper(merged_profile, self, other)
         BaseColumnPrimitiveTypeProfiler._add_helper(merged_profile, self, other)
+        self._merge_calculations(merged_profile.__calculations,
+                                 self.__calculations,
+                                 other.__calculations)
+        if "vocab" in merged_profile.__calculations:
+            merged_profile.vocab = self.vocab.copy()
+            merged_profile._update_vocab([other.vocab])
```
JGSweets, line 54: is this right that it should be a list of lists? or can it just be: `merged_profile._update_vocab(other.vocab)`

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -711,3 +711,46 @@ def test_profile_merge_no_bin_overlap(self):
                                     'Profiles have no overlapping bin methods '
                                     'and therefore cannot be added together.'):
             profiler1 + profiler2
+
+    def test_profile_merge_with_different_options(self):
+        # Creating first profiler with default options
+        options = FloatOptions()
+        options.max.is_enabled = False
+        options.min.is_enabled = False
```
JGSweets, line 719: this doesn't look like default

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -711,3 +711,46 @@ def test_profile_merge_no_bin_overlap(self):
                                     'Profiles have no overlapping bin methods '
                                     'and therefore cannot be added together.'):
             profiler1 + profiler2
+
+    def test_profile_merge_with_different_options(self):
+        # Creating first profiler with default options
+        options = FloatOptions()
```
JGSweets, line 717: what happens if we have multiple options? do we get two warnings? should we test for that?

JGSweets: will order cause test to fail?

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -299,3 +299,48 @@ def test_merge_timing(self):
...
             self.assertDictEqual(expected_times, profiler3.profile['times'])
 
+    def test_profile_merge_with_different_options(self):
+        # Creating first profiler with default options
+        options = TextOptions()
+        options.max.is_enabled = False
+        options.min.is_enabled = False
+
+        df = pd.Series(
+            ["pancake", "banana", "lighthouse", "aa", "b", "4", "3", "2", "dfd", "2"]
```
JGSweets, line 309: banana pancake sounds more appetizing than pancake banana

JGSweets (changes requested):
added some concerns

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -711,3 +711,46 @@ def test_profile_merge_no_bin_overlap(self):
...
+        # Asserting warning when adding 2 profilers with different options
+        with self.assertWarnsRegex(RuntimeWarning,
+                                   "precision is disabled because it is not "
+                                   "enabled in both profiles."):
+            profiler3 = profiler1 + profiler2
+
+        # Assert that these features are still merged
+        self.assertIsNotNone(profiler3.histogram_selection)
+        self.assertIsNotNone(profiler3.variance)
+        self.assertIsNotNone(profiler3.sum)
```
AnhTruong, line 746: can we actually test the exact values instead of NotNone?

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -299,3 +299,48 @@ def test_merge_timing(self):
...
+        # Asserting warning when adding 2 profilers with different options
+        with self.assertWarnsRegex(RuntimeWarning,
+                                   "vocab is disabled because it is not "
+                                   "enabled in both profiles."):
+            profiler3 = profiler1 + profiler2
+
+        # Assert that these features are still merged
+        self.assertIsNotNone(profiler3.histogram_selection)
+        self.assertIsNotNone(profiler3.variance)
+        self.assertIsNotNone(profiler3.sum)
```
AnhTruong, line 336: same comment as above

AnhTruong (commented):
two small comments

dataprofiler/profilers/base_column_profilers.py
```diff
@@ -103,9 +103,9 @@ def _filter_properties_w_options(calculations, options):
         :param options: Contains all the options.
         :type options: BaseColumnOptions
         """
-        for property in list(calculations):
-            if options and not options.is_prop_enabled(property):
-                del calculations[property]
+        for prop in list(calculations):
+            if options and not options.is_prop_enabled(prop):
+                del calculations[prop]
```
lettergram, line 108: Addresses issue #65

dataprofiler/tests/profilers/test_float_column_profile.py
lettergram: ^ seconded, shouldn't it be "False" or "True" `self.assertIsNotNone` seems to leave a lot to be desired.

dataprofiler/tests/profilers/test_text_column_profile.py
lettergram: banana lighthouse #yum

dataprofiler/profilers/text_column_profile.py
gme5078: This was the precedence set, so I'm not sure, I can try without and see what happens

dataprofiler/tests/profilers/test_float_column_profile.py
gme5078: correct 👍

dataprofiler/tests/profilers/test_float_column_profile.py
gme5078: It will print both warnings, but from what I can tell, it is not possible to test for more than one warning

gme5078: what do you mean by order?

dataprofiler/profilers/text_column_profile.py
gme5078: I took it out and it works. Idk why we had that

dataprofiler/profilers/float_column_profile.py
gme5078: Updated this to be min. Actually updated precision in general to be more appropriate.
I don't see a need to create a one line function tbh

dataprofiler/tests/profilers/test_float_column_profile.py
gme5078: Fixed to handle multiple warnings

dataprofiler/profilers/float_column_profile.py
```diff
@@ -110,7 +118,7 @@ def _get_float_precision(cls, df_series):
             value_len = len(value)
             value_precision = value_len - decimal_loc - 1
 
-            if float_precision < value_precision:
+            if float_precision is None or float_precision > value_precision:
```
lettergram, line 121: Nice bug fix.

lettergram (approved):

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -54,25 +55,20 @@ def test_profiled_precision(self):
...
-
+        float_profiler = FloatColumn("Name")
         float_profiler.update(df_3)
         self.assertEqual(3, float_profiler.precision)
 
         float_profiler.update(df_2)
-        self.assertEqual(3, float_profiler.precision)
+        self.assertEqual(2, float_profiler.precision)
 
-        float_profiler.update(df_mix)
```
AnhTruong, line 70: so we don't test df_mix ?

AnhTruong: oh, I see below. Should the precision be 5 as we have this number 2.13143. If the precision is 1, does that mean the number will be round up to 2.1?

AnhTruong (commented):
one small comment left

dataprofiler/profilers/float_column_profile.py
```diff
@@ -150,9 +158,11 @@ def _update_precision(self, df_series, prev_dependent_properties,
         :type df_series: pandas.DataFrame
         :return: None
         """
-        self.precision = max(
-            self.precision, self._get_float_precision(df_series)
-        )
+        subset_precision = self._get_float_precision(df_series)
+        if self.precision and subset_precision:
+            self.precision = min(self.precision, subset_precision)
```
ChrisWallace2020, line 163: Is the switch from max to min intentional or an error here?

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -299,3 +299,48 @@ def test_merge_timing(self):
...
 
+    def test_profile_merge_with_different_options(self):
+        # Creating first profiler with default options
+        options = TextOptions()
+        options.max.is_enabled = False
+        options.min.is_enabled = False
+
+        df = pd.Series(
+            ["pancake", "banana", "lighthouse", "aa", "b", "4", "3", "2", "dfd", "2"]
+        ).apply(str)
```
ChrisWallace2020, line 310: why do you apply str to a series full of strings?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -711,3 +707,53 @@ def test_profile_merge_no_bin_overlap(self):
...
+        # Creating first profiler with default options
+        options = FloatOptions()
+        options.max.is_enabled = False
+        options.min.is_enabled = False
+
+        data = [2, 4, 6, 8]
+        df = pd.Series(data).apply(str)
+        profiler1 = FloatColumn("Float", options=options)
+        profiler1.update(df)
+        profiler1.match_count = 0
```
ChrisWallace2020, line 721: Its 0 here but its 4 ints?

dataprofiler/profilers/float_column_profile.py
gme5078: Intentional. We want the lowest common denominator for precision.

dataprofiler/tests/profilers/test_float_column_profile.py
gme5078: We are taking the lowest common denomination for precision. So the number is 1 since 3. exists in the dataset

dataprofiler/tests/profilers/test_text_column_profile.py
gme5078: Precedence. but really, I guess no reason at all. Took it out.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -711,3 +707,53 @@ def test_profile_merge_no_bin_overlap(self):
...
+
+            message = str(w.pop().message)
+            self.assertEqual("precision is disabled because it "
+                             "is not enabled in both profiles.",
+                             message)
+
+            message = str(w.pop().message)
+            self.assertEqual("max is disabled because it is not enabled in both"
+                             " profiles.",
+                             message)
```
JGSweets, line 743: i am concerned that requires proper order

lettergram (approved):

AnhTruong (approved):

dataprofiler/profilers/float_column_profile.py
```diff
@@ -50,6 +50,14 @@ def __add__(self, other):
         merged_profile = FloatColumn(None)
         BaseColumnPrimitiveTypeProfiler._add_helper(merged_profile, self, other)
         NumericStatsMixin._add_helper(merged_profile, self, other)
+        
+        self._merge_calculations(merged_profile.__calculations,
+                                 self.__calculations,
+                                 other.__calculations)
+
+        if "precision" in merged_profile.__calculations:
+            merged_profile.precision = min(self.precision, other.precision)
```
ChrisWallace2020, line 59: Is there a reason this isn't handled in self._merge_calculations?

lettergram (changes requested):
@JGSweets comment regarding the warning messages appears correct - there's a risk that those warnings will not appear in the correct order using `pop()` across the board might not be the best approach.

dataprofiler/profilers/float_column_profile.py
ChrisWallace2020, line 59: Is there a reason this isn't done in _merge_calculations?

ChrisWallace2020: Nvm, explained over standup mtg (_merge_calculations is generic and top level and just for instantiating dict, not populating it)

ChrisWallace2020 (dismissed):

lettergram (dismissed):

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -711,3 +707,53 @@ def test_profile_merge_no_bin_overlap(self):
...
+            list_of_warning_messages = []
+            for warning in w:
+                list_of_warning_messages.append(str(warning.message))
+            
+            warning1 = "precision is disabled because it is not enabled in both" \
+                       " profiles."
+            warning2 = "max is disabled because it is not enabled in both " \
+                       "profiles."
+            self.assertTrue(warning1 in list_of_warning_messages)
+            self.assertTrue(warning2 in list_of_warning_messages)
```
JGSweets, line 743: assertIn as opposed to assertTrue will provide more insight into failures if they occur

JGSweets (approved):

Issue #69, opened 2021-03-16
Profiling is slow


lettergram:
**General Information:**
 - OS: Arch Linux
 - Python version: 3.7
 - Library version: 0.3.4

**Describe the bug:**

Profiling `diamonds.csv` (2.5Mb) or really any dataset is much slower than expected. Even with the data labeler disabled.

**To Reproduce:**

The following code takes:

* Base run: 38 seconds and ~785Mb

* No labeler: 24.48 seconds to execute and ~67Mb
* No labeler, no histogram: 23.47 seconds to execute and ~67Mb.
* No labeler, no datetime: 23.35 seconds and ~67Mb

```python
import sys
import json
import time
import dataprofiler as dp

filename = sys.argv[1]

def profile_test(filename):
    data = dp.Data(filename)

    profile_options = dp.ProfilerOptions()
    profile_options.structured_options.data_labeler.is_enabled = False                       
    profile_options.set({"histogram_and_quantiles.is_enabled": False})                       
    profile_options.set({"datetime.is_enabled": False})  

    profile = dp.Profiler(data, profiler_options=profile_options)

    human_readable_report = profile.report(report_options={"output_format":"pretty"})

    print(json.dumps(human_readable_report, indent=4))

start_time = time.time()
profile_test(filename)
end_time = time.time()

print("Profile runtime for "+filename, end_time-start_time, 'seconds')
```

In the output, there's no reason in the "timing" for this to be occuring:
```
{
    "global_stats": {
        "samples_used": 10788,
        "column_count": 10,
        "unique_row_ratio": 0.9973,
        "row_has_null_ratio": 0.0,
        "duplicate_row_count": 146,
        "file_type": "csv",
        "encoding": "utf-8",
        "data_classification": null,
        "covariance": null
    },
    "data_stats": {
        "carat": {
            "column_name": "carat",
            "data_type": "float",
            "categorical": true,
            "order": "random",
            "samples": "['2.52', '1', '0.55', '1.75', '2.01']",
            "statistics": {
                "min": 0.2,
                "max": 4.01,
                "mean": 0.7976,
                "median": null,
                "variance": 0.2244,
                "stddev": 0.4738,
                "histogram": {
                    "bin_counts": "[  70,  205,  606, 1224,  443, ... , 0, 0, 0, 0, 1]",
                    "bin_edges": "[0.2       , 0.23663462, ... , 3.97336538, 4.01      ]"
                },
                "quantiles": {
                    "0": 0.3832,
[...]

PR #70, opened 2021-03-16
Update unique row detection method, reduce processing time by >91%

lettergram:
Addresses issue #69 

**NOTE**: This does increase memory requirements by the number of `4 bytes * rows` when reviewing the row statistics. In terms of what that means, 10 million rows is only 40Mb of memory.

However, the processing time is <9% previous to this update.

Previously, `diamonds.csv` dataset was taking ~24 seconds. 22 seconds were spent on detecting duplicate / null rows. The updated function reduces the duplicate / null row detection to 0.11 seconds.


AnhTruong (dismissed):
the code lgtm. The result is given on a reasonable-size file.

JGSweets (approved):
good fix

gme5078 (approved):

Issue #57, opened 2021-03-15
Delimiter detection uses digits in a number


lettergram:
**General Information:**
 - OS: Arch Linux
 - Python version: 3.9
 - Library version: 0.3.4

**Describe the bug:**

File:
```
-123
234
345
534
231
```

Delimiter detected: `3`

Delimiter should probably not detect inside numbers

PR #75, opened 2021-03-18
Improve delimiter detection

lettergram:
Added multiple tests that don't work with the current delimiter detection. This method is messy, but improved. 

Addresses issue #57


dataprofiler/tests/data/csv/daily-activity-sheet-@.csv
```diff
@@ -0,0 +1,32 @@
...
+02:00pm - 03:00pm@Potty + Diapers@Potty@Primary parent
+03:00pm - 03:30pm@Snack Time@Food@
+03:30pm - 04:00pm@Craft or Game@"Playdough, Board game, etc."@
+04:00pm - 05:00pm@Free Play@Inside@
+05:00pm - 05:30pm@Dinner@Food@
+05:30pm - 05:45pm@Potty + Diapers@Potty@Both parents split
+05:45pm - 07:00pm@Bath(s) + Book & Bed @Bathtub + Books + Bed@
+07:00pm - 07:30pm@Books + Bed @Books + Bed@
+@@@
+Master@Stationary Activiy@Structured Play@
```
AnhTruong, line 22: imo, this looks like the second part of the csv file. Should we consider this example?

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,152 @@ def _check_and_return_options(self, options):
...
+        ordered_vocab = []
+        sorted_keys = sorted(vocab, key=vocab.get)
+        sorted_keys.reverse()
+        for c in sorted_keys:
+            ordered_vocab.append(c)
+
+        # Evaluate vocab, reviewing rows and columns
+        delimiter = None
+        validated_proposed_delimiters = {}
+        for proposed_delim in ordered_vocab[0:12]:
```
ChrisWallace2020, line 144: Should we look into maybe making the upper limit here relative to the length of data_as_a_str or vocab? Like what if there are very few, but very long cells, and many characters appear more than twice, then they would appear more than the delimiter? Just a thought/discussion, a pretty edge one at that.

dataprofiler/tests/data/csv/daily-activity-sheet-@.csv
AnhTruong: and, when we take the first 5 lines for the detection, this won't affect the results. Did you increase 5 to 50 to see the difference?

dataprofiler/tests/data/csv/num-negative-title-large.csv
```diff
@@ -0,0 +1,5 @@
+-12345
+11234
+34555
+12534
+02311
```
AnhTruong, line 5: why this file is different from the other file with smaller numbers?

dataprofiler/tests/data/csv/daily-activity-sheet-tab.csv
```diff
@@ -0,0 +1,31 @@
+Time	Activity 	Supplies	Notes
```
AnhTruong, line 1: what if we have the line `COVID19 Activity Sheet` as in the other file?

AnhTruong (commented):
some first comments

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,152 @@ def _check_and_return_options(self, options):
...
+            proposed_dataset = []
+            for row in data_as_str.split('\n')[:-1]:
```
ChrisWallace2020, line 146: What is the reason we ignore the last row?

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,152 @@ def _check_and_return_options(self, options):
...
+                        if col_id == len(proposed_row)-1:
+                            cell_end_safe_flag = True
+                        if col_id == 0:
+                            cell_start_safe_flag = True
+                    
+                    prior_cell_type = proposed_cell['type']
+
+            edge_safe_flag = False
+            if cell_end_safe_flag or cell_start_safe_flag:
+                edge_safe_flag = True
```
ChrisWallace2020, line 223: This would look cleaner and easier to read if it were just `edge_safe_flag = cell_end_safe_flag or cell_start_safe_flag`

dataprofiler/data_readers/csv_data.py
lettergram: Say we have 1000 columns, with 1000 words per column. 

The vocab is stored in a dictionary and a count is kept per character. So it's bound by the number of unique characters. Then I truncate to the top 12. How was 12 determined? The idea was 10 common digits + two special characters for eval (`0-9`, `' '`, `,`). 

Worked on all the files 🤷‍♂️

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,152 @@ def _check_and_return_options(self, options):
...
+                    delimiter = proposed_delim
+                if prior_col_count \
+                   and prior_col_count > validated_proposed_delimiters[delimiter]:
+                    delimiter = proposed_delim
+
+
+        # Use preferred delimiters with highest count, if possible
+        largest_delim_count = 0
+        for proposed_delim in validated_proposed_delimiters.keys():
+            w_delim_count = validated_proposed_delimiters[delimiter]
```
ChrisWallace2020, line 240: What does w_delim_count mean? Perhaps name the variable a bit clearer

dataprofiler/data_readers/csv_data.py
ChrisWallace2020: Oh I suppose thats a good point, alright then.

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,150 @@ def _check_and_return_options(self, options):
...
+
+            # Add a delimiter candidate safe to do so
+            if not incorrect_delimiter_flag and cell_type_safe_flag and edge_safe_flag:
+                validated_proposed_delimiters[proposed_delim] = prior_col_count
+                
+                # Find one with most columns 
+                if not delimiter:
+                    delimiter = proposed_delim
+                if prior_col_count \
+                   and prior_col_count > validated_proposed_delimiters[delimiter]:
```
JGSweets, line 231: should be tabbed over 1

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,150 @@ def _check_and_return_options(self, options):
...
+
+        # Detect vocabulary (and count)
+        vocab = {}
+        for i in range(len(data_as_str)):
+            c = data_as_str[i]
+            if c not in vocab:
+                vocab[c] = 0
+            vocab[c] += 1            
+        if '\n' in vocab: vocab.pop('\n')        
+        for char in omitted:
```
JGSweets, line 130: this is a little confusing format-wise

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,150 @@ def _check_and_return_options(self, options):
...
-            return None
-        return dialect.delimiter
+
+        # Detect vocabulary (and count)
+        vocab = {}
+        for i in range(len(data_as_str)):
+            c = data_as_str[i]
+            if c not in vocab:
+                vocab[c] = 0
+            vocab[c] += 1
```
JGSweets, line 128: python has a builtin counter in collections. Might be better to use or potentially optimized somehow?

dataprofiler/data_readers/csv_data.py
```diff
@@ -103,24 +103,149 @@ def _check_and_return_options(self, options):
...
+                   and prior_col_count > validated_proposed_delimiters[delimiter]:
+                    delimiter = proposed_delim
+
+
+        # Use preferred delimiters with highest count, if possible
+        largest_delim_count = 0
+        for proposed_delim in validated_proposed_delimiters.keys():
+            weighted_delim_count = validated_proposed_delimiters[proposed_delim]
+            if proposed_delim in preferred:
+                weighted_delim_count = 100 * validated_proposed_delimiters[proposed_delim]
```
ChrisWallace2020, line 240: Might be more clear to say `weighted_delim_count *= 100`

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,23 +104,163 @@ def _check_and_return_options(self, options):
...
+            try:
+                dialect = sniffer.sniff(data_as_str)
+                quotechar = dialect.quotechar
+            except csv.Error as exc:
+                quotechar = '"'
+
+        # Evaluate vocab, reviewing rows and columns
+        delimiter = None
+        validated_proposed_delimiters = {}
+        for proposed_delim in ordered_vocab[0:12]:
```
gme5078, line 158: This is likely to cause failure if a csv exists with paragraphs in each cell I think, right?

dataprofiler/data_readers/csv_data.py
gme5078: agree

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,23 +104,163 @@ def _check_and_return_options(self, options):
...
+                   and prior_col_count > validated_proposed_delimiters[delimiter]:
+                    delimiter = proposed_delim
+
+
+        # Use preferred delimiters with highest count, if possible
+        largest_delim_count = 0
+        for proposed_delim in validated_proposed_delimiters.keys():
+            weighted_delim_count = validated_proposed_delimiters[proposed_delim]
+            if proposed_delim in preferred:
+                weighted_delim_count = 100 * validated_proposed_delimiters[proposed_delim]
```
gme5078, line 255: why even multiply by 100 when its all arbitrary? I think we can just look at which preferred delimiters exist in the validated delimiters and take the highest count.

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,23 +104,163 @@ def _check_and_return_options(self, options):
...
+
+            # Add a delimiter candidate safe to do so
+            if not incorrect_delimiter_flag and cell_type_safe_flag and edge_safe_flag:
+                validated_proposed_delimiters[proposed_delim] = prior_col_count
+                
+                # Find one with most columns 
+                if not delimiter:
+                    delimiter = proposed_delim
+                if prior_col_count \
+                   and prior_col_count > validated_proposed_delimiters[delimiter]:
```
gme5078, line 246: Why is the highest column count the basis for which delimiter we use. Couldnt there be a case where that isnt always true?

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,23 +104,163 @@ def _check_and_return_options(self, options):
...
+                    proposed_cell = proposed_row[col_id]
+
+                    col_types[col_id] = proposed_cell['type']
+
+                    # Handle if alpha character are seperator                     
+                    # NOTE: delimiter needs two ajoining cells to flag
+                    if proposed_cell['type'] in ['str', 'none'] \
+                       and prior_cell_type in ['str', 'none']:
+                        if proposed_delim.isalpha():
+                            cell_type_safe_flag = False
```
gme5078, line 216: Can we break from the loop if any of these flags would cause failure?

gme5078 (changes requested):

dataprofiler/data_readers/csv_data.py
AnhTruong: can we make it work for multiple-character delimiters?

dataprofiler/data_readers/csv_data.py
ChrisWallace2020: +1

dataprofiler/data_readers/csv_data.py
lettergram: Doesn't meaningfully impact results

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,26 +106,147 @@ def _check_and_return_options(self, options):
...
+            # If large dataset, select first 25 rows then random sampling
+            proposed_dataset = data_as_str.split('\n')
+            if len(proposed_dataset) > 25:
+                sample_count = min(int(0.1*(len(proposed_dataset)-25)), 1000)
+                proposed_dataset = proposed_dataset[:25] \
+                    + random.choices(proposed_dataset[:25], k=sample_count)
+
+            # Reverse to start at bottom
+            # Fewer columns are okay, as long as earlier in file
+            proposed_dataset.reverse()
```
gme5078, line 182: What is the reason for this?

lettergram: If you look from the bottom of a spreadsheet up, it's clearer what the delimiters are. Typically, the bottom of a file will contain the most accurate rows, then you may have descriptions, notes, etc. at the top of a sheet.

gme5078 (approved):

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,26 +106,145 @@ def _check_and_return_options(self, options):
...
-        return dialect.delimiter
+
+        # Detect vocabulary (and count)
+        vocab = {}
+        for i in range(len(data_as_str)):
+            c = data_as_str[i]
+            if c not in vocab:
+                vocab[c] = 0
+            vocab[c] += 1
+        if '\n' in vocab: vocab.pop('\n')
```
JGSweets, line 137: Not blocking, but previously mentioned this could be using the builtin python counter, might be faster? not sure it matters if we end up limiting char inputs anyways. Ultimately could be cleaner.
```python
from collections import Counter


vocab = Counter(data_as_str)
vocab.pop('\n', None)
``` 
below with other pops could be same.

lettergram: Yeah, had considered. Initially I was going to do a couple other tricks - didn't end up doing it. I'm fine changing it, but I don't think anything will materially change in runtime

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,26 +106,145 @@ def _check_and_return_options(self, options):
...
+                sample_count = min(int(0.1*(len(proposed_dataset)-25)), 1000)
+                proposed_dataset = proposed_dataset[:25] \
+                    + random.choices(proposed_dataset[:25], k=sample_count)
+
+            # Reverse to start at bottom where likely the most columns
+            # Fewer columns are okay, as long as earlier in file
+            for row_idx in range(len(proposed_dataset)-1, -1, -1):
+                row = proposed_dataset[row_idx]
+                proposed_cells = list(csv.reader(
+                    [row], delimiter=proposed_delim, quotechar=quotechar))[0]
```
JGSweets, line 184: i wonder if a better solution to quote detection would be to also loop through quotechars like we do delimiters, except limit quotechars to those specified that already exist in the vocab, i.e. `(``, ', ")`

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,26 +106,145 @@ def _check_and_return_options(self, options):
...
+                            row_start_safe_flag = True
+                            
+                    prior_cell_type = cell_type
+                    
+            edge_safe_flag = row_end_safe_flag or row_start_safe_flag
+            
+            if not incorrect_delimiter_flag and cell_type_safe_flag and edge_safe_flag:
+                validated_proposed_delimiters[proposed_delim] = prior_col_count
+                if prior_col_count \
+                   and prior_col_count > validated_proposed_delimiters[proposed_delim]:
```
JGSweets, line 229: Please tab this over to not be in line with line below

dataprofiler/data_readers/data_utils.py
```diff
@@ -302,3 +303,41 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
+        try:
+            if dateutil.parser.parse(cell, fuzzy=False):
+                cell_type = 'date'
+                
+        except ValueError:
+            pass
+        except OverflowError:
+            pass
+        except TypeError:
+            pass
```
JGSweets, line 330: I think this can be simplified into:
```python
except (ValueError, OverflowError, TypeError):
    pass
```

JGSweets (commented):
I am concerned that this isn't using an equation solve like the regex in csv. I think a solution that solves for the equation would ideally be a better solution but practically would take longer to implement in the proposed time.

I also think, allowing the iteration of quotechars with the delims inside the loop with respect to vocab and allowed quotechars (unless specified by user) would be a potential solution as well that would be quicker to implement.

Overall, the stats ran by @lettergram provided better results (hopefully on general pop dataset as well). Hence, I'm inclined to approve after the other comments have been addressed.

dataprofiler/data_readers/csv_data.py
lettergram: Added some breaks

JGSweets (approved):

dataprofiler/data_readers/csv_data.py
```diff
@@ -104,26 +106,145 @@ def _check_and_return_options(self, options):
...
+                    data_as_str, ordered_vocab[:20])[0]
+            except csv.Error as exc:
+                quotechar = None
+            if not quotechar or len(quotechar) == 0:
+                quotechar = '"'
+                
+        # Evaluate vocab, reviewing rows and columns
+        delimiter = None
+        validated_proposed_delimiters = {}
+        for proposed_delim in ordered_vocab[:20]:
```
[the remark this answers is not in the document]
lettergram: Fixed

AnhTruong (approved):

dataprofiler/tests/data/csv/num-negative-title-large.csv
lettergram: These numbers could have `4` selected as a number, there's also checks related to small fields (no / single character) vs more than one.

dataprofiler/tests/data/csv/daily-activity-sheet-tab.csv
lettergram: That's not a problem, I just left it this way to try various things (tabs in this case)

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -91,6 +91,11 @@ def setUpClass(cls):
             dict(path=os.path.join(test_dir, 'csv/quote-test.txt'),
                  count=10, delimiter=' ', has_header=[0, None],
                  num_columns=3, encoding='utf-8'),
+            
+            #dict(path=os.path.join(test_dir, 'csv/quote-test-singlequote.txt'),
+            #     count=9, delimiter=' ', has_header=[0, None],
+            #     num_columns=3, encoding='utf-8'),
```
AnhTruong, line 97: can you remove this?

dataprofiler/data_readers/data_utils.py
```diff
@@ -302,3 +304,66 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
+    delimiter_regex = re.escape(str(delimiter))    
+    quotechar_regex = "(?=" 
+    quotechar_regex +=  "(?:"
+    quotechar_regex +=    "[^"+re.escape(quotechar)+"]*"
+    quotechar_regex +=    re.escape(quotechar)
+    quotechar_regex +=    "[^"+re.escape(quotechar)+"]*"
+    quotechar_regex +=    re.escape(quotechar)
+    quotechar_regex +=   ")*"
+    quotechar_regex +=   "[^"+re.escape(quotechar)+"]*"
+    quotechar_regex += "$)"
```
JGSweets, line 367: any reason why not create var for `re.escape(quotechar)` instead of calling `re.escape` multiple times?

dataprofiler/tests/data_readers/test_csv_data.py
lettergram: Left to denote an issue, will need to remove next PR. Currently, there's an issue *reloading* (not the initial load) for a file with a single quote.

JGSweets (approved):
There's an understanding that this still contains an error with the quote test file that will be fixed in the next PR. 

There will not be a release made until that is fixed.

AnhTruong (approved):

Issue #63, opened 2021-03-15
Suppress / Remove / Fix warnings from tensorflow


lettergram:
Currently, I am seeing:

> WARNING:tensorflow:5 out of the last 5 calls to <function recreate_function.<locals>.restored_function_body at 0x7f27803a6790> triggered tf.function retracing. Tracing is expensive and the excessive number of tracings could be due to (1) creating @tf.function repeatedly in a loop, (2) passing tensors with different shapes, (3) passing Python objects instead of tensors. For (1), please define your @tf.function outside of the loop. For (2), @tf.function has experimental_relax_shapes=True option that relaxes argument shapes that can avoid unnecessary retracing. For (3), please refer to https://www.tensorflow.org/tutorials/customization/performance#python_or_tensor_args and https://www.tensorflow.org/api_docs/python/tf/function for  more details.

Every time a column is profiled / labeled. It would be nice if the warnings were suppressed and / or the issue was fixed.

PR #78, opened 2021-03-19
[WIP] Fix repeated loading data labeler for each column

AnhTruong:
- will reduce memory usage for new model graph
- reduce running time for populating new graph
- fix the issue https://github.com/capitalone/DataProfiler/issues/63

### DIFF
--- a/dataprofiler/labelers/character_level_cnn_model.py
+++ b/dataprofiler/labelers/character_level_cnn_model.py
@@ -272,6 +272,7 @@ class CharacterLevelCnnModel(BaseTrainableModel,
 
         # load tf model
         # Use TFA to add f1 score to output
+        tf.keras.backend.clear_session()
         import tensorflow_addons as tfa
         custom_objects = {
             "F1Score": tfa.metrics.F1Score(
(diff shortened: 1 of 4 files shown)


dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -272,6 +272,7 @@ def load_from_disk(cls, dirpath):
 
         # load tf model
         # Use TFA to add f1 score to output
+        tf.keras.backend.clear_session()
```
lettergram, line 275: Careful doing that, I believe it releases the memory

dataprofiler/profilers/profile_builder.py
```diff
@@ -535,3 +536,14 @@ def _update_profile_from_chunk(df, profile=None, sample_size=None,
...
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.close()
+
+    def __enter__(self):
+        return self
+
+    def close(self):
+        config.data_labeler_loaded = False
+        config.existing_data_labeler = None
```
JGSweets, line 548: I don't believe this should be associated with the profiles as the existence of the labeler should be independent of profiles.

PR #79, opened 2021-03-19
Add tests for Numerical, Int, Float, and Text Options

sagars729:
- Tests added for numerical options
- Tests added for int, float, and text options (with override + super call to enhance clarity)
- Updated variable paths in Profiler Options


dataprofiler/tests/profilers/test_base_column_options.py
```diff
@@ -1,10 +1,13 @@
...
 
-
 class TestBaseColumnOptions(TestBooleanOption):
     
-    option_class = BaseColumnOptions
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.option_class = BaseColumnOptions
+        cls.keys = []
```
sagars729, line 10: I think this is a safer method of setting the option_class variable. If you just set the option_class = BaseColumnOptions without the setUpClass method, it would mean that if anyone extended the TestBaseColumnOptions class (i.e. the TestNumericalOptions class) but forgot to specify the option_class, it would default to BaseColumnOptions instead of None. The desired behavior is None as the Abstract Class has a check for that.

dataprofiler/tests/profilers/test_int_options.py
```diff
@@ -0,0 +1,34 @@
...
+        super().test_set(*mocks)
+    
+    def test_validate_helper(self, *mocks):
+        super().test_validate_helper(*mocks)
+    
+    def test_validate(self, *mocks):
+        super().test_validate(*mocks)
+    
+    def test_is_numeric_stats_enabled(self, *mocks):
+        super().test_is_numeric_stats_enabled(*mocks)
```
sagars729, line 34: Similar to approach in BaseColumnOptions, taking the approach of overriding the super to call the super to explicitly state which tests are being run (for readability and future modification).

dataprofiler/tests/profilers/test_base_column_options.py
```diff
@@ -1,10 +1,12 @@
...
 from dataprofiler.tests.profilers.test_boolean_option import TestBooleanOption
 
-
 class TestBaseColumnOptions(TestBooleanOption):
     
-    option_class = BaseColumnOptions
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.option_class = BaseColumnOptions
```
sagars729, line 9: I think this is a safer method of setting the option_class variable. If you just set the option_class = BaseColumnOptions without the setUpClass method, it would mean that if anyone extended the TestBaseColumnOptions class (i.e. the TestNumericalOptions class) but forgot to specify the option_class, it would default to BaseColumnOptions instead of None. The desired behavior is None as the Abstract Class has a check for that.

JGSweets: The reason we changed it was because internally tests were failing when done in parallel.

This created a competing setup for who is assigning the class value and resulted in it being set to None and the abstract check for the option_class value failed.

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -3,10 +3,11 @@
 from dataprofiler.profilers.profiler_options import BaseOption
 from dataprofiler.tests.profilers.abstract_test_options import AbstractTestOptions
 
-
 class TestBaseOption(AbstractTestOptions, unittest.TestCase):
     
-    option_class = BaseOption
+    @classmethod
+    def setUpClass(cls):
+        cls.option_class = BaseOption
```
JGSweets, line 10: same as above

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -1,11 +1,13 @@
 from dataprofiler.profilers.profiler_options import BooleanOption
 from dataprofiler.tests.profilers.test_base_option import TestBaseOption
 
-
 class TestBooleanOption(TestBaseOption):
+    
+    @classmethod
+    def setUpClass(cls):
+        cls.option_class = BooleanOption    
+        cls.keys = []
```
JGSweets, line 9: as above

dataprofiler/tests/profilers/test_float_options.py
```diff
@@ -0,0 +1,28 @@
+from dataprofiler.profilers.profiler_options import FloatOptions
+from dataprofiler.tests.profilers.test_numerical_options import TestNumericalOptions
+
+class TestFloatOptions(TestNumericalOptions):
+    
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.keys += ["precision"]
+        cls.option_class = FloatOptions
```
JGSweets, line 10: In a case like this, I would expect it to be the top level, so having set outside of the setup class seems reasonable to me

JGSweets, line 10: as above

dataprofiler/tests/profilers/test_int_options.py
```diff
@@ -0,0 +1,27 @@
+from dataprofiler.profilers.profiler_options import IntOptions
+from dataprofiler.tests.profilers.test_numerical_options import TestNumericalOptions
+
+class TestIntOptions(TestNumericalOptions):
+    
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.option_class = IntOptions
```
JGSweets, line 9: as above

dataprofiler/tests/profilers/test_base_column_options.py
sagars729: Does this mean that any class variable would have to be defined outside of the setUpClass method (i.e. keys)?

dataprofiler/tests/profilers/test_numerical_options.py
```diff
@@ -0,0 +1,105 @@
+from dataprofiler.profilers.profiler_options import BooleanOption, NumericalOptions
+from dataprofiler.tests.profilers.test_base_column_options import TestBaseColumnOptions
+
+class TestNumericalOptions(TestBaseColumnOptions):
+    
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.keys += ["min", "max", "sum", "variance", "histogram_and_quantiles"]
+        cls.option_class = NumericalOptions
```
JGSweets, line 10: we will have to validate these pass when conducted in parallel.
try running: `python3 -m pytest dataprofiler/tests/ --cov=dataprofiler --cov-fail-under=80 --cov-report=xml:dist/coverage.xml --forked`

JGSweets (changes requested):
We need to address the parallelization issue of the tests.

dataprofiler/tests/profilers/test_numerical_options.py
sagars729: The keys are causing issues when running in parallel - I will try to resolve and update the PR

sagars729:
> We need to address the parallelization issue of the tests.

@JGSweets By moving the keys outside, I was able to fix the parallelization issue of the tests.

Would an alternative be to define a setUpClass method in the Abstract Class?
```python
def setUpClass(cls, *mocks):
    cls.option_class = None
```
It seems like this would still fail because setUpClass isn't being called at all.

dataprofiler/tests/profilers/test_numerical_options.py
```diff
@@ -0,0 +1,102 @@
...
+        optpth = self.get_options_path()
+    
+        # Set BooleanOptions' is_enabled to a non-boolean value
+        for key in self.keys:
+            skey = '{}.is_enabled'.format(key)
+            expected_error = "{}.{}.is_enabled must be a Boolean.".format(optpth, key)
+            
+            options.set({skey: "Hello World"})
+            self.assertEqual([expected_error], options._validate_helper())
+            options.set({skey: True})
```
lettergram, line 48: Why are you setting skey to true?

lettergram (commented):
Curious

dataprofiler/tests/profilers/test_numerical_options.py
sagars729: Only setting one key.is_enabled to "Hello World" at a time (rather than setting all of them).

dataprofiler/tests/profilers/abstract_test_options.py
```diff
@@ -1,6 +1,5 @@
 from dataprofiler.profilers.profiler_options import BaseOption
 
-
 class AbstractTestOptions():
```
ChrisWallace2020, line 3: There should be 2 lines between imports and code

dataprofiler/tests/profilers/test_base_column_options.py
```diff
@@ -1,7 +1,6 @@
 from dataprofiler.profilers.profiler_options import BaseColumnOptions
 from dataprofiler.tests.profilers.test_boolean_option import TestBooleanOption
 
-
 class TestBaseColumnOptions(TestBooleanOption):
```
ChrisWallace2020, line 4: same as above

dataprofiler/tests/profilers/test_base_option.py
```diff
@@ -3,11 +3,10 @@
 from dataprofiler.profilers.profiler_options import BaseOption
 from dataprofiler.tests.profilers.abstract_test_options import AbstractTestOptions
 
-
 class TestBaseOption(AbstractTestOptions, unittest.TestCase):
```
ChrisWallace2020, line 6: same as above

dataprofiler/profilers/profiler_options.py
```diff
@@ -355,7 +365,7 @@ def _validate_helper(self, variable_path='FloatOptions'):
         if not isinstance(self.precision, BooleanOption):
```
ChrisWallace2020, line 365: I'm a bit confused, I just looked at another PR from @JGSweets where precision was was an integer denoting sig figs, is this something different?

dataprofiler/tests/profilers/test_boolean_option.py
```diff
@@ -1,11 +1,11 @@
 from dataprofiler.profilers.profiler_options import BooleanOption
 from dataprofiler.tests.profilers.test_base_option import TestBaseOption
 
-
 class TestBooleanOption(TestBaseOption):
```
ChrisWallace2020, line 4: Need 2 blank lines

dataprofiler/tests/profilers/abstract_test_options.py
sagars729: Oh! I did not know that, let me fix all of those.

dataprofiler/profilers/profiler_options.py
sagars729: Yes, I believe this is different: self.precision is a BooleanOption that will indicate if precision is enabled or not enabled.

sagars729:
> There should be 2 lines between imports and code

Fixed in new commit @ChrisWallace2020 - all files have been updated with 2 lines between imports and code

lettergram (approved):

dataprofiler/tests/profilers/test_text_options.py
```diff
@@ -0,0 +1,26 @@
+from dataprofiler.profilers.profiler_options import TextOptions
+from dataprofiler.tests.profilers.test_numerical_options import TestNumericalOptions
+
+
+class TestTextOptions(TestNumericalOptions):
+    
+    option_class = TextOptions
+    keys = ["min", "max", "sum", "variance", "histogram_and_quantiles", "vocab"]
```
JGSweets, line 8: i think you could do `keys = TestNumericalOptions.keys + ["vocab"]`
or do you think that becomes unclear?

dataprofiler/tests/profilers/test_float_options.py
```diff
@@ -0,0 +1,26 @@
+from dataprofiler.profilers.profiler_options import FloatOptions
+from dataprofiler.tests.profilers.test_numerical_options import TestNumericalOptions
+
+
+class TestFloatOptions(TestNumericalOptions):
+    
+    option_class = FloatOptions    
+    keys = ["min", "max", "sum", "variance", "histogram_and_quantiles", "precision"]
```
JGSweets, line 8: i think you could do keys = TestNumericalOptions.keys + ["precision"]
or do you think that becomes unclear?

dataprofiler/tests/profilers/test_text_options.py
sagars729: I think that's a good idea - if NumericalOptions keys change in the future, you would only need to edit one test file

lettergram (approved):

dataprofiler/tests/profilers/test_numerical_options.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # Disable Sum and Enable Variance
+        options.set({"sum.is_enabled": False, "variance.is_enabled": True})
+        expected_error = "{}: The numeric stats must toggle on the sum if the variance is toggled on.".format(optpth)
+        self.assertEqual([expected_error], options._validate_helper())
```
JGSweets, line 54: I'm going to approve as we can fix as tech debt later in a style cleanup, but make sure we limit or width to < 80

JGSweets (approved):

JGSweets:
@sagars729 please update the branch with `main` Just need to click the button below

PR #81, opened 2021-03-19
Update precision calculation to address scientific notation, ints, and spaces

JGSweets:
Previously was precision after the decimal, now is complete sig figs of number. 
Also handles sig figs

### DIFF
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -96,21 +96,22 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         return None
 
     @classmethod
-    def _get_float_precision(cls, df_series):
+    def _get_float_precision(cls, df_series_clean):
         """
-        Determines the precision of the numeric value
+        Determines the precision of the numeric value.
         
-        :param df_series: a given column
-        :type df_series: pandas.core.series.Series
+        :param df_series_clean: df series with nulls removed, assumes all values
+            are floats as well
+        :type df_series_clean: pandas.core.series.Series
         :return: string representing its precision print format
         :rtype: int
         """
-        if df_series.empty:
+        if df_series_clean.empty:
             return 0
 
         # set it to first value length, at very least will be smaller since min.
-        precision = len(df_series.iloc[0])
-        for value in df_series:
+        precision = len(df_series_clean.iloc[0])
+        for value in df_series_clean:
 
             # remove all spaces around value
             value = value.strip()
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -106,7 +106,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :return: string representing its precision print format
         :rtype: int
         """
-        if df_series_clean.empty:
+        if not len(df_series_clean):
             return 0
 
         # set it to first value length, at very least will be smaller since min.
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -105,23 +105,39 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :return: string representing its precision print format
         :rtype: int
         """
-        integer_decimal_loc = -1
-        float_precision = None
+        if df_series.empty:
+            return 0
+
+        # set it to first value length, at very least will be smaller since min.
+        precision = len(df_series.iloc[0])
         for value in df_series:
-            decimal_loc = value.rfind('.')
 
-            # integers will not have a '.'
-            if decimal_loc == integer_decimal_loc:
-                continue
+            # remove all spaces around value
+            value = value.strip()
 
-            # since has a '.', subtract the str len from the position,
-            # since indexes start at 0: len - pos - 1
-            value_len = len(value)
-            value_precision = value_len - decimal_loc - 1
+            # if scientific notation, remove e and everything after
+            e_ind = value.find('e')
+            if e_ind > -1:
+                value = value[:e_ind]
 
-            if float_precision is None or float_precision > value_precision:
-                float_precision = value_precision
-        return float_precision
+            # remove negative of positive characters
+            if value[0] in ['+', '-']:
+                value = value[1:]
+
+            # if int: remove 0s on right, otherwise if float remove the decimal.
+            if '.' not in value:
+                value = value.rstrip('0')
+            else:
+                value = value.replace('.', '')
+
+            # strip all zeros on left
+            value = value.lstrip('0')
+            temp_precision = len(value)
+
+            # take the minimum precision
+            if temp_precision < precision:
+                precision = temp_precision
+        return precision
 
     @classmethod
     def _is_each_row_float(cls, df_series):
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/float_column_profile.py
```diff
@@ -105,23 +105,39 @@ def _get_float_precision(cls, df_series):
         :return: string representing its precision print format
         :rtype: int
         """
-        integer_decimal_loc = -1
-        float_precision = None
+        if df_series.empty:
```
lettergram, line 108: If it's empty, should it be None?

dataprofiler/profilers/float_column_profile.py
```diff
@@ -105,23 +105,39 @@ def _get_float_precision(cls, df_series):
...
+            value = value.strip()
+
+            # if scientific notation, remove e and everything after
+            e_ind = value.find('e')
+            if e_ind > -1:
+                value = value[:e_ind]
+
+            # remove negative of positive characters
+            if value[0] in ['+', '-']:
+                value = value[1:]
```
lettergram, line 125: Couldn't this technically be behind the number as well, i.e. `1+e3`?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -69,6 +69,36 @@ def test_profiled_precision(self):
...
+        self.assertEqual(2, float_profiler.precision)
+
+        # minus
+        df_minus = pd.Series(['-1.3234e-3', '-0.244', '-1.3324e4'])
+        float_profiler = FloatColumn("Name")
+        float_profiler.update(df_minus)
+        self.assertEqual(3, float_profiler.precision)
+
+        # spaces around values
+        df_spaces = pd.Series(['  -1.3234e-3  ', '  -0.244  '])
```
gme5078, line 98: What precision does it think '  -1.3234e-3  ' is?
Can we have a test for something like 323000.? 
I'm slightly worried that some of these tests actually fail, but since we are testing multiple floats at the same time, it still passes. Like maybe it is treating '  -1.3234e-3  ' as 12 sig figs (exaggeration i know), but the test passes because it properly recognized the common denominator of 3 with '  -0.244  '. Thoughts? I think we might want to separate many of these tests out

gme5078 (changes requested):

dataprofiler/profilers/float_column_profile.py
JGSweets: Currently, our default value is 0. Personally, I think `None` is best for default as well if we want to change.

dataprofiler/profilers/float_column_profile.py
JGSweets: Is that a valid notation? That looks more like an equation `1 + 1000`

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: fair, I will have to loop through since the function takes a series and provides the minimum.

dataprofiler/profilers/float_column_profile.py
lettergram: My concern is this may actually error as well:

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.empty.html

```python
>>> df_empty = pd.DataFrame({'A' : []})

>>> df_empty
Empty DataFrame
Columns: [A]
Index: []

>>> df_empty.empty
True
```

gme5078 (dismissed):

dataprofiler/profilers/float_column_profile.py
lettergram: Sorry, misplaced the sign on the wrong side of the `e`, but yes this is: `1e+3`

https://encyclopedia2.thefreedictionary.com/E+notation

> A type of scientific notation in which the phrase “times 10 to the power of” is replaced by the letter E; for example, 3.1 × 10^7 is written 3.1E+7 and 5.1 × 10-9 is written 5.1E-9.

So I was slightly off, but very close and is a common notation. All through school I used the notation in the style. `5.1e-9` notation.

dataprofiler/profilers/float_column_profile.py
lettergram:
```python
>>> df = pd.DataFrame({'A' : [np.nan]})

>>> df
    A
0 NaN

>>> df.empty
False

>>> df.dropna().empty
True
```

dataprofiler/profilers/float_column_profile.py
```diff
@@ -105,23 +105,39 @@ def _get_float_precision(cls, df_series):
         :return: string representing its precision print format
         :rtype: int
         """
-        integer_decimal_loc = -1
-        float_precision = None
+        if df_series.empty:
+            return 0
+
+        # set it to first value length, at very least will be smaller since min.
+        precision = len(df_series.iloc[0])
```
lettergram, line 112: what if `nan` is the first value?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -69,6 +69,57 @@ def test_profiled_precision(self):
...
+            ['10.01', 4],
+            ['.01', 1],
+            ['0.01', 1],
+            ['-0.01', 1],
+            ['+0.01', 1],
+            [' +0.013', 2],
+            ['  -1.3234e-3  ', 5],
+            ['  0012345600  ', 6],
+            ['  0012345600.  ', 8],
+            ['  -0012345600.  ', 8],
```
lettergram, line 115: Should toss a full nan's and none's in here.

lettergram (changes requested):
Need to test for the possibilities of `nans`

dataprofiler/profilers/float_column_profile.py
JGSweets: This function is created under the assumption that it is receiving a cleaned dataset, i.e. no `nan`s are provided to it.

The profilers should never receive a nan into its update profile as the series is cleaned prior to that point.

If we want to account for `nan`s it will make the function slower (maybe not much, unsure)? However, we would have to do that for all the profiles as all of them make the same assumption. If this is desired, maybe it should be in a separate PR cleaning up all profiles for this case.

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: as above

JGSweets:
@lettergram I don't believe this function is intended to receive `nan`s at any point as all of the profilers are supposed to receive data which is cleaned of `nan`s. If we want to make all profiler's handle `nan`s as well, I think that's outside of the scope of this PR as it will be many profile's which will need to be updated.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -105,23 +105,39 @@ def _get_float_precision(cls, df_series):
         :return: string representing its precision print format
```
lettergram, line 105: Add df_series description for required input

dataprofiler/profilers/float_column_profile.py
lettergram: Plz switch off `.empty`

lettergram (approved):
Recommend removing 'nans'

gme5078 (approved):

dataprofiler/profilers/float_column_profile.py
JGSweets: updated

PR #82, opened 2021-03-19
Fix issue of loading data labeler multiple times (50% faster)

AnhTruong:
- will reduce memory usage for new model graph
- reduce running time for populating new graph
- fix issue https://github.com/capitalone/DataProfiler/issues/63


lettergram:
Clearly there are still issues, but thus far this change is 50%.

I ran cProfiler on both the main and @AnhTruong branch using the following: 

https://gist.github.com/lettergram/d8f7d9f3d19856d4a0187462445382a0 

**Master repo**:
```
         33013991 function calls (30305353 primitive calls) in 20.547 seconds

   Ordered by: cumulative time
   List reduced from 9981 to 20 due to restriction <20>

        1    0.000    0.000   20.656   20.656 cprofile_test.py:81(Profile)
        1    0.000    0.000   20.656   20.656 profile_builder.py:309(__init__)
        1    0.000    0.000   20.650   20.650 profile_builder.py:456(update_profile)
        1    0.000    0.000   20.606   20.606 profile_builder.py:492(_update_profile_from_chunk)
       10    0.001    0.000   20.606    2.061 profile_builder.py:29(__init__)
       30    0.000    0.000   19.729    0.658 column_profile_compilers.py:18(__init__)
       30    0.002    0.000   19.729    0.658 column_profile_compilers.py:31(_create_profile)
       10    0.000    0.000   12.471    1.247 data_labeler_column_profile.py:14(__init__)
       10    0.000    0.000   12.469    1.247 data_labelers.py:73(__new__)
       10    0.000    0.000   12.469    1.247 base_data_labeler.py:23(__init__)
       10    0.000    0.000   12.469    1.247 base_data_labeler.py:518(_load_data_labeler)
       10    0.000    0.000   12.464    1.246 base_data_labeler.py:459(_load_model)
       10    0.000    0.000   11.179    1.118 character_level_cnn_model.py:253(load_from_disk)
       10    0.052    0.005   11.073    1.107 save.py:160(load_model)
       10    0.001    0.000   10.889    1.089 load.py:97(load)
       10    0.000    0.000    8.712    0.871 load.py:671(load_partial)
       10    0.000    0.000    8.711    0.871 load.py:862(load_internal)
       10    0.003    0.000    8.582    0.858 load.py:122(__init__)
       10    0.076    0.008    6.533    0.653 function_deserialization.py:294(load_function_def_library)
      180    0.001    0.000    6.445    0.036 base_column_profilers.py:77(wrapper)
```

**Anh's fork**:
```
         21868307 function calls (19369223 primitive calls) in 9.962 seconds

   Ordered by: cumulative time
   List reduced from 9955 to 20 due to restriction <20>

        1    0.000    0.000   10.068   10.068 cprofile_test.py:81(Profile)
        1    0.000    0.000   10.068   10.068 profile_builder.py:311(__init__)
        1    0.000    0.000   10.063   10.063 profile_builder.py:458(update_profile)
        1    0.000    0.000   10.017   10.017 profile_builder.py:498(_update_profile_from_chunk)
       10    0.001    0.000   10.017    1.002 profile_builder.py:30(__init__)
       30    0.000    0.000    9.150    0.305 column_profile_compilers.py:18(__init__)
       30    0.002    0.000    9.150    0.305 column_profile_compilers.py:31(_create_profile)
      180    0.001    0.000    5.868    0.033 base_column_profilers.py:77(wrapper)
       10    0.000    0.000    4.721    0.472 data_labeler_column_profile.py:275(update)
       10    0.035    0.003    4.717    0.472 data_labeler_column_profile.py:230(_update_predictions)
       10    0.016    0.002    4.663    0.466 base_data_labeler.py:250(predict)
       10    0.001    0.000    2.688    0.269 data_processing.py:1417(process)
2263330/530    1.254    0.000    2.535    0.005 copy.py:132(deepcopy)
 5959/530    0.010    0.000    2.534    0.005 copy.py:236(_deepcopy_dict)
       10    0.000    0.000    2.474    0.247 data_labeler_column_profile.py:15(__init__)
        1    0.000    0.000    2.473    2.473 data_labelers.py:73(__new__)
        1    0.000    0.000    2.473    2.473 base_data_labeler.py:23(__init__)
        1    0.000    0.000    2.473    2.473 base_data_labeler.py:518(_load_data_labeler)
        1    0.000    0.000    2.472    2.472 base_data_labeler.py:459(_load_model)
 85413/40    0.541    0.000    2.443    0.061 copy.py:210(_deepcopy_list)
```

dataprofiler/config.py
```diff
@@ -0,0 +1,4 @@
+global data_labeler_loaded, existing_data_labeler, data_labeler_called_from_profile
+data_labeler_loaded = False
+existing_data_labeler = None
+data_labeler_called_from_profile = False
```
lettergram, line 4: I don't think this is appropriate?

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -5,6 +5,7 @@
 from . import BaseColumnProfiler
 from ..labelers.data_labelers import DataLabeler
 from .profiler_options import DataLabelerOptions
+from .. import config
```
lettergram, line 8: I'd remove this

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -5,6 +5,7 @@
...
 
 
 class DataLabelerColumn(BaseColumnProfiler):
```
lettergram, line 11: IMO a `datalabeler` should be passed in as opposed to a `data_labeler_dirpath`

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -788,3 +788,7 @@ def load_with_components(cls, preprocessor, model, postprocessor):
         data_labeler.set_model(model)
         data_labeler.set_postprocessor(postprocessor)
         return data_labeler
+
+    def close(self):
+        if hasattr(self.model, 'close'):
+            self.model.close()
```
gme5078, line 794: I don't think this is ever executed because I believe it is self._model. But in eithercase, can we have tests for the close functions?

gme5078 (changes requested):

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,20 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
...
+            self.data_labeler = data_labeler
+        else:
+            data_labeler_dirpath = None
+            if options and options.data_labeler_dirpath:
+                    data_labeler_dirpath = options.data_labeler_dirpath
+
+            self.data_labeler = DataLabeler(
+                labeler_type='structured',
+                dirpath=data_labeler_dirpath,
+                load_options=None)
```
JGSweets, line 43: It might be cleaner to do:
```python
if data_labeler is None:
    data_labeler_dirpath = None
    if options and options.data_labeler_dirpath:
        data_labeler_dirpath = options.data_labeler_dirpath

    data_labeler = DataLabeler(
            labeler_type='structured',
            dirpath=data_labeler_dirpath,
            load_options=None)
            
self.data_labeler = data_labeler
```

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,20 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
             if not isinstance(options, DataLabelerOptions):
                 raise ValueError("DataLabelerColumn parameter 'options' must be"
                                  " of type DataLabelerOptions.")
-            if options.data_labeler_dirpath:
-                data_labeler_dirpath = options.data_labeler_dirpath
```
JGSweets, line 31: while we remove this check, we now need a new check for the `data_labeler` object

dataprofiler/profilers/profile_builder.py
```diff
@@ -80,15 +81,32 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
             'data_stats_profile':
                 ColumnStatsProfileCompiler(clean_sampled_df, self.options)}
 
-        # use the data labeler by default
-        use_data_labeler = True
-        if options and isinstance(options, StructuredOptions):
-            use_data_labeler = options.data_labeler.is_enabled
-
-        if use_data_labeler:
```
JGSweets, line 88: IMO, the options should still dictate the usage of the data_labeler whether someone passes it in or not.

That way we have one source of control. Having multiple ways to enable/disable could be confusing.

dataprofiler/labelers/base_data_labeler.py
JGSweets: I don't think all models will need this functionality and therefore should not be in base. Hence why this is checking to begin with, it may not be necessary. Otherwise, models that don't need it will have functions that are not used / intended.

dataprofiler/labelers/base_model.py
```diff
@@ -329,3 +329,12 @@ def fit(self, train_data, val_data, batch_size=32, epochs=1,
...
         raise NotImplementedError()
+
+    @abc.abstractmethod
+    def close(self):
+        """
+        Close the graph of the model.
+
+        :return: None
+        """
+        raise NotImplementedError()
```
JGSweets, line 340: something like the regex model does not need to close and this would cause errant functionality, IMO. We should ideate more on how to implement this if it is truly necessary.

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -788,3 +788,7 @@ def details(self):
         print("\nModel Label Mapping:")
         for key, value in self.label_mapping.items():
             print("{}: {}".format(key, value))
+
+    def close(self):
+        tf.keras.backend.clear_session()
+        print('closing model =====================')
```
JGSweets, line 794: why is closing necessary? Did we have to do it before?

Also, we should remove the print

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -15,20 +15,20 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
     # NOTE: these profilers are ordered. Test functionality if changed.
     _profilers = list()
 
-    def __init__(self, df_series, options=None):
+    def __init__(self, df_series, options=None, data_labeler=None):
```
JGSweets, line 18: This seems problematic as it forces all compilers, whether they need a datalabeler or not to have that as an input. I think we might want to implement the option only for the datalabeler profilers by building out its init maybe.

dataprofiler/profilers/profile_builder.py
```diff
@@ -365,6 +383,34 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
...
+        if profiler_options and profiler_options.structured_options:
+            structured_options = profiler_options.structured_options
+
+        if structured_options and isinstance(
+                structured_options, StructuredOptions):
+            data_labeler_options = structured_options.data_labeler
+
+        if data_labeler_options and isinstance(
+                data_labeler_options, DataLabelerOptions):
+            use_data_labeler = data_labeler_options.is_enabled
```
gme5078, line 401: My two cents is that data_labeler should be added to the data labeler column options and that there is no need for all the function signatures to change at all. 

Also, some/all these if statements can be combined

gme5078 (changes requested):

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -15,20 +15,20 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
...
         self._profiles = OrderedDict()
-        self._create_profile(df_series, options)
+        self._create_profile(df_series, options, data_labeler)
 
     @property
     @abc.abstractmethod
     def profile(self):
         raise NotImplementedError()
 
-    def _create_profile(self, df_series, options=None):
```
JGSweets, line 31: If possible we should think of another option to pass the data labeler instead of changing the functionality for all.

Potentially if it gets sent in options and passed to all.

JGSweets (changes requested):
Haven't completed full review yet, but posted initial thoughts.

dataprofiler/labelers/base_model.py
lettergram: If true, shouldn't this fail some test?

dataprofiler/labelers/base_data_labeler.py
AnhTruong: This method works, I double checked. The reason I added it is to make sure we can close session after done with the model. I can move to char-cnn model if needed. Whether to keep this, I'm open for the discussion.

dataprofiler/labelers/character_level_cnn_model.py
AnhTruong: The reason I added it is to make sure we can close session after done with the model. Whether to keep this, I'm open for the discussion.

lettergram: IMO Definitely remove the "print"

I do think you have to close the session. A tad concerned that closing will make it so it can't be reran

dataprofiler/profilers/column_profile_compilers.py
lettergram: Isn't that a large, intended aspect built into the library?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -56,8 +56,14 @@ def _create_profile(self, df_series, options=None):
...
-                    self._profiles[col_profile_type.col_type] = \
-                        col_profile_type(df_series.name, options=col_profile_options)
+                    if not data_labeler:
+                        self._profiles[col_profile_type.col_type] = \
+                            col_profile_type(df_series.name, options=col_profile_options)
+                    else:
+                        self._profiles[col_profile_type.col_type] = \
+                            col_profile_type(df_series.name,
+                                             options=col_profile_options,
+                                             data_labeler=data_labeler)
```
lettergram, line 66: Isn't it safe to pass a data_labeler with `none` as that's the default?

dataprofiler/labelers/base_data_labeler.py
lettergram: I'm not sure this specifically needs to be here

lettergram (changes requested):
I think there are definitely significant refactoring that can be done. 

Particularly, I'm not sure if close needs to be at the top level

dataprofiler/labelers/base_data_labeler.py
AnhTruong: this's only for `TrainableDataLabeler` and `BaseTrainableModel`, which are only used for ml models

lettergram: Then shouldn't it be in the model class?

AnhTruong: i can move it to the model class if we're all fine with that

dataprofiler/profilers/column_profile_compilers.py
AnhTruong: even we can add additional parameters for the datalabeler profiler, we still need some sort of the self.data_labeler saved in order to run the create_profile() function. If we don't want that self.data_labeler, we have to copy the function create_profile() to overwrite in the datalabeler profiler. I'm not sure there's a better way.

I'd suggest adding a new argument here, called model_associated. So, if the profiler doesn't need a model, it is None. Otherwise, datalabeler profiler needs a model. If later on, we have a ml model for histogram, it will needs a model as well. Thoughts? (I don't really like adding the model into the options. That sounds confusing to me.)

dataprofiler/labelers/base_data_labeler.py
JGSweets: A trainable model may or may not need close, hence I still don't think it would be appropriate here.

dataprofiler/labelers/base_model.py
JGSweets: This is new functionality so new tests would need to exist for it to fail.

dataprofiler/labelers/character_level_cnn_model.py
JGSweets: Why do we need to close the session? I thought TF2 was supposed to correct the need for session handling.

IIRC, using `tf.keras.backend.clear_session()` is like using a TF1 technique in `TF2`. I maybe wrong about this.

dataprofiler/profilers/column_profile_compilers.py
JGSweets: Either way we choose to do it, `BaseColumnProfileCompiler` should not need to have an altered input argument as only one compiler requires this functionality. Either we need a new class derived form it, or use an existing class already derived from it.

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -284,3 +289,8 @@ def update(self, df_series):
             self, self.__calculations, df_series=df_series,
             prev_dependent_properties={}, subset_properties=profile)
         self._update_helper(df_series, profile)
+
+    def close(self):
+        if hasattr(self.data_labeler, 'close'):
+            self.data_labeler.close()
```
lettergram, line 295: I think we can delete this

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,20 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
...
+            data_labeler_dirpath = None
+            if options and options.data_labeler_dirpath:
+                data_labeler_dirpath = options.data_labeler_dirpath
+
+            data_labeler = DataLabeler(
+                labeler_type='structured',
+                dirpath=data_labeler_dirpath,
+                load_options=None)
+
+        self.data_labeler = data_labeler
```
lettergram, line 43: This is super weird.

AnhTruong: what are multiple ways? Can you give an example?

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,21 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
...
-            if options.data_labeler_dirpath:
-                data_labeler_dirpath = options.data_labeler_dirpath
             if options.max_sample_size:
                 self._max_sample_size = options.max_sample_size
 
-        self.data_labeler = DataLabeler(
-            labeler_type='structured',
-            dirpath=data_labeler_dirpath,
-            load_options=None)
+        data_labeler = options.data_labeler_model
```
lettergram, line 33: It's not a model it's a full `DataLabeler` which includes a pipeline

dataprofiler/profilers/data_labeler_column_profile.py
lettergram: We need to throw a warning if there's a `data_labeler_dirpath` which differs from the `data_labeler`. `data_labeler` will also need to keep track of its source directory. Tests will need to be written.

dataprofiler/profilers/profile_builder.py
```diff
@@ -80,16 +81,31 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
             'data_stats_profile':
                 ColumnStatsProfileCompiler(clean_sampled_df, self.options)}
 
-        # use the data labeler by default
-        use_data_labeler = True
-        if options and isinstance(options, StructuredOptions):
-            use_data_labeler = options.data_labeler.is_enabled
-
-        if use_data_labeler:
+        data_labeler = self.options.data_labeler.data_labeler_model
```
lettergram, line 84: This isn't just the `data_labeler_model`, the data labeler contains a pre/post processor and a model. Recommend renaming to `data_labeler`, also why are there two levels to this? 

`data_labeler.data_labeler`?

dataprofiler/profilers/profile_builder.py
```diff
@@ -80,16 +81,31 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+        if data_labeler:
             self.profiles.update(
                 {'data_label_profile':
                      ColumnDataLabelerCompiler(clean_sampled_df, self.options)})
 
+        else:
+            # use the data labeler by default
+            use_data_labeler = True
+            if options and isinstance(options, StructuredOptions):
+                use_data_labeler = options.data_labeler.is_enabled
```
lettergram, line 94: Should be set at the top level in a default state?

dataprofiler/profilers/profile_builder.py
```diff
@@ -80,16 +81,31 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+
+            if use_data_labeler:
+                data_labeler_dirpath = None
+                if options and options.data_labeler_dirpath:
+                    data_labeler_dirpath = options.data_labeler_dirpath
+
+                self.options.data_labeler.data_labeler_model = \
+                    DataLabeler(labeler_type='structured',
+                                dirpath=data_labeler_dirpath,
+                                load_options=None)
```
lettergram, line 104: Isn't this effectively the same code as the column: `data_labeler_column_profile.py`

dataprofiler/profilers/profile_builder.py
```diff
@@ -365,6 +381,38 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
...
+                    data_labeler_options.data_labeler_dirpath
+
+            self.data_labeler = DataLabeler(
+                labeler_type='structured',
+                dirpath=data_labeler_dirpath,
+                load_options=None)
+
+        if self.data_labeler:
+            self.options.structured_options.data_labeler.data_labeler_model = \
+                self.data_labeler
```
lettergram, line 414: Couldn't this create an issue with the options?

lettergram (changes requested):

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -267,6 +267,92 @@ def test_structured_data_labeler_fit_predict_take_data_obj(self):
...
+        results = profile.report()
+
+        # print(json.dumps(human_readable_report, indent=4))
+        columns = []
+        predictions = []
+        for col in results['data_stats']:
+            columns.append(col)
+            predictions.append(results['data_stats'][col]['data_label'])
+        print(columns)
+        print(predictions)
```
lettergram, line 295: remove print statements

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,21 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
...
-        self.data_labeler = DataLabeler(
-            labeler_type='structured',
-            dirpath=data_labeler_dirpath,
-            load_options=None)
+        self.data_labeler = None
+        if options and options.data_labeler_object:
+            self.data_labeler = options.data_labeler_object
+        if self.data_labeler is None:
+            data_labeler_dirpath = None
+            if options and options.data_labeler_dirpath:
```
JGSweets, line 38: I don't believe the`options.data_labeler_dirpath` in the if statement is necessary as at worst case it sets it to `None` which it already would be. so maybe just `if options:`

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -284,3 +290,5 @@ def update(self, df_series):
             self, self.__calculations, df_series=df_series,
             prev_dependent_properties={}, subset_properties=profile)
         self._update_helper(df_series, profile)
+
+
```
JGSweets, line 294: Looks like an extra line got added, please remove.

dataprofiler/profilers/profile_builder.py
```diff
@@ -80,7 +81,31 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+        #         if options and options.data_labeler_dirpath:
+        #             data_labeler_dirpath = options.data_labeler_dirpath
+        #
+        #         self.options.data_labeler.data_labeler_object = \
+        #             DataLabeler(labeler_type='structured',
+        #                         dirpath=data_labeler_dirpath,
+        #                         load_options=None)
+        #         self.profiles.update(
+        #             {'data_label_profile':
+        #                  ColumnDataLabelerCompiler(clean_sampled_df, self.options)})
```
JGSweets, line 106: remove comments if these are not needed

dataprofiler/profilers/profile_builder.py
```diff
@@ -365,6 +390,38 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
         if isinstance(data, data_readers.text_data.TextData):
             raise TypeError("Cannot provide TextData object to Profiler")
 
+        # assign data labeler
+        use_data_labeler = True
+        data_labeler_dirpath = None
+        self.data_labeler = None
```
JGSweets, line 396: if the data_labeler is set in the options, is it necessary to store it here?

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -267,6 +267,92 @@ def test_structured_data_labeler_fit_predict_take_data_obj(self):
...
+                                 "float.is_enabled": False,
+                                 "order.is_enabled": False,
+                                 "category.is_enabled": False,
+                                 "datetime.is_enabled": False, })
+
+            profile = dp.Profiler(data, profiler_options=profile_options)
+
+            results = profile.report()
+
+            # print(json.dumps(human_readable_report, indent=4))
```
lettergram, line 318: remove

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -267,6 +267,92 @@ def test_structured_data_labeler_fit_predict_take_data_obj(self):
...
+            columns = []
+            predictions = []
+            for col in results['data_stats']:
+                columns.append(col)
+                predictions.append(results['data_stats'][col]['data_label'])
+            print(columns)
+            print(predictions)
```
lettergram, line 325: remove

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -267,6 +267,92 @@ def test_structured_data_labeler_fit_predict_take_data_obj(self):
...
+        profile_options.set({"text.is_enabled": False,
+                             "int.is_enabled": False,
+                             "float.is_enabled": False,
+                             "order.is_enabled": False,
+                             "category.is_enabled": False,
+                             "datetime.is_enabled": False, })
+        print('running dp2')
+        profile2 = dp.Profiler(data, profiler_options=profile_options)
+
+        #profile = profile1 + profile2
```
lettergram, line 354: I think these should be turned on?

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -267,6 +267,92 @@ def test_structured_data_labeler_fit_predict_take_data_obj(self):
...
+
```
lettergram, line 355: I think you need to also add a test where you do:

profile1 - create
profile2 - create
profile1 - update

To see if that triggers anything

dataprofiler/profilers/data_labeler_column_profile.py
AnhTruong: fixed

dataprofiler/profilers/data_labeler_column_profile.py
AnhTruong: removed

dataprofiler/profilers/profile_builder.py
AnhTruong: removed

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: done

dataprofiler/profilers/profile_builder.py
```diff
@@ -365,6 +390,38 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
...
+        if profiler_options and profiler_options.structured_options:
+            structured_options = profiler_options.structured_options
+
+        if structured_options and isinstance(
+                structured_options, StructuredOptions):
+            data_labeler_options = structured_options.data_labeler
+
+        if data_labeler_options and isinstance(
+                data_labeler_options, DataLabelerOptions):
+            use_data_labeler = data_labeler_options.is_enabled
```
JGSweets, line 383: If we receive options and they are `profiler_options` we can assume it has `structured_options` and should not need these checks. However, at the top we guarantee options to be ProfilerOptions by setting it to the class.

Additionally, here we only want to add a default_labeler if none exists in the options

Hence, I think we can simplify to:
```python
# assign data labeler
data_labeler_options = self.options.structured_options.data_labeler
if data_labeler_options.is_enabled and data_labeler_options.data_labeler_object is None:
    data_labeler = DataLabeler(
        labeler_type='structured',
        dirpath=data_labeler_options.data_labeler_dirpath,
        load_options=None)
    self.options.set({'data_labeler.data_labeler_object': data_labeler})
```

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: done

dataprofiler/profilers/profile_builder.py
AnhTruong: no, i removed self

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: yes, turned on now

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,21 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
...
-        self.data_labeler = DataLabeler(
-            labeler_type='structured',
-            dirpath=data_labeler_dirpath,
-            load_options=None)
+        self.data_labeler = None
+        if options and options.data_labeler_object:
+            self.data_labeler = options.data_labeler_object
+        if self.data_labeler is None:
+            data_labeler_dirpath = None
+            if options:
```
gme5078, line 38: think it should be: if options and options.data_labeler_dirpath

gme5078 (changes requested):

dataprofiler/profilers/profiler_options.py
```diff
@@ -461,6 +463,18 @@ def __init__(self):
         BaseColumnOptions.__init__(self)
         self.data_labeler_dirpath = None
         self.max_sample_size = None
+        self.data_labeler_object = None
+
+    def __deepcopy__(self, memo):
```
lettergram, line 468: Add comment to this function

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -27,15 +27,21 @@ def __init__(self, name, data_labeler_dirpath=None, options=None):
...
+                data_labeler_dirpath = options.data_labeler_dirpath
```
ChrisWallace2020, line 39: Just because options is not none doesn't necessarily imply it contains `data_labeler_dirpath`, right?

lettergram (approved):
Added my requested tests, looks good for now.

dataprofiler/profilers/profiler_options.py
AnhTruong: added

dataprofiler/profilers/profile_builder.py
AnhTruong: I tried but it has some error. But I think it can definitely be simplified later.

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: added

dataprofiler/profilers/data_labeler_column_profile.py
AnhTruong: in case options.data_labeler_dirpath is None, the data_labeler_dirpath is already None. So, it works.

dataprofiler/profilers/data_labeler_column_profile.py
AnhTruong: it contains data_labeler_dirpath as None by default

lettergram (approved):

gme5078 (approved):

dataprofiler/profilers/data_labeler_column_profile.py
gme5078: true

JGSweets (approved):
I believe there needs to be continued refactor of the Profiler __init__ but the current functionality seems to work as intended.

PR #89, opened 2021-03-22
Validate Options for CSVData

sagars729:
- Modified _check_and_return_options to validate delimiter, data_format, and selected_columns
- Added test_options to test valid and invalid delimiter, data_format, selected_columns, and header


dataprofiler/data_readers/csv_data.py
```diff
@@ -107,6 +107,17 @@ def _check_and_return_options(self, options):
                                  'none for no header, or a non-negative '
                                  'integer for the row that represents the '
                                  'header (0 based index)')
+            if key == "delimiter" and value != None and not isinstance(value, str):
```
gme5078, line 110: Because it's starting to look like there will be more if statements than dictionary length, i think this should be changed from a forloop to several "if <key> in options.keys()" statements. Will look a little cleaner too.

Also, should use "is not None" instead of "!= None"

gme5078 (changes requested):

dataprofiler/data_readers/csv_data.py
```diff
@@ -100,13 +100,31 @@ def _check_and_return_options(self, options):
...
         """
         options = super()._check_and_return_options(options)
-        for key, value in options.items():
-            if key == 'header' and value != 'auto' and value is not None \
-                    and not (isinstance(value, int) and value > -1):
+        
+        if 'header' in options.keys():
+            value = options["header"]
+            if value != 'auto' and value is not None \
+                and not (isinstance(value, int) and value > -1):
```
sagars729, line 107: @gme5078 Do you think I should remove the `value = ...` line and just collapse it into a single if statement? I wasn't sure which one was more readable

ChrisWallace2020: In my opinion, thats a bit much, and I think this is more readable

ChrisWallace2020 (approved):

gme5078 (approved):

PR #90, opened 2021-03-22
Add progress bar for profiler

sagars729:
- Used tqdm to add a progress bar for profiler
- Currently, doesn't look as clean with all the warnings, but it would generally give a single progress bar with steps taken/remaining and an eta
- Quick warning that a requirement was added and existing environments would need to be updated


dataprofiler/profilers/profile_builder.py
```diff
@@ -23,6 +23,7 @@
 from .helpers.report_helpers import calculate_quantiles, _prepare_report
```
gme5078, line 23: Can we see a picture of how it looks now?
Also, does this cause any slowdown?

gme5078 (changes requested):

dataprofiler/profilers/profile_builder.py
sagars729: In terms of slowdown,
with progress bar on guns.csv file: `20.00s user 2.37s system 107% cpu 20.799 total`
without progress bar on guns.csv file: `19.84s user 2.37s system 107% cpu 20.603 total`

requirements.txt
```diff
@@ -11,3 +11,4 @@ chardet>=3.0.4
 fastavro>=1.0.0.post1
 python-snappy>=0.5.4
 h5py>=2.10.0
+tqdm>=4.59.0
```
lettergram, line 14: I don't necessarily like adding another requirement. Also, this will be executed over SSH and non-standard terminal environments.

dataprofiler/profilers/profile_builder.py
```diff
@@ -23,6 +23,7 @@
...
 from .profiler_options import ProfilerOptions, StructuredOptions
 
+from tqdm import tqdm
```
lettergram, line 26: Import only needs to happen right before use IMO

dataprofiler/profilers/profile_builder.py
```diff
@@ -515,7 +516,7 @@ def _update_profile_from_chunk(df, profile=None, sample_size=None,
             raise ValueError('`Profiler` does not currently support data which '
                              'contains columns with duplicate names.')
 
-        for col in df.columns:
+        for col in tqdm(df.columns):
```
lettergram, line 519: tqdm shouldn't be a hard requirement IMO. Attempt import, if fails use another method.

lettergram (commented):
@JGSweets thoughts?

dataprofiler/profilers/profile_builder.py
sagars729: Screen Shot 2021-03-22 at 12 38 02 PM

dataprofiler/profilers/profile_builder.py
```diff
@@ -23,6 +23,14 @@
...
 from .profiler_options import ProfilerOptions, StructuredOptions
 
+try:
+	from tqdm import tqdm
+except:
+    def tqdm(l):
+        for i, e in enumerate(l):
+            print("Processing Column {}/{}".format(i, len(l)))
+            yield e
+
```
sagars729, line 33: @lettergram here is an alternative to importing tqdm

dataprofiler/profilers/profile_builder.py
sagars729: Does this mean I should move this to inside the _update_profile_from_chunk method?

dataprofiler/profilers/profile_builder.py
gme5078: ah I see, this was my fear. Looks kind of janky. But still seems to be an improvement. We'll have to look into that tensorflow warning.

dataprofiler/profilers/profile_builder.py
gme5078: Yup

sagars729: Moved it inside

dataprofiler/profilers/profile_builder.py
sagars729: Yeah, for sure; there does seem to be an issue open for it though - https://github.com/capitalone/DataProfiler/issues/63

dataprofiler/profilers/profile_builder.py
```diff
@@ -515,7 +515,15 @@ def _update_profile_from_chunk(df, profile=None, sample_size=None,
...
                              'contains columns with duplicate names.')
 
-        for col in df.columns:
+        try:
+            from tqdm import tqdm
+        except:
+            def tqdm(l):
+                for i, e in enumerate(l):
+                    print("Processing Column {}/{}".format(i+1, len(l)))
+                    yield e
```
lettergram, line 524: Interesting solution, I like

requirements.txt
lettergram: Plz remove requirement

sagars729: Removed the requirement
This makes sense as I'm not sure what would happen if tqdm tried to output to a non-standard terminal environment, but it does take away from a user who is unaware of this feature and installs to a virtual environment (since it probably wouldn't have tqdm).

lettergram (approved):

gme5078 (approved):

Issue #58, opened 2021-03-15
Constant memory for CSVData Match / Header check


JGSweets:
Currently the CSVData object reads in X rows which is non-deterministic for the amount of bytes being read.

Instead, we should have a max bytes / max rows to process for header/csv check. This can be done by reading in bytes at a time until X rows are read or X bytes have been read.

PR #96, opened 2021-03-24
Constrain memory for CSVData Match / Header check

lettergram:
Addresses #58


dataprofiler/data_readers/data_utils.py
```diff
@@ -368,3 +368,102 @@ def get_delimiter_regex(delimiter=",", quotechar=","):
...
+    """
+
+    data_as_str = ""
+    total_occurances = 0
+    with open(file_path, encoding=file_encoding) as csvfile:
+        
+        # Read the file until the appropriate number of occurances                    
+        for byte_count in range(0, max_bytes, chunk_size_bytes):
+            
+            sample_lines = csvfile.read(chunk_size_bytes)
```
JGSweets, line 455: not going to block this but these are more like sample_bytes than lines as it may not have any lines at all

dataprofiler/data_readers/data_utils.py
```diff
@@ -368,3 +368,102 @@ def get_delimiter_regex(delimiter=",", quotechar=","):
...
+            if len(sample_lines) == 0:
+                break
+        
+            remaining_lines = max_lines-total_occurances
```
JGSweets, line 459: non-blocking: but space around `-`

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -161,7 +161,7 @@ def test_auto_file_identification(self):
                 self.assertEqual(input_data_obj.delimiter, input_file['delimiter'],
                                  input_file['path'])
                 self.assertEqual(len(input_data_obj.data.columns),
-                                 input_file['num_columns'])
+                                 input_file['num_columns'], input_file['path'])
```
JGSweets, line 164: should this be on the next line? format wise.

dataprofiler/data_readers/csv_data.py
```diff
@@ -581,33 +581,32 @@ def is_match(cls, file_path, options=None):
...
-                    count = len([i.start() for i in re.finditer(delimiter_regex, line)])
-                else:                    
-                    # If no delimiter, see if spaces are regular intervals
-                    count = len([i.start() for i in re.finditer(space_regex, line)])
+            # Find the location(s) where each delimiter was detected
+            if delimiter:
+                count = len([i.start() for i in re.finditer(delimiter_regex, line)])
+            else:                    
+                # If no delimiter, see if spaces are regular intervals
+                count = len([i.start() for i in re.finditer(space_regex, line)])
```
JGSweets, line 599: not saying change now, but maybe a cleanup issue, as it seems we already no if a delimiter exists above. we could remove the if statement and just do

```python
delimiter_regex = data_utils.get_delimiter_regex(" ", quotechar)
if delimiter:
    delimiter_regex = data_utils.get_delimiter_regex(delimiter, quotechar)
for ....
    count = len([i.start() for i in re.finditer(delimiter_regex, line)])
```

dataprofiler/data_readers/data_utils.py
```diff
@@ -368,3 +368,97 @@ def get_delimiter_regex(delimiter=",", quotechar=","):
...
+        for byte_count in range(0, max_bytes, sample_size_bytes):
+            
+            sample_lines = csvfile.read(sample_size_bytes)
+            if len(sample_lines) == 0:                
+                break # No more bytes in file
+        
+            remaining_lines = max_lines - total_occurances
+            loc, occurance = find_nth_loc(sample_lines,
+                                          search_query='\n',
+                                          n=remaining_lines)
```
JGSweets, line 457: Maybe add a comment, but i think what this is doing is seeing how many lines it will search up to max are left and tell the find nth to look for all `\n`s, even if the remaining does not exist, tell me the last index of a `\n`?

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -36,3 +37,125 @@ def test_file_UTF_encoding_detection(self):
...
+                 query="\n", n=0, index=-1, occurances=0),
+            dict(string="This is a test\nThis is a second test\n",
+                 query="\n", n=2, index=36, occurances=2),
+            dict(string='t', query="t", n=1, index=0, occurances=1),
+            dict(string='s', query="t", n=1, index=1, occurances=0),
+            dict(string="This is a test\nThis is a second test\n\n",
+                 query="\n", n=3, index=37, occurances=3),
+            dict(string="This is a test\nThis is a second test\n\nTest\n",
+                 query="\n", n=5, index=43, occurances=4),
+            dict(string="", query="\n", n=3, index=-1, occurances=0),
```
JGSweets, line 61: ideally i think it would be best if they are broken into dicts of inputs / expected to know exactly via code, but this is fine.
e.g. `dict(input_args=dict(string=.....), expected_results=dict(index=.....)`

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -36,3 +37,125 @@ def test_file_UTF_encoding_detection(self):
...
+                results = ""
+            ),dict(
+                path=os.path.join(test_dir, 'csv/iris-utf-8.csv'),
+                encoding='utf-8', max_lines=5, max_bytes=65536,
+                chunk_size_bytes=1024,
+                results = """Id,SepalLengthCm,SepalWidthCm,PetalLengthCm,PetalWidthCm,Species
+1,5.1,3.5,1.4,0.2,Iris-setosa
+2,4.9,3.0,1.4,0.2,Iris-setosa
+3,4.7,3.2,1.3,0.2,Iris-setosa
+4,4.6,3.1,1.5,0.2,Iris-setosa"""
```
JGSweets, line 97: i find this format extremely confusing. did we have to use `"""` to do this?
something like:
```python
"Id,SepalLengthCm,SepalWidthCm,PetalLengthCm,PetalWidthCm,Species\n"
"1,5.1,3.5,1.4,0.2,Iris-setosa\n"
"2,4.9,3.0,1.4,0.2,Iris-setosa\n"
"3,4.7,3.2,1.3,0.2,Iris-setosa\n"
"4,4.6,3.1,1.5,0.2,Iris-setosa\n"
```
might be more readable

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -36,3 +37,125 @@ def test_file_UTF_encoding_detection(self):
...
+            dict(string="This is a test\nThis is a second test\n\nTest\n",
+                 query="\n", n=5, index=43, occurances=4),
+            dict(string="", query="\n", n=3, index=-1, occurances=0),
+        ]
+
+        for q in test_queries:
+            self.assertEqual(
+                data_utils.find_nth_loc(q['string'], q['query'], q['n']),
+                (q['index'], q['occurances'])
+            )
```
JGSweets, line 68: i believe the proper order is `self.assertEqual(expected, actual)`

dataprofiler/data_readers/data_utils.py
```diff
@@ -368,3 +368,97 @@ def get_delimiter_regex(delimiter=",", quotechar=","):
...
+    :param n: The number of occurances to iterate through
+    :type n: int
+    
+    :return idx: Index of the nth or last occurance of the search_query
+    :rtype idx: int
+    :return id_count: Number of identifications prior to idx
+    :rtype id_count: int
+    """
+    
+    if not string or not search_query or 0 >= n:
```
ChrisWallace2020, line 392: This could use a comment to explain intent of this case.

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -36,3 +37,125 @@ def test_file_UTF_encoding_detection(self):
...
+        ]
+
+        for f in input_files:
+            self.assertEqual(f['results'],
+                             data_utils.load_as_str_from_file(
+                                 file_path=f['path'], file_encoding=f['encoding'],
+                                 max_lines=f['max_lines'], max_bytes=f['max_bytes'],
+                                 chunk_size_bytes=f['chunk_size_bytes']
+                             ),
+                             f['path']
```
JGSweets, line 160: this is kinda difficult to understand,

might be easier to read if you do:
```python
expected = f['results']
output_str = data_utils.load_as_str_from_file(
     file_path=f['path'], file_encoding=f['encoding'],
     max_lines=f['max_lines'], max_bytes=f['max_bytes'],
     chunk_size_bytes=f['chunk_size_bytes']
 ),
self.assertEqual(expected, output_str, f['path'])
```

I think would be helpful for future code

dataprofiler/data_readers/data_utils.py
```diff
@@ -368,3 +368,103 @@ def get_delimiter_regex(delimiter=",", quotechar=","):
...
+            
+            sample_lines = csvfile.read(sample_size_bytes)
+            if len(sample_lines) == 0:                
+                break # No more bytes in file
+
+            # Number of lines remaining to be added to data_as_str
+            remaining_lines = max_lines - total_occurances
+            
+            # Return either the last index of sample_lines OR
+            # the index of the newline char that matches max_lines
```
JGSweets, line 459: matches max lines or remaining lines?

JGSweets: what if  last \n is less occurrences than remaining lines?

lettergram: > matches max lines or remaining lines?

Overall, this is in a for-loop so it finds the number of max_lines, if less lines occur then less lines occur 🤷‍♂️ 

> what if last \n is less occurrences than remaining lines?

Then the line break above happens:
```python
if len(sample_lines) == 0:                
    break # No more bytes in file
```

lettergram: Changed from "max_lines" to "remaining_lines" for clarity, but overall *the function itself* identifies up-to the max_lines

JGSweets (approved):

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -36,3 +37,127 @@ def test_file_UTF_encoding_detection(self):
...
+        Tests the ability for the `data_utils.find_nth_location` to detect the
+        nth index of a search_query in a string. 
+        """
+        # Input args: string, query, n
+        # Expected results: index, occurances
+        test_queries = [
+            dict(string="This is a test.", query=".", n=1, index=14, occurances=1),
+            dict(string="This is a test\n", query="\n", n=1, index=14, occurances=1),
+            dict(string="This is a test\nThis is a second test\n",
+                 query="\n", n=0, index=-1, occurances=0),
```
ChrisWallace2020, line 52: Returning an index of -1 seems somewhat misleading here, since this implies the 0th occurrence of the query was at the end of the string, which isn't true as that is the 2nd occurrence (in this case), would it be better to return None instead?

dataprofiler/data_readers/data_utils.py
```diff
@@ -368,3 +368,103 @@ def get_delimiter_regex(delimiter=",", quotechar=","):
...
+            remaining_lines = max_lines - total_occurances
+            
+            # Return either the last index of sample_lines OR
+            # the index of the newline char that matches remaining_lines
+            loc, occurance = find_nth_loc(sample_lines,
+                                          search_query='\n',
+                                          n=remaining_lines)
+
+            # Add sample_lines to data_as_str no more than max_lines
+            data_as_str += sample_lines[:loc]
```
ChrisWallace2020, line 465: Here, if loc is -1 (base case return of find_nth_loc) then data_as_str will add on sample_lines up to the last character, which feels somewhat incorrect. What would the intent be in this case (n = 0)?

lettergram: That could occur if there were no remaining lines... but that would break.

ChrisWallace2020 (approved):

PR #97, opened 2021-03-24
Add Tests for Order, Categorical, Datetime, Data Labeler Options

sagars729:
- Tests for Order, Categorical, and DateTime Options (Set Option Class, Leave framework)
- Tests for Data Labeler Options 
- Modified profiler_options to better validate all properties


dataprofiler/tests/profilers/test_categorical_options.py
```diff
@@ -0,0 +1,25 @@
...
+        super().test_set(*mocks)
+    
+    def test_validate_helper(self, *mocks):
+        super().test_validate_helper(*mocks)
+    
+    def test_validate(self, *mocks):
+        super().test_validate(*mocks)
+
+    def test_is_prop_enabled(self, *mocks):
+        super().test_is_prop_enabled(*mocks)
```
ChrisWallace2020, line 25: I'm not super familiar with mock but if you don't have any on this class why are you passing *mocks through?

ChrisWallace2020: This comment goes for all the other test files too, you aren't mocking anything out in any of them so is this kwarg still necessary?

dataprofiler/profilers/profiler_options.py
```diff
@@ -472,12 +503,15 @@ def _validate_helper(self, variable_path='DataLabelerOptions'):
         :rtype: list(str)
         """
         errors = super()._validate_helper(variable_path=variable_path)
-        if self.data_labeler_dirpath and \
+        if self.data_labeler_dirpath is not None and \
                 not isinstance(self.data_labeler_dirpath, str):
```
lettergram, line 507: What happens if the `self.data_labeler_dirpath` is none?

dataprofiler/tests/profilers/test_categorical_options.py
lettergram: @JGSweets would know more, I'm not super familiar

dataprofiler/profilers/profiler_options.py
JGSweets: no error occurs bc the datalabeler is set internally to the default when none

dataprofiler/tests/profilers/test_categorical_options.py
JGSweets: You know, that is a good catch. I missed that in the previous PRs too I believe. There would be no need for the `*mocks` if no mock decorator was applied (it may be required if it were applied to the base class).

dataprofiler/tests/profilers/test_order_options.py
```diff
@@ -0,0 +1,25 @@
+from dataprofiler.profilers.profiler_options import OrderOptions
+from dataprofiler.tests.profilers.test_base_column_options import TestBaseColumnOptions
+
+
+class TestOrderOptions(TestBaseColumnOptions):
+    
+    options_class = OrderOptions
+    
+    def test_init(self, *mocks):
```
JGSweets, line 9: same thing with the other mocks

dataprofiler/tests/profilers/test_categorical_options.py
sagars729: Makes sense, it was included in the original test_profiler_options since mocks were being used in those files, but there hasn't been a need for them in these tests - I will go ahead and delete them from all files.

dataprofiler/tests/profilers/test_datalabeler_options.py
```diff
@@ -0,0 +1,101 @@
...
+class TestDataLabelerOptions(TestBaseColumnOptions):
+    
+    option_class = DataLabelerOptions
+
+    def test_init(self):
+        options = self.get_options()
+        expected_val = {'data_labeler_dirpath': None, 
+            'max_sample_size': None, 
+            'is_enabled': True,
+            'data_labeler_object': None}
```
sagars729, line 14: This is new and currently missing tests - will add in future commits

sagars729: Added validation tests for data_labeler_object in new commits

dataprofiler/tests/profilers/test_datalabeler_options.py
```diff
@@ -0,0 +1,126 @@
...
+        super().test_set()
+    
+    def test_validate_helper(self):
+        super().test_validate_helper()
+        optpth = self.get_options_path() 
+
+        # Test valid dirpath
+        options = self.get_options()
+        options.set({'data_labeler_dirpath': ''})
+        self.assertEqual([], options._validate_helper())
```
AnhTruong, line 32: so, _validate_helper returns empty list if there's no error? If so, can you add a comment here for the clarity

dataprofiler/tests/profilers/test_datalabeler_options.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Test valid sample size
+        options = self.get_options()
+        options.set({'max_sample_size': 1})
+        self.assertEqual(None, options.validate()) 
+
+        # Test valid sample size
+        options = self.get_options()
+        options.set({'max_sample_size': 1})
+        self.assertEqual(None, options.validate())
```
ChrisWallace2020, line 87: Copied the same test case twice, and while I'm here I noticed you don't have a test for valid DataLabeler object into validate

sagars729: My bad, I meant to add a test for valid DataLabeler and must have added a valid sample size test - fixed in new commit!

dataprofiler/tests/profilers/test_datalabeler_options.py
sagars729: Added a comment to both test_validate and test_validate_helper that should help clarify this.

AnhTruong (approved):

ChrisWallace2020 (approved):

PR #98, opened 2021-03-24
Fix data labeler set bug

JGSweets:
Data labeler Options will fail if the data labeler is set and properties is called.
This causes failures in validation during the setting of the data labeler.

### DIFF
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -327,6 +327,7 @@ class IntOptions(NumericalOptions):
         """
         return super()._validate_helper(variable_path) 
 
+
 class FloatOptions(NumericalOptions):
     def __init__(self):
         """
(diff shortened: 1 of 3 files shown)


dataprofiler/tests/profilers/test_profiler_options.py
```diff
@@ -359,4 +357,21 @@ def test_data_labeler(self, *mocks):
                                     load_options=None)
         actual_sample_size = profile._profile[0].profiles['data_label_profile'] \
             ._profiles["data_labeler"]._max_sample_size
-        self.assertEqual(actual_sample_size, 50)
\ No newline at end of file
+        self.assertEqual(actual_sample_size, 50)
```
JGSweets, line 360: the following adds the test for the checking if data labeler can be set properly via the options

gme5078 (approved):
Very nice. Very clean.

dataprofiler/profilers/profiler_options.py
```diff
@@ -483,6 +484,19 @@ def __deepcopy__(self, memo):
...
+    def properties(self):
+        """
+        Returns a copy of the option properties.
+
+        :return: dictionary of the option's properties attr: value
+        :rtype: dict
+        """
+        props = {k: copy.deepcopy(v)
+                 for k,v in self.__dict__.items() if k != 'data_labeler_object'}
+        props['data_labeler_object'] = self.data_labeler_object
```
lettergram, line 497: I'm not 100% sure what would happen, but I'm guessing an error could happen here if you had multiple active instances of the data_labeler going on (i.e. multiple profiling). This could cause errors if the data_labeler_object was active; if it's not active it'll maybe be okay.

lettergram (approved):
We need to revisit the comment I mentioned. I'm not 100% sure it'll error, but I highly suspect that it could cause issues. It's fine as long as you only have one profiler active at a given time.

dataprofiler/profilers/profiler_options.py
JGSweets: I agree. However, the current code didn't allow setting or validation if a data labeler was set. 

Additionally, the previous code had the same issue where it was overwriting a mutable object, but this maintains the options having priority which may not be ideal.

Issue #94, opened 2021-03-22
Number of rows / objects in a file should be a global stat


lettergram:
**Is your feature request related to a problem? Please describe.**

The global stats should return with the number of rows or objects in the dataset.

Issue #71, opened 2021-03-16
Rows containing no data should either be included as a null row OR should be counted as an empty row


lettergram:
If you were to ingest the following file:

```
NAME, VALUE
test1, 1
test2, 2
test3, 3
,,
test5, 5
test6, 6
,,
test7, 7
```

I would expect there to be two null rows. Currently, the profiler identifies them as zero null rows.

gme5078:
Looking into this

PR #101, opened 2021-03-25
Fixed the Null Count Bug

gme5078:
AC:
- Solve the bug to make null count operate properly.
- Don't cause massive slowdown or failure.


dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -446,28 +446,42 @@ def test_null_count(self):
 
 class TestProfilerNullValues(unittest.TestCase):
 
-    @classmethod
-    def setUpClass(cls):
```
gme5078, line 450: Got rid of this/reorganized it because it doesnt really make sense to have a setupClass function for one test.

dataprofiler/profilers/profile_builder.py
```diff
@@ -504,7 +504,29 @@ def _update_row_statistics(self, data):
         self.hashed_row_dict = dict.fromkeys(
             pd.util.hash_pandas_object(data, index=False), True
         )
-        self.null_in_row_count = data.isnull().any(axis=1).sum()
```
gme5078, line 507: This was the bug. IsNull doesnt catch all null types, thats the issue.

dataprofiler/profilers/profile_builder.py
```diff
@@ -459,7 +460,8 @@ def report(self, report_options=None):
                 "samples_used": columns[0].sample_size if columns else 0,
                 "column_count": len(columns),
                 "unique_row_ratio": self._get_unique_row_ratio(),
-                "row_has_null_ratio": self._get_null_row_ratio(),
+                "row_has_null_ratio": self._get_null_in_row_ratio(),
+                "row_null_ratio": self._get_null_row_ratio(),
```
lettergram, line 464: `null_in_row_ratio`
`null_full_row_ratio`

`row_has_null_ratio`
`row_is_null_ratio`

lettergram:
Appears to be addressing:

#94 
#71

dataprofiler/profilers/profile_builder.py
```diff
@@ -504,7 +511,33 @@ def _update_row_statistics(self, data):
...
-        self.null_in_row_count = data.isnull().any(axis=1).sum()
+
+        # Calculate Null Column Count
+        null_rows = {}
+        null_in_row_count = {}
+        first_col_flag = True
+        for column in self.profile:
+            if "statistics" in self.profile[column].profile and \
+                    "null_types_index" in self.profile[column]\
+                    .profile["statistics"]:
```
JGSweets, line 522: are there cases where this was untrue? Is it possible to turn off in options?

JGSweets: Instead of calling from the profile, we can use the internals which will be cheaper than building an entire profile, 
i.e.: `self._profile[column].null_types_index`

JGSweets: I would assume then we wouldn't need the if statement, i maybe wrong.

dataprofiler/tests/data/csv/iris-with-null-rows.csv
```diff
@@ -0,0 +1,151 @@
...
+141,6.7,3.1,5.6,2.4,Iris-virginica
+142,6.9,3.1,5.1,2.3,Iris-virginica
+143,5.8,2.7,5.1,1.9,Iris-virginica
+144,6.8,3.2,5.9,2.3,Iris-virginica
+,,,,,
+146,6.7,3.0,5.2,2.3,Iris-virginica
+147,6.3,2.5,5.0,1.9,Iris-virginica
+148,6.5,3.0,5.2,2.0,Iris-virginica
+149,6.2,3.4,5.4,2.3,Iris-virginica
+150,5.9,3.0,5.1,1.8,Iris-virginica
```
AnhTruong, line 151: should we have a row with the last cell missing?

dataprofiler/tests/data/csv/empty_rows.txt
```diff
@@ -0,0 +1,9 @@
+NAME, VALUE
```
JGSweets, line 1: this is more ticky tacky but this file has nulls across a row while the iris one below has nulls randomly which is good but the file is super long.

it would be good to just truncate the iris null rows and it would make it more readable IMO. Is there an intention to having such  a long file?

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Since we are updating this, I think the profiler should not run the datalabeler, will make it faster.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -446,28 +449,44 @@ def test_null_count(self):
...
+        self.assertEqual(0.25, profile._get_row_is_null_ratio())
+        
+        
+        file_path = os.path.join(test_root_path, 'data','csv/iris-with-null-rows.csv')
+        data = pd.read_csv(file_path)
+        profile = dp.Profiler(data, profiler_options=profiler_options)
+        self.assertEqual(15, profile.row_has_null_count)
+        self.assertAlmostEqual(0.1, profile._get_row_has_null_ratio())
+        self.assertEqual(7, profile.row_is_null_count)
+        self.assertAlmostEqual(0.04666666, profile._get_row_is_null_ratio())
```
JGSweets, line 489: if we know the fraction, which I believe we should, we could use the fraction instead of `assertAlmostEqual`

dataprofiler/profilers/profile_builder.py
```diff
@@ -504,7 +511,33 @@ def _update_row_statistics(self, data):
...
+
+                # Cycle through each column and get all the null indices 
+                null_type_dict = self.profile[column].profile["statistics"]\
+                                 ["null_types_index"]
+                null_row_indices = {item for sublist in null_type_dict.values()
+                                    for item in sublist}
```
JGSweets, line 528: now i'm not sure about optimization, but this in of itself is creating a set.

Is it best to conduct the set operation on all these smaller lists and then a final operation merging all sets, or is it better to combine as a list and perform a single set operation?

Just thoughts

dataprofiler/profilers/profile_builder.py
```diff
@@ -504,7 +511,33 @@ def _update_row_statistics(self, data):
...
+
+                # Find the common null indices between the columns
+                if first_col_flag:
```
JGSweets, line 531: this if statement is only needed once. Another way to do this would be to set the initial values to the first column outside of the loop and then loop over all other columns.

Another option would be to set the null_rows to be all rows (perhaps memory inefficient) and null_in_row to an empty set.

just some thoughts, none of the necessarily better

dataprofiler/profilers/profile_builder.py
```diff
@@ -504,7 +511,33 @@ def _update_row_statistics(self, data):
...
+                    null_rows = null_row_indices
+                    null_in_row_count = null_row_indices
+                    first_col_flag = False
+                else:
+                    null_rows = null_rows.intersection(null_row_indices)
+                    null_in_row_count = null_in_row_count.union(null_row_indices)
```
JGSweets, line 537: technically we have a function which does this called `_combine_unique_sets` which is supposed to be faster I believe.

dataprofiler/profilers/profile_builder.py
```diff
@@ -504,7 +511,33 @@ def _update_row_statistics(self, data):
...
+        first_col_flag = True
+        for column in self.profile:
+            if "statistics" in self.profile[column].profile and \
+                    "null_types_index" in self.profile[column]\
+                    .profile["statistics"]:
+
+                # Cycle through each column and get all the null indices 
+                null_type_dict = self.profile[column].profile["statistics"]\
+                                 ["null_types_index"]
+                null_row_indices = {item for sublist in null_type_dict.values()
```
JGSweets, line 527: a link on combining sets of a dict need to discern whether this is truly different: https://stackoverflow.com/questions/60077538/how-to-speed-up-intersection-of-dict-of-sets-in-python

def intersect_sets(sets):
    return set.intersection(*sorted(sets, key=len))

dataprofiler/profilers/profile_builder.py
gme5078: 1. There were cases in the tests where it caused errors but using the internals got rid of the if statement :)

dataprofiler/tests/data/csv/iris-with-null-rows.csv
gme5078: I can make that happen

dataprofiler/tests/data/csv/empty_rows.txt
gme5078: I wanted this file because it the direct file that caused the issue. And I wanted a modified iris (with mix of null rows and null in rows) because it is a much larger and would cover more bases in my opinion. I will truncate iris somewhat.

dataprofiler/profilers/profile_builder.py
gme5078: thats for lists, kind of ironic name I think

JGSweets (approved):

lettergram (approved):

Issue #104, opened 2021-03-26
Consider renaming some of report variables 


lettergram:
Consider converting:
- [x] `total_samples` -> `data_object_count` or `samples_total` or `row_count`
- [x]  `BACKGROUND` -> `UNKNOWN`

Remove from pretty:
- [x]  `data_label_representation` -> remove
- [x]  `avg_predictions` -> remove
- [x]  `times` -> remove

Remove from all:
- [x]  `data_label_probability` -> remove
- [x]  `covariance` -> remove

Possibly remove (always null):
- [x]  `median` -> remove - can be approximated
- [x] `data_classification` -> remove - could easily be implemented for PII, NPI, etc

PR #110, opened 2021-03-27
Removing unnecessary keys in reports

lettergram:
Starts tackling issue #104

Changing the names of some of our functions which we hope to make more legible.


README.md
```diff
@@ -1,4 +1,4 @@
-# Data Profiler | What's in your data?
+\65;6203;1c# Data Profiler | What's in your data?
```
AnhTruong, line 1: What's `\65;6203;1c` for?

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -41,37 +41,105 @@ def flat_dict(od, separator='_', key=''):
...
+                      report a '.' represents a level of recursion example: 
+                      report: { 'test1': { 'test2': val, 'test3': val }, 
+                      to omit key 'test3' from report: omit_keys=['test1.test3']
+                      wildcards are also possible, so: omit_keys=['*.test3']
+    :type omit_keys: list(str)
+    :return report: handle to the updated report
+    :type report: dict()
+    """
+    
+    format_options = ['pretty', 'serializable', 'flat', 'compact']
```
AnhTruong, line 62: should we bring it outside of the function?

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -41,37 +41,105 @@ def flat_dict(od, separator='_', key=''):
...
     max_array_len = 5
+
+    if output_format == 'compact':
+        omit_keys.extend([
+            "data_stats.*.statistics.times",
+            "data_stats.*.statistics.avg_predictions",
+            "data_stats.*.statistics.data_label_representation",
+            "data_stats.*.statistics.null_types_index",
+            "data_stats.*.statistics.histogram.bin_counts",
+            "data_stats.*.statistics.histogram.bin_edges"
```
AnhTruong, line 79: did we decide to remove histogram results as well?

dataprofiler/tests/profilers/helpers/test_report_helpers.py
```diff
@@ -27,5 +28,157 @@ def test_serializable_report(self):
...
+        omit_keys = None
+        prepared_report = _prepare_report(report, output_format, omit_keys)
+        self.assertDictEqual(prepared_report, report)
+
+
+        # Test wildcard removals
+        wild_report1 = {
+            "test0": 0,
+            "test1": 1,
+            "test2": {
```
AnhTruong, line 133: should we remove test2 in this case, as it's an empty dict now.

AnhTruong (commented):
some comments

dataprofiler/profilers/helpers/report_helpers.py
ChrisWallace2020: Also this list isn't accessed at all within the function, what is its purpose exactly?

lettergram: Fair point @ChrisWallace2020, I removed the requirement - I'll add to the description. 

@AnhTruong These are just the modifications that can occur in the function. The function itself used to check if one of those options existed.

dataprofiler/profilers/helpers/report_helpers.py
lettergram: The histogram result was already truncated and inaccessible via the "pretty" format. 

I simply deleted it for the "compact" format. The reality is that, if a user wants to use the histogram they'll need to use none of the format options.

dataprofiler/tests/profilers/helpers/test_report_helpers.py
lettergram: I could probably recurse and do that, I was thinking I'd leave it unless otherwise specified.

ChrisWallace2020 (approved):

AnhTruong (approved):

ChrisWallace2020 (approved):

gme5078 (changes requested):
still reviewing.

gme5078 (commented):
Dismissing my previous review

gme5078 (approved):

PR #112, opened 2021-03-29
Use CharsetNormalizer alternative for Chardet

sagars729:
- Uses the charset_normalizer library instead of the chardet UniversalDetector
- Pros: More robust than chardet UniversalDetector
- Cons: Much slower
  - Chardet on guns.csv file: 0.000698089599609375 seconds
  - Charset Normalizer on guns.csv file: 0.3633871078491211 seconds


dataprofiler/data_readers/data_utils.py
```diff
@@ -288,17 +288,21 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
-    line_count = 0
-    with open(file_path, 'rb') as input_file:
-        chunk = input_file.read(buffer_size)
-        while chunk and line_count < max_lines:
-            detector.feed(chunk)
+    try:
+        from charset_normalizer import CharsetNormalizerMatches as CnM
+        result = CnM.from_path(file_path, steps=max_lines, chunk_size=buffer_size)
+        encoding = result.best().first().encoding
+    except:
```
sagars729, line 295: Used try except rather than adding new requirement

lettergram: Can you add a test that fails without this and works with this?

lettergram (changes requested):
I think a test needs to be added.

In terms of detection, does it need the whole file?

And I hate to say it, but would this be better?

https://github.com/chomechome/charamel

It doesn't appear to be maintained, which is a bit scary though.

dataprofiler/data_readers/data_utils.py
sagars729: Sure, let me look into that

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -37,7 +37,17 @@ def test_file_UTF_encoding_detection(self):
...
+        
+        try:
+            from charset_normalizer import CharsetNormalizerMatches as CnM
+            detected_encoding = \
+                data_utils.detect_file_encoding(file_path=os.path.join(test_dir, 'txt/utf8.txt'))
+            self.assertEqual("utf-8", detected_ecnoding.lower()) 
+        except:
+            detected_encoding = \
+                data_utils.detect_file_encoding(file_path=os.path.join(test_dir, 'txt/utf8.txt'))
+            self.assertEqual("windows-1254", detected_encoding.lower())
```
sagars729, line 49: @lettergram I added a test that uses the utf8.txt file from https://github.com/stain/encoding-test-files. Currently, I have it set to use CharsetNormalizer if it can and assert that it's detected as utf-8. Otherwise, it should be detected as a windows-1254 file by chardet.

ChrisWallace2020: I would leave a comment on this case to explain that one of the detections is incorrect, but it is knowingly incorrect in order to show that the CharsetNormalizer is superior and should be tried first.

sagars729: Added a comment to clarify that we are looking to verify that chardet returns an incorrect detection as compared to charset_normalizer

sagars729:
> In terms of detection, does it need the whole file?

It doesn't seem like it does, both chardet and charset_normalizer support detecting in chunks of bytes and only using a max number of chunks. But, both libraries also provide a simple command to run on the entire file.

lettergram:
@sagars729 it appears you can read in a set number of bytes - https://github.com/chomechome/charamel/blob/f5d664f19b8be70c2361f4037a4f065959e050bd/scripts/benchmark.py#L385

All you'd have to do is wrap it in a light weight function& load say 5kb or something.

I know it's more work, but would you mind trying that out? The reason I ask is really performance, taking half a second to just sample for encoding seems expensive. I'd like to avoid it if we can.

sagars729:
> @sagars729 it appears you can read in a set number of bytes - https://github.com/chomechome/charamel/blob/f5d664f19b8be70c2361f4037a4f065959e050bd/scripts/benchmark.py#L385
> 
> All you'd have to do is wrap it in a light weight function& load say 5kb or something.
> 
> I know it's more work, but would you mind trying that out? The reason I ask is really performance, taking half a second to just sample for encoding seems expensive. I'd like to avoid it if we can.

Sure! I will take a look and add a new commit for it

sagars729:
Also, one thing to note is that charset_normalizer may be failing on the iris-utf-8.csv file - it is being recognized as "gb18030" because of the last line in the file. I'm not sure if this is an error or not though. I'll see what happens with charamel

sagars729:
> Also, one thing to note is that charset_normalizer may be failing on the iris-utf-8.csv file - it is being recognized as "gb18030" because of the last line in the file. I'm not sure if this is an error or not though. I'll see what happens with charamel

With charamel, the guns.csv file is recognized as CP_775 encoding and the iris-utf-8.csv is recognized as TIS_620. Again, this may not be an error, but the expected ascii or utf-8 encoding

lettergram:
@sagars729 what is the current implementation doing?

sagars729:
> @sagars729 what is the current implementation doing?

@lettergram The existing implementation with chardet creates a UniversalDetector object every time detect_file_encoding is called and returns the following encodings for the three files mentioned:
- guns.csv: ascii
- iris-utf-8.csv: utf-8
- utf8.txt: windows-1254

lettergram:
@sagars729 Why don't we add a few more file type checks. IMO it might be easier to fix chardet (or hack around it) than implement something new. In all the cases you mentioned accuracy appeared to decrease?

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -37,7 +37,18 @@ def test_file_UTF_encoding_detection(self):
...
+        try:
+            from charset_normalizer import CharsetNormalizerMatches as CnM
+            detected_encoding = \
+                data_utils.detect_file_encoding(file_path=os.path.join(test_dir, 'txt/utf8.txt'))
+            self.assertEqual("utf-8", detected_encoding.lower()) 
+        except:
+            # Verify that detected encoding is incorrect using chardet instead of charset_normalizer
+            detected_encoding = \
+                data_utils.detect_file_encoding(file_path=os.path.join(test_dir, 'txt/utf8.txt'))
+            self.assertEqual("windows-1254", detected_encoding.lower())
```
AnhTruong, line 50: why don't we use assertIn for a list of utf-8 and windows-1254. So, we don't have to run detect_file_encoding in both try and except.

sagars729: Well, we wouldn't want windows-1254 as a result for charset_normalizer. Instead, it might be a better idea to only do the self.assertEqual in the try and except

sagars729:
> @sagars729 Why don't we add a few more file type checks. IMO it might be easier to fix chardet (or hack around it) than implement something new. In all the cases you mentioned accuracy appeared to decrease?

@lettergram Maybe, but I am not sure why chardet fails for the utf8.txt and I'm not sure how robust a hack would be.

I would say that the accuracy with the normalizer seemed to increase - it was able to correctly identify the utf8.txt case and the iris-utf-8.csv has a line "150,5.9,3.0,5.1,λεί τη μ,Iris-virginica" which causes it to be identified as gb18030, which is very similar to unicode with legacy compatibility: https://stackoverflow.com/questions/223360/how-does-gb18030-differ-from-unicode.

For charamel, the detector was able to correctly identify the utf8.txt case but I'm not sure what to make out of the other two cases; CP_775 and TIS_620 might work to decode the file, but I don't have as much confidence in those results.

JGSweets:
@sagars729, @lettergram were talking about potentially implementing charsetnormalizer as a fallback if chardet fails. charsetnormalizer is much slower, so ideally we would only use it if necessary.

lettergram:
I sampled 10k bytes for each of the examples below. All successfully worked with chaset_normalizer in the following method:

```python
rawdata = ""
in_file = open(filename, "rb") 
rawdata = in_file.read(10000)
in_file.close()

start_time = time.time()
results = CnM.from_bytes(
    rawdata,
    steps=5,  # Number of steps/block to extract from my_byte_str
    chunk_size=512,  # Set block size of each extraction
    threshold=0.2,
    cp_isolation=None,  # Finite list of encoding to use when searching for a match
    cp_exclusion=None,  # Finite list of encoding to avoid when searching for a match
    preemptive_behaviour=True,  # Determine if we should look into my_byte_str (ASCII-Mode) for pre-defined encoding
    explain=False  # Print on screen what is happening when searching for a match
)
charset_normalizer_encoding = results.best().first().encoding
```

```
reddit_wsb.csv
chardet_encoding Windows-1254 Time (sec): 2.86102294921875e-06
cchardet_encoding TIS-620 Time (sec): 0.000240325927734375
charamel_encoding Encoding.CP_932 Time (sec): 0.6641716957092285
charset_normalizer_encoding utf_8 Time (sec): 0.06576323509216309

zomato.csv
chardet_encoding Windows-1254 Time (sec): 2.6226043701171875e-06
cchardet_encoding ISO-8859-1 Time (sec): 0.0002841949462890625
charamel_encoding Encoding.CP_932 Time (sec): 0.6530959606170654
charset_normalizer_encoding iso8859_10 Time (sec): 0.06933283805847168

diamonds.csv
chardet_encoding ascii Time (sec): 3.0994415283203125e-06
cchardet_encoding ASCII Time (sec): 2.574920654296875e-05
charamel_encoding Encoding.UTF_7 Time (sec): 0.6447408199310303
charset_normalizer_encoding ascii Time (sec): 0.12702059745788574

iris.csv
chardet_encoding ascii Time (sec): 2.6226043701171875e-06
cchardet_encoding ASCII Time (sec): 2.002716064453125e-05
charamel_encoding Encoding.TIS_620 Time (sec): 0.6429488658905029
charset_normalizer_encoding ascii Time (sec): 0.008540630340576172

iris-utf-8.csv
chardet_encoding utf-8 Time (sec): 3.337860107421875e-06
cchardet_encoding UTF-8 Time (sec): 5.817413330078125e-05
charamel_encoding Encoding.TIS_620 Time (sec): 0.6465256214141846
charset_normalizer_encoding cp1125 Time (sec): 0.036332130432128906

iris-utf-16.csv
chardet_encoding UTF-16 Time (sec): 3.337860107421875e-06
cchardet_encoding UTF-16 Time (sec): 1.2636184692382812e-05
charamel_encoding Encoding.UTF_16 Time (sec): 0.6414165496826172
charset_normalizer_encoding utf_16 Time (sec): 0.01061701774597168

iris-utf-32.csv
chardet_encoding UTF-32 Time (sec): 2.384185791015625e-06
cchardet_encoding UTF-32 Time (sec): 1.33514404296875e-05
charamel_encoding Encoding.UTF_32 Time (sec): 0.7965893745422363
charset_normalizer_encoding utf_32 Time (sec): 0.10565662384033203

aws_honeypot_marx_geo.csv
chardet_encoding ascii Time (sec): 3.0994415283203125e-06
cchardet_encoding ASCII Time (sec): 2.6226043701171875e-05
charamel_encoding Encoding.CP_775 Time (sec): 0.6681275367736816
charset_normalizer_encoding ascii Time (sec): 0.012526512145996094
```

sagars729:
> 1. `chardet` first, map `Windows-1254` -> `utf_8` if it appears
> 2. If that fails, do a sample with `charset_normalizer`, say 10k chars
> 3. If that fails try larger sample, (say 100k chars) then fail

@lettergram sounds good, what does it mean for chardet to fail (`fail if encoding == None`)?

JGSweets:
@sagars729 it means that if the encoding selected does not work for reading the file or it it errors during encoding detection

dataprofiler/data_readers/data_utils.py
```diff
@@ -303,6 +302,43 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
     # Typical file representation is utf-8 instead of ascii, treat as such.
     if not encoding or encoding == 'ascii':
         encoding = 'utf-8'
+
+    # Check if encoding can be used to decode without throwing an error
+    def _decode_is_valid(encoding):
+        try: 
+            open(file_path, "rb").read().decode(encoding.lower())
+            return True
+        except: return False
```
sagars729, line 311: @lettergram @JGSweets this is how I am checking to see if I should use CharsetNormalizer/use more bytes - does this solution look better?

dataprofiler/data_readers/data_utils.py
```diff
@@ -303,6 +302,43 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
+
+            # Try again with full sample
+            if not _decode_is_valid(encoding): 
+                with open(file_path, 'rb') as input_file:
+                    raw_data = input_file.read(max_lines*buffer_size)
+                    result = CnM.from_bytes(raw_data, steps=max_lines, 
+                                            chunk_size=buffer_size, threshold=0.2,
+                                            cp_isolation=None, cp_exclusion=None,
+                                            preemptive_behaviour=True, explain=False)
+                encoding = result.best().first().encoding
```
JGSweets, line 334: @lettergram should this be a while loop that doubles the byte size each time if fails until feof or success?

dataprofiler/data_readers/data_utils.py
```diff
@@ -303,6 +302,43 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
+
+        except:
+            print("Install charset_normalizer for improved file encoding detection")
+
+    # If no encoding is still found, default to utf-8
+    if not encoding:
```
JGSweets, line 340: is this not covered by lines 303-304?

JGSweets: can CnM return None?

dataprofiler/data_readers/data_utils.py
```diff
@@ -303,6 +302,43 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
     # Typical file representation is utf-8 instead of ascii, treat as such.
     if not encoding or encoding == 'ascii':
```
JGSweets, line 303: @lettergram here do we want to set the if statement for if windows-1254 set to 'utf-8'?

JGSweets: we should also add the data that fails if this isn't true to check for it

sagars729: It might be a better idea to just leave it to CnM since I would assume there could be a file that actually does have the windows-1254 encoding (or alternatively put it in the except if CnM is not installed).

JGSweets: i think the idea is that the error was common enough that we should do that by default as a known error of the library and then CnM will correct it if we actually are incorrect.

dataprofiler/data_readers/data_utils.py
sagars729: I'm not sure but I think result can be, let me fix that

dataprofiler/data_readers/data_utils.py
sagars729: I used the alternative solution in the new commit (and changed the test accordingly)

dataprofiler/data_readers/data_utils.py
sagars729: Fixed, but I left the last if statement since I am not sure if CnM encoding can be None

dataprofiler/data_readers/data_utils.py
sagars729: @JGSweets, I can swap it and if windows-1254 is actually valid then CnM will correct it back to windows-1254 (given it is installed)

dataprofiler/data_readers/data_utils.py
lettergram: This will never fail, as far as I can gather. @JGSweets can you share a couple files we know fail (if you haven't already)?

Then, I think you're going to have to try probably just check the encoding to see if it's a known failure state. 

At the end of the day, you'll probably have to try loading it (see the section of the code which loads the files). If there's a failure at that point, reinitialize the process with this method or something to that effect.

dataprofiler/data_readers/data_utils.py
```diff
@@ -303,6 +302,47 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
+            return True
+        except: return False
+
+    if not _decode_is_valid(encoding):
+        try:
+            from charset_normalizer import CharsetNormalizerMatches as CnM
+            
+            # Try with small sample 
+            with open(file_path, 'rb') as input_file:
+                raw_data = input_file.read(2560)
```
lettergram, line 319: How was that selected?

sagars729: @lettergram In your code block you had tested for the small sample with

rawdata = in_file.read(10000)
...
results = CnM.from_bytes(
    rawdata,
    steps=5,  # Number of steps/block to extract from my_byte_str
    chunk_size=512,  # Set block size of each extraction
...

5 steps * 512 byte chunk size = 2560 bytes? Which is why I only read in 2560 bytes. Should I change this to do 10k?

sagars729:
@JGSweets moved the Windows-1254 bug check before CnM

dataprofiler/data_readers/data_utils.py
lettergram: Interesting, I'm not sure how it works, I recall I was getting less accurate results when I read in smaller amounts.

I'm fine leaving it, but was curious.

dataprofiler/data_readers/data_utils.py
```diff
@@ -301,7 +300,46 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
     encoding = detector.result["encoding"]
 
     # Typical file representation is utf-8 instead of ascii, treat as such.
-    if not encoding or encoding == 'ascii':
+    if not encoding or encoding == 'ascii' or encoding == 'Windows-1254':
```
lettergram, line 303: Why is ascii not okay?

dataprofiler/data_readers/data_utils.py
```diff
@@ -301,7 +300,46 @@ def detect_file_encoding(file_path, buffer_size=1024, max_lines=20):
...
+        encoding = 'utf-8'
+
+    # Check if encoding can be used to decode without throwing an error
+    def _decode_is_valid(encoding):
+        try: 
+            open(file_path, "rb").read().decode(encoding.lower())
```
lettergram, line 309: If you're going to use this, you should limit the number of bytes (could be 1000Gb). Recommend no more than 1Mb

dataprofiler/data_readers/data_utils.py
lettergram: I'd read in 10k samples, above you're already reading the entire file

dataprofiler/data_readers/data_utils.py
JGSweets: Additionally, I think we might want to consider either opening using `with` or saving the open so we can close the file when done rather than leaving open.

maybe
```python
bytes_to_read = 1024*1024  # 1MB
with open(file_path, 'r', encoding=encoding.lower()) as fp:
    fp.read(bytes_to_read)
```

dataprofiler/data_readers/data_utils.py
JGSweets: I would suggest using lower here as well and using an in statement to clean up the if statement:
```python
if not encoding or encoding.lower() in ['ascii', 'windows-1254']:
    ...
```

sagars729: Added in new commit

dataprofiler/data_readers/data_utils.py
sagars729: I read in 1Mb = 1024*1024 bytes rather the entire file and switched to using encoding=encoding in the open command in the new commit

lettergram (dismissed):

dataprofiler/tests/data_readers/test_data_utils.py
```diff
@@ -31,13 +31,25 @@ def test_file_UTF_encoding_detection(self):
                  encoding="utf-16"),
             dict(path=os.path.join(test_dir, 'json/iris-utf-32.json'),
                  encoding="utf-32"),
+            dict(path=os.path.join(test_dir, 'txt/utf8.txt'),
+                 encoding='utf-8'),
+            # Failing Test
+            #dict(path=os.path.join(test_dir, 'csv/zomato.csv'),
+            #     encoding='ISO-8859-1'),
```
lettergram, line 38: This should work, correct?

sagars729: Here were the results I got,
Chardet 20L, 1024B - Windows-1254 => utf-8; Fails to Decode
Chardet Full File - Windows-1254 => utf-8; Fails to Decode
Charset Normalizer 10k - iso8859_10; Decodes but content is different
Charset Normalizer Full File - cp1125; Decodes but content is different

sagars729: Charset Normalizer with 10k samples is very close though. It gets 99.987 % of characters correctly decoded.

sagars729: I could modify the test to check for getting 99.9% of characters, but it would still fail with Chardet (and cause the GitHub actions to fail too) as Charset Normalizer isn't a required package. That's why I just commented out the test and added a comment for now

lettergram (approved):

JGSweets (approved):

Ousret:
Hi there!
Sorry to dig up this closed PR.

Thanks for considering Charset-Normalizer. Since your initial merge, a lot has happened. Version 2.0+ is now faster than Chardet. (Twice in avg). Would be awesome to collect your throughs.
Do you have any feedback that I could use for future improvement? 

Regards,

JGSweets:
@Ousret I tried running `Charset-Normalizer` on the `zomato.csv` dataset in our library. However, it is not detecting an encoding.

The code I tried:
```python
import charset_normalizer

file_path = 'dataprofiler/tests/data/csv/zomato.csv'
print(str(charset_normalizer.from_path(file_path).best()))
# output: None
```

If this is not the appropriate usage, please LMK.

Ousret:
Hi @JGSweets 

Thanks for bringing that to my attention. I will look into it as soon as possible.
In the meantime, you may increase the default maximum chaos/mess threshold. Indeed using this file should return != None.

Ousret:
I have a patch ready. See https://github.com/Ousret/charset_normalizer/pull/72

Ousret:
Fixed in version 2.0.4, feel free to give us feedback.

PR #115, opened 2021-03-29
Add tests for Structured and Profiler Options

sagars729:
- Tests added for Structured and Profiler Options
- StructuredOption _validate_helper slightly modified to include variable path in error message
- ProfilerOptions _validate_helper to type check StructuredOptions and variable_path


dataprofiler/tests/profilers/test_profiler_class_options.py
```diff
@@ -0,0 +1,111 @@
...
+            for key in option.structured_options.properties]
+        expected_error = set(expected_error)
+        self.assertSetEqual(expected_error, expected_error.intersection(set(option._validate_helper())))
+
+        # Wrong Class Type
+        option = self.get_options()
+        option.structured_options = ProfilerOptions()
+
+        expected_error = set()
+        expected_error.add('{}.structured_options must be a StructuredOptions.'.format(optpth,))
```
ChrisWallace2020, line 84: This is a tad strange, would be more readable if you didn't use sets but just asserted that a singleton list was equal to the _validate_helper output, since if theres only a 1 element set, using a set at all feels unnecessary. Unless I am missing the motivation here.

dataprofiler/tests/profilers/test_profiler_class_options.py
```diff
@@ -0,0 +1,111 @@
...
+        self.assertEqual(None, option.validate())
+        
+        # Option is_enabled is not a boolean
+        for key in option.structured_options.properties:
+            option.set({'structured_options.{}.is_enabled'.format(key): "Hello World"}) 
+
+        expected_error = ['{}.structured_options.{}.is_enabled must be a Boolean.'.format(optpth, key) 
+            for key in option.structured_options.properties]
+        expected_error = set(expected_error)
+        self.assertSetEqual(expected_error, expected_error.intersection(set(option.validate(raise_error=False))))
```
ChrisWallace2020, line 103: If raise_error is False, then this is essentially the same as _validate_helper, since it literally just returns the return of the helper. I think it would be more productive to have the test cases for validate raise errors and you verify the right type of error is raised with the right messaging, while the test cases for the helper just check the lists returned.

dataprofiler/tests/profilers/test_profiler_class_options.py
```diff
@@ -0,0 +1,111 @@
...
+
+        # Wrong Class Type
+        option = self.get_options()
+        option.structured_options = ProfilerOptions()
+
+        expected_error = set()
+        expected_error.add('{}.structured_options must be a StructuredOptions.'.format(optpth,))
+        self.assertSetEqual(expected_error, set(option.validate(raise_error=False)))
```
ChrisWallace2020, line 111: Again, see my other 2 comments about sets not being necessary for this case and raise_error being False not being as informative of a test case.

ChrisWallace2020:
As a general request/change, now that you're about done with all your profiler tests, and they're all they're own files, I think it might be worthwhile to make an "options" folder within "tests/profilers" because they're all very related to one another and there are quite a few of them. Idk, other people can chime in on whether this is a good idea or not, it just seems like we have 10 test modules for 1 module so we should organize somehow.

dataprofiler/tests/profilers/test_profiler_class_options.py
sagars729: That is true. I had kept it consistent with the other assertions since they all used sets but it is very unnecessary. I will add a commit to fix this

dataprofiler/tests/profilers/test_profiler_class_options.py
sagars729: The only problem with this is that I don't want to assume what order the errors will be. If there is a list of errors then all of them are thrown but the regex assertion is only performed on the first line. I could hard code the first error thrown but I felt it was better just to compare the sets that are returned when raise_error is set to False.

sagars729: Alternatively, I can indent this assertion in to the for loop and have it check each error (which would make the raised assertion predictable) - I think this will be a better solution

dataprofiler/tests/profilers/test_structured_options.py
```diff
@@ -0,0 +1,153 @@
...
+        self.assertEqual(None, option.validate())
+        
+        # Option is_enabled is not a boolean
+        for key in self.keys:
+            option.set({'{}.is_enabled'.format(key): "Hello World"}) 
+        
+        expected_error = ["{}.{}.is_enabled must be a Boolean.".format(optpth, key) 
+            for key in self.keys]
+        expected_error = set(expected_error)
+        self.assertSetEqual(expected_error, expected_error.intersection(set(option.validate(raise_error=False))))
```
ChrisWallace2020, line 113: What is the point of asserting that set A is equal to set A intersect set B? Why not just ensure that set A equals set B?

ChrisWallace2020: This pattern is present several times, I think all should be fixed unless I'm missing something.

sagars729: Setting is_enabled doesn't just change is_enabled for each of those options but also the properties of those options as well. This means that the cardinality of set B >= set A, but we are only concerned with whether set A is a subset of set B

dataprofiler/tests/profilers/test_profiler_class_options.py
sagars729: I actually can't do this for this specific case because of the different cardinalities of the sets, but I will do it for the cases where it's only one error being raised.

ChrisWallace2020: Then is there some way to capture output and turn the string of errors separated by \n into a set and check that way? I understand thats a hassle for verifying a very small extra piece of logic but this test is essentially identical to the validate_helper ones.

dataprofiler/tests/profilers/test_structured_options.py
ChrisWallace2020: Ah I see, that makes sense, and I see there is no .assertIsSubset or anything, so I suppose this is fine, I would maybe just appreciate a comment above to explain what exactly you are trying to verify.

dataprofiler/tests/profilers/test_profiler_class_options.py
```diff
@@ -0,0 +1,110 @@
...
+    @classmethod
+    def get_options(self, **params):
+        options = ProfilerOptions()
+        options.set(params)
+        return options
+    
+    def test_init(self):
+        options = self.get_options()
+        for key in self.keys:
+            self.assertTrue(key in options.properties)
```
JGSweets, line 19: for this, can do `self.assertIn(key, options.properties)`
will have better error reporting

dataprofiler/tests/profilers/test_profiler_class_options.py
```diff
@@ -0,0 +1,110 @@
...
+            option._set_helper({'structured_options.{}.is_enabled'.format(key): False}, '')
+            self.assertFalse(option.structured_options.properties[key].is_enabled)        
+            option._set_helper({'structured_options.{}.is_enabled'.format(key): True}, '')
+            self.assertTrue(option.structured_options.properties[key].is_enabled)        
+
+        # Treat is_enabled as a BooleanOption
+        for key in option.structured_options.properties:
+            expected_error = "type object 'structured_options.{}.is_enabled' has no attribute 'is_enabled'".format(key)
+            with self.assertRaisesRegex(AttributeError, expected_error):
+                option._set_helper({'structured_options.{}.is_enabled.is_enabled'.format(key): True}, '')
```
JGSweets, line 37: I might suggest having tests that cross multiple layers like this in its own test class as this seems to test more than just the profileroptions.

Also, not sure this needs to be in test_set_helper as `_set_helper` is called by `set` which is where the same test also exists and is more of a duplicate.

dataprofiler/tests/profilers/test_profiler_class_options.py
```diff
@@ -0,0 +1,110 @@
...
+            option.set({'structured_options.{}.is_enabled'.format(key): False})
+            self.assertFalse(option.structured_options.properties[key].is_enabled)        
+            option.set({'structured_options.{}.is_enabled'.format(key): True})
+            self.assertTrue(option.structured_options.properties[key].is_enabled)        
+    
+        # Treat is_enabled as a BooleanOption
+        for key in option.structured_options.properties:
+            expected_error = "type object 'structured_options.{}.is_enabled' has no attribute 'is_enabled'".format(key)
+            with self.assertRaisesRegex(AttributeError, expected_error):
+                option.set({'structured_options.{}.is_enabled.is_enabled'.format(key): True})
```
JGSweets, line 55: similar to above

dataprofiler/tests/profilers/test_profiler_class_options.py
JGSweets: Though I guess for each class we only make the helper. I'm open to ideas. Ultimately I don't think we need to duplicate the tests since the set is more like a wrapper which calls it.

However, i do think we need an overall api tests which does check multiple levels.

dataprofiler/tests/profilers/test_structured_options.py
sagars729: Sure, I will add a comment to clarify in the next push

dataprofiler/tests/profilers/test_profiler_class_options.py
sagars729: Makes sense, modified in profiler, structured, and numerical options tests

dataprofiler/tests/profilers/test_profiler_class_options.py
sagars729: Took a little bit of tinkering, but yes was able to capture the error output and check that way - pushed that change in the new commit

sagars729:
> "options" folder within "tests/profilers"

@ChrisWallace2020 waiting on OpenSource office to approve and then I can push this update

dataprofiler/tests/profilers/test_profiler_class_options.py
sagars729: That makes sense - it does test more than just the profiler options. The aim of this test was more to get the type object '...' has no attribute error from the BaseOption class. I didn't want to assume that the BooleanOption Test had been run so this would be the first time this error would be raised and checked for. 

And yes, the set and _set_helper are identical with the two differences being the function and variable path. I thought it might make sense to leave it this way to 
  - test that both return the same results when called the same way
  - explicitly write it out for anyone who may need to update this test in the future

sagars729: @JGSweets do you think that justification is sufficient or do you think I should delete most of the _set_helper tests?

sagars729:
@ChrisWallace2020 
- created a profiler_options directory inside of "tests/profilers" that has all of the tests 
- changed all import statements to use new directory
- modified test_histogram_options to fix import statement

dataprofiler/tests/profilers/test_profiler_class_options.py
JGSweets: I think the PR is fine to merge, but I would suggest that since it is a unit test, we should be able to compartmentalize to the unit itself. BooleanOption test should be assumed to run, and a separate integration test should be able to catch if there's an issue when the two overlap.

JGSweets: I think catching the attribute error could be validated in the location where the test for that validate occurs

JGSweets: also might make the tests more modular based on keys etc

JGSweets (dismissed):

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
```diff
@@ -0,0 +1,116 @@
...
+
+        expected_error = ['{}.structured_options.{}.is_enabled must be a Boolean.'.format(optpth, key) 
+            for key in option.structured_options.properties]
+        expected_error = set(expected_error)
+        # Verify expected errors are a subset of all errors
+        with self.assertRaises(ValueError) as cm:
+            option.validate(raise_error=True)
+        raised_error = set(str(cm.exception).split("\n"))
+        self.assertEqual(expected_error, expected_error.intersection(raised_error))
+        self.assertSetEqual(expected_error, expected_error.intersection(set(option.validate(raise_error=False))))
```
ChrisWallace2020, line 107: This feels unnecessary as this is essentially the validate_helper test and you have a more robust version of this that handles the error being raised in the lines above, I think you can remove this because it makes this block a bit overwhelming.

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # Wrong Class Type
+        option = self.get_options()
+        option.structured_options = ProfilerOptions()
+
+        expected_error = '{}.structured_options must be a StructuredOptions.'.format(optpth,)
+        with self.assertRaisesRegex(ValueError, expected_error):
+            option.validate()
+        self.assertEqual([expected_error], option.validate(raise_error=False))
```
ChrisWallace2020, line 116: Why don't you catch the message here?

sagars729: the assertRaisesRegex will catch the message and assert it

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
+        
+        expected_error = ["{}.{}.is_enabled must be a Boolean.".format(optpth, key) 
+            for key in self.keys]
+        expected_error = set(expected_error)
+        # Verify expected errors are a subset of all errors
+        with self.assertRaises(ValueError) as cm:
+            option.validate(raise_error=True)
+        raised_error = set(str(cm.exception).split("\n"))
+        self.assertEqual(expected_error, expected_error.intersection(raised_error))
+        self.assertSetEqual(expected_error, expected_error.intersection(set(option.validate(raise_error=False))))
```
ChrisWallace2020, line 123: Again if you already do this subset check with the error output 2 lines above, and you do this check in test_validate_helper, I don't think this is necessary and it makes it a bit more cumbersome to understand the relationship between the 2 (from these tests).

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
+            if key == "data_labeler": ckey = "DataLabeler"
+            elif key == "category": ckey = "Categorical"
+            elif key == "datetime": ckey = "DateTime"
+            if key == "multiprocess":
+                expected_error.add('{}.{} must be a(n) BooleanOption.'.format(optpth, key, ckey))
+            else:
+                expected_error.add('{}.{} must be a(n) {}Options.'.format(optpth, key, ckey))
+        expected_error = set(expected_error)
+        # Verify expected errors are a subset of all errors
+        self.assertSetEqual(expected_error, set(option.validate(raise_error=False)))
```
ChrisWallace2020, line 148: I'm confusing why you are doing set equality here but subsets in other assertions? This seems redundant with the assertions in the following lines too

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
+        with self.assertRaises(ValueError) as cm:
+            option.validate(raise_error=True)
+        raised_error = set(str(cm.exception).split("\n"))
+        self.assertEqual(expected_error, expected_error.intersection(raised_error))
+        self.assertSetEqual(expected_error, expected_error.intersection(set(option.validate(raise_error=False))))
```
ChrisWallace2020, line 153: Once again I don't think this is necessary in test_validate

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
sagars729: It is a different method though - for the same reason the set and set_helper look very similar, it's just to test that both _validate_helper and validate are both returning the same errors (and if something changes it will be caught here).

sagars729: As long as validate calls _validate_helper correctly though, you are correct in that it is a bit unnecessary

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
```diff
@@ -0,0 +1,116 @@
...
+        # Variable Path Is Not A String
+        expected_error = "The variable path must be a string."
+        with self.assertRaisesRegex(ValueError, expected_error):
+            option._validate_helper(1)
+        
+        # Option is_enabled is not a boolean
+        for key in option.structured_options.properties:
+            option.set({'structured_options.{}.is_enabled'.format(key): "Hello World"}) 
+
+        expected_error = ['{}.structured_options.{}.is_enabled must be a Boolean.'.format(optpth, key)
```
ChrisWallace2020, line 74: This indentation feels wrong given the for loop below it is indented further in

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
```diff
@@ -0,0 +1,116 @@
...
+        expected_error = "The variable path must be a string."
+        with self.assertRaisesRegex(ValueError, expected_error):
+            option._validate_helper(1)
+        
+        # Option is_enabled is not a boolean
+        for key in option.structured_options.properties:
+            option.set({'structured_options.{}.is_enabled'.format(key): "Hello World"}) 
+
+        expected_error = ['{}.structured_options.{}.is_enabled must be a Boolean.'.format(optpth, key) 
+            for key in option.structured_options.properties]
```
ChrisWallace2020, line 75: Indent this `for` to line up with the open bracket on the line above it

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
```diff
@@ -0,0 +1,116 @@
...
+    
+        # Default Configuration Is Valid
+        self.assertEqual(None, option.validate())
+        
+        # Option is_enabled is not a boolean
+        for key in option.structured_options.properties:
+            option.set({'structured_options.{}.is_enabled'.format(key): "Hello World"}) 
+
+        expected_error = ['{}.structured_options.{}.is_enabled must be a Boolean.'.format(optpth, key) 
+            for key in option.structured_options.properties]
```
ChrisWallace2020, line 100: Fix this for indent

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
+        # Variable Path Is Not A String
+        expected_error = "The variable path must be a string."
+        with self.assertRaisesRegex(ValueError, expected_error):
+            option._validate_helper(1)
+        
+        # Option is_enabled is not a boolean
+        for key in self.keys:
+            option.set({'{}.is_enabled'.format(key): "Hello World"}) 
+        expected_error = ['{}.{}.is_enabled must be a Boolean.'.format(optpth, key) 
+            for key in self.keys]
```
ChrisWallace2020, line 74: Fix this indent (move to open bracket's column)

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
+    
+        # Default Configuration Is Valid
+        self.assertEqual(None, option.validate())
+        
+        # Option is_enabled is not a boolean
+        for key in self.keys:
+            option.set({'{}.is_enabled'.format(key): "Hello World"}) 
+        
+        expected_error = ["{}.{}.is_enabled must be a Boolean.".format(optpth, key) 
+            for key in self.keys]
```
ChrisWallace2020, line 116: Fix this indent (see other comments above)

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
+        for key in self.keys:
+            ckey = key.capitalize()
+            if key == "data_labeler": ckey = "DataLabeler"
+            elif key == "category": ckey = "Categorical"
+            elif key == "datetime": ckey = "DateTime"
+            if key == "multiprocess":
+                expected_error.add('{}.{} must be a(n) BooleanOption.'.format(optpth, key, ckey))
+            else:
+                expected_error.add('{}.{} must be a(n) {}Options.'.format(optpth, key, ckey))
+        expected_error = set(expected_error)
```
ChrisWallace2020, line 146: What is the point of this? expected_error was defined as a set and added to, why do we need to cast it to a set once again?

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,173 @@
...
```
ChrisWallace2020, line 100: Is this casting necessary?

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
sagars729: That is a good point, let me fix it in the future assertion - I think I got used to doing the subset assertions

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
ChrisWallace2020: But in my mind if they don't return the same thing it'll get caught in the fact that the same test cases and expected output across 2 test functions will have one pass and one not, so it feels unnecessary. This isn't critical or anything, just trying to find places to slim these functions down because they are quite lengthy.

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
ChrisWallace2020: Oh whoops, my mistake I thought this was just an exception catch not a assertRaisesRegex.

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
sagars729: Got it, will do

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
sagars729: Good point, will fix in next commit

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
sagars729: Nope, will fix

dataprofiler/tests/profilers/profiler_options/test_profiler_class_options.py
sagars729: That's true, just to clarify are you calling for removing all occurrences of raise_error=False or this entire block?

ChrisWallace2020 (dismissed):

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -0,0 +1,170 @@
...
+        expected_error = set()
+        for key in self.keys:
+            ckey = key.capitalize()
+            if key == "data_labeler": ckey = "DataLabeler"
+            elif key == "category": ckey = "Categorical"
+            elif key == "datetime": ckey = "DateTime"
+            if key == "multiprocess":
+                expected_error.add('{}.{} must be a(n) BooleanOption.'.format(optpth, key, ckey))
+            else:
+                expected_error.add('{}.{} must be a(n) {}Options.'.format(optpth, key, ckey))
```
JGSweets, line 99: not asking for a fix, but in future, since we have to specify each one already, instead of doing a loop, might be more readable anyways to just specify each error. Then no if / else either. 

Otherwise, technically you could get the class name of the expected object and do a loop.

sagars729:
I need to update this PR with all lines <= 80 characters. I will push this update soon

JGSweets (approved):

JGSweets (approved):

gme5078 (approved):

Issue #107, opened 2021-03-26
null ratio, rows ingested may not accurately reflect the sampling providing an incorrect


JGSweets:
**Describe the bug:**
self.rows_ingested = len(data) overwrites the rows ingested each time which means it does not allow for streaming data.

**To Reproduce:**
```python
import pandas as pd
import dataprofiler as dp


data = pd.DataFrame([1,2,3,4])

profiler = dp.Profiler(data[:2])
proifler.update(data[2:])
assert profiler.rows_ingested == 4
```

Additionally, these rows ingested don't represent the rows sampled during acquisition of the null parameters which is described by the column's samples. 

Since each column may get a differing sample set, this could cause issues because one col may have higher samples count that the other e.g.
```python
# presuming > 1 col
list(profiler.profile.values())[0].sample_size 
# may not equal, (notice the column index change)
list(profiler.profile.values())[1].sample_size 
```

PR #117, opened 2021-03-30
Fixed the null count and total samples, and more.

gme5078:
AC: 
- Address issue 107 (https://github.com/capitalone/DataProfiler/issues/107)
- Fix issue.

Notes:
Rows_ingested and total_samples were the same, got rid of rows_ingested.
The nulls were not being updated properly with multiple batches of data, fixed.
The null ratios were not being calculated properly, fixed.
Setting the seed for tests was faulty, fixed.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -447,11 +447,13 @@ class Profiler(object):
 
     @property
     def samples_used(self):
+        """
+        Calculates and returns the maximum samples used in any of the columns.
+        """
         samples_used = 0
         columns = list(self._profile.values())
         for col in columns:
-            if col.sample_size:
-                samples_used = max(samples_used, col.sample_size)
+            samples_used = max(samples_used, col.sample_size)
         return samples_used
 
     def report(self, report_options=None):
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -219,7 +219,7 @@ class StructuredDataProfile(object):
     def _get_sample_size(self, df_series):
         """
         Determines the minimum sampling size for detecting column type.
-        
+
         :param df_series: a column of data
         :type df_series: pandas.core.series.Series
         :return: integer sampling size
@@ -357,7 +357,6 @@ class Profiler(object):
         self.row_has_null_count = 0
         self.row_is_null_count = 0
         self.hashed_row_dict = dict()
-        self.rows_ingested = 0
         self._samples_per_update = samples_per_update
         self._min_true_samples = min_true_samples
         self._profile = dict()
@@ -432,7 +431,7 @@ class Profiler(object):
             self.row_has_null_count + other.row_has_null_count
         merged_profile.row_is_null_count = \
             self.row_is_null_count + other.row_is_null_count
-        merged_profile.rows_ingested = self.rows_ingested + other.rows_ingested
+        merged_profile.total_samples = self.total_samples + other.total_samples
         merged_profile.hashed_row_dict.update(self.hashed_row_dict)
         merged_profile.hashed_row_dict.update(other.hashed_row_dict)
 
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -483,16 +482,20 @@ def report(self, report_options=None):
         return _prepare_report(report, output_format, omit_keys)
 
     def _get_unique_row_ratio(self):
-        return len(self.hashed_row_dict) / self.rows_ingested
+        return len(self.hashed_row_dict) / self.total_samples
 
     def _get_row_is_null_ratio(self):
-        return self.row_is_null_count / self.rows_ingested
+        columns = list(self._profile.values())
+        samples_used = columns[0].sample_size
```
JGSweets, line 489: this value needs to be the max of all `column[0].sample_size` otherwise the ratio could be > 1

dataprofiler/profilers/profile_builder.py
```diff
@@ -483,16 +482,20 @@ def report(self, report_options=None):
...
+        return self.row_is_null_count / samples_used
 
     def _get_row_has_null_ratio(self):
-        return self.row_has_null_count / self.rows_ingested
+        columns = list(self._profile.values())
+        samples_used = columns[0].sample_size
```
JGSweets, line 494: this value needs to be the max of all `column[0].sample_size` otherwise the ratio could be > 1

dataprofiler/profilers/profile_builder.py
```diff
@@ -446,6 +445,15 @@ def __add__(self, other):
     def profile(self):
         return self._profile
 
+    @property
+    def samples_used(self):
```
JGSweets, line 449: description

dataprofiler/profilers/profile_builder.py
```diff
@@ -446,6 +445,15 @@ def __add__(self, other):
...
+        samples_used = 0
+        columns = list(self._profile.values())
+        for col in columns:
+            if col.sample_size:
```
JGSweets, line 453: why the if?

JGSweets (approved):

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -473,6 +472,9 @@ def test_null_count(self):
 
 class TestProfilerNullValues(unittest.TestCase):
 
+    def setUp(self):
+        test_utils.set_seed(0)
```
lettergram, line 476: This breaks the set seed does it not?

dataprofiler/tests/profilers/utils.py
```diff
@@ -11,8 +11,7 @@ def set_seed(seed=None):
     :param seed:
     :return:
     """
-    np.random.seed(seed)
-    random.seed(seed)
```
lettergram, line 15: Doesn't this still need to be set?

lettergram (changes requested):
I think modifying the environment variable is not appropriate

dataprofiler/tests/profilers/utils.py
JGSweets: ahh good catch, i do believe these would need to stay.

JGSweets:
@lettergram can you elaborate further on the env vars issue? The env var would only be temporary to the session in which the python tests being run and would not persist outside.

This also makes it easier on users as one may not realize they need to set the environment seed external to the tests.

Just would like more explanation to understand. Thanks!

lettergram:
@JGSweets the tests are already ran with the environment variable set, there's no reason to set it again (differently). We should be checking before setting.

Further, shouldn't this function still set the random seed?

I'm not sure how the forking works here (environment variables are copied), but there could be impacts to other forks.

JGSweets:
@lettergram What do you mean should be checking before setting? As in if it is already set, the test shouldn't overwrite? I thought you wanted to remove it entirely.

Forks shouldn't be impacted because it maintains the state at the start of the fork. 

However, I do understand the fork concern.

I'm fine approving on the revert, but I still would appreciate further explanation as I feel like I'm missing something.

Yes, seeds should still be set.

Either way, I think it would be nice if there was a mechanism in the long term where users don't have to set variables externally to a test in order for tests to be successful.

JGSweets (changes requested):
Seeds still need to be set.

lettergram:
> Either way, I think it would be nice if there was a mechanism in the long term where users don't have to set variables externally to a test in order for tests to be successful.

I agree, I think we dont' have a good solution though. The current environment variable solution lets seeds to be set outside the tests (which IMO is good). I don't know of a better solution at the moment because numpy doesn't let us set it another way. Only other solution would be to set a seed inside the library and pass it through (doable, but a larger effort).

dataprofiler/profilers/profile_builder.py
```diff
@@ -446,6 +445,17 @@ def __add__(self, other):
     def profile(self):
         return self._profile
 
+    @property
+    def samples_used(self):
+        """
+        Calculates and returns the maximum samples used in any of the columns.
```
AnhTruong, line 451: should we rename the function as `max_samples_used` then?

JGSweets (approved):

lettergram (approved):

PR #122, opened 2021-04-01
Added multiprocessing of column profiles (40% runtime reduction)

lettergram:
For comparison, see the two runs below (ran `United_States_Drought_Monitor__2000-2016.csv`)

I think there's still plenty of room to improve, but this reduces runtime by about 40% on this file. 

-------------------------

For `United_States_Drought_Monitor__2000-2016.csv`

Multiprocess:
```
50419680 function calls (50392635 primitive calls) in 122.032 seconds

   Ordered by: internal time
   List reduced from 1108 to 50 due to restriction <50>

      312   64.649    0.207   64.649    0.207 {method 'acquire' of '_thread.lock' objects}
       42   18.371    0.437   20.995    0.500 utils.py:53(shuffle_in_chunks)
  597/489   11.262    0.019   11.262    0.023 {built-in method numpy.array}
 16716443    3.991    0.000    3.991    0.000 {method 'search' of 're.Pattern' objects}
       14    3.897    0.278    3.897    0.278 {built-in method builtins.sorted}
       18    3.776    0.210    3.782    0.210 {pandas._libs.lib.map_infer}
        6    3.351    0.558   51.707    8.618 profile_builder.py:259(get_base_props_and_clean_null_params)
```

Single Process
```
189132806 function calls (188964701 primitive calls) in 191.153 seconds

   Ordered by: internal time
   List reduced from 1414 to 50 due to restriction <50>

      126   50.061    0.397   52.994    0.421 numerical_column_stats.py:239(_total_histogram_bin_variance)
       42   18.545    0.442   21.149    0.504 utils.py:53(shuffle_in_chunks)
10513/9311   14.103    0.001   14.114    0.002 {built-in method numpy.array}
        6   11.487    1.914   16.918    2.820 float_column_profile.py:97(_get_float_precision)
 16716414   10.113    0.000   21.751    0.000 base_column_profilers.py:47(_combine_unique_sets)
```
------------------------------

For `diamonds.csv` (no labeler)

Multiprocess:
```
2013326 function calls (1972475 primitive calls) in 5.615 seconds

   Ordered by: internal time
   List reduced from 1108 to 50 due to restriction <50>

      492    3.943    0.008    3.943    0.008 {method 'acquire' of '_thread.lock' objects}
       60    0.475    0.008    0.559    0.009 utils.py:53(shuffle_in_chunks)
```

Single process 
```
7273657 function calls (7192892 primitive calls) in 7.827 seconds

   Ordered by: internal time
   List reduced from 1473 to 50 due to restriction <50>

      304    1.945    0.006    1.945    0.006 {method 'acquire' of '_thread.lock' objects}
    23976    0.640    0.000    0.794    0.000 numerical_column_stats.py:386(_get_percentile)
       60    0.470    0.008    0.554    0.009 utils.py:53(shuffle_in_chunks)

```

### DIFF
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -94,21 +94,22 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
                     single_process_list.append(col_profile_type.col_type)
 
         # Loop through remaining multiprocesses and close them out
-        for col in multi_process_dict.keys():
-            try:
-                f = multi_process_dict[col]
-                returned_profile = f.get()
-                if returned_profile is not None:
-                    self._profiles[col] = returned_profile                
-            except Exception as e:
-                # Attempt again as a single process
-                single_process_list.append(col_profile_type.col_type)
-                    
-        # Close pool for new tasks        
-        pool.close()
-        
-        # Wait for all workers to complete
-        pool.join()
+        if multiprocess_flag:
+            for col in multi_process_dict.keys():
+                try:
+                    f = multi_process_dict[col]
+                    returned_profile = f.get()
+                    if returned_profile is not None:
+                        self._profiles[col] = returned_profile                
+                except Exception as e:
+                    # Attempt again as a single process
+                    single_process_list.append(col_profile_type.col_type)
+
+            # Close pool for new tasks  
+            pool.close()
+            
+            # Wait for all workers to complete
+            pool.join()
 
 
         # Single process thread to loop through
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -165,7 +165,8 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
         """
         df_series = df_series.apply(str)
         for column_profile in self._profiles:
-            self._profiles[column_profile].update(df_series)
+            self._profiles[column_profile].update(df_series)            
+        return self
 
 
 class ColumnPrimitiveTypeProfileCompiler(BaseColumnProfileCompiler):
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -9,6 +9,7 @@ from . import OrderColumn, CategoricalColumn
 from . import DataLabelerColumn
 from .profiler_options import StructuredOptions
 
+import multiprocessing as mp
 
 class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
 
@@ -40,41 +41,90 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
         :rtype: None
         """
 
+        if len(self._profilers) == 0:
+            return
+
         # convert all the values to string
         df_series = df_series.apply(str)
-        
+
+        multiprocess_flag = True
         selected_col_profiles = None
         if options and isinstance(options, StructuredOptions):
             selected_col_profiles = options.enabled_columns
+            multiprocess_flag = options.multiprocess.is_enabled
 
+        pool = mp.Pool(len(self._profilers))
+        single_process_list = []
+        multi_process_dict = {}
+
+        # smm = mp.managers.SharedMemoryManager()        
+        
         for col_profile_type in self._profilers:
+            
             # Create profile if options allow for it or if there are no options
             if selected_col_profiles is None or \
-                    col_profile_type.col_type in selected_col_profiles:
-                col_profile_options = None
+               col_profile_type.col_type in selected_col_profiles:
+
+                col_options = None
                 if options and options.properties[col_profile_type.col_type]:
-                    col_profile_options = options.properties[col_profile_type.col_type]
-
-                try:
-                    self._profiles[col_profile_type.col_type] = \
-                        col_profile_type(df_series.name, options=col_profile_options)
-                    self._profiles[col_profile_type.col_type].update(df_series)
-                except Exception as e:
-                    import warnings
-                    warning_msg = "\n\n!!! WARNING Partial Profiler Failure !!!\n\n"
-                    warning_msg += "Profiling Type: {}".format(col_profile_type.col_type)
-                    warning_msg += "\nException: {}".format(type(e).__name__)
-                    warning_msg += "\nMessage: {}".format(e)
-
-                    # This is considered a major error
-                    if type(e).__name__ == "ValueError":
-                        raise ValueError(e)
+                    col_options = options.properties[col_profile_type.col_type]
                     
-                    warning_msg += "\n\nFor labeler errors, try installing "
-                    warning_msg += "the extra ml requirements via:\n\n"
-                    warning_msg += "$ pip install dataprofiler[ml] --user\n\n"
+                self._profiles[col_profile_type.col_type] = \
+                    col_profile_type(df_series.name, options=col_options)
+                
+                # MULTIPROCESSING
+                single_process_flag = True
+                if multiprocess_flag:
+                    single_process_flag = False
+                    try:                        
+                        f = pool.apply_async(
+                            self._profiles[col_profile_type.col_type].update,
+                            (df_series,)
+                        )
+                        multi_process_dict[col_profile_type.col_type] = f
+                    except Exception as e:
+                        # Occurs when object cannot be pickled
+                        single_process_flag = True
 
-                    warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
+                # Add to single process list to execute later
+                if single_process_flag:
+                    single_process_list.append(col_profile_type.col_type)
+
+        for col in multi_process_dict.keys():
+            try:
+                f = multi_process_dict[col]
+                returned_profile = f.get()
+                if returned_profile is not None:
+                    self._profiles[col] = returned_profile                
+            except Exception as e:
+                # Attempt again as a single process
+                single_process_list.append(col_profile_type.col_type)
+                    
+        # Close pool for new tasks        
+        pool.close()
+        
+        # Wait for all workers to complete
+        pool.join()
+
+        for col_profile_type_col_type in single_process_list:
+            try:
+                self._profiles[col_profile_type_col_type].update(df_series)
+            except Exception as e:
+                import warnings
+                warning_msg = "\n\n!!! WARNING Partial Profiler Failure !!!\n\n"
+                warning_msg += "Profiling Type: {}".format(col_profile_type_col_type)
+                warning_msg += "\nException: {}".format(type(e).__name__)
+                warning_msg += "\nMessage: {}".format(e)
+
+                # This is considered a major error
+                if type(e).__name__ == "ValueError":
+                    raise ValueError(e)
+            
+                warning_msg += "\n\nFor labeler errors, try installing "
+                warning_msg += "the extra ml requirements via:\n\n"
+                warning_msg += "$ pip install dataprofiler[ml] --user\n\n"
+                
+                warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
             
 
     def __add__(self, other):
@@ -120,6 +170,9 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
 
 class ColumnPrimitiveTypeProfileCompiler(BaseColumnProfileCompiler):
 
+    def __repr__(self):
+        return 'ColumnPrimitiveTypeProfileCompiler'
+    
     # NOTE: these profilers are ordered. Test functionality if changed.
     _profilers = [
         DateTimeColumn,
@@ -136,7 +189,9 @@ class ColumnPrimitiveTypeProfileCompiler(BaseColumnProfileCompiler):
             "statistics": dict()
         }
         has_found_match = False
+        
         for _, profiler in self._profiles.items():
+
             if not has_found_match and profiler.data_type_ratio == 1.0:
                 profile.update(
                     {
@@ -153,6 +208,9 @@ class ColumnPrimitiveTypeProfileCompiler(BaseColumnProfileCompiler):
 
 class ColumnStatsProfileCompiler(BaseColumnProfileCompiler):
 
+    def __repr__(self):
+        return 'ColumnStatsProfileCompiler'
+    
     # NOTE: these profilers are ordered. Test functionality if changed.
     _profilers = [
         OrderColumn,
@@ -169,6 +227,10 @@ class ColumnStatsProfileCompiler(BaseColumnProfileCompiler):
 
 class ColumnDataLabelerCompiler(BaseColumnProfileCompiler):
 
+    def __repr__(self):
+        return 'ColumnDataLabelerCompiler'
+
+    
     # NOTE: these profilers are ordered. Test functionality if changed.
     _profilers = [
         DataLabelerColumn
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -16,6 +16,9 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
     # NOTE: these profilers are ordered. Test functionality if changed.
     _profilers = list()
 
+    def __repr__(self):
+        return self.__class__.__name__
+
     def __init__(self, df_series, options=None):
         if not self._profilers:
             raise NotImplementedError("Must add profilers.")
@@ -53,12 +56,12 @@ class BaseColumnProfileCompiler(with_metaclass(abc.ABCMeta, object)):
             selected_col_profiles = options.enabled_columns
             multiprocess_flag = options.multiprocess.is_enabled
 
-        pool = mp.Pool(len(self._profilers))
+
         single_process_list = []
         multi_process_dict = {}
+        if multiprocess_flag: # only create pool when necessary
+            pool = mp.Pool(len(self._profilers))
 
-        # smm = mp.managers.SharedMemoryManager()        
-        
         for col_profile_type in self._profilers:
             
             # Create profile if options allow for it or if there are no options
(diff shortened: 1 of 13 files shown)


dataprofiler/tests/profilers/test_base_column_profilers.py
```diff
@@ -272,9 +272,6 @@ def test_created_profile(self):
                 zip(self.column_profile._profiles.values(), profiler_mocks):
             self.assertIn(profile.__class__, self.profilers)
             self.assertEqual(1, profiler_mock.call_count)
-            self.assertEqual(
-                1, profiler_mock.return_value.update.call_count
-            )
```
lettergram, line 277: Can't keep a count of something which isn't called (mock the profiler) and mock's can't be pickled so this errors in multiprocess mode.

dataprofiler/tests/profilers/test_base_column_profilers.py
```diff
@@ -284,9 +281,6 @@ def test_updated_profile(self):
 
         for profiler_mock in profiler_mocks:
             self.assertEqual(1, profiler_mock.call_count)
-            self.assertEqual(
-                2, profiler_mock.return_value.update.call_count
-            )
```
lettergram, line 289: Can't keep a count of something which isn't called (mock the profiler) and mock's can't be pickled so this errors in multiprocess mode.

dataprofiler/tests/profilers/test_profiler_options.py
```diff
@@ -124,7 +124,9 @@ def test_disabling_vocab(self, vocab_mock, *mocks):
         vocab_mock.assert_not_called()
 
         # Check to see default options enable vocab
-        profile = Profiler(self.data)
+        multi_options = ProfilerOptions()
+        multi_options.structured_options.multiprocess.is_enabled = False
```
AnhTruong, line 128: so, this test doesn't work when the multiprocess is enabled?

dataprofiler/tests/profilers/test_profiler_options.py
```diff
@@ -296,11 +298,14 @@ def test_invalid_options_type(self, *mocks):
...
         options = ProfilerOptions()
         options.structured_options.float.precision.is_enabled = False
+        options.structured_options.multiprocess.is_enabled = False
 
         profile = Profiler(self.data, profiler_options=options)
         update_precision.assert_not_called()
 
-        profile = Profiler(self.data)
+        multi_options = ProfilerOptions()
+        multi_options.structured_options.multiprocess.is_enabled = False
```
AnhTruong, line 307: same as above

dataprofiler/tests/profilers/test_profiler_options.py
AnhTruong: it that's true can we set default value for multiprocess as False?

dataprofiler/profilers/int_column_profile.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import copy
```
AnhTruong, line 2: do we need it?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
         selected_col_profiles = None
         if options and isinstance(options, StructuredOptions):
             selected_col_profiles = options.enabled_columns
+            multiprocess_flag = options.multiprocess.is_enabled
+
+        pool = mp.Pool(len(self._profilers))
+        single_process_list = []
+        multi_process_dict = {}
 
+        # smm = mp.managers.SharedMemoryManager()
```
AnhTruong, line 60: remove?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+
+                # This is considered a major error
+                if type(e).__name__ == "ValueError":
+                    raise ValueError(e)
+            
+                warning_msg += "\n\nFor labeler errors, try installing "
+                warning_msg += "the extra ml requirements via:\n\n"
+                warning_msg += "$ pip install dataprofiler[ml] --user\n\n"
+                
+                warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
```
AnhTruong, line 127: can we add those warning messages to the multiprocess tasks?

AnhTruong (commented):
some comments. Do we have a wait to test if multiprocess is working?

dataprofiler/tests/profilers/test_profiler_options.py
lettergram: Multiprocess is a Python standard library. I don't see a reason to set it to false as, if you have anything >3 cores at the moment this should be faster.

Regarding why this test is single threaded: I set it to false because as it's a sub process the test cannot evaluate the number of times a function is executed (it doesn't own the process). You have to "own" the process for this library to see how often a given function is executed.

dataprofiler/profilers/int_column_profile.py
lettergram: No, I should remove.

dataprofiler/profilers/column_profile_compilers.py
lettergram: Indeed

dataprofiler/profilers/column_profile_compilers.py
lettergram: I'm unclear on what you mean.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                            (df_series,)
+                        )
+                        multi_process_dict[col_profile_type.col_type] = f
+                    except Exception as e:
+                        # Occurs when object cannot be pickled
+                        single_process_flag = True
+
+                # Add to single process list to execute later
+                if single_process_flag:
+                    single_process_list.append(col_profile_type.col_type)
```
JGSweets, line 91: not asking for change, but i'm not sure another flag was necessary, instead, the exception could have just set 
`multiprocess_flag = False` and continue to use that, no?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                # MULTIPROCESSING
+                single_process_flag = True
+                if multiprocess_flag:
+                    single_process_flag = False
+                    try:                        
+                        f = pool.apply_async(
+                            self._profiles[col_profile_type.col_type].update,
+                            (df_series,)
+                        )
+                        multi_process_dict[col_profile_type.col_type] = f
```
JGSweets, line 84: does this mean if you don't have tensorflow installed you have to run as a single processes?

JGSweets: actually the call above on 72 is the one that would technically error.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
-                        raise ValueError(e)
+                    col_options = options.properties[col_profile_type.col_type]
                     
-                    warning_msg += "\n\nFor labeler errors, try installing "
-                    warning_msg += "the extra ml requirements via:\n\n"
-                    warning_msg += "$ pip install dataprofiler[ml] --user\n\n"
-
-                    warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
+                self._profiles[col_profile_type.col_type] = \
+                    col_profile_type(df_series.name, options=col_options)
```
JGSweets, line 73: this has been removed from the try statement which is a change in functionality and removes the protection for profiles. I'm not sure we want that change.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                # Add to single process list to execute later
+                if single_process_flag:
+                    single_process_list.append(col_profile_type.col_type)
+
+        for col in multi_process_dict.keys():
+            try:
+                f = multi_process_dict[col]
+                returned_profile = f.get()
+                if returned_profile is not None:
+                    self._profiles[col] = returned_profile
```
JGSweets, line 98: is the object changing when it goes into and comes out of multiprocessing? This is confusing to me that we have to validate we get an object out and now have to return self for every update which it should return none be default.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -115,11 +165,15 @@ def update_profile(self, df_series):
         """
         df_series = df_series.apply(str)
         for column_profile in self._profiles:
-            self._profiles[column_profile].update(df_series)
+            self._profiles[column_profile].update(df_series)            
+        return self
```
JGSweets, line 169: why do the compiler updates need self as well? is that for consistency?

dataprofiler/profilers/column_profile_compilers.py
ChrisWallace2020: I think he means that currently the warnings only appear for tasks that are single process, and he wants to add them for when you iterate through the multiprocess ones, but @AnhTruong if you notice, when an exception occurs in multiprocess, the task gets added to single process, so in those cases if something truly were wrong it would go from multi_process_dict -> single_process_list -> warning, so I think its covered.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -115,11 +165,15 @@ def update_profile(self, df_series):
...
 
 
 class ColumnPrimitiveTypeProfileCompiler(BaseColumnProfileCompiler):
 
+    def __repr__(self):
+        return 'ColumnPrimitiveTypeProfileCompiler'
```
JGSweets, line 175: instead in the base could technically do:
```python
def __repr__(self):
    return self.__class__.__name__
```

would reduce some duplicate code

dataprofiler/profilers/float_column_profile.py
```diff
@@ -194,7 +194,24 @@ def _update_helper(self, df_series_clean, profile):
         if self._NumericStatsMixin__calculations:
             NumericStatsMixin._update_helper(self, df_series_clean, profile)
         self._update_column_base_properties(profile)
-
+        
+    def _update_numeric_stats(self, df_series, prev_dependent_properties,
```
JGSweets, line 198: good move

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -123,7 +123,7 @@ def test_categorical_mapping(self):
         num_null_types = 4
         num_nan_count = 2
         categories = pd.concat([df1, df2]).apply(str).unique().tolist()
-        column_profile.update_profile(df2)
+        column_profile.update_profile(df2)
```
JGSweets, line 126: why the added spaces?

JGSweets (changes requested):
I see a discussion on speed, but what are the implications on memory. last we talked, we thought it was possible that this could effectively duplicate the memory per process.

Do you have info on that?

dataprofiler/profilers/column_profile_compilers.py
lettergram: Next stage, but yes consistency. Goal was all updates return (no harm in doing so)

JGSweets (commented):
@lettergram mentioned that the size shouldn't matter as it is only ran on a single col. I am concerned that this will be a problem with large files even if it is a single column, but I will not block the PR.

dataprofiler/profilers/column_profile_compilers.py
JGSweets: I think the intent is that if multiprocess fails, it switches over to single anyways and if that fails will get this error so @lettergram  didn't feel the need to duplicate it if it was just a fault of multiprocessing failing, which I think is fair.

JGSweets: I believe @ChrisWallace2020 is right in his assessment: `multi_process_dict -> single_process_list -> warning`

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
-                    warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
+                self._profiles[col_profile_type.col_type] = \
+                    col_profile_type(df_series.name, options=col_options)
+                
+                # MULTIPROCESSING
+                single_process_flag = True
+                if multiprocess_flag:
+                    single_process_flag = False
+                    try:                        
+                        f = pool.apply_async(
```
gme5078, line 80: Why is this called f?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+
+                # Add to single process list to execute later
+                if single_process_flag:
+                    single_process_list.append(col_profile_type.col_type)
+
+        for col in multi_process_dict.keys():
+            try:
+                f = multi_process_dict[col]
+                returned_profile = f.get()
+                if returned_profile is not None:
```
gme5078, line 97: Why have this outside of the other for loop? Is it to wait for the process to finish and validate after the thread has finished? If so, shouldnt this be after  pool.close() and pool.join()?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                    single_process_flag = False
+                    try:                        
+                        f = pool.apply_async(
+                            self._profiles[col_profile_type.col_type].update,
+                            (df_series,)
+                        )
+                        multi_process_dict[col_profile_type.col_type] = f
+                    except Exception as e:
+                        # Occurs when object cannot be pickled
+                        single_process_flag = True
```
gme5078, line 87: Hmmm, is there a common case where a certain column is often ran as a single process?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
         df_series = df_series.apply(str)
-        
+
+        multiprocess_flag = True
         selected_col_profiles = None
         if options and isinstance(options, StructuredOptions):
             selected_col_profiles = options.enabled_columns
+            multiprocess_flag = options.multiprocess.is_enabled
+
+        pool = mp.Pool(len(self._profilers))
```
gme5078, line 56: This creates the pool workers even if the flag is set to false, I think that might not be proper.

dataprofiler/profilers/column_profile_compilers.py
gme5078: I would like that

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                    
+        # Close pool for new tasks        
+        pool.close()
+        
+        # Wait for all workers to complete
+        pool.join()
+
+        for col_profile_type_col_type in single_process_list:
+            try:
+                self._profiles[col_profile_type_col_type].update(df_series)
```
gme5078, line 111: This is supposed to cover the creation of the profile more than the update, so this warning isnt operating properly anymore

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                # Attempt again as a single process
+                single_process_list.append(col_profile_type.col_type)
+                    
+        # Close pool for new tasks        
+        pool.close()
+        
+        # Wait for all workers to complete
+        pool.join()
+
+        for col_profile_type_col_type in single_process_list:
```
gme5078, line 109: Maybe just name it profile_type or col_profile_type or something

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -40,41 +41,90 @@ def _create_profile(self, df_series, options=None):
...
+                    except Exception as e:
+                        # Occurs when object cannot be pickled
+                        single_process_flag = True
+
+                # Add to single process list to execute later
+                if single_process_flag:
+                    single_process_list.append(col_profile_type.col_type)
+
+        for col in multi_process_dict.keys():
+            try:
```
gme5078, line 94: Why would this fail and need a try-catch?

gme5078 (changes requested):

dataprofiler/profilers/column_profile_compilers.py
AnhTruong: Ok, makes sense

dataprofiler/profilers/column_profile_compilers.py
AnhTruong: I'd prefer keeping single_process_flag as even in the multiprocessing mode, if we have errors, we still need to run the tasks in the single_process_list. It looks more clear to me with that extra flag.

dataprofiler/profilers/column_profile_compilers.py
lettergram: Yes, you do - that will always be an issue with TF, but that's why we have a seperate "single-process" operation section

dataprofiler/profilers/column_profile_compilers.py
lettergram: f.get() returns the error from tensorflow, if it exists.

dataprofiler/profilers/profiler_options.py
```diff
@@ -585,6 +585,7 @@ def __init__(self):
         :ivar data_labeler: option set for data_labeler profiling.
         :vartype data_labeler: DataLabelerOptions
         """
+        self.multiprocess = BooleanOption()
```
gme5078, line 588: The ease at which this is added to options pleases me greatly 👍

AnhTruong (commented):
lgtm now, but it'd be nice if we have the test for multiprocess

AnhTruong (approved):

gme5078 (approved):

PR #123, opened 2021-04-01
Create HistogramOption object to allow user to specify Histogram method via options

ChrisWallace2020:
Wrote accompanying unit and integration tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,7 +497,8 @@ Below is an breakdown of all the options.
                 * is_enabled - (Boolean) Enables or disables variance
             * histogram_and_quantiles - Generates a histogram and quantiles
             from the column values
-                * method - (String) Designates preferred method for calculating histogram ('auto', 'fd', 'doane', 'scott', 'rice', 'sturges', 'sqrt'). If left unspecified all methods will be used to calculate and optimal one will be chosen.
+                * method - (String) Designates preferred method for calculating histogram. If left unspecified the optimal method will be chosen.  
+                options: 'auto', 'fd', 'doane', 'scott', 'rice', 'sturges', 'sqrt'
                 * is_enabled - (Boolean) Enables or disables histogram and quantiles  
         * datetime - Options for the datetime columns
             * is_enabled - (Boolean) Enables or disables the datetime operations
--- a/README.md
+++ b/README.md
@@ -497,6 +497,7 @@ Below is an breakdown of all the options.
                 * is_enabled - (Boolean) Enables or disables variance
             * histogram_and_quantiles - Generates a histogram and quantiles
             from the column values
+                * method - (String) Designates preferred method for calculating histogram
                 * is_enabled - (Boolean) Enables or disables histogram and quantiles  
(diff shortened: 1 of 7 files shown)


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -53,8 +53,15 @@ def __init__(self, options=None):
...
-        self.histogram_bin_method_names = ['auto', 'fd', 'doane', 'scott',
-                                           'rice', 'sturges', 'sqrt']
+        if options is not None:
+            self.histogram_selection = options.histogram_and_quantiles.method
+        else:
+            self.histogram_selection = None
+        if self.histogram_selection is not None:
+            self.histogram_bin_method_names = [self.histogram_selection]
+        else:
+            self.histogram_bin_method_names = ['auto', 'fd', 'doane', 'scott',
```
lettergram, line 63: Is there an "all" option or something or do you explicitly state the list?

ChrisWallace2020: So if nothing is specified via the option then it stays None, and then all are attempted as normal, if you want I can easily add support for an "all" keyword or something like that, but right now leaving it unspecified or `None` handles that.

AnhTruong: 'all' or 'infer' may be more explicit than None

AnhTruong:
also, should we merge histograms with different methods?

dataprofiler/profilers/numerical_column_stats.py
JGSweets: typically to avoid the if else, what we've done in the past is set the defaults before the if statement and hten allow the if to change it if met, like:
```python
self.histogram_bin_method_names = ['auto', 'fd', 'doane', 'scott',
                                           'rice', 'sturges', 'sqrt']
self.histogram_selection = None
if options is not None and isinstance(options, NumericalOptions):
    self.histogram_selection = options.histogram_and_quantiles.method
    if self.histogram_selection:
        self.histogram_bin_method_names = [self.histogram_selection]
```

would need to double check it hits all edge cases

JGSweets (approved):

JGSweets (dismissed):

JGSweets (approved):

README.md
```diff
@@ -497,6 +497,7 @@ Below is an breakdown of all the options.
                 * is_enabled - (Boolean) Enables or disables variance
             * histogram_and_quantiles - Generates a histogram and quantiles
             from the column values
+                * method - (String) Designates preferred method for calculating histogram
```
gme5078, line 500: You'll have to list out the possible options.

gme5078 (changes requested):

README.md
JGSweets: Agreed, should list out the options that are possible.

JGSweets (approved):

JGSweets (approved):

gme5078 (approved):

PR #127, opened 2021-04-02
Refactor to ensure multiprocessing happens on profile updates as well

lettergram:


dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -124,25 +121,22 @@ def test_categorical_mapping(self):
         num_nan_count = 2
         categories = pd.concat([df1, df2]).apply(str).unique().tolist()
         column_profile.update_profile(df2)
+        cat_profiler = column_profile.profiles['data_stats_profile']._profiles["category"]
```
AnhTruong, line 124: is cat_profiler generated from the beginning of this test? If it's regenerated, does it change the test meaning?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -124,25 +121,22 @@ def test_categorical_mapping(self):
...
+        print('real', cat_profiler.categories + column_profile.null_types)
+        print('cat_profiler', cat_profiler.categories + column_profile.null_types)
```
AnhTruong, line 126: also remove the print statement

dataprofiler/tests/profilers/test_categorical_column_profile.py
lettergram: Previously the test had a bug, because it assumed the object reference was consistent. Since we switched to multiprocessing the object is no longer going to have the same reference. Because of that we have to see the current properties of the class.

gme5078 (dismissed):

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -47,21 +51,12 @@ def _create_profile(self, df_series, options=None):
...
         if options and isinstance(options, StructuredOptions):
             selected_col_profiles = options.enabled_columns
-            multiprocess_flag = options.multiprocess.is_enabled
-
-
-        single_process_list = []
-        multi_process_dict = {}
-        if multiprocess_flag: # only create pool when necessary
-            pool = mp.Pool(len(self._profilers))
+            self.multiprocess_flag = options.multiprocess.is_enabled
```
JGSweets, line 57: any reason this is set here? I think would make the most sense int `__init__` as that's where options previously were set and it doesn't impact the profile creation.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -111,10 +111,7 @@ def test_categorical_mapping(self):
         num_nan_count = 1
         categories = df1.apply(str).unique().tolist()
         six.assertCountEqual(
-            self,
-            categories,
-            cat_profiler.categories +
-            column_profile.null_types)
+            self,categories,cat_profiler.categories+column_profile.null_types)
```
ChrisWallace2020, line 114: Why did we change this formatting? It seemed cleaner the other way?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -160,18 +104,60 @@ def __add__(self, other):
...
+        
         df_series = df_series.apply(str)
-        for column_profile in self._profiles:
-            self._profiles[column_profile].update(df_series)            
+                
+        # If single process, loop and return
+        if not self.multiprocess_flag:
+            for col_profile in self._profiles:
+                self._profiles[col_profile].update(df_series)
+            return self
```
JGSweets, line 127: this no longer is being caught by errors and specifying the `warn_on_profile`

Could we put hte other code in the if and then the code below isn't duplicated

dataprofiler/profilers/profile_builder.py
```diff
@@ -400,23 +399,7 @@ def __init__(self, data, samples_per_update=None, min_true_samples=None,
...
-                if type(e).__name__ == "ValueError":
-                    raise ValueError(e)
-                
-                warning_msg += "\n\nFor labeler errors, try installing "
-                warning_msg += "the extra ml requirements via:\n\n"
-                warning_msg += "$ pip install dataprofiler[ml] --user\n\n"
-                
-                warnings.warn(warning_msg, RuntimeWarning, stacklevel=2)
-
+                utils.warn_on_profile('data_labeler', e)
```
JGSweets, line 402: nice

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets, line 114: any reason why these are not spaced?

JGSweets (changes requested):
One of the profile updates is no longer try protected for import errors

JGSweets (approved):
since update makes sense to error on failure. Looks cleaner

dataprofiler/profilers/column_profile_compilers.py
JGSweets: will be fixed in future PR anyways so ignoring

gme5078 (approved):

Issue #129, opened 2021-04-02
Null rows are being incorrectly calculated when sampling


lettergram:
**General Information:**
 - OS: Arch Linux
 - Python version: 3.7 
 - Library version: 0.4.2

**Describe the bug:**

Because indexes are randomly generated *per column* 

https://github.com/capitalone/DataProfiler/blob/5dfa3b000d8f20b6921a2bdf36470c7fe543511f/dataprofiler/profilers/profile_builder.py#L280-L281

Those indexes are random and may not align with one another. They have to align to get correct results. The inaccurate results would be: `row_has_null` and `row_is_null`

**To Reproduce:**

Every time the data profiler is ran.

**Expected behavior:**

The solution is to:

1. Generate all the random indicates for sampling first. 
2. Each profile stop after they obtain the number of samples (`min_true_samples`) needed (return / store said location)
3. Do an intersection up-to the smallest number of samples needed for a given column - this should give you the `row_is_null` calculation

As part of this PR, tests should be written to check the `row_has_null` and `row_is_null` counts.

lettergram:
@ChrisWallace2020 will take this

lettergram:
`col1 = [0, 1, 2, NULL, 4, 5]` (actual nulls={3})
`col2 = [6, 7, NULL, 9, NULL]` (actual nulls={2, 4})
`sample_ids = [0, 1, 2, 3, 4]
If min_true_samples=4

-> col1 nulls = 1, stopped at index 4
-> col2 nulls = 2, stopped at index 5
(Take intersection for the smallest sample `intersection(col1.nulls, sample_ids[:4]), intersection(col2.nulls, sample_ids[:4])`
-> global row_has_nulls = 1
-> global row_is_null = 1

PR #130, opened 2021-04-03
Shuffle sampling indices once, when possible (reduced runtime by 15%)

lettergram:
Starts to tackle #129 Null rows are not being properly identified, the method in this PR attempts to address the issues, in part, by pre-generating all of the indices used for sampling, then sending them to the given column profilers. This (a) saves a lot of time and (b) starts limiting potential errors.  

Test file (56Mb): `United_States_Drought_Monitor__2000-2016.csv`

PR runtime:
```
36496459 function calls (36467890 primitive calls) in 105.407 seconds

   Ordered by: internal time
   List reduced from 1110 to 50 due to restriction <50>

      312   65.508    0.210   65.508    0.210 {method 'acquire' of '_thread.lock' objects}
  562/454    9.914    0.018    9.914    0.022 {built-in method numpy.array}
 16716443    3.985    0.000    3.985    0.000 {method 'search' of 're.Pattern' objects}
       18    3.950    0.219    3.956    0.220 {pandas._libs.lib.map_infer}
       14    3.937    0.281    3.937    0.281 {built-in method builtins.sorted}
```

Prior runtime: 
```
50428572 function calls (50400007 primitive calls) in 122.518 seconds

   Ordered by: internal time
   List reduced from 1111 to 50 due to restriction <50>

      312   64.863    0.208   64.863    0.208 {method 'acquire' of '_thread.lock' objects}
       42   18.592    0.443   21.230    0.505 utils.py:53(shuffle_in_chunks)
  597/489   11.308    0.019   11.308    0.023 {built-in method numpy.array}
 16716443    4.038    0.000    4.038    0.000 {method 'search' of 're.Pattern' objects}
       14    3.886    0.278    3.886    0.278 {built-in method builtins.sorted}
```

Regarding the next steps for task #129 - track the location in the `sample_ids` array where the first column obtains enough samples. Then to use that index to drop all nulls in each column *when calculating the null ratios*. As there are already a set containing all the null indices per column, it should be straight forward to create a new set containing the indices that need to be dropped. This can be accomplished taking the `sample_ids[:slice_loc]` and using it as an intersection for the other columns.


AnhTruong (approved):

JGSweets (approved):
This may make a list the size of the original dataset which could be a memory issue, but the current priority is speed.

dataprofiler/profilers/profile_builder.py
```diff
@@ -277,16 +287,22 @@ def get_base_props_and_clean_null_params(self, df_series, sample_size,
         # Pandas reads empty values in the csv files as nan
         df_series = df_series.apply(str)
 
-        sample_ind_generator = utils.shuffle_in_chunks(
-            len_df, chunk_size=sample_size)
 
+        # Select generator depending if sample_ids availablity
+        if sample_ids is None:
```
gme5078, line 292: I dont think we have any tests that explore this condition. Can we make one and just make sure the profile is operating appropriately? Think the PR looks good, just needs some kind of test

gme5078 (changes requested):

dataprofiler/profilers/profile_builder.py
```diff
@@ -309,7 +325,8 @@ def get_base_props_and_clean_null_params(self, df_series, sample_size,
                 break
             
         # close the generator in case it is not exhausted.
-        sample_ind_generator.close()
+        if sample_ids is None:
```
ChrisWallace2020, line 328: Why is this checking if `sample_ids` is None and not if `sample_ind_generator` is None?

ChrisWallace2020: Furthermore why is any check necessary since either way a sample_ind_generator is created?

dataprofiler/profilers/profile_builder.py
```diff
@@ -617,14 +634,19 @@ def tqdm(l):
                 for i, e in enumerate(l):
                     print("Processing Column {}/{}".format(i+1, len(l)))
                     yield e
-			
+
+
+        # Shuffle indices ones and share with columns
```
ChrisWallace2020, line 639: I think you meant "once" not "ones"?

Issue #151, opened 2021-04-13
Min, Max & Avg precision for float & integer columns


lettergram:
**Is your feature request related to a problem? Please describe.**

There are cases where users may know the given dataset contains measurements from a device with a given precision. If that's the case, our measured precision is likely highly inaccurate. Take the case of the following array `[15000, 39023, 94201401]`, if we know the measurement 15000 is accurate - the significant figures are not 2, they are really 5. We should take that into account and provide both to the users.

**Describe the outcome you'd like:**

I'd like to see the minimum measured precision and the maximum measured precision.

Precision should also be shown if it's an integer column.

PR #153, opened 2021-04-13
Expanded precision report (and 40% reduced runtime on numeric columns)

lettergram:
Addresses #151 

File for assessment: `United_States_Drought_Monitor__2000-2016.csv` (6 integer columns)

**Current** (in master) 
```
         39720854 function calls (39580112 primitive calls) in 96.714 seconds

   Ordered by: internal time
   List reduced from 9663 to 50 due to restriction <50>

     1328   57.742    0.043   57.742    0.043 {method 'acquire' of '_thread.lock' objects}
  496/478    9.885    0.020    9.885    0.021 {built-in method numpy.array}
     8375    3.957    0.000    3.958    0.000 {built-in method builtins.sorted}
 16716568    3.947    0.000    3.947    0.000 {method 'search' of 're.Pattern' objects}
        6    3.211    0.535   29.649    4.941 profile_builder.py:256(get_base_props_and_clean_null_params)
        2    3.140    1.570    3.591    1.796 utils.py:73(shuffle_in_chunks)
        6    2.461    0.410    8.344    1.391 {pandas._libs.lib.map_infer_mask}
```

**This PR**
```
        39715338 function calls (39575343 primitive calls) in 57.925 seconds

   Ordered by: internal time
   List reduced from 9663 to 50 due to restriction <50>

     1304   19.129    0.015   19.129    0.015 {method 'acquire' of '_thread.lock' objects}
  496/478    9.885    0.020    9.885    0.021 {built-in method numpy.array}
     8375    3.991    0.000    3.992    0.000 {built-in method builtins.sorted}
 16716568    3.854    0.000    3.854    0.000 {method 'search' of 're.Pattern' objects}
        6    3.157    0.526   29.592    4.932 profile_builder.py:256(get_base_props_and_clean_null_params)
        2    3.084    1.542    3.524    1.762 utils.py:73(shuffle_in_chunks)
        6    2.437    0.406    8.223    1.371 {pandas._libs.lib.map_infer_mask}
```


dataprofiler/profilers/float_column_profile.py
```diff
@@ -28,7 +30,22 @@ def __init__(self, name, options=None):
...
         NumericStatsMixin.__init__(self, options)
         BaseColumnPrimitiveTypeProfiler.__init__(self, name)
-        self.precision = 0
+
+        self.precision = {
+            'min': None,
+            'max': None,
+            'mean': None,
+            'var': None,
+            'std': None,
```
JGSweets, line 39: are both these needed to calculate at run time? i would assume std is only needed at profile along with margin_of_error, hence I'm not sure they need storage so much as calc when needed. Could save some speed, but maybe not much.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -28,7 +30,22 @@ def __init__(self, name, options=None):
...
+            'sum': None,
+            'sample_size': None,
+            'margin_of_error': None,
+            'confidence_level': 0.999
+        }
+        
+        # https://www.calculator.net/confidence-interval-calculator.html
+        self.z_value_precision = 3.291
```
JGSweets, line 47: maybe make it private? not sure user needs this

dataprofiler/profilers/float_column_profile.py
```diff
@@ -106,15 +155,35 @@ def _get_float_precision(cls, df_series_clean):
...
+            'max': float(len_per_float.max()),
+            'mean': precision_sum / sample_size,
+            'var': float(len_per_float.var()),
+            'sum': precision_sum,
+            'sample_size': sample_size
+        }
+        
+        return subset_precision
 
+
```
JGSweets, line 186: remove space

dataprofiler/profilers/float_column_profile.py
```diff
@@ -148,12 +217,56 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
+        if self.precision['max'] is None:
+            self.precision['max'] = subset_precision['max']
+        else:
+            self.precision['max'] = max(
+                self.precision['max'], subset_precision['max'])
+            
+        if self.precision['sum'] is None:
+            self.precision['sum'] = subset_precision['sum']
+        else:
+            self.precision['sum'] = self.precision['sum'] + subset_precision['sum']
```
JGSweets, line 241: cloud just do the `+=`

dataprofiler/profilers/float_column_profile.py
```diff
@@ -148,12 +217,56 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
+        # (min, max, var, sum, sample_size)
         subset_precision = self._get_float_precision(df_series)
-        if self.precision and subset_precision:
-            self.precision = min(self.precision, subset_precision)
-        elif subset_precision:
-            self.precision = subset_precision
+        if subset_precision is None: return
+        
+        # Update values
+        if self.precision['min'] is None:
```
JGSweets, line 226: if any one of these values are `None` for precision, can we not assume the rest are also `None`?
e.g.
```python
if self.precision['min'] is None:
    self.precision = subset_precision
    return
self.precision['min']  = min(self.precision['min'], subset_precision['min'])
self.precision['max']  = min(self.precision['max'], subset_precision['max'])
self.precision['sum']  += subset_precision['sum'])
self.precision['var'] = self._merge_variance(...)
.
.
.
```

JGSweets: similarly, this code and the code above in the `__add__` are near identical

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -603,9 +622,8 @@ def test_profile(self):
             quantiles = profile.pop('quantiles')
             expected_quantiles = expected_profile.pop('quantiles')
             actual_quartiles = {0: quantiles[249], 1: quantiles[499], 2: quantiles[749]}
-
-            self.assertDictEqual(expected_profile, profile)
-            self.assertEqual(expected_profile['precision'], 1.0)
```
JGSweets, line 608: good fix

JGSweets (changes requested):
We check min / max, but there's only 1 test on the rest of the properties for precision. Additionally, it's a small test which only tests against 3 numbers. I think it should be extended such that we have a 3 tests with least 10 numbers:
1. Constant value
2. Constant Step: [1, 2, 3.3, .44, 555, 6.66, 7.777e-1, 8888e10, +99.999, -10000]
3. random Sampling of numbers with precision: [ + 9, -.3, -1e-3, 3.2343, 1230, 0.33, 4.3, 302.1, -4.322, 0]. (I randomly polled precision between length 1,6 for 10 numbers using numpy).

JGSweets:
With the above tests, we should have the expected value out from `_get_float_precision`

dataprofiler/profilers/float_column_profile.py
```diff
@@ -28,7 +30,26 @@ def __init__(self, name, options=None):
...
+        self.precision = {
+            'min': None,
+            'max': None,
+            'mean': None,
+            'var': None,
+            'std': None,
+            'sum': None,
+            'sample_size': None,
+            'margin_of_error': None,
+            'confidence_level': 0.999
```
AnhTruong, line 43: usually, the confidence level is 95% or 99%. Can we just set either of these as the default?

dataprofiler/profilers/float_column_profile.py
```diff
@@ -57,7 +78,39 @@ def __add__(self, other):
...
+            if self.precision['min'] is None:
+                merged_profile.precision = copy.deepcopy(other.precision)
+            elif other.precision['min'] is None:
+                merged_profile.precision = copy.deepcopy(self.precision)                
+            else:
+                merged_profile.precision['min'] = min(
+                    self.precision['min'], other.precision['min'])
+                merged_profile.precision['max'] = max(
+                    self.precision['max'], other.precision['max'])
+                merged_profile.precision['sum'] = \
```
AnhTruong, line 91: do we really need sum?

AnhTruong (commented):
some small comments

dataprofiler/profilers/profiler_options.py
```diff
@@ -416,8 +421,17 @@ def _validate_helper(self, variable_path='FloatOptions'):
...
             errors.append("{}.precision must be a BooleanOption."
+                          .format(variable_path))        
+        errors += self.precision._validate_helper(
+            variable_path + '.precision')
+        
+        
+        if self.precision_sample_ratio is not None \
+           and isinstance(self.precision_sample_ratio, float) \
+           and self.precision_sample_ratio < 0 and self.precision_sample_ratio > 1.0:
+            errors.append("{}.precision_sample_ratio must be a float between 0 and 1."
```
ChrisWallace2020, line 432: Should have error get raised if precision sample ratio isn't float at all (what if they passed a string or boolean)? In that case this if condition won't pick it up and add the error. You could do this via 
`if <not None> and (<not float> or <not btwn 0,1>)`

dataprofiler/profilers/float_column_profile.py
ChrisWallace2020: Although it isn't a useful metric it has to get stored somewhere for the sake of calculating mean.

dataprofiler/profilers/float_column_profile.py
ChrisWallace2020: I agree, in `__add__` if you use just a check of `['min']` being none to determine whether or not precision has been calculated it feels fair to do the same here.

dataprofiler/profilers/float_column_profile.py
lettergram: Yes! Wont need it returned, but need it to calculated mean & variance between chunks.

dataprofiler/profilers/float_column_profile.py
lettergram: I think 99.9% is fine given this is an application that would probably want higher accuracy

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -123,13 +140,16 @@ def test_profiled_precision(self):
...
+        
         for sample in samples:
             df_series = pd.Series([sample[0]])
             min_expected_precision = sample[1]
             precision = FloatColumn._get_float_precision(df_series)
             self.assertEqual(min_expected_precision, precision['min'],
                              msg='Errored for: {}'.format(sample[0]))
 
+        
+
```
ChrisWallace2020, line 149: Should only be 2 blank lines between functions

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -109,6 +109,23 @@ def test_profiled_precision(self):
...
+        print(float_profiler.precision)
+        print()
+        self.assertEqual(3, float_profiler.precision['min'])
+        self.assertEqual(3, float_profiler.precision['max'])
+        self.assertEqual(3, float_profiler.precision['mean'])
+        self.assertEqual(10, float_profiler.precision['sample_size'])
+        self.assertEqual(0, float_profiler.precision['var'])
+        self.assertEqual(0, float_profiler.precision['std'])
+
+
```
ChrisWallace2020: Should only have 1 blank line between logic within a function

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -57,52 +57,75 @@ def test_profiled_precision(self):
...
+                                 '+.342', '-123e1', '23.1'])
+        float_profiler = FloatColumn("Name")
+        float_profiler.update(df_constant)
+        self.assertEqual(3, float_profiler.precision['min'])
+        self.assertEqual(3, float_profiler.precision['max'])
+        self.assertEqual(3, float_profiler.precision['mean'])
+        self.assertEqual(10, float_profiler.precision['sample_size'])
+        self.assertEqual(0, float_profiler.precision['var'])
+        self.assertEqual(0, float_profiler.precision['std'])
+
```
ChrisWallace2020, line 124: One too many blank lines

dataprofiler/profilers/float_column_profile.py
```diff
@@ -148,12 +233,46 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
+        self.precision['std'] = math.sqrt(self.precision['var'])
+
+        # Margin of error, 99.9% confidence level
+        self.precision['margin_of_error'] = round(
+            self.__z_value_precision * (
+                self.precision['std'] / math.sqrt(self.precision['sample_size'])
+            ), int(self.precision['max'])
+        )            
+        
+
```
ChrisWallace2020, line 275: This should be 1 blank line

dataprofiler/tests/profilers/test_float_column_profile.py
ChrisWallace2020, line 124: 1 too many blank lines

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -114,13 +137,16 @@ def test_profiled_precision(self):
...
             df_series = pd.Series([sample[0]])
-            expected_precision = sample[1]
+            min_expected_precision = sample[1]
             precision = FloatColumn._get_float_precision(df_series)
-            self.assertEqual(expected_precision, precision,
+            self.assertEqual(min_expected_precision, precision['min'],
                              msg='Errored for: {}'.format(sample[0]))
 
+        
+
```
ChrisWallace2020, line 149: Don't need these blank lines

dataprofiler/profilers/float_column_profile.py
```diff
@@ -262,15 +266,21 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
                 self.precision['std'] / math.sqrt(self.precision['sample_size'])
             ), int(self.precision['max'])
-        )            
+        )
+
+        # Set the significant figures
+        sigfigs = int(self.precision['max'])
+        for key in ['mean', 'var', 'std', 'margin_of_error']:
+            self.precision[key] = \
+                float('{:.{p}g}'.format(self.precision[key], p=sigfigs))
```
JGSweets, line 286: I'm not sure why, but this incorrectly rounds: `float('{:.{p}g}'.format(.15, p=1))` out: `0.1` expected: `0.2`
what's weird is `float('{:.{p}g}'.format(15, p=1))` is correct and becomes `20`. Not sure diff

dataprofiler/profilers/profiler_options.py
```diff
@@ -376,6 +376,43 @@ def _validate_helper(self, variable_path='IntOptions'):
         return super()._validate_helper(variable_path) 
 
 
+class PrecisionOptions(BooleanOption):
+    
+    def __init__(self, is_enabled=True, sample_ratio=None):
```
JGSweets, line 381: should this instead have the default instead of `None`?

dataprofiler/profilers/float_column_profile.py
```diff
@@ -57,7 +78,39 @@ def __add__(self, other):
...
+            
+                merged_profile.precision['std'] = math.sqrt(
+                    merged_profile.precision['var'])
+
+                # Margin of error, 99.9% confidence level
+                merged_profile.precision['margin_of_error'] = \
+                    merged_profile.__z_value_precision * (
+                        merged_profile.precision['std']
+                        / math.sqrt(merged_profile.precision['sample_size'])
+                )
```
JGSweets, line 113: in the `_update` sig figs are rounded but it doesn't look like it is done here as well.

dataprofiler/profilers/profiler_options.py
```diff
@@ -376,6 +376,44 @@ def _validate_helper(self, variable_path='IntOptions'):
...
+        """
+        Options for precision
+
+        :ivar is_enabled: boolean option to enable/disable the column.
+        :vartype is_enabled: bool
+        :ivar precision: boolean option to enable/disable precision
+        :vartype precision: BooleanOption
```
JGSweets, line 388: remove

dataprofiler/profilers/float_column_profile.py
```diff
@@ -96,25 +164,49 @@ def data_type_ratio(self):
...
+            to_replace=r, value='').map(len)
+
+        # Determine statistics precision
+        precision_sum = float(len_per_float.sum())
+        subset_precision = {
+            'min': float(len_per_float.min()),
+            'max': float(len_per_float.max()),
+            'mean': precision_sum / sample_size,
+            'var': float(len_per_float.var()),
+            'sum': precision_sum,
```
AnhTruong, line 204: why don't use `float(len_per_float.sum())` to be consistent with other above calculations?

AnhTruong: oh, I see it's used for mean. Nvm.

dataprofiler/tests/profilers/profiler_options/test_precision_options.py
```diff
@@ -0,0 +1,31 @@
...
+            {"is_enabled": False, "sample_ratio": None}, option.properties)
+        option = self.get_options(is_enabled=False, sample_ratio=0.5)
+        self.assertDictEqual(
+            {"is_enabled": False, "sample_ratio": 0.5}, option.properties)
+    
+    def test_set_helper(self):
+        super().test_set_helper()
+
+    def test_set(self):
+        super().test_set()
```
JGSweets, line 25: this needs tests for setting `sampling_ratio`

dataprofiler/tests/profilers/profiler_options/test_precision_options.py
```diff
@@ -0,0 +1,31 @@
...
+    
+    def test_validate_helper(self):
+        super().test_validate_helper()
+    
+    def test_validate(self):
```
JGSweets, line 30: needs to test the validate features added

JGSweets (changes requested):
Missing tests, some doc issues, and rounding done in only 1 location

dataprofiler/profilers/float_column_profile.py
```diff
@@ -148,12 +240,51 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
+            
+            self.precision['sample_size'] += subset_precision['sample_size']            
+            self.precision['mean'] = self.precision['sum'] \
+                / self.precision['sample_size']
+
+        # Calculated outside
+        self.precision['std'] = math.sqrt(self.precision['var'])
+
+        # Margin of error, 99.9% confidence level
+        self.precision['margin_of_error'] = round(
```
AnhTruong, line 276: do we really need to round the margin of error?

dataprofiler/profilers/profiler_options.py
lettergram: None is the correct default - see the rest of the section it is used.

Basically, if none the default mechanism is used max(20k, len(data)*0.05). If it's set by the user it'll use exactly what the user defines: len(data)*0.05

dataprofiler/profilers/float_column_profile.py
lettergram: There are no good rounding functions...

```
>>> float('{:.{p}g}'.format(.15, p=1))
0.1
>>> float('{:.{p}g}'.format(.155, p=1))
0.2
>>> float('{:.{p}g}'.format(15, p=1))
20.0
>>> float('{:.{p}g}'.format(25, p=1))
20.0
>>> float('{:.{p}g}'.format(35, p=1))
40.0
```

Attempted an alternative approach:
```
from math import log10, floor
>>> def round_sig(x, sig=99):
...     return round(x, sig-int(floor(log10(abs(x))))-1)
... 
>>> round_sig(0.0232)
0.0232
>>> round_sig(0.0232, sig=1)
0.02
>>> round_sig(0.15, sig=1)
0.1
>>> round_sig(0.155, sig=1)
0.2
>>> round_sig(0.151, sig=1)
0.2
>>> round_sig(15, sig=1)
20
>>> round_sig(35, sig=1)
40
``

Comment (lettergram) [reply dataprofiler/profilers/float_column_profile.py:286]: Think this is the issue, just `surprised` - https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html 

In the scheme of things I believe this is a minimal issue. Numbers under zero probably aren't going to happen (where the issue primarily arises) and often we wont see 15 sigfigs. 5 rounds appropriately via their current method(s).

File: dataprofiler/profilers/float_column_profile.py (hunk 17 above)
Comment (lettergram) [reply dataprofiler/profilers/float_column_profile.py:113]: Removed the round in the other location - shouldn't be rounding pre-output.

File: dataprofiler/profilers/profiler_options.py [hunk 23]
```diff
@@ -376,6 +376,44 @@ def _validate_helper(self, variable_path='IntOptions'):
         return super()._validate_helper(variable_path) 
 
 
+class PrecisionOptions(BooleanOption):
```
ChrisWallace2020, line 379: Not saying this is a problem, but just something to note, in `profiler_options`, some of them are plural (e.g. NumericalOptions) and some are singular (e.g. BooleanOption). It was my understanding that the singular ones were for smaller scale options that exist in other options (boolean, histogram, etc) and plural was for data type columns (float, int, text, etc). Whether or not that is correct, this may be something we should address as given how complicated the options are already, this naming inconsistency doesn't help either. Does anyone have any insight or thoughts on this to explain, or is it an oversight that I am applying unintended meaning to?

dataprofiler/profilers/profiler_options.py
```diff
@@ -376,6 +376,42 @@ def _validate_helper(self, variable_path='IntOptions'):
...
```
ChrisWallace2020, line 379: Not for this PR, but we may want to look into consolidating naming in `profiler_options` currently base, boolean, and histogram are in singular form (i.e. BooleanOption) while all others are plural (FloatOptions). This isn't a super big deal but just something to note.

dataprofiler/profilers/profiler_options.py
JGSweets: I see, so sample_ratio doesn't have a cap

dataprofiler/profilers/float_column_profile.py
```diff
@@ -71,6 +124,12 @@ def profile(self):
         if self.histogram_selection is not None:
             histogram_method = self.histogram_selection
 
+        protected_precision = {}
+        for key in ['min', 'max']:
+            protected_precision[key] = None
+        if self.precision[key] is not None:
+            protected_precision[key] = int(self.precision[key])
```
ChrisWallace2020, line 131: This needs to be nested, right?

ChrisWallace2020: Also what exactly is the purpose of this dictionary?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -603,9 +657,8 @@ def test_profile(self):
             quantiles = profile.pop('quantiles')
             expected_quantiles = expected_profile.pop('quantiles')
             actual_quartiles = {0: quantiles[249], 1: quantiles[499], 2: quantiles[749]}
-
-            self.assertDictEqual(expected_profile, profile)
-            self.assertEqual(expected_profile['precision'], 1.0)
+            self.assertCountEqual(expected_profile, profile)
+            self.assertCountEqual(expected_profile['precision'], profile['precision'])
```
JGSweets, line 661: I don't believe we actually need the second assert, unless we want to be specific, but it needs to be `assertDictEqual`. This test should fail bc the current code has an error, but bc this is `assertCountEqual` it is passing.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -603,9 +657,8 @@ def test_profile(self):
...
             self.assertCountEqual(expected_histogram['bin_counts'],
                                   histogram['bin_counts'])
             self.assertCountEqual(np.round(expected_histogram['bin_edges'], 12),
```
JGSweets, line 664: these technically should be:
```python
self.assertDictEqual(expected_profile['precision'], profile['precision'])
            self.assertTrue(np.all(expected_histogram['bin_counts'] ==
                                  histogram['bin_counts']))
            self.assertTrue(np.all(np.round(expected_histogram['bin_edges'], 12) ==
                                  np.round(histogram['bin_edges'], 12)))

            self.assertDictEqual(actual_quartiles, expected_quantiles)
```

but i can fix these in a subsequent PR

dataprofiler/profilers/float_column_profile.py
JGSweets: this should cause tests to fail, but the tests aren't actually checking this. I referenced below

JGSweets (changes requested):
Error in code, but tests passing because they don't check for the error

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: Also, `assertDictEqual` is necessary bc it checks all the values in the profile that this test is no longer checking

dataprofiler/profilers/profiler_options.py
```diff
@@ -376,6 +376,45 @@ def _validate_helper(self, variable_path='IntOptions'):
...
+        :param variable_path: current path to variable set.
+        :type variable_path: str
+        :return: list of errors (if raise_error is false)
+        :rtype: List of strings
+        """
+        errors = super()._validate_helper(variable_path=variable_path)    
+        if self.sample_ratio is not None:
+            if not isinstance(self.sample_ratio, float) \
+               and not isinstance(self.sample_ratio, int):
+                errors.append("{}.sample_ratio must be a float and cannot be a string."
```
ChrisWallace2020, line 408: I don't think there is any point specifying that it "cannot be a string", just saying it must be a float seems sufficient

dataprofiler/tests/profilers/test_float_column_profile.py
lettergram: Assertcountequal checks all the values, but doesn't care about order

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -628,6 +706,7 @@ def test_option_timing(self):
         options.set({"min.is_enabled": False})
 
         profiler = FloatColumn(df.name, options=options)
+
```
ChrisWallace2020, line 709: Shouldn't have 2 blank lines within function

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -803,7 +882,21 @@ def test_profile_merge_with_different_options(self):
         # Assert that these features are not calculated
         self.assertIsNone(profiler3.max)
         self.assertIsNone(profiler3.min)
-        self.assertEqual(0, profiler3.precision)
+        self.assertEqual(None, profiler3.precision['min'])
+        self.assertEqual(None, profiler3.precision['max'])
+
+
```
ChrisWallace2020, line 888: Shouldn't have 2 blank lines within function

dataprofiler/profilers/profiler_options.py
lettergram: The validation function checks its between 0 and 1, but by default it'll use `max(20k, len(data)*0.05)`

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: disregard this, countequal actually works and prevents disagreements between int/float which is very interesting.

JGSweets (dismissed):

dataprofiler/profilers/profile_builder.py
```diff
@@ -321,20 +322,17 @@ def get_base_props_and_clean_null_params(self, df_series, sample_size,
                     # need to make new var
                     row_value = str(df_series_subset[row])
                     na_columns.setdefault(row_value, list()).append(row)
-                    
-            # Drop the values that matched regex_na
-            df_series_subset = df_series_subset[~matching_na_elements]
-            
-            true_sample_list += df_series_subset.index.tolist()
+                else:
+                    true_sample_list.add(row)
```
lettergram, line 326: This change shaves off ~5% overall runtime:

**Current** (in master):
```
         39712807 function calls (39573098 primitive calls) in 58.290 seconds

   Ordered by: internal time
   List reduced from 9663 to 50 due to restriction <50>

     1304   19.516    0.015   19.516    0.015 {method 'acquire' of '_thread.lock' objects}
  496/478    9.761    0.020    9.761    0.020 {built-in method numpy.array}
 16716568    3.961    0.000    3.961    0.000 {method 'search' of 're.Pattern' objects}
     8375    3.960    0.000    3.961    0.000 {built-in method builtins.sorted}
        6    3.231    0.538   29.630    4.938 profile_builder.py:254(get_base_props_and_clean_null_params)
```

**New in PR**
```
         56424765 function calls (56285092 primitive calls) in 55.109 seconds

   Ordered by: internal time
   List reduced from 9607 to 50 due to restriction <50>

     1304   19.222    0.015   19.222    0.015 {method 'acquire' of '_thread.lock' objects}
  412/406    6.258    0.015    6.258    0.015 {built-in method numpy.array}
        6    4.642    0.774   26.368    4.395 profile_builder.py:254(get_base_props_and_clean_null_params)
```

dataprofiler/profilers/profile_builder.py
```diff
@@ -334,6 +332,7 @@ def get_base_props_and_clean_null_params(self, df_series, sample_size,
         if sample_ids is None:
             sample_ind_generator.close()
 
+        # iloc should work here, there appears to be a bug
```
ChrisWallace2020, line 335: What is the bug exactly?

dataprofiler/profilers/float_column_profile.py
```diff
@@ -71,6 +124,12 @@ def profile(self):
         if self.histogram_selection is not None:
             histogram_method = self.histogram_selection
 
+        protected_precision = {}
+        for key in ['min', 'max']:
+            protected_precision[key] = None
+            if self.precision[key] is not None:
+                protected_precision[key] = int(self.precision[key])
```
ChrisWallace2020, line 131: So what exactly is the point of this? If the max and min are calculated by taking the min and max over a list of string lengths, isn't it already guaranteed that they'll be integers?

lettergram: They're all of float type

dataprofiler/profilers/profile_builder.py
lettergram: I can't use iloc, but should be able to

ChrisWallace2020 (approved):

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -620,14 +673,39 @@ def test_profile(self):
...
+            
     def test_option_timing(self):
         data = [2.0, 12.5, 'not a float', 6.0, 'not a float']
         df = pd.Series(data).apply(str)
 
         options = FloatOptions()
         options.set({"min.is_enabled": False})
 
-        profiler = FloatColumn(df.name, options=options)
+        profiler = FloatColumn(df.name, options=options)
```
JGSweets, line 708: spaces

JGSweets (approved):

ChrisWallace2020 (approved):

Issue #128, opened 2021-04-02
Save & Load profiles from disk


lettergram:
**Is your feature request related to a problem? Please describe.**

I'd like to be able to do: `priofile.save(filename=<optional>)` and `Profiler.load(filename=<profile_filename>)`, after which I should be able to do `profile.update(data)` or `profile.report()`.

This would be an amazing feature as it would enable distributing profiling generating and merging.

PR #154, opened 2021-04-14
Save and Load from Disk

sagars729:
- First attempt at saving and loading from disk using pickle
- Added methods for saving and loading to a specific filepath with "profile.pkl" as default file path
- Added methods for deleting and restoring DataLabeler models
- https://github.com/capitalone/DataProfiler/issues/128


dataprofiler/profilers/profile_builder.py
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
             
         return profile
 
+    def _delete_data_labelers(self):
```
sagars729, line 690: Is there a better way to delete and restore all data labelers that have been created?

dataprofiler/profilers/profile_builder.py
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
...
+                "total_samples": self.total_samples,
+                "encoding": self.encoding,
+                "file_type": self.file_type,
+                "row_has_null_count": self.row_has_null_count,
+                "row_is_null_count": self.row_is_null_count,
+                "hashed_row_dict": self.hashed_row_dict,
+                "_samples_per_update": self._samples_per_update,
+                "_min_true_samples": self._min_true_samples,
+                "options": self.options,
+                "_profile": self.profile
```
sagars729, line 779: The options classes and numpy arrays are two of obstacles in saving the entire profile as a JSON. Any recommendations on how I might be able to save (and later load) these two objects to JSON?

dataprofiler/profilers/profile_builder.py
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
...
+        # Restore data labeler for profiler
+        data_labeler_options = self.options.structured_options.data_labeler
+        if data_labeler_options.is_enabled \
+                and data_labeler_options.data_labeler_object is None:
+            try:
+                data_labeler = DataLabeler(
+                    labeler_type='structured',
+                    dirpath=data_labeler_options.data_labeler_dirpath,
+                    load_options=None)
+                self.options.set({'data_labeler.data_labeler_object': data_labeler})
```
lettergram, line 726: Isn't there a name or something we have to load these (aka we can match with)? @JGSweets

dataprofiler/profilers/profile_builder.py
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
...
+                    if val.options:
+                        data_labeler_dirpath = val.options.data_labeler.data_labeler_dirpath
+                
+                    data_labeler_profile.data_labeler = DataLabeler(
+                        labeler_type='structured',
+                        dirpath=data_labeler_dirpath,
+                        load_options=None)
+
+
+    def save(self, filepath="profile.pkl"):
```
JGSweets, line 757: based on how the DataLabeler components work, I would expect the api mechanism for loading/saving a profile to be using the class itself, i.e.:
```python
profiler = dp.Profiler(...)
profiler.save(filepath)
.
.
.
# now need to load 
profiler = dp.Profiler.load(filepath)

File: dataprofiler/profilers/profile_builder.py (hunk 2 above)
Comment (lettergram) [reply dataprofiler/profilers/profile_builder.py:779]: Only option for saving as a JSON is really writing them as lists or dicts to the profile.

File: dataprofiler/profilers/profile_builder.py [hunk 5]
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
...
 
+    def _delete_data_labelers(self):
+        """
+        Helper method for deleting all data labelers before saving to disk.
+        """
+        # Delete data labeler for profiler
+        data_labeler_options = self.options.structured_options.data_labeler
+        if data_labeler_options.is_enabled \
+                and data_labeler_options.data_labeler_object is not None:
+                    data_labeler_options.data_labeler_object = None
```
lettergram, line 698: I know I had mentioned something like this, but does it work if we leave the data_labeler as well?

Also I don't believe this function in the current state actually deletes the labeler, you'd have to actually call the `del data_labaler` for it to get cleaned up.

dataprofiler/profilers/profile_builder.py
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
...
+            val = self._profile[key]
+            use_data_labeler = True
+            if val.options and isinstance(val.options, StructuredOptions):
+                use_data_labeler = val.options.data_labeler.is_enabled
+
+            if use_data_labeler:
+                val.profiles['data_label_profile']._profiles['data_labeler'] \
+                   .data_labeler = None
+
+    def _restore_data_labelers(self):
```
JGSweets, line 712: @lettergram can correctly if i'm wrong, but for now, we assume that each column would have the same DL. Hence, if we need the DL to be loaded, we should load it just once.

dataprofiler/profilers/profile_builder.py
AnhTruong: we shouldn't delete `data_labeler_object ` when saving the profile, as we might need it after saving.

dataprofiler/profilers/profile_builder.py
AnhTruong: We should load data labeler once for all columns. IMO we should rename this function to load_data_labelers or something, as we probably don't need the _delete_data_labeler function as per my above comment.

dataprofiler/profilers/profile_builder.py
AnhTruong: agree

dataprofiler/profilers/profile_builder.py
```diff
@@ -683,3 +687,127 @@ def tqdm(l):
...
+                "options": self.options,
+                "_profile": self.profile
+               } 
+
+        # Pickle and save profile to disk
+        with open(filepath, "wb") as outfile:
+            pickle.dump(data, outfile)
+
+        # Restore all data labelers
+        self._restore_data_labelers()
```
AnhTruong, line 787: maybe I misunderstand your code, but why we need to delete labeler first, and then restore it at the end when we save the profile.

AnhTruong (commented):
some comments

dataprofiler/profilers/profile_builder.py
sagars729: The Data Labeler objects were deleted because they can't be pickled. That is why I added both a delete and restore function. The restore function is called after saving as well so the data_labeler_object can still be used after saving the profile.

dataprofiler/profilers/profile_builder.py
lettergram: That's a fair observation - we should definitely avoid deleting!

dataprofiler/profilers/profile_builder.py
AnhTruong: interesting. Then, we should only save other components except data labeler (the data labeler by default is just to predict, not supposed to be updated, so the resource folder should be good to load it when needed). We should somehow bundle with the resource containing data labeler to a zip file or something.

Then, when we load the profiler, we load all other components, along with the data labeler loaded from the resource in that zip file.
Any thought?

AnhTruong: @lettergram @JGSweets

dataprofiler/profilers/profile_builder.py
JGSweets: Options at some point should get serializable capability would be extremely helpful

dataprofiler/profilers/profile_builder.py
JGSweets: I'm not sure what you mean. 

in the old version before we updated, it used to be `name` was associated to the labeler in our our repo, but that was since removed I believe. Best we can do now is serialize the DL being used (recognize if lib DL and store / reload it with that knowledge)

dataprofiler/profilers/profile_builder.py
JGSweets: restore seems reasonable if it is just swapping them and not reloading the actual DL.

dataprofiler/profilers/profile_builder.py
JGSweets: I don't think a labeler should be bundled with a profile as that would be a lot of memory. We should just save references to where that labeler exists.

dataprofiler/profilers/profile_builder.py
JGSweets: actually, we can load the processors/models that way. And if we send in the info to the DL it should be able to be loaded with just names if it is within the lib

JGSweets: `BaseDataLabeler.load_from_library(name=...)`

JGSweets: So we can save the DL name if it is from our own lib. And on reload use this.

dataprofiler/profilers/profile_builder.py
AnhTruong: That's kind of my thought process. Saving the reference to data labeler is sufficient if we work on the same machine. If we copy the saved profiler to another machine, I think we need to bundle all components, but I'm not sure we should tackle that usage for now.

JGSweets: you only need bundled components if it isn't part of the library itself though

JGSweets: Also, we might want to save profiler versions with the save

gme5078:
Can you add some tests?

dataprofiler/profilers/profile_builder.py
sagars729: That makes sense, I will update the load to make it a class method

sagars729:
> Can you add some tests?

Yes, that's a good idea - I will add some tests to make sure this works correctly

dataprofiler/profilers/profile_builder.py
sagars729: @AnhTruong As far as saving the references, the way I am "deleting" and "restoring" is just to set the references to None and then use the file paths and reload them based on whether those options were set or not. Is the current solution sufficient?

dataprofiler/profilers/profile_builder.py
sagars729: @AnhTruong @JGSweets do all columns share the same Data Labeler? I wasn't sure which is why I took two steps here - one for the profile, one for all the columns. Because I am setting the pointer to None, I would still need to do this (set all pointers to None), but if I use `del`, I could probably circumvent this. I would only want to make this assumption if all columns and the profiler always share the same Data Labeler object (otherwise the save would break when pickling if multiple data labelers were used)

dataprofiler/profilers/profile_builder.py
sagars729: Changed this in new commit, but I did have to start by making an empty profiler object:
`profile = Profiler(pd.DataFrame([]))`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -568,5 +568,36 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
...
+            save_profile = dp.Profiler(data)
+            
+            # Save and Load profile
+            save_profile.save()
+            load_profile = dp.Profiler.load()
+
+            # Check that reports are equivalent
+            save_report = _clean_report(save_profile.report())
+            load_report = _clean_report(load_profile.report())
+            self.assertDictEqual(save_report, load_report)
```
sagars729, line 600: @gme5078 This is probably the cleanest way to test that the profile saved and loaded; do you think there should be more coverage?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -597,5 +597,36 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
         self.assertEqual(col_one_len, len(data['NAME']))
         self.assertEqual(col_two_len, len(data[' VALUE']))
 
+    def test_save_and_load(self):
+        datapth = "dataprofiler/tests/data/"
```
JGSweets, line 601: when we unit test for save, we shouldn't actually create a file or load a file. We ideally mock the actual IO.
On save, we validate what would be sent to save is what we would expect. On load, we intercept the open and give it what it would be expected to open.

I can see load being more of an issue, but at least it is a static file. However, when we test right now, a save would actually create a file which would need cleanup, no?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -597,5 +597,36 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
...
+            # Save and Load profile
+            save_profile.save()
+            load_profile = dp.Profiler.load()
+
+            # Check that reports are equivalent
+            save_report = _clean_report(save_profile.report())
+            load_report = _clean_report(load_profile.report())
+            self.assertDictEqual(save_report, load_report)       
+
 if __name__ == '__main__':
```
JGSweets, line 631: should be two spaces between the class and the if main

dataprofiler/profilers/profile_builder.py
```diff
@@ -19,6 +19,8 @@
 import pandas as pd
 import numpy as np
 
+import pickle
```
JGSweets, line 22: should be grouped with the built-in imports above, can just go on line 16 so no space between warnings and mp

dataprofiler/profilers/profile_builder.py
```diff
@@ -774,3 +776,136 @@ def tqdm(l):
             pool.join() # Wait for all workers to complete
             
         return profile
+
```
JGSweets, line 779: only 1 space between class functions

dataprofiler/tests/profilers/test_profile_builder.py
gme5078: I agree with Jeremy in that the test should have the IO mocked out

dataprofiler/tests/profilers/test_profile_builder.py
sagars729: That makes sense, currently save would create a file which would require cleanup - I'll look into how I can mock the IO

dataprofiler/profilers/profile_builder.py
sagars729: Fixed in new commit

dataprofiler/profilers/profile_builder.py
sagars729: Fixed in new commit

dataprofiler/tests/profilers/test_profile_builder.py
sagars729: Fixed in new commit

dataprofiler/profilers/profile_builder.py
```diff
@@ -825,3 +826,159 @@ def tqdm(l):
...
+                "options": self.options,
+                "_profile": self.profile
+               } 
+
+        # Pickle and save profile to disk
+        with open(filepath, "wb") as outfile:
+            pickle.dump(data, outfile)
+
+        # Restore all data labelers
+        self._restore_data_labelers(data_labelers)
```
sagars729, line 951: @AnhTruong @JGSweets - here is an implementation that "swaps" data labelers rather than deleting and reloading them. Since I already had the code written out for working with multiple data labelers, I left the same logic in place just in case that behavior changes in the future.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -729,6 +730,42 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
         self.assertEqual(col_one_len, len(data['NAME']))
         self.assertEqual(col_two_len, len(data['VALUE']))
 
+    def test_save_and_load(self):
```
JGSweets, line 733: These tests should probably be under profiler builder rather than a specifically for testing nulls

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -729,6 +730,42 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
...
+            save_profile = dp.Profiler(data)
+            
+            # Save and Load profile with Mock IO
+            save_profile.save()
+            load_profile = dp.Profiler.load()
+            #with mock.patch('builtins.open', mock.mock_open()) as m: 
+            #    save_profile.save()
+            #with mock.patch('builtins.open', mock.mock_open(
+            #    read_data = m().write.call_args[0][0])) as m:
+            #    load_profile = dp.Profiler.load()
```
JGSweets, line 762: These mocks should go around the function at the top so they can apply throughout the calls.
Likely you will have to mock the DataLabeler as well as it uses the `open` command too. 

We should avoid creating objects during the tests

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -729,6 +730,42 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
         self.assertEqual(col_one_len, len(data['NAME']))
         self.assertEqual(col_two_len, len(data['VALUE']))
 
+    def test_save_and_load(self):
+        datapth = "dataprofiler/tests/data/"
+        test_files = ["csv/guns.csv", "csv/iris.csv"]
```
JGSweets, line 735: as opposed to loading files, we can just create our own mock data inside the function. Then we no longer are dependent on external files which could disappear or change

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -729,6 +730,42 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
...
+            return report
+
+        for test_file in test_files:
+            # Create Data and Profiler objects
+            data = dp.Data(os.path.join(datapth, test_file))
+            save_profile = dp.Profiler(data)
+            
+            # Save and Load profile with Mock IO
+            save_profile.save()
+            load_profile = dp.Profiler.load()
```
JGSweets, line 757: it might be better to test save and load in their own tests

JGSweets: Also, what is `load_profile = dp.Profiler.load()` calling load on? I would have assumed it would be the path to the profiler to load, but currently it is empty.

dataprofiler/profilers/profile_builder.py
```diff
@@ -825,3 +826,160 @@ def tqdm(l):
...
+
+    @staticmethod
+    def load(filepath="profile.pkl"):
```
JGSweets, line 954: I think the user should be explicit about what they are loading, hence no default value. @lettergram ?

dataprofiler/profilers/profile_builder.py
JGSweets: I think the user should be explicit about what they are loading, hence no default value. @lettergram ?

dataprofiler/tests/profilers/test_profile_builder.py
sagars729: It currently has a default path enabled (as does save). Should I remove the default filepath?

dataprofiler/tests/profilers/test_profile_builder.py
sagars729: Moved to TestProfiler

dataprofiler/tests/profilers/test_profile_builder.py
sagars729: Since the same test data files are being used for other tests, it might be unnecessary to create mock data - it's still a good point though and I could replace these with pandas data frames

dataprofiler/tests/profilers/test_profile_builder.py
sagars729: Fixed in new commit

dataprofiler/profilers/profile_builder.py
sagars729: Changed load to require filepath and save to use "profile-{datetime}.pkl" format

JGSweets (approved):
We should add more tests around this in another PR:
1. validate no DL added if DL turned off
2. Validate that all props are transferred, not just report ==
3. future should load internal DL if known to be internal (issue if someone set object, but not internal)
4. Look into JSON load / save instead of pickle
5. versioning

lettergram (approved):

PR #144, opened 2021-04-08
update get_percentiles to take in an array

JGSweets:
Reduction in histogram time by refactoring get_percentiles:

TLDR;

--profile time--
Old method, optimized old method, new method
all bin methods: 8.37,7.80,6.78
auto only method: 7.18,6.81,6.66

--merge time--
Old method, optimized old method, new method
all bin methods: 6.61,5.17,0.46
auto only method: 1.56,0.84,0.39

===========================================================================

All tests are ran on a 23MB file with 1 col of each type plus a NAN col (index was accidentally included so extra int).
Old time, all binning:
```
profile time: 8.37269401550293
merge time: 6.606444835662842
Wrote profile results to profile.py.lprof
Timer unit: 1e-06 s

Total time: 6.46363 s
data-profiler/dataprofiler/profilers/numerical_column_stats.py
Function: _add_helper_merge_profile_histograms at line 100

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   100                                               @BaseColumnProfiler._timeit(name="histogram_and_quantiles")
   101                                               @profile
   102                                               def _add_helper_merge_profile_histograms(self, other1, other2):
   103                                                   """
   104                                                   Adds histogram of two profiles together
   105
   106                                                   :param other1: profile1 being added to self
   107                                                   :type other1: BaseColumnProfiler
   108                                                   :param other2: profile2 being added to self
   109                                                   :type other2: BaseColumnProfiler
   110                                                   :return: None
   111                                                   """
   112                                                   # get available bin methods and set to current
   113        28         90.0      3.2      0.0          bin_methods = list(set(other1.histogram_bin_method_names) &
   114        14         24.0      1.7      0.0                             set(other2.histogram_bin_method_names))
   115        14          8.0      0.6      0.0          if not bin_methods:
   116                                                       raise ValueError('Profiles have no overlapping bin methods and '
   117                                                                        'therefore cannot be added together.')
   118        14         16.0      1.1      0.0          elif other1.user_set_histogram_bin and other2.user_set_histogram_bin:
   119                                                       if other1.user_set_histogram_bin != other2.user_set_histogram_bin:
   120                                                           warnings.warn('User set histogram bin counts did not match. '
   121                                                                         'Choosing the larger bin count.')
   122                                                       self.user_set_histogram_bin = max(other1.user_set_histogram_bin,
   123                                                                                         other2.user_set_histogram_bin)
   124
   125                                                   # initial creation of the profiler creates all methods, but
[...]

PR #155, opened 2021-04-14
Allow user to set the histogram bin count (moving towards speeding up histograms)

JGSweets:
The intent of this PR is two fold:
1. Add the ability for the user to set the bin count.
2. A stepping stone for https://github.com/capitalone/DataProfiler/pull/144

Once approved / merged, https://github.com/capitalone/DataProfiler/pull/144 will be the next focus which will significantly speed up the percentile portion of the library.


dataprofiler/profilers/profiler_options.py
```diff
@@ -143,16 +143,21 @@ def _validate_helper(self, variable_path='BooleanOption'):
 
 class HistogramOption(BooleanOption):
 
-    def __init__(self, method=['auto']):
+    def __init__(self, is_enabled=True, method=None, hist_bin_count=None):
```
AnhTruong, line 145: i think it should be more clear if method is defaulted as 'auto', as we implied the whole list of methods if it is None.

dataprofiler/profilers/profiler_options.py
```diff
@@ -6,7 +6,6 @@
 import warnings
 import abc
 import copy
-from ..labelers.data_labelers import DataLabeler
```
JGSweets, line 9: unused

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
```diff
@@ -0,0 +1,146 @@
...
+            option._set_helper({'hist_bin_count.is_enabled': True}, 'test')
+
+    def test_set(self):
+        option = self.get_options()
+
+        params_to_check = [
+            dict(prop='is_enabled', value_list=[False, True]),
+            dict(prop='method',
+                 value_list=['auto', 'fd', 'doane', 'scott', 'rice', 'sturges',
+                             'sqrt', ['sturges', 'doane']]),
```
AnhTruong, line 34: can we add None to the list?

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
```diff
@@ -0,0 +1,146 @@
...
+                     "strings from the following: ['auto', 'fd', 'doane', "
+                     "'scott', 'rice', 'sturges', 'sqrt']."
+                 ]),
+            dict(prop='method', value_list=['whoops', ["doane", "incorrect"]],
+                 errors=[
+                     "HistogramOption.method must be a subset or selection "
+                     "from: ['auto', 'fd', 'doane', 'scott', 'rice', 'sturges',"
+                     " 'sqrt']."
+                 ]),
+            dict(prop='hist_bin_count', value_list=[-1, False, [], 1.2, 1.0],
```
AnhTruong, line 92: can we add a case with string, e.g., 'one', or '1'?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -736,13 +736,41 @@ def test_profile_merge_edge_case(self):
...
+
+        data2 = [10.0, 'not a float', 15.0, 'not a float']
+        df2 = pd.Series(data2).apply(str)
+        profiler2 = FloatColumn("Float", options)
+        profiler2.update(df2)
+
+        # no warning should occur
+        with warnings.catch_warnings(record=True) as w:
+            profiler1 + profiler2
+        self.assertListEqual([], w)
```
AnhTruong, line 758: can we add a test to assert `profiler2.user_set_histogram_bin = 10`? and also for profiler1?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -109,7 +115,24 @@ def _add_helper_merge_profile_histograms(self, other1, other2):
...
+        self.histogram_methods = dict()
+        for method in self.histogram_bin_method_names:
+            self.histogram_methods[method] = {
+                'total_loss': 0,
+                'current_loss': 0,
+                'histogram': {
+                    'bin_counts': None,
+                    'bin_edges': None
+                }
+            }
```
ChrisWallace2020, line 135: This could use a comment explaining why this is repeated for the case of 'custom', and also maybe an if statement to ensure it is being done in that case, and signal the intent for that case.

ChrisWallace2020: Actually this still makes sense in the non-'custom' case, but still a comment would be helpful to clarify the need for seemingly-redundant computation.

AnhTruong (commented):
some comments

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -321,11 +346,14 @@ def _get_histogram(self, values, bin_method):
                 unique_value = values.iloc[0]
             bin_edges = np.array([unique_value, unique_value])
         else:
-            values, weights = histogram_utils._ravel_and_check_weights(
-                values, None)
-            _, n_equal_bins = histogram_utils._get_bin_edges(
-                values, bin_method, None, None)
-            n_equal_bins = min(n_equal_bins, self.max_histogram_bin)
+            n_equal_bins = suggested_bin_count = self.user_set_histogram_bin
```
ChrisWallace2020, line 349: There is no reason to define suggested_bin_count here since it is only used within the if statement after a second definition. Also I think the naming between n_equal_bins and suggested_bin_count isn't super clear considering suggested_bin_count seems to just be an intermediary value to ensure n_equal_bins lands between the min and max histogram bins.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -736,13 +736,41 @@ def test_profile_merge_edge_case(self):
...
+
+        # make bin counts different and get warning
+        profiler2.user_set_histogram_bin = 120
+        with self.assertWarnsRegex(UserWarning,
+                                   'User set histogram bin counts did not '
+                                   'match. Choosing the larger bin count.'):
+            merged_profile = profiler1 + profiler2
+        self.assertEqual(120, merged_profile.user_set_histogram_bin)
```
ChrisWallace2020, line 766: Would a user ever do it this way or is the correct usage to set this via options?

dataprofiler/tests/profilers/test_float_column_profile.py
ChrisWallace2020: ^ I agree

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -821,4 +850,28 @@ def test_histogram_option_integration(self):
...
-        self.assertEqual(["sturges", "doane"], num_profiler.histogram_bin_method_names)
+        self.assertEqual(["sturges", "doane"],
+                         num_profiler.histogram_bin_method_names)
+
+        # test hisogram bin count set
+        options.histogram_and_quantiles.method = None
+        options.histogram_and_quantiles.hist_bin_count = 100
+        num_profiler = FloatColumn(name="test3", options=options)
+        self.assertIsNone(num_profiler.histogram_selection)
+        self.assertEqual(['custom'], num_profiler.histogram_bin_method_names)
```
ChrisWallace2020, line 861: Should we make `.histogram_selection = 'custom'` in this case? I'm just thinking for in terms of accessing that dictionary that is indexed via names. I suppose this will eventually get picked by picking the best method out of a list of 1, but currently my implementation for user picking 1 method is it sets `.histogram_selection` in `__init__`. I'm all for changing that to be None to match the custom case, I just think they should be consistent.

dataprofiler/profilers/profiler_options.py
JGSweets: The reason I chose this route is that if someone were to write:
```python
HistogramOptions(hist_bin_count=100)
```
How do you know what to choose? Maybe the user specified `'auto'` as well? 
Maybe they wrote `HistogramOptions(method='auto', hist_bin_count=100)`
Ultimately, we need to protect against both from being set simultaneously.
If we revert it, the user would then have to write:
```python
HistogramOptions(method=None, hist_bin_count=100)
```

Unless we want to always say if `hist_bin_count` is specified to  choose `hist_bin_count`. But to not get a warning, a user will have to write as above.

I'm open to suggestions.

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
JGSweets: adding

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
JGSweets: adding

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: the intent of this function is to test merging when custom bin is set, not individual options.
However, i did add: `self.assertEqual(10, merge_profile.user_set_histogram_bin)`

Additionally, the test to check if options propagate and effect final results occurs in: `test_histogram_option_integration` in the area commented: `# test histogram bin count set`

dataprofiler/profilers/numerical_column_stats.py
JGSweets: This is actually an intermediary step before the next PR which technically exists: https://github.com/capitalone/DataProfiler/pull/144

```python
else:
    n_equal_bins = suggested_bin_count = self.user_set_histogram_bin
    if n_equal_bins is None:
        values, weights = histogram_utils._ravel_and_check_weights(
            values, None)
        _, n_equal_bins = histogram_utils._get_bin_edges(
            values, bin_method, None, None)
        suggested_bin_count = min(n_equal_bins, self.max_histogram_bin)
        n_equal_bins = max(self.min_histogram_bin, suggested_bin_count)
    bin_counts, bin_edges = np.histogram(values, bins=n_equal_bins)
return bin_counts, bin_edges, suggested_bin_count
```
the reason being is if we don't have a user set number of bins. we use the min/max bin count to choose the bin count and the suggested bin count by each method (which has to be less than the max) is passed up to be stored. 

That way at report time, we return the best histogram rather than a really sparse one that may have more bins than necessary. e.g. min bins = 1000, but only 5 numbers. we probably only need 3-5 bins at report otherwise the histogram would be no bueno. So we need to know the suggested to convert from the 1000 -> 3-5 bins

JGSweets: I'm open to naming changes, but maybe we change in next PR since we will be adapting this code anyways and it is not functionally harming anything

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: the intent is not that a user would do this, the intent is to not need to create options every time to get a test to be functional for eval

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: Hmm, thinking about this. I wonder if it should be `None`, bc technically nothing has been computed yet and would also fit what the list is like.

I'm open to change, but rn i'm coding such that default is none till update is run

ChrisWallace2020: Well then do you think we should make the case of the user picking one method match this? Or do you think that isn't as important.

ChrisWallace2020 (approved):

dataprofiler/profilers/profiler_options.py
AnhTruong: Actually, if we combine `method` with `hist_bin_count` to have only something like `bins`, that'd make the most sense. So, it can be str, list(str), int, or None.
https://numpy.org/doc/stable/reference/generated/numpy.histogram.html

AnhTruong:
Currently, the code works as is. But I have a concern that, if we combine `method` with `hist_bin_count` to have only something like `bins`, that'd make the most sense. So, it can be str, list(str), int, or None.
https://numpy.org/doc/stable/reference/generated/numpy.histogram.html

ChrisWallace2020:
^ I think this is a good idea for a future refactor but since there are other PRs predicated on this one I don't think we should expand the scope.

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def test_set(self):
+        option = self.get_options()
+
+        params_to_check = [
+            dict(prop='is_enabled', value_list=[False, True]),
+            dict(prop='bin_count_or_method',
+                 value_list=[None, 'auto', 'fd', 'doane', 'scott', 'rice',
+                             'sturges', 'sqrt', ['sturges', 'doane'], '1',
+                             '10', '100', '1000', '99', '10000000'])
```
ChrisWallace2020, line 35: Should these be ints, not strings

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
```diff
@@ -0,0 +1,125 @@
...
+            dict(prop='is_enabled', value_list=[False, True], errors=[]),
+            dict(prop='bin_count_or_method',
+                 value_list=['auto', 'fd', 'doane', 'scott', 'rice', 'sturges',
+                             'sqrt', ['sturges', 'doane'], 1, 10, 100, 1000,
+                             99, 10000000],
+                 errors=[]),
+
+            # errors
+            dict(prop='bin_count_or_method',
+                 value_list=[-1, 1.2, 1.0, [], False],
```
ChrisWallace2020, line 73: Can we add tests for strings of ints? (`"1", "100"`)

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
JGSweets: fixed

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
```diff
@@ -0,0 +1,125 @@
...
+                     r"HistogramOption.bin_count_or_method must be a subset or "
+                     r"selection from: ['auto', 'fd', 'doane', 'scott', 'rice', "
+                     r"'sturges', 'sqrt']."
+                 ]),
+        ]
+
+        # # this code can be abstracted to limit code everywhere else
+        # # AKA, params_to_check would be the only needed code
+        # def _assert_validate_helper(prop, value):
+        #
```
ChrisWallace2020, line 92: Should this get deleted?

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
JGSweets: added a single test case

dataprofiler/tests/profilers/profiler_options/test_histogram_option.py
JGSweets: no bc the code below could be abstracted, but updated the comment

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -45,25 +45,31 @@ def __init__(self, options=None):
...
+        if options:
+            bin_count_or_method = \
+                options.histogram_and_quantiles.bin_count_or_method
+            if isinstance(bin_count_or_method, str):
+                self.histogram_bin_method_names = [bin_count_or_method]
+            elif isinstance(bin_count_or_method, list):
+                self.histogram_bin_method_names = bin_count_or_method
+            elif isinstance(bin_count_or_method, int):
+                self.user_set_histogram_bin = \
+                    options.histogram_and_quantiles.bin_count_or_method
```
AnhTruong, line 71: can be replaced by `bin_count_or_method ` to be consistent with the above

dataprofiler/profilers/profiler_options.py
```diff
@@ -164,25 +164,26 @@ def _validate_helper(self, variable_path='HistogramOption'):
...
+                pass  # use errors below if not a passing int
+            elif (not isinstance(value, list) or len(value) < 1
+                  or not all([isinstance(item, str) for item in value])):
+                errors.append("{}.bin_count_or_method must be an integer more "
+                              "than 1, a string, or list of strings from the "
+                              "following: {}.".format(variable_path,
+                                                      valid_methods))
+            elif not set(value).issubset(set(valid_methods)):
+                errors.append("{}.bin_count_or_method must be a subset or "
+                              "selection from: {}.".format(variable_path,
```
ChrisWallace2020, line 185: This doesn't account for `bin_count_or_method` being able to be an integer, and my be useful for the user to know. Should the error message be the same for all cases, or should this one just be extended to say "if providing a bin method name, it must be a subset ..."

dataprofiler/profilers/numerical_column_stats.py
JGSweets: fixing

AnhTruong (approved):

ChrisWallace2020 (dismissed):

ChrisWallace2020 (approved):

PR #157, opened 2021-04-16
Mitigated type bug in json output & runtime improvements 

lettergram:
File tested: `United_States_Drought_Monitor__2000-2016.csv` (6 int columns)

Should be directly related to how many columns there are in the dataset, so the more columns there are the more this improvement will be apparent

**Master** (current):
```
         56426689 function calls (56287010 primitive calls) in 58.344 seconds

   Ordered by: internal time
   List reduced from 9647 to 50 due to restriction <50>

     1304   19.401    0.015   19.401    0.015 {method 'acquire' of '_thread.lock' objects}
  466/460    9.783    0.021    9.783    0.021 {built-in method numpy.array}
        6    4.674    0.779   29.565    4.928 profile_builder.py:254(get_base_props_and_clean_null_params)
```

**This PR**:
```
         73147049 function calls (73007354 primitive calls) in 53.684 seconds

   Ordered by: internal time
   List reduced from 9655 to 50 due to restriction <50>

     1304   19.312    0.015   19.312    0.015 {method 'acquire' of '_thread.lock' objects}
        6    6.153    1.026   22.929    3.822 profile_builder.py:254(get_base_props_and_clean_null_params)
```

### DIFF
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -123,27 +123,35 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         histogram_method = self.histogram_bin_method_names[0]
         if self.histogram_selection is not None:
             histogram_method = self.histogram_selection
-            
+
+        def np_type_to_type(val):
+            if isinstance(val, np.integer):
+                return int(val)
+            if isinstance(val, np.float):
+                return float(val)
+            return val
+        
         profile = dict(
-            min=self.min,
-            max=self.max,
-            mean=self.mean,
-            variance=self.variance,
-            stddev=self.stddev,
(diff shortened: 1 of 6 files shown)


dataprofiler/profilers/float_column_profile.py
```diff
@@ -123,27 +123,35 @@ def profile(self):
...
         if self.histogram_selection is not None:
             histogram_method = self.histogram_selection
-            
+
+        def np_type_to_type(val):
+            if isinstance(val, np.integer):
+                return int(val)
+            if isinstance(val, np.float):
+                return float(val)
+            return val
```
ChrisWallace2020, line 132: Should we move this to a util since its used in multiple places

ChrisWallace2020 (approved):

gme5078 (approved):

PR #159, opened 2021-04-19
Improved null finder and Multiprocessing - Oh my! (35% runtime reduction)

lettergram:
Prior regex query was utilizing a grouping mechanism that was slower, this improves that.

Test file: `United_States_Drought_Monitor__2000-2016.csv` (6 integer columns)

Current Master:
```
         73147049 function calls (73007354 primitive calls) in 53.684 seconds

   Ordered by: internal time
   List reduced from 9655 to 50 due to restriction <50>

     1304   19.312    0.015   19.312    0.015 {method 'acquire' of '_thread.lock' objects}
        6    6.153    1.026   22.929    3.822 profile_builder.py:254(get_base_props_and_clean_null_params)
 16716568    3.957    0.000    3.957    0.000 {method 'search' of 're.Pattern' objects}
        2    3.095    1.547    3.543    1.772 utils.py:73(shuffle_in_chunks)
  520/514    2.664    0.005    2.664    0.005 {built-in method numpy.array}
        6    2.465    0.411    8.360    1.393 {pandas._libs.lib.map_infer_mask}
     8370    2.302    0.000    2.304    0.000 {built-in method builtins.sorted}
 16716402    1.938    0.000    5.894    0.000 object_array.py:120(<lambda>)
        6    1.283    0.214    1.285    0.214 {pandas._libs.lib.map_infer}
```

-----------------

Improving only the `get_base_props_and_clean_null_params` gets the following performance improvements

```
         39714085 function calls (39574368 primitive calls) in 47.602 seconds

   Ordered by: internal time
   List reduced from 9656 to 50 due to restriction <50>

     1304   19.476    0.015   19.476    0.015 {method 'acquire' of '_thread.lock' objects}
 16717923    3.612    0.000    3.612    0.000 {method 'match' of 're.Pattern' objects}
  520/490    3.459    0.007    3.459    0.007 {built-in method numpy.array}
        2    3.112    1.556    3.568    1.784 utils.py:73(shuffle_in_chunks)
        6    2.396    0.399    7.935    1.322 {pandas._libs.lib.map_infer_mask}
     8370    2.343    0.000    2.344    0.000 {built-in method builtins.sorted}
 16716402    1.928    0.000    5.538    0.000 object_array.py:216(<lambda>)
      549    1.748    0.003    1.748    0.003 {method 'update' of 'set' objects}
        6    1.238    0.206    1.240    0.207 {pandas._libs.lib.map_infer}
```
-----------------------

Separating `get_base_props_and_clean_null_params` into a static function and processing the columns in a pool provides the following speedup:

```
         6271339 function calls (6131870 primitive calls) in 34.671 seconds

   Ordered by: internal time
   List reduced from 9554 to 50 due to restriction <50>

     1343   22.891    0.017   22.891    0.017 {method 'acquire' of '_thread.lock' objects}
        2    3.081    1.541    3.523    1.761 utils.py:73(shuffle_in_chunks)
     8364    2.293    0.000    2.294    0.000 {built-in method builtins.sorted}
      353    0.687    0.002    0.687    0.002 {built-in method numpy.array}
      500    0.602    0.001    0.602    0.001 {built-in method fromkeys}
        1    0.476    0.476    0.541    0.541 {method 'read' of 'pandas._libs.parsers.TextReader' objects}
        6    0.344    0.057    0.344    0.057 {method 'choice' of 'numpy.random.mtrand.RandomState' objects}
       22    0.289    0.013    0.290    0.013 {pandas._libs.lib.infer_dtype}
        1    0.284    0.284   30.385   30.385 profile_builder.py:618(_update_profile_from_chunk)
```


dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -304,6 +305,31 @@ def __init__(self, label_mapping=None, parameters=None):
         :return: None
         """
 
+        # https://github.com/tensorflow/tensorflow/issues/34697
```
AnhTruong, line 308: did you try another method down below on the same url? It doesn't require saving a file, and we can remove tempfile

lettergram: I think this is appropriate as we may be transferring the model between processes which don't' share a memory space.

dataprofiler/profilers/profile_builder.py
```diff
@@ -639,57 +661,108 @@ def tqdm(l):
...
-            else:
+        # Numpy arrays allocate to heap and can be shared between processes
+        # Non-locking multiprocessing fails on machines without POSIX (windows)
+        # The function handles that situation, but will be single process
+        # Newly introduced features (python3.8) improves the situation
+        sample_ids = np.array(sample_ids)        
+
+        # Create structured profile objects
+        new_cols = set()
+        for col in df.columns:
```
ChrisWallace2020, line 672: Are we not using tqdm anymore? If not we should remove it from requirements and/or the "tqdm" function defined above.

dataprofiler/profilers/profile_builder.py
```diff
@@ -639,57 +661,108 @@ def tqdm(l):
...
                     options=structured_options
                 )
+                new_cols.add(col)
+                        
+        # Format the data
+        notification_str = "Finding the Null values in the columns..."
+        pool = None
+        if options.structured_options.multiprocess.is_enabled:
+            if len(new_cols) > 0:
+                pool, pool_size = utils.generate_pool(max_pool_size=len(new_cols))
```
ChrisWallace2020, line 690: Should we make the max_pool_size `min(len(new_cols), 4/8)` so that we still have the upper bound from before?

dataprofiler/profilers/profile_builder.py
lettergram: We are, but this is just creating them it's very very fast. tqdm is further below, when processing occurs.

dataprofiler/profilers/profile_builder.py
```diff
@@ -639,57 +661,108 @@ def tqdm(l):
...
+
+        # Create structured profile objects
+        new_cols = set()
+        for col in df.columns:
+            if col not in profile:
                 structured_options = None
                 if options and options.structured_options:
                     structured_options = options.structured_options
                 profile[col] = StructuredDataProfile(
-                    df[col],
```
ChrisWallace2020, line 682: Why do we not pass this to the StructuredDataProfile here?

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -304,6 +305,30 @@ def __init__(self, label_mapping=None, parameters=None):
...
+                    fd.flush()
+                model = tf.keras.models.load_model(fd.name)
+                self.__dict__ = model.__dict__
+
+            cls = tf.keras.models.Model
+            cls.__getstate__ = __getstate__
+            cls.__setstate__ = __setstate__
+
+        # Run the function
+        make_keras_picklable()
```
gme5078, line 330: Some questions purely because I'm curious:
Okay so if I understand this correctly, you are overwriting the get and set state for a keras model so that it is in the form of a dictionary so that you can pickle it, but what I don't understand is how does the tempfile work? like how is the tempfile going to be replaced by the actual model? Is that just how the tempfile library works? it is designed to get replaced with whatever is in the keras model?
Secondly, what about models that dont have the .hdf5 format?
Thirdly, are models that have already been created, still going to be pickleable?
Fourthly, is there ever a time when you wouldnt want the model to be pickleable?

gme5078:
So on a high level, why are so many things being moved around? Particularly the profile compilers and min_sample_size?

Some of this is a refactor but some of this is new features so you need tests. Tests I would need to approve this:
Test pickling a model
Apparently the null finder is improved? How so? Just speed or is it something you can test?
Test all the possible notifications in the multiprocessing (also, do these prints interrupt the tqdm?)

dataprofiler/profilers/profile_builder.py
lettergram: Because we don't have the data yet and thus cannot pass it to the profile. I made it so the class could not take in data and later can be updated.

dataprofiler/profilers/profile_builder.py
```diff
@@ -51,56 +52,80 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+        :param clean_sampled_df: sampled series with none types dropped
+        :type clean_sampled_df: Pandas.Series
+        :param pool: pool utilized for multiprocessing
+        :type pool: multiprocessing.pool
+        """
+
+        if self.name is None:
+            self.name = clean_sampled_df.name
+        if self.name != clean_sampled_df.name:
+            raise ValueError('Column names have changed!')
```
lettergram, line 101: Should specify column differences

dataprofiler/profilers/profile_builder.py
```diff
@@ -51,56 +52,80 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+            clean_sampled_df, base_stats  = \
+                self.clean_data_and_get_base_stats(
+                    df_series=df_series, sample_size=sample_size,
+                    min_true_samples=self._min_true_samples, sample_ids=sample_ids)
+            self.update_type_statistics(clean_sampled_df, pool)
+            self._update_base_stats(base_stats)
+
+    def update_type_statistics(self, clean_sampled_df, pool):
+        """
+        Calculates type statistics and labels dataset
```
lettergram, line 91: add newline

dataprofiler/profilers/profile_builder.py
```diff
@@ -51,56 +52,80 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
-
         self.profiles = {
-            'data_type_profile':
-            ColumnPrimitiveTypeProfileCompiler(
-                clean_sampled_df, self.options, pool),
-            'data_stats_profile':
-            ColumnStatsProfileCompiler(
-                clean_sampled_df, self.options, pool)
+            'data_type_profile': None,
+            'data_stats_profile': None
```
gme5078, line 70: I think this should just be None

dataprofiler/profilers/profile_builder.py
```diff
@@ -51,56 +52,80 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+                              "not the whole dataset.".format(sample_size))
+                
+            clean_sampled_df, base_stats  = \
+                self.clean_data_and_get_base_stats(
+                    df_series=df_series, sample_size=sample_size,
+                    min_true_samples=self._min_true_samples, sample_ids=sample_ids)
+            self.update_type_statistics(clean_sampled_df, pool)
+            self._update_base_stats(base_stats)
+
+    def update_type_statistics(self, clean_sampled_df, pool):
```
gme5078, line 89: nitpick: can we call this update profile compilers or update_compilers or something along those lines

dataprofiler/profilers/profile_builder.py
```diff
@@ -639,57 +666,108 @@ def tqdm(l):
...
+                        min_true_samples=min_true_samples, sample_ids=sample_ids)
+                profile[col]._update_base_stats(base_stats)
+            
+        # Process and label the data
+        notification_str = "Calculating the statistics... "
+        pool = None
+        if options.structured_options.multiprocess.is_enabled:
+            pool, pool_size = utils.generate_pool(max_pool_size=4)
+            if pool:
+                notification_str += " (with " + str(pool_size) + " processes)"
```
gme5078, line 764: test all the notifications are working?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -570,7 +587,7 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
             self.assertEqual(0.5, profile._get_row_is_null_ratio())
             self.assertEqual(0.4375, profile._get_unique_row_ratio())
             self.assertEqual(9, profile._get_duplicate_row_count())
-        
+            
         self.assertEqual(col_one_len, len(data['NAME']))
```
gme5078, line 591: test pickling

lettergram: > Pickle files don't have a header, so there's no standard way of identifying them short of trying to unpickle one and seeing if any exceptions are raised while doing so.

https://stackoverflow.com/questions/13939913/how-to-test-if-a-file-has-been-created-by-pickle

lettergram: Removed, will add to a different PR

dataprofiler/profilers/profile_builder.py
lettergram: There's not a good way to do that in this case with multiple different notification types (tqdm, print) and multiple sets of text.

Further, this isn't critical and we weren't testing it before?

JGSweets (dismissed):
Remove the new if statement checking profiles, the check should have occurred previously above.

dataprofiler/profilers/profile_builder.py
```diff
@@ -51,56 +52,82 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+                        clean_sampled_df, self.options, pool)
+                })
+        else:
 
-        use_data_labeler = True
-        if options and isinstance(options, StructuredOptions):
-            use_data_labeler = options.data_labeler.is_enabled
+            # Profile compilers being updated
+            for profile in self.profiles.values():
+                profile.update_profile(clean_sampled_df, pool)
```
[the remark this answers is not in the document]
JGSweets: I see there were some indents, but I am confused bc the if statement creates profiles if they don't exist, then why not just update them since they now exist?

dataprofiler/profilers/profile_builder.py
```diff
@@ -51,56 +52,82 @@ def __init__(self, df_series, sample_size=None, min_sample_size=5000,
...
+            use_data_labeler = True
+            if self.options and isinstance(self.options, StructuredOptions):
+                use_data_labeler = self.options.data_labeler.is_enabled
+
+            if use_data_labeler:
+                self.profiles.update({
+                    'data_label_profile':
+                    ColumnDataLabelerCompiler(
+                        clean_sampled_df, self.options, pool)
+                })
```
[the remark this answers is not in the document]
JGSweets: Can you elaborate on:
> it's not in consistent to say "there are no statistical profiles for float because there are no floats.
I'm not following what it means.

dataprofiler/profilers/profile_builder.py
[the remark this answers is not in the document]
lettergram: That's what's happening?

dataprofiler/profilers/profile_builder.py
```diff
@@ -143,18 +170,24 @@ def __add__(self, other):
         samples = list(dict.fromkeys(self.sample + other.sample))
         merged_profile.sample = random.sample(samples, min(len(samples), 5))
         for profile_name in self.profiles:
-            merged_profile.profiles[profile_name] = (
-                self.profiles[profile_name] + other.profiles[profile_name]
-            )
+            if profile_name in other.profiles:
+                merged_profile.profiles[profile_name] = (
+                    self.profiles[profile_name] + other.profiles[profile_name]
+                )
```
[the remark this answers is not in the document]
lettergram: If both are none they are equal.

dataprofiler/profilers/profile_builder.py
```diff
@@ -593,10 +622,10 @@ def update_profile(self, data, sample_size=None, min_true_samples=None):
                 "Data must either be imported using the data_readers or "
                 "pd.DataFrame."
             )
-
+   
     @staticmethod
     def _update_profile_from_chunk(df, profile=None, sample_size=None,
-                                   min_true_samples=None, options=None):
+                                   min_true_samples=0, options=None):
```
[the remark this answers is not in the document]
lettergram: Based on my understanding of the previous code, in the two places this was used we replaced it with 0 immediately if it was None with no checks. I'm happy to look around, but the default behavior seems appropriately "use all the samples you can" and the tests pass.

Is there a missing test you'd like me to add?

dataprofiler/profilers/profile_builder.py
```diff
@@ -639,57 +669,108 @@ def tqdm(l):
...
+        if pool is not None:
 
+            # Create a bunch of simultaneous column conversions
+            for col in df.columns:
+                if min_true_samples is None:
+                    min_true_samples = profile[col]._min_true_samples
+                try:
+                    multi_process_dict[col] = pool.apply_async(
+                        profile[col].clean_data_and_get_base_stats,
+                        (df[col], sample_size, min_true_samples, sample_ids))
```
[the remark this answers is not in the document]
lettergram: You can try running this yourself. It's just adding stuff to sub processeses already created. Super fast.

dataprofiler/profilers/profile_builder.py
lettergram: It's a non-critical component that has never been tested. A bug here is not likely nor important. I don't see a reason to test it and add more time to our 15 min of tests

dataprofiler/profilers/profile_builder.py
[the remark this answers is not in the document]
JGSweets: it is inside an else statement contrary to creating the profiles:
```
if self.profiles is None or len(self.profiles) == 0:
    # create profiles
else:
    # Profile compilers being updated
    for profile in self.profiles.values():
        profile.update_profile(clean_sampled_df, pool)
```

dataprofiler/profilers/profile_builder.py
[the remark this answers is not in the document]
JGSweets: This doesn't seem to be checking `None` it appears to check if the profile contains a key and if so add it.
`elif set(self.profiles) != set(other.profiles):`
already checks to see if they have the same keys.

Am I missing something?

dataprofiler/profilers/profile_builder.py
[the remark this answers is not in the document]
JGSweets: We should validate if we have any tests regarding min_true_samples for ensuring the correct sampling.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -262,8 +263,7 @@ def test_data_label_assigned(self):
             self.fail(
                 "Dataset tested did not have a non-null column and therefore "
                 "could not validate the test.")
-
-    @mock.patch('dataprofiler.profilers.profile_builder.StructuredDataProfile')
```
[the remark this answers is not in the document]
lettergram: Mocking is an issue actually, you can't pickle mocks. It expects the real classes so it doesn't appropriate test the functions.

dataprofiler/profilers/profile_builder.py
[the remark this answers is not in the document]
lettergram: I missed something, don't need this.

gme5078 (dismissed):

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -4,6 +4,7 @@
 import sys
 import time
 import logging
+import tempfile
```
ChrisWallace2020, line 7: What is the point of this?

ChrisWallace2020 (approved):

gme5078 (approved):

PR #166, opened 2021-04-20
Fix Null Row calculations

ChrisWallace2020:
Addresses issue #129 

Solution is to only calculate null indices in each column up-to the minimum (final) sampling index across all columns.


dataprofiler/profilers/profile_builder.py
```diff
@@ -158,12 +159,14 @@ def __add__(self, other):
...
+             "sample_size": self.sample_size,
              "null_count": self.null_count,
              "null_types": copy.deepcopy(self.null_types_index)}
         )
         merged_profile._update_base_stats(
-            {"sample": other.sample, 'sample_size': other.sample_size,
+            {"sample": other.sample, "sample_ids": other.sample_ids,
+             "sample_size": other.sample_size,
              "null_count": other.null_count,
              "null_types": copy.deepcopy(other.null_types_index)}
```
lettergram, line 171: @gme5078 and @ChrisWallace2020 this doesn't seem correct, at least without some modifications...

For instance: 
```
df_series1 = [None, 1, 2, 3]
df_series2 = [0, None, 2, 3]
profile1 = StructuredDataProfile(df_series1)
profile2 = StructuredDataProfile(df_series2)
profile3 = profile1 + profile2
profile3.profile['null_types'] # Will be [0, 1]
```

IMO should be [0, 5] because the indices should be an extension of the dataset. Currently, the null locations will be [0, 1]

dataprofiler/profilers/profile_builder.py
```diff
@@ -317,12 +320,13 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
         if not len_df:
             return df_series, {
-                "sample_size": 0, "null_count": 0,
+                "sample_size": 0, "sample_ids": [], "null_count": 0,
                 "null_types": dict(), "sample": []
             }
 
         # Pandas reads empty values in the csv files as nan
         df_series = df_series.apply(str)
+        original_index = df_series.index
```
lettergram, line 329: I don't really like this approach because if this happens outside the function it'll require less memory. Just select the series with the smallest set of sample_ids used, then utilize that

dataprofiler/profilers/profile_builder.py
```diff
@@ -335,11 +339,15 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
         total_sample_size = 0
+        sample_ids_used = []
         query = '|'.join(null_values_and_flags.keys())
         regex = f"^(?:{(query)})$"
         
         for chunked_sample_ids in sample_ind_generator:
+            if isinstance(chunked_sample_ids, np.ndarray):
+                chunked_sample_ids = chunked_sample_ids.tolist()
             total_sample_size += len(chunked_sample_ids)
+            sample_ids_used += chunked_sample_ids
```
lettergram, line 350: This creates an unnecessary copy increasing memory by columns * rows * int64

Instead you should be able to select the location of the slice and pass that outside, assuming it's not dynamically creating sample_ids. If it is generating sample_ids then this is fine. But I'm not 100% sure when that happens any more anyway

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: The problem is that df_series is mutated within this function to contain only the non-null values, so pulling out its index after the fact won't have the index values which are null, resulting in a potentially empty intersection every time it is calculated in `_update_row_statistics`

ChrisWallace2020:
Just a thought, if we are only taking the minimum row for calculating null row statistics, then the ratio should be updated to use this too instead right? Currently it uses max col samples used, but I think it should probably be updated to minimum to reflect how the rows are counted. Thoughts?

dataprofiler/profilers/profile_builder.py
lettergram: I believe sample_ids should contain the entry locations and is external to this function, then you can get the index locations external to this function by doing:

`df.index[sample_ids[:threshold]]`

lettergram:
@ChrisWallace2020 I believe you are correct

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: See issue #167

dataprofiler/profilers/profile_builder.py
```diff
@@ -352,9 +353,9 @@ def clean_data_and_get_base_stats(df_series, sample_size,
                 na_columns.setdefault(cell, list()).append(index)
             
             # Ensure minimum number of true samples met
-            # and if total_sample_size > samplesize exit
+            # and if total_sample_size >= sample size, exit
             if len(true_sample_list) >= min_true_samples \
-               and total_sample_size > sample_size:                
+                    and total_sample_size >= sample_size:
```
lettergram, line 358: Good call

lettergram (commented):
Please add a test case for lined up null rows

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -294,6 +294,7 @@ def test_sample_size_warning_in_the_profiler(self, *mocks):
         # structure data profile mock
         sdp_mock = mock.Mock()
         sdp_mock.clean_data_and_get_base_stats.return_value = (None, None)
+        sdp_mock.sample_size = 3
```
JGSweets, line 297: why is this necessary?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -653,6 +654,24 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
         self.assertEqual(col_one_len, len(data['NAME']))
         self.assertEqual(col_two_len, len(data[' VALUE']))
 
+    def test_null_calculation_with_differently_sampled_cols(self):
+        opts = ProfilerOptions()
+        opts.structured_options.multiprocess.is_enabled = False
+        data = pd.DataFrame({"full": [1, 2, 3, 4, 5, 6, 7, 8, 9],
+                             "sparse": [1, None, 3, None, 5, None, 7, None, 9]})
```
lettergram, line 661: Should test if there's rows as well:

```
data = pd.DataFrame({
    "full": [1, None, 3, None, 5, None, 7, None, 9],
    "sparse": [1, None, 3, None, 5, 6, 7, 8, 9]})
```

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: This test case kept throwing mock errors after I added some logic to update_from_chunk, because when I tried to calculate `self._complete_rows_sampled` for the Profile the underlying `StructuredDataProfile` was mocked out, but I still needed its sample size for a calculation.

lettergram (dismissed):
Will do a performance eval after this PR to see where we are at.

Looks good though.

dataprofiler/profilers/profile_builder.py
```diff
@@ -561,17 +572,17 @@ def _get_unique_row_ratio(self):
...
-        return 0 if self._max_col_samples_used == 0 \
-            else self.row_has_null_count / self._max_col_samples_used
+        return 0 if self._min_col_samples_used in {0, None} \
+            else self.row_has_null_count / self._min_col_samples_used
 
     def _get_duplicate_row_count(self):
         return self.total_samples - len(self.hashed_row_dict)
 
-    def _update_row_statistics(self, data):
+    def _update_row_statistics(self, data, sample_ids):
```
lettergram, line 585: default should be `None`

dataprofiler/profilers/profile_builder.py
```diff
@@ -580,13 +591,23 @@ def _update_row_statistics(self, data):
...
+                             "not a DataFrame")
         
         self.total_samples += len(data)
         self.hashed_row_dict = dict.fromkeys(
             pd.util.hash_pandas_object(data, index=False), True
         )
 
+        # Calculate complete rows read
+        # Only consider nulls in fully sampled rows for calculations
+        complete_row_ids = sample_ids[:self._min_col_samples_used]
```
lettergram, line 609: Probably should check if `sample_ids` exists

JGSweets: if it doesn't exist, then what do we do?

dataprofiler/profilers/profile_builder.py
```diff
@@ -521,6 +522,16 @@ def _max_col_samples_used(self):
...
 
+    @property
+    def _min_col_samples_used(self):
+        """
+        Calculates and returns the number of rows that were completely sampled
+        i.e. every column in the Profile was read up to this row (possibly
+        further in some cols)
+        """
+        return min([self._profile[col].sample_size
+                    for col in self._profile], default=0)
```
JGSweets, line 533: Not asking for change in this PR, but i do find it less clear as opposed to:
```python
if not self._profiles:
    return 0
return min([profile.sample_size for profile 
            in self._profile.values()])
```
I find it more explicit about saying it defaults to 0 when there are no profiles

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -303,6 +303,23 @@ def test_sample_size_warning_in_the_profiler(self, *mocks):
...
+        self.assertEqual(0, empty_profile._min_col_samples_used)
+
+        # Every column fully sampled
+        full_df = pd.DataFrame([[1, 2, 3], [4, 5, 6], [7, 8, 9]])
+        full_profile = dp.Profiler(full_df)
+        self.assertEqual(3, full_profile._min_col_samples_used)
+
+        # First col sampled only once, so that is min
+        sparse_df = pd.DataFrame([[1, None, None], [1, 1, None], [1, None, 1]])
+        sparse_profile = dp.Profiler(sparse_df, min_true_samples=1,
```
JGSweets, line 319: i think we should set min_true_samples to 2 as it is possible in this case for it to randomly select non-null sample_ids.
Additionally, this test is not isolated to just testing `test_min_col_samples_used` and requires running the entire profiler which may not be necessary.

lettergram (dismissed):

dataprofiler/profilers/profile_builder.py
```diff
@@ -597,6 +620,10 @@ def _update_row_statistics(self, data):
             if null_type_dict:
                 null_row_indices = set.union(*null_type_dict.values())
 
+            if complete_row_ids is not None:
+                null_row_indices = null_row_indices.intersection(
+                    data.index[complete_row_ids])
```
lettergram, line 625: Just check for `sample_ids is not None` here then do this IMO. 

So:

```
if sample_ids is not None:
    null_row_indices = null_row_indices.intersection(
        data.index[sample_ids[:self._min_col_samples_used]])
```

dataprofiler/profilers/profile_builder.py
```diff
@@ -792,6 +813,14 @@ def tqdm(l):
...
+
         if pool is not None:
-            pool.close() # Close pool for new tasks
-            pool.join() # Wait for all workers to complete
+            pool.close()  # Close pool for new tasks
+            pool.join()  # Wait for all workers to complete
+
+        # Only pass along sample ids if min col samples exists
+        samples_for_row_stats = None
+        if self._min_col_samples_used > 0:
```
lettergram, line 823: `if (min_true_samples is None or min_true_samples == 0) and len(sample_ids) > 0:`

dataprofiler/profilers/profile_builder.py
```diff
@@ -792,6 +813,14 @@ def tqdm(l):
...
+
         if pool is not None:
-            pool.close() # Close pool for new tasks
-            pool.join() # Wait for all workers to complete
+            pool.close()  # Close pool for new tasks
+            pool.join()  # Wait for all workers to complete
+
+        # Only pass along sample ids if necessary
+        samples_for_row_stats = None
+        if min_true_samples not in [None, 0] and len(sample_ids) > 0:
```
lettergram, line 823: can drop the len(sample_ids) check

dataprofiler/profilers/profile_builder.py
```diff
@@ -682,13 +704,14 @@ def tqdm(l):
             # If there's a sample size, truncate
```
lettergram, line 704: Can you modify the above to match the statement check below?

lettergram (approved):

JGSweets (approved):

Issue #163, opened 2021-04-20
Create the TextProfiler for unstructured profiling


JGSweets:
**Is your feature request related to a problem? Please describe.**
Can't profile unstructured text

**Describe the outcome you'd like:**
Need class for text profiling of unstructured data which mimics the formats of the structured profiling.
Include the following statistics:
word counts
vocab counts
line_lengths (min,max,...)

**Additional context:**
Starting point, final code may change:
```python
class TextProfiler(object):
    
    type = 'general_info'

    def __init__(options):

        self.sample_size = 0
        self.times = defaultdict(float)
        self.vocab = set()
        self.words = defaultdict(int)
        self.line_length = {'max': None, 'min': None, ...} # should use numeric stats mixin?

        # options values

        # these stop words are from nltk
        self._stop_words = {
            'i', 'me', 'my', 'myself', 'we', 'our', 'ours', 'ourselves', 'you',
            "you're", "you've", "you'll", "you'd", 'your', 'yours', 'yourself',
            'yourselves', 'he', 'him', 'his', 'himself', 'she', "she's", 'her',
            'hers', 'herself', 'it', "it's", 'its', 'itself', 'they', 'them',
            'their', 'theirs', 'themselves', 'what', 'which', 'who', 'whom',
            'this', 'that', "that'll", 'these', 'those', 'am', 'is', 'are',
            'was', 'were', 'be', 'been', 'being', 'have', 'has', 'had',
            'having', 'do', 'does', 'did', 'doing', 'a', 'an', 'the', 'and',
            'but', 'if', 'or', 'because', 'as', 'until', 'while', 'of', 'at',
            'by', 'for', 'with', 'about', 'against', 'between', 'into',
            'through', 'during', 'before', 'after', 'above', 'below', 'to',
            'from', 'up', 'down', 'in', 'out', 'on', 'off', 'over', 'under',
            'again', 'further', 'then', 'once', 'here', 'there', 'when',
            'where', 'why', 'how', 'all', 'any', 'both', 'each', 'few', 'more',
            'most', 'other', 'some', 'such', 'no', 'nor', 'not', 'only', 'own',
            'same', 'so', 'than', 'too', 'very', 's', 't', 'can', 'will',
            'just', 'don', "don't", 'should', "should've", 'now', 'd', 'll',
            'm', 'o', 're', 've', 'y', 'ain', 'aren', "aren't", 'couldn',
[...]

PR #170, opened 2021-04-20
Adding Unstructured Text Profiler

gme5078:
Adds the unstructured text profiler and test file.

PRing before its totally finished to allow for group contribution
Known "TODOs":
- Need to add line_length
- Need to add options to the options class

Do you like how I've named things? I am open to suggestions.

### DIFF
--- a/dataprofiler/profilers/unstructured_text_profile.py
+++ b/dataprofiler/profilers/unstructured_text_profile.py
@@ -1,6 +1,8 @@
 from collections import defaultdict
-from . import utils, BaseColumnProfiler
 import itertools
+
+from . import utils, BaseColumnProfiler
+
 
 class TextProfiler(object):
     type = 'unstructured_text'
@@ -20,6 +22,7 @@ class TextProfiler(object):
         self.vocab = set()
         self.word_count = defaultdict(int)
         self.metadata = dict()
+        self.case_sensitive = True
 
         # TODO: Add line length
         #self.line_length = {'max': None, 'min': None,...} #numeric stats mixin?
@@ -121,10 +124,8 @@ class TextProfiler(object):
         if "words" in merged_profile.__calculations:
             merged_profile.word_count = self.word_count.copy()
             for word in other.word_count:
-                if word in merged_profile.word_count:
+                if word.lower() not in self._stop_words:
                     merged_profile.word_count[word] += other.word_count[word]
-                elif word not in self._stop_words:
-                    merged_profile.word_count[word] = other.word_count[word]
                     
         merged_profile.sample_size = self.sample_size + other.sample_size
 
@@ -180,11 +181,10 @@ class TextProfiler(object):
         :type subset_properties: dict
         :return: None
         """
-        for word in list(" ".join(data).split(" ")):
-            if word in self.word_count:
-                self.word_count[word] += 1
-            elif word not in self._stop_words:
-                self.word_count[word] = 1
+        for row in data:
+            for word in row.split(" "):
+                if word.lower() not in self._stop_words:
+                    self.word_count[word] += 1
 
     def _update_helper(self, data, profile):
         """
--- a/dataprofiler/profilers/unstructured_text_profile.py
+++ b/dataprofiler/profilers/unstructured_text_profile.py
@@ -1,5 +1,6 @@
 from collections import defaultdict
 import itertools
+import re
 
 from . import utils, BaseColumnProfiler
 
@@ -22,10 +23,11 @@ class TextProfiler(object):
         self.vocab = set()
         self.word_count = defaultdict(int)
         self.metadata = dict()
-        self.case_sensitive = True
 
         # TODO: Add line length
         #self.line_length = {'max': None, 'min': None,...} #numeric stats mixin?
+
+        self._case_sensitive = True
 
         # these stop words are from nltk
         self._stop_words = {
@@ -60,31 +62,35 @@ class TextProfiler(object):
         }
         BaseColumnProfiler._filter_properties_w_options(self.__calculations, options)
 
-    def _add_helper(self, other1, other2):
+    def _merge_words(self, other, merged_profile):
         """
-        Merges the properties of two BaseColumnProfile objects
+        Merges the words of two TextProfiler profiles
 
-        :param other1: first profile
-        :param other2: second profile
-        :type other1: BaseColumnProfiler
-        :type other2: BaseColumnProfiler
+        :param self: first profile
+        :param other: second profile
+        :param merged_profile: merged profile
+        :type self: TextProfiler
+        :type other: TextProfiler
+        :type merged_profile: TextProfiler
+        :return:
         """
-        if other1.name == other2.name:
-            self.name = other1.name
+        if not self._case_sensitive:
+            merged_profile.word_count = self.word_count.copy()
+            additive_words = other.word_count
         else:
-            raise ValueError("Text names unmatched: {} != {}"
-                             .format(other1.name, other2.name))
-
-        self.times = defaultdict(
-            float, {key: (other1.times.get(key, 0)
-                          + other2.times.get(key, 0)
-                          + self.times.get(key, 0))
-                    for key in (set(other1.times) | set(other2.times)
-                                | set(self.times))}
-        )
-
-        self.sample_size = other1.sample_size + other2.sample_size
-
+            merged_profile.word_count = other.word_count.copy()
+            additive_words = self.word_count
+            
+        if merged_profile._case_sensitive:
+            for word in additive_words:
+                if word.lower() not in self._stop_words:
+                    merged_profile.word_count[word] += additive_words[word]
+        else:
+            for word in additive_words:
+                if word.lower() not in self._stop_words:
+                    merged_profile.word_count[word.lower()] += additive_words[
+                            word]
+    
     def __add__(self, other):
         """
         Merges the properties of two TextProfiler profiles
@@ -113,6 +119,10 @@ class TextProfiler(object):
                     for key in (set(self.times) | set(other.times))}
         )
         
+        merged_profile._case_sensitive = False
+        if self._case_sensitive and other._case_sensitive:
+            merged_profile._case_sensitive = True
+        
         BaseColumnProfiler._merge_calculations(merged_profile.__calculations,
                                  self.__calculations,
                                  other.__calculations)
@@ -122,11 +132,8 @@ class TextProfiler(object):
             merged_profile._update_vocab(other.vocab)
             
         if "words" in merged_profile.__calculations:
-            merged_profile.word_count = self.word_count.copy()
-            for word in other.word_count:
-                if word.lower() not in self._stop_words:
-                    merged_profile.word_count[word] += other.word_count[word]
-                    
+            self._merge_words(other, merged_profile)
+
         merged_profile.sample_size = self.sample_size + other.sample_size
 
         return merged_profile
@@ -138,10 +145,13 @@ class TextProfiler(object):
 
         :return:
         """
+        word_count = sorted(self.word_count.items(),
+                            key=lambda x: x[1],
+                            reverse=True)
         profile = dict(
             vocab=self.vocab,
             words=list(self.word_count.keys()),
-            word_count=dict(self.word_count),
+            word_count=dict(word_count),
             times=self.times,
         )
         return profile
@@ -181,10 +191,17 @@ class TextProfiler(object):
         :type subset_properties: dict
         :return: None
         """
-        for row in data:
-            for word in row.split(" "):
-                if word.lower() not in self._stop_words:
-                    self.word_count[word] += 1
+        if self._case_sensitive:
+            for row in data:
+                for word in re.findall(r'\w+', row):
+                    if word.lower() not in self._stop_words:
+                            self.word_count[word] += 1
+        else:
+            for row in data:
+                for word in re.findall(r'\w+', row):
+                    if word.lower() not in self._stop_words:
+                        self.word_count[word.lower()] += 1
+
 
     def _update_helper(self, data, profile):
         """
--- a/dataprofiler/profilers/unstructured_text_profile.py
+++ b/dataprofiler/profilers/unstructured_text_profile.py
@@ -27,36 +27,108 @@ class TextProfiler(object):
         # TODO: Add line length
         #self.line_length = {'max': None, 'min': None,...} #numeric stats mixin?
 
-        self._case_sensitive = True
+        self._is_case_sensitive = True
         if options:
-            self._case_sensitive = options.case_sensitive
+            self._is_case_sensitive = options.is_case_sensitive
 
         # these stop words are from nltk
         self._stop_words = {
-            'i', 'me', 'my', 'myself', 'we', 'our', 'ours', 'ourselves', 'you',
-            "you're", "you've", "you'll", "you'd", 'your', 'yours', 'yourself',
-            'yourselves', 'he', 'him', 'his', 'himself', 'she', "she's", 'her',
-            'hers', 'herself', 'it', "it's", 'its', 'itself', 'they', 'them',
-            'their', 'theirs', 'themselves', 'what', 'which', 'who', 'whom',
-            'this', 'that', "that'll", 'these', 'those', 'am', 'is', 'are',
-            'was', 'were', 'be', 'been', 'being', 'have', 'has', 'had',
-            'having', 'do', 'does', 'did', 'doing', 'a', 'an', 'the', 'and',
-            'but', 'if', 'or', 'because', 'as', 'until', 'while', 'of', 'at',
-            'by', 'for', 'with', 'about', 'against', 'between', 'into',
-            'through', 'during', 'before', 'after', 'above', 'below', 'to',
-            'from', 'up', 'down', 'in', 'out', 'on', 'off', 'over', 'under',
-            'again', 'further', 'then', 'once', 'here', 'there', 'when',
-            'where', 'why', 'how', 'all', 'any', 'both', 'each', 'few', 'more',
-            'most', 'other', 'some', 'such', 'no', 'nor', 'not', 'only', 'own',
-            'same', 'so', 'than', 'too', 'very', 's', 't', 'can', 'will',
-            'just', 'don', "don't", 'should', "should've", 'now', 'd', 'll',
-            'm', 'o', 're', 've', 'y', 'ain', 'aren', "aren't", 'couldn',
-            "couldn't", 'didn', "didn't", 'doesn', "doesn't", 'hadn', "hadn't",
-            'hasn', "hasn't", 'haven', "haven't", 'isn', "isn't", 'ma',
-            'mightn', "mightn't", 'mustn', "mustn't", 'needn', "needn't",
-            'shan', "shan't", 'shouldn', "shouldn't", 'wasn', "wasn't",
-            'weren', "weren't", 'won', "won't", 'wouldn', "wouldn't"
-        }
+            'thereafter', 'several', 'whenever', 'round', "she'd", 'shall', 
+            'adj', 'front', 'abroad', 'specify', 'namely', 'lest', 'among', 
+            'other', "wont", "couldn't", "i've", 'notwithstanding', 'isn', 
+            'recently', 'seemed', 'therefore', 'believe', 'twenty', "c'mon", 
+            'enough', 'fifth', 'otherwise', 'necessary', "weren't", '‘d', 
+            'all', 'thereupon', 'becomes', 'et', 'different', 'top', 'forty', 
+            'past', "daren't", 'rd', 'although', 'edu', 'again', 'it', 'three', 
+            'hadn', 'my', "you'll", 'makes', 'though', 'fewer', 'four', '’d', 
+            'neednt', 'isnt', 'meantime', 'hes', 'qv', 'wasn', 'very', 
+            'perhaps', 'who', 'meanwhile', 'sent', 'fifteen', 'has', 'plus', 
+            'ago', 'respectively', 'considering', 'provided', 'at', 'few', 
+            'elsewhere', 'ten', 'new', 'seven', 'unless', "i'm", 'how', 
+            'nothing', 'each', 'hence', 'are', 'some', 'selves', 'already', 
+            'behind', 'thanx', 'apart', "he'll", 'in', 'sometime', 'yours', 
+            'looks', 'always', 'five', 'with', 'need', 'most', "'ll", "'ve", 
+            'com', 'k', 'cause', 'those', 'or', 'toward', 'afterwards', 'ex', 
+            'wouldn', 'let', 'out', 'gives', 'there', 'alongside', 'own', 
+            'himself', 'mightve', 'her', 'wherein', 'both', 'caption', 'than', 
+            'nearly', 'better', 'thereby', 'anyhow', 'itself', 'without', 
+            'bottom', "mustn't", 'look', 'shouldn', 'show', 'wasnt', 'mightn', 
+            'whose', 'herself', 'hasnt', 'definitely', 'if', 'neverf', 'th', 
+            'whence', 'doesn', 'inner', '‘s', 'concerning', 'n’t', 'amid', 
+            'ending', '‘m', 'wouldnt', 'appear', 'within', 'greetings', 
+            'associated', 'clearly', "shouldn't", 'corresponding', 'o', 'first',
+            'novel', 'right', "didn't", "can't", 'non', 'latter', 'whereas', 
+            'and', 'under', 'comes', 'mainly', 'these', 've', 'dare', 'towards',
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+            'where', 'why', 'how', 'all', 'any', 'both', 'each', 'few', 'more',
+            'most', 'other', 'some', 'such', 'no', 'nor', 'not', 'only', 'own',
+            'same', 'so', 'than', 'too', 'very', 's', 't', 'can', 'will',
+            'just', 'don', "don't", 'should', "should've", 'now', 'd', 'll',
+            'm', 'o', 're', 've', 'y', 'ain', 'aren', "aren't", 'couldn',
+            "couldn't", 'didn', "didn't", 'doesn', "doesn't", 'hadn', "hadn't",
+            'hasn', "hasn't", 'haven', "haven't", 'isn', "isn't", 'ma',
+            'mightn', "mightn't", 'mustn', "mustn't", 'needn', "needn't",
+            'shan', "shan't", 'shouldn', "shouldn't", 'wasn', "wasn't",
+            'weren', "weren't", 'won', "won't", 'wouldn', "wouldn't"
```
lettergram, line 51: NLTK is probably not the selection of words like this I'd use - reason:

1. Look at `shouldn't`, I typically type that `shouldnt`
2. Their lists are often less inclusive (as shown above)

I believe this is already in somewhere, but bringing up my concern for poor selection.

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+        :type data: Union[list, numpy.array, pandas.DataFrame]
+        :param prev_dependent_properties: Contains all the previous properties
+            that the calculations depend on.
+        :type prev_dependent_properties: dict
+        :param subset_properties: Contains the results of the properties of the
+            subset before they are merged into the main data profile.
+        :type subset_properties: dict
+        :return: None
+        """
+        for word in list(" ".join(data).split(" ")):
```
lettergram, line 183:
```
data = ["this is a test", "sentence"]
" ".join(data) # "This is a test sentence"
" ".join(data).split(" ") # ["this", "is", "a", "test", "sentence"]
```

I don't think you need to put it in a `list`?

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+            if word in self.word_count:
+                self.word_count[word] += 1
+            elif word not in self._stop_words:
+                self.word_count[word] = 1
```
lettergram, line 187:
```
if word not in self._stop_words:
    self.word_count[word] = 0
self.word_count[word] += 1
```

dataprofiler/profilers/unstructured_text_profile.py
gme5078: Ah, good points, do you have a suggestion? I am unfamiliar with stop word datasets

dataprofiler/profilers/unstructured_text_profile.py
gme5078: Oh, maybe. I'll try without and see what happens

dataprofiler/profilers/unstructured_text_profile.py
gme5078: This would make every valid word count equal to 1 and every non-valid word count equal to however many counts there are of it

dataprofiler/profilers/unstructured_text_profile.py
JGSweets: @lettergram my assumption is that NLTK would use this as a base, but also do a strip on the `'` to include in the set so they wouldn't repeat the words multiple times. maybe i'm wrong?

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+        :param other: second profile
+        :type self: TextProfiler
+        :type other: TextProfiler
+        :return: New TextProfiler merged profile
+        """
+        if not isinstance(other, TextProfiler):
+            raise TypeError("Unsupported operand type(s) for +: "
+                            "'TextProfiler' and '{}'".format(
+                other.__class__.__name__))
+        merged_profile = TextProfiler(None)
```
JGSweets, line 99: we will have a TODO here for sending in the options potentially depending on what might need to be set on instantiation

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
+from collections import defaultdict
+from . import utils, BaseColumnProfiler
+import itertools
+
+class TextProfiler(object):
+    type = 'unstructured_text'
+
+    def __init__(self, name, options=None):
```
JGSweets, line 8: should this take in name? is it the name of the file or dataset? not sure how we are dealing with this since textdata

JGSweets: should this even have name?

AnhTruong: i'd vote for removing the name

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+        if "vocab" in merged_profile.__calculations:
+            merged_profile.vocab = self.vocab.copy()
+            merged_profile._update_vocab(other.vocab)
+            
+        if "words" in merged_profile.__calculations:
+            merged_profile.word_count = self.word_count.copy()
+            for word in other.word_count:
+                if word in merged_profile.word_count:
+                    merged_profile.word_count[word] += other.word_count[word]
+                elif word not in self._stop_words:
```
JGSweets, line 126: stop words are constant time, we might want to check it first which might make it faster if self.word_count is very large. Additionally, if word_count is a defaultdict(int) we can just add it no matter if it was there previously or not.
i.e.
```python
for word in other.word_count:
    if word not in self._stop_words:
        merged_profile.word_count[word] += other.word_count[word]
```

dataprofiler/profilers/unstructured_text_profile.py
AnhTruong: imo, we should have a txt file saving all stopwords, which should be updated every time we want

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+    @property
+    def profile(self):
+        """
+        Property for profile. Returns the profile of the column.
+
+        :return:
+        """
+        profile = dict(
+            vocab=self.vocab,
+            words=list(self.word_count.keys()),
```
JGSweets, line 142: if we have word_count, do we also need words?
Should word_count be ordered by value?

dataprofiler/profilers/unstructured_text_profile.py
JGSweets: is it not costly to join everything and then split as opposed to just looping through each item in data applying split and adding to the dict?

JGSweets: also, would `" ".join(data).split(" ")`  need `list()` around it as doesn't `split` create a list?

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+            'mightn', "mightn't", 'mustn', "mustn't", 'needn', "needn't",
+            'shan', "shan't", 'shouldn', "shouldn't", 'wasn', "wasn't",
+            'weren', "weren't", 'won', "won't", 'wouldn', "wouldn't"
+        }
+
+        self.__calculations = {
+            "vocab": TextProfiler._update_vocab,
+            "words": TextProfiler._update_words,
+        }
+        BaseColumnProfiler._filter_properties_w_options(self.__calculations, options)
```
AnhTruong, line 58: not for this PR, but later should we move `_filter_properties_w_options ` to some other generic class to be used for both structured and unstructured profilers.

AnhTruong (commented):
some comments

dataprofiler/profilers/unstructured_text_profile.py
JGSweets:
```python
if word not in self._stop_words:
    self.word_count[word] += 1
```
@lettergram no need to initialize first because it is a defaultdict(int)

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+
+from dataprofiler.profilers.unstructured_text_profile import TextProfiler
+
+class TestUnstructuredTextProfile(unittest.TestCase):
+    
+    
+    def test_text_profile_update_and_name(self):
+        text_profile = TextProfiler("Name")
+        sample = pd.Series(["Hello my name is Grant",
+                            "Bob and Grant are friends"])
```
JGSweets, line 15: what about bob?

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+        self.assertNotIn("is", profile['words'])
+
+        # Assert word counts are correct
+        expected_word_count = {'Hello': 1, 'name': 1, 'Grant': 2, 'Bob': 1,
+                               'friends': 1}
+        self.assertDictEqual(expected_word_count, profile['word_count'])
+        
+        # Assert timing is occurring
+        self.assertIn("vocab", profile["times"])
+        self.assertIn("words", profile["times"])
```
JGSweets, line 141: add line at end

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+        text_profile.update(sample)
+        profile = text_profile.profile
+
+        # Assert vocab is correct
+        expected_vocab = [' ', 'B', 'G', 'H', 'a', 'b', 'c', 'd', 'e',
+                          'f', 'h', 'i', 'k', 'l', 'm', 'n', 'o', 'r',
+                          's', 't', 'w', 'y']
+        self.assertListEqual(sorted(expected_vocab), sorted(profile['vocab']))
+
+
```
JGSweets, line 44: extra line here

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+class TestUnstructuredTextProfile(unittest.TestCase):
+    
+    
+    def test_text_profile_update_and_name(self):
+        text_profile = TextProfiler("Name")
+        sample = pd.Series(["Hello my name is Grant",
+                            "Bob and Grant are friends"])
+        text_profile.update(sample)
+        self.assertEqual("Name", text_profile.name)
+
```
JGSweets, line 18: extra line here

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+from unittest import mock
+from collections import defaultdict
+
+import pandas as pd
+
+from dataprofiler.profilers.unstructured_text_profile import TextProfiler
+
+class TestUnstructuredTextProfile(unittest.TestCase):
+    
+
```
JGSweets, line 11: extra line here

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
+import unittest
+from unittest import mock
+from collections import defaultdict
+
+import pandas as pd
+
+from dataprofiler.profilers.unstructured_text_profile import TextProfiler
+
```
JGSweets, line 8: space needed here

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
+from collections import defaultdict
+from . import utils, BaseColumnProfiler
+import itertools
+
```
JGSweets, line 4: space needed here

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
+from collections import defaultdict
+from . import utils, BaseColumnProfiler
+import itertools
```
JGSweets, line 3: 1. python builtin first
2. 3rd party
3. our lib

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+        :type name: String
+        :param options: Options for the Text Profiler
+        :type options: UnstructuredTextOptions
+        """
+        self.name = name
+        self.sample_size = 0
+        self.times = defaultdict(float)
+        self.vocab = set()
+        self.word_count = defaultdict(int)
+        self.metadata = dict()
```
JGSweets, line 22: we should consider an option for setting case sensitvity

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,223 @@
...
+            'weren', "weren't", 'won', "won't", 'wouldn', "wouldn't"
+        }
+
+        self.__calculations = {
+            "vocab": TextProfiler._update_vocab,
+            "words": TextProfiler._update_words,
+        }
+        BaseColumnProfiler._filter_properties_w_options(self.__calculations, options)
+
+    def _add_helper(self, other1, other2):
```
AnhTruong, line 60: do we need this function?

dataprofiler/profilers/unstructured_text_profile.py
AnhTruong: actually, another way is to allow stopwords as an input argument for the user

AnhTruong (commented):
other comments

dataprofiler/profilers/unstructured_text_profile.py
JGSweets: agreed, however, we haven't built out the options for it yet, so this should work for now. I will create an issue to make the options class

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+        expected_vocab = [' ', 'B', 'G', 'H', 'a', 'b', 'c', 'd', 'e',
+                          'f', 'h', 'i', 'k', 'l', 'm', 'n', 'o', 'r',
+                          's', 't', 'w', 'y']
+        self.assertListEqual(sorted(expected_vocab), sorted(profile['vocab']))
+
+ 
+    def test_words_and_word_count(self):
+        text_profile = TextProfiler("Name")
+        sample = pd.Series(["Hello my name is Grant",
+                            "Bob and Grant are friends"])
```
AnhTruong, line 48: can we add some punctuations to these tests?

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -0,0 +1,141 @@
...
+
+        # Assert word counts are correct
+        expected_word_count = {'Hello': 1, 'name': 1, 'Grant': 2, 'Bob': 1,
+                               'friends': 1}
+        self.assertDictEqual(expected_word_count, profile['word_count'])
+
+        # Update the data again
+        sample = pd.Series(["Grant knows how to code",
+                            "Grant will code with Bob"])
+        text_profile.update(sample)
```
AnhTruong, line 65: can TextProfiler work with str or list(str) as the input?

dataprofiler/profilers/unstructured_text_profile.py
gme5078: fixed

dataprofiler/profilers/unstructured_text_profile.py
gme5078: fixed

dataprofiler/profilers/unstructured_text_profile.py
gme5078: Fixed

dataprofiler/profilers/unstructured_text_profile.py
gme5078: Will order word count by value. 
Words, I'm indifferent about. It was in our pseudo code so I left it

dataprofiler/profilers/unstructured_text_profile.py
gme5078: Agree

dataprofiler/tests/profilers/test_unstructured_text_profile.py
gme5078: He's my friend. good guy

dataprofiler/profilers/unstructured_text_profile.py
gme5078: Just maintaining consistency

gme5078: Oh wait no, that has to go. Good catch

dataprofiler/tests/profilers/test_unstructured_text_profile.py
gme5078: Good idea

dataprofiler/tests/profilers/test_unstructured_text_profile.py
gme5078: It is not supposed to

dataprofiler/profilers/unstructured_text_profile.py
gme5078: done.

dataprofiler/profilers/unstructured_text_profile.py
gme5078: After talking, we decided to leave it for now

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,242 @@
...
+        """
+        if self._case_sensitive:
+            for row in data:
+                for word in re.findall(r'\w+', row):
+                    if word.lower() not in self._stop_words:
+                            self.word_count[word] += 1
+        else:
+            for row in data:
+                for word in re.findall(r'\w+', row):
+                    if word.lower() not in self._stop_words:
```
JGSweets, line 204: word_lower = word.lower() so don't have to compute twice

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,242 @@
...
+        :type data: Union[list, numpy.array, pandas.DataFrame]
+        :param prev_dependent_properties: Contains all the previous properties
+            that the calculations depend on.
+        :type prev_dependent_properties: dict
+        :param subset_properties: Contains the results of the properties of the
+            subset before they are merged into the main data profile.
+        :type subset_properties: dict
+        :return: None
+        """
+        if self._case_sensitive:
```
JGSweets, line 196: less pretty bc somewhat repeated, but faster bc the if isn't computed each loop. I'm okay with it

JGSweets: not sure the cost if it were internal, might be miniscule in comparison to lower or findall

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,314 @@
...
+
+        # TODO: Add line length
+        #self.line_length = {'max': None, 'min': None,...} #numeric stats mixin?
+
+        self._is_case_sensitive = True
+        if options:
+            self._is_case_sensitive = options.is_case_sensitive
+
+        # these stop words are from nltk
+        self._stop_words = {
```
gme5078, line 35: I created this myself from a merge of several different stop word datasets. If you see anything missing or things that shouldnt be in there, let me know, and I will change it.

dataprofiler/profilers/unstructured_text_profile.py
JGSweets: did some timings, and this actually can result in faster code:
```
if statement in loop: [0.718934227, 0.691763384, 0.7284096360000001, 0.7014499450000002, 0.6951934310000003]
if statement NOT in loop: [0.6208281739999997, 0.6166344240000008, 0.6279374830000002, 0.617970788, 0.6203451770000008]
```
```python
import re
from collections import defaultdict
import timeit


def test_if_speed2(data, is_case_sensitive, stop_words):
    word_count = defaultdict(int)
    if is_case_sensitive:
        for row in data:
            for word in re.findall(r'\w+', row):
                if word.lower() not in stop_words:
                    word_count[word] += 1
    else:
        for row in data:
            for word in re.findall(r'\w+', row):
                word_lower = word.lower()
                if word_lower not in stop_words:
                    word_count[word_lower] += 1


def test_if_speed(data, is_case_sensitive, stop_words):
    word_count = defaultdict(int)
    for row in data:
        for word in re.findall(r'\w+', row):
            word_lower = word.lower()
            if word.lower() not in stop_words:
                if is_case_sensitive:
                    word = word_lower
                word_count[word] += 1
                
    return word_count
            
            
stop_words = {
    'i', 'me', 'my', 'myself', 'we', 'our', 'ours', 'ourselves', 'you',
    "you're", "you've", "you'll", "you'd", 'your', 'yours', 'yourself',
    'yourselves', 'he', 'him', 'his', 'himself', 'she', "she's", 'her',
    'hers', 'herself', 'it', "it's", 'its', 'itself', 'they', 'them',
    'their', 'theirs', 'themselves', 'what', 'which', 'who', 'whom',
    'this', 'that', "that'll", 'these', 'those', 'am', 'is', 'are',
    'was', 'were', 'be', 'been', 'being', 'have', 'has', 'had',
    'having', 'do', 'does', 'did', 'doing', 'a', 'an', 'the', 'and',
    'but', 'if', 'or', 'because', 'as', 'until', 'while', 'of', 'at',
    'by', 'for', 'with', 'about', 'against', 'between', 'into',
    'through', 'during', 'before', 'after', 'above', 'below', 'to',
    'from', 'up', 'down', 'in', 'out', 'on', 'off', 'over', 'under',
    'again', 'further', 'then', 'once', 'here', 'there', 'when',
    'where', 'why', 'how', 'all', 'any', 'both', 'each', 'few', 'more',
    'most', 'other', 'some', 'such', 'no', 'nor', 'not', 'only', 'own',
    'same', 'so', 'than', 'too', 'very', 's', 't', 'can', 'will',
    'just', 'don', "don't", 'should', "should've", 'now', 'd', 'll',
    'm', 'o', 're', 've', 'y', 'ain', 'aren', "aren't", 'couldn',
    "couldn't", 'didn', "didn't", 'doesn', "doesn't", 'hadn', "hadn't",
    'hasn', "hasn't", 'haven', "haven't", 'isn', "isn't", 'ma',
    'mightn', "mightn't", 'mustn', "mustn't", 'needn', "needn't",
    'shan', "shan't", 'shouldn', "shouldn't", 'wasn', "wasn't",
    'weren', "weren't", 'won', "won't", 'wouldn', "wouldn't"
}
data = (['a', 'asdf', 'asdfljds', 'sdlfkjldsa', 'ksldjfkd sldkjf sdf']
        * int(1e2))
            
# if __name__ == '__main__':
print(timeit.repeat(
    'test_if_speed(data, is_case_sensitive=True, stop_words=stop_words)',
    setup='from __main__ import test_if_speed, stop_words, data',
    number=1000,
    repeat=5
))

print(timeit.repeat(
    'test_if_speed2(data, is_case_sensitive=True, stop_words=stop_words)',
    setup='from __main__ import test_if_speed2, stop_words, data',
    number=1000,
    repeat=5
))
```

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -0,0 +1,314 @@
...
+            'youre', 'dont', 'eighty', 'whatever', 'they', 'secondly', 'ones', 
+            'regardless', 'whereby', 'eight', 'she', 'maybe', 'looking', 
+            'thatllwas', 'needn', 'sometimes', 'put', 'others', 'amidst', 
+            'empty', 'next', 'tell', 'must', 'consider', 'backward', 'any', 
+            'into', 'don', 'happens', 'ours', 'side', 'especially', 'follows', 
+            'throughout', "'m", 'forth', 'were', "don't", 'much', 'amount', 
+            'taken', 'mustn', 'may', 'make', 'miss', 'sup', 'becoming', 'still',
+            "a's", 'certain', 'eg', 'never', 'course', 'inside', 'same', 
+            'backwards', 'actually', "n't", '‘ve', "needn't", 'contains', 
+            'somebody', 'm', 'probably', 'onto', 'just', 'two', 'known',
```
JGSweets, line 82: `two` not sure numbers should be stopped words

JGSweets (approved):

AnhTruong (approved):

gme5078:
This PR addresses this issue:
https://github.com/capitalone/DataProfiler/issues/163

Issue #171, opened 2021-04-20
After PR #159, The same test file took 3-4x longer to profile


JGSweets:
**General Information:**
 - OS: macOS
 - Python version:  3.8.7
 - Library version: PR before and after 159

**Describe the bug:**
Profiling now takes 3-4x

**To Reproduce:**
```python
import dataprofiler as dp
import time


data = dp.Data('time_profile.csv')

options = dp.ProfilerOptions()
options.structured_options.multiprocess.is_enabled = False
options.structured_options.data_labeler.is_enabled = False

start_time = time.time()
profiler1 = dp.Profiler(data, profiler_options=options)
profile_time = time.time() - start_time
print()
print()
print('profile time:', profile_time)

start_time = time.time()
profiler2 = profiler1 + profiler1
merge_time = time.time() - start_time
print('merge time:', merge_time)
```

**Expected behavior:**
Should either be quicker or no different.

**Screenshots:**
After PR #159 is on top, before is on bottom
```
(venv3) data-profiler % python3 profile.py
Finding the Null values in the columns...
Processing Column 1/6
Processing Column 2/6
Processing Column 3/6
Processing Column 4/6
Processing Column 5/6
Processing Column 6/6
Calculating the statistics...
Processing Column 1/6
Processing Column 2/6
dataprofiler/profilers/datetime_column_profile.py:221: RuntimeWarning: Years provided were in two digit format. As a result, datetime assumes dates < 69 are for 2000s and above are for the 1990s. https://stackoverflow.com/questions/37766353/pandas-to-datetime-parsing-wrong-year
  warnings.warn(
Processing Column 3/6
Processing Column 4/6
Processing Column 5/6
Processing Column 6/6


profile time: 27.837885856628418
Finding the Null values in the columns...
Calculating the statistics...
merge time: 3.6805949211120605


(venv3) data-profiler % python3 profile.py
Processing Column 1/6
dataprofiler/profilers/profile_builder.py:78: UserWarning: The data will be profiled with a sample size of 100000. All statistics will be based on this subsample and not the whole dataset.
  warnings.warn("The data will be profiled with a sample size of {}. "
venv3/lib/python3.8/site-packages/pandas/core/strings/accessor.py:101: UserWarning: This pattern has match groups. To actually get the groups, use str.extract.
  return func(self, *args, **kwargs)
Processing Column 2/6
[...]

PR #172, opened 2021-04-21
Ensure sample size limited for large data

JGSweets:
Address the issue: https://github.com/capitalone/DataProfiler/issues/171

Speeds are now below the original speed and the speed from the PR previous to PR #159 .

Profile time for 26MB file:
Current: 24+s
Prior to PR #159: ~8s
This PR: ~6.5s

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -126,7 +126,6 @@ class StructuredDataProfile(object):
             for profile in self.profiles.values():
                 profile.update_profile(clean_sampled_df, pool)
 
-
     def __add__(self, other):
         """
         Merges two Structured profiles together overriding the `+` operator.
@@ -441,7 +440,8 @@ class Profiler(object):
                 utils.warn_on_profile('data_labeler', e)
                 self.options.set({'data_labeler.is_enabled': False})
 
-        self.update_profile(data)
+        if len(data):
+            self.update_profile(data)
 
     def __add__(self, other):
         """
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -484,9 +484,10 @@ class Profiler(object):
         :return: integer sampling size
         :rtype: int
         """
-        len_data = len(data)
         if self._samples_per_update:
             return self._samples_per_update
+
+        len_data = len(data)
         if len_data <= self._min_sample_size:
             return int(len_data)
         return max(int(self._sampling_ratio * len_data), self._min_sample_size)
@@ -604,24 +605,26 @@ class Profiler(object):
         :type min_true_samples
         :return: None
         """
-        if not min_true_samples:
-            min_true_samples = self._min_true_samples
         if isinstance(data, data_readers.base_data.BaseData):
-            self._update_profile_from_chunk(
-                data.data, sample_size, min_true_samples, self.options)
-            self._update_row_statistics(data.data)
             self.encoding = data.file_encoding
             self.file_type = data.data_type
+            data = data.data
         elif isinstance(data, pd.DataFrame):
-            self._update_profile_from_chunk(
-                data, sample_size, min_true_samples, self.options)
-            self._update_row_statistics(data)
             self.file_type = str(data.__class__)
         else:
             raise ValueError(
                 "Data must either be imported using the data_readers or "
                 "pd.DataFrame."
             )
+
+        if not min_true_samples:
+            min_true_samples = self._min_true_samples
+        if not sample_size:
+            sample_size = self._get_sample_size(data)
+
+        self._update_profile_from_chunk(
+            data, sample_size, min_true_samples, self.options)
+        self._update_row_statistics(data)
 
     def _update_profile_from_chunk(self, df, sample_size=None,
                                    min_true_samples=None, options=None):
@@ -669,9 +672,6 @@ class Profiler(object):
                 for i, e in enumerate(l):
                     print("Processing Column {}/{}".format(i+1, len(l)))
                     yield e
-
-        if not sample_size:
-            sample_size = self._get_sample_size(df)
 
         # Shuffle indices ones and share with columns
         sample_ids = [*utils.shuffle_in_chunks(len(df), len(df))]
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -14,8 +14,6 @@ import re
 from collections import OrderedDict
 import warnings
 
-import multiprocessing as mp
-
 import pandas as pd
 import numpy as np
 
@@ -610,13 +608,13 @@ class Profiler(object):
             min_true_samples = self._min_true_samples
         if isinstance(data, data_readers.base_data.BaseData):
             self._update_profile_from_chunk(
-                data.data, self._profile, sample_size, min_true_samples, self.options)
+                data.data, sample_size, min_true_samples, self.options)
             self._update_row_statistics(data.data)
             self.encoding = data.file_encoding
             self.file_type = data.data_type
         elif isinstance(data, pd.DataFrame):
             self._update_profile_from_chunk(
-                data, self._profile, sample_size, min_true_samples, self.options)
+                data, sample_size, min_true_samples, self.options)
             self._update_row_statistics(data)
             self.file_type = str(data.__class__)
         else:
@@ -632,8 +630,6 @@ class Profiler(object):
         
         :param df: a dataset
         :type df: pandas.DataFrame
-        :param profile: list of profiled columns [BaseColumnProfiler subclasses]
-        :type profile: list
         :param sample_size: number of samples for df to use for profiling
         :type sample_size: int
         :param min_true_samples: minimum number of true samples required
@@ -648,6 +644,24 @@ class Profiler(object):
             raise ValueError('`Profiler` does not currently support data which '
                              'contains columns with duplicate names.')
 
+        # assign data labeler
+        data_labeler_options = self.options.structured_options.data_labeler
+        if data_labeler_options.is_enabled \
+                and data_labeler_options.data_labeler_object is None:
+
+            try:
+
+                data_labeler = DataLabeler(
+                    labeler_type='structured',
+                    dirpath=data_labeler_options.data_labeler_dirpath,
+                    load_options=None)
+                self.options.set(
+                    {'data_labeler.data_labeler_object': data_labeler})
+
+            except Exception as e:
+                utils.warn_on_profile('data_labeler', e)
+                self.options.set({'data_labeler.is_enabled': False})
+
         try:
             from tqdm import tqdm
         except:
@@ -657,7 +671,7 @@ class Profiler(object):
                     yield e
 
         if not sample_size:
-            self._get_sample_size(df)
+            sample_size = self._get_sample_size(df)
 
         # Shuffle indices ones and share with columns
         sample_ids = [*utils.shuffle_in_chunks(len(df), len(df))]
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -426,6 +426,24 @@ class Profiler(object):
         if isinstance(data, data_readers.text_data.TextData):
             raise TypeError("Cannot provide TextData object to Profiler")
 
+        # assign data labeler
+        data_labeler_options = self.options.structured_options.data_labeler
+        if data_labeler_options.is_enabled \
+                and data_labeler_options.data_labeler_object is None:
+
+            try:
+
+                data_labeler = DataLabeler(
+                    labeler_type='structured',
+                    dirpath=data_labeler_options.data_labeler_dirpath,
+                    load_options=None)
+                self.options.set(
+                    {'data_labeler.data_labeler_object': data_labeler})
+
+            except Exception as e:
+                utils.warn_on_profile('data_labeler', e)
+                self.options.set({'data_labeler.is_enabled': False})
+
         if len(data):
             self.update_profile(data)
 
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -421,27 +418,16 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
...
-                    labeler_type='structured',
-                    dirpath=data_labeler_options.data_labeler_dirpath,
-                    load_options=None)
-                self.options.set({'data_labeler.data_labeler_object': data_labeler})
-                
-            except Exception as e:
-                utils.warn_on_profile('data_labeler', e)
-                self.options.set({'data_labeler.is_enabled': False})
-
-        self.update_profile(data)
```
JGSweets, line 444: moved to inside `update_profile`, was done initially in case merging of profiles could duplicate the labeler, but looks like merge handles it by passing its options through.

JGSweets: Technically this allows a profile to be instantiated without creating the labeler until profile time making it faster. Not sure if this is desirable.

dataprofiler/profilers/profile_builder.py
```diff
@@ -602,37 +604,32 @@ def update_profile(self, data, sample_size=None, min_true_samples=None):
...
+            self._update_profile_from_chunk(
+                data.data, sample_size, min_true_samples, self.options)
             self._update_row_statistics(data.data)
             self.encoding = data.file_encoding
             self.file_type = data.data_type
         elif isinstance(data, pd.DataFrame):
-            self._profile = self._update_profile_from_chunk(
-                data, self._profile, sample_size, min_true_samples, self.options)
+            self._update_profile_from_chunk(
+                data, sample_size, min_true_samples, self.options)
```
JGSweets, line 617: previously `_update_profile_from_chunk` was static, but it doesn't really make sense and means we no longer have to pass profiles and can just use `self._profile` within the function

dataprofiler/profilers/profile_builder.py
```diff
@@ -602,37 +604,32 @@ def update_profile(self, data, sample_size=None, min_true_samples=None):
...
         else:
             raise ValueError(
                 "Data must either be imported using the data_readers or "
                 "pd.DataFrame."
             )
-   
-    @staticmethod
-    def _update_profile_from_chunk(df, profile=None, sample_size=None,
+
+    def _update_profile_from_chunk(self, df, sample_size=None,
```
JGSweets, line 626: previously `_update_profile_from_chunk` was static, but it doesn't really make sense and means we no longer have to pass profiles and can just use `self._profile` within the function

This allows us to get the sample_size within `_update_profile_from_chunk`

dataprofiler/profilers/profile_builder.py
JGSweets: This was done so that `_get_sample_size` could be called within `_update_profile_from_chunk`. If we did it here, the function would have to be rearranged in order to account for data checks which previously were not needed.

Currently the data check is the `else` in the conditionals, but if sample size was acquired here, it would need to move to the top  followed by the `_get_sample_size` and the instance check conditionals. Feasible if desired.

dataprofiler/profilers/profile_builder.py
```diff
@@ -642,14 +639,29 @@ def _update_profile_from_chunk(df, profile=None, sample_size=None,
         :return: list of column profile base subclasses
         :rtype: list(BaseColumnProfiler)
         """
-        
-        if not profile:
-            profile = OrderedDict()
```
JGSweets, line 647: just use the `self._profile`

dataprofiler/profilers/profile_builder.py
JGSweets: I'm no recalling the original intent was because if something error'ed within update, we wouldn't update anything. We can revert to keep this intent if desired.

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -351,6 +350,7 @@ def test_data_labeler(self, *mocks):
         options.structured_options.data_labeler.data_labeler_dirpath \
             = "Test_Dirpath"
         options.structured_options.data_labeler.max_sample_size = 50
+        options.structured_options.multiprocess.is_enabled = False
```
JGSweets, line 353: speeds up test

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -288,13 +286,16 @@ def test_text_data_raises_error(self):
...
+    @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler')
+    @mock.patch('dataprofiler.profilers.profile_builder.Profiler.'
+                '_update_row_statistics')
+    @mock.patch('dataprofiler.profilers.profile_builder.StructuredDataProfile')
     def test_sample_size_warning_in_the_profiler(self, *mocks):
+        # structure data profile mock
+        sdp_mock = mock.Mock()
+        sdp_mock.clean_data_and_get_base_stats.return_value = (dict(a=1), None)
+        mocks[0].return_value = sdp_mock
+
```
JGSweets, line 298: speeds up test

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -493,12 +493,31 @@ def test_null_count(self):
...
-
     def test_generating_report_ensure_no_error(self):
         file_path = os.path.join(test_root_path, 'data', 'csv/diamonds.csv')
         data = pd.read_csv(file_path)
         profile = dp.Profiler(data[:1000])
-        readable_report = profile.report(report_options={"output_format":"compact"})
+        readable_report = profile.report(
+            report_options={"output_format": "compact"})
+
+    def test_get_sample_size(self):
```
JGSweets, line 503: new sample size check for tests

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -493,12 +493,64 @@ def test_null_count(self):
...
+
+        # test data size * 0.20 < min_sample_size < data size
+        profile = dp.Profiler(data[:10000], profiler_options=profiler_options)
+        self.assertIn('sample_size', mocks[0].call_args[1])
+        self.assertEqual(5000, mocks[0].call_args[1]['sample_size'])
+
+        # test min_sample_size > data size * 0.20
+        profile = dp.Profiler(data, profiler_options=profiler_options)
+        self.assertIn('sample_size', mocks[0].call_args[1])
+        self.assertEqual(10000, mocks[0].call_args[1]['sample_size'])
```
JGSweets, line 552: this test validates it is actually passed as expected

dataprofiler/profilers/profile_builder.py
```diff
@@ -489,6 +475,22 @@ def __add__(self, other):
...
+        :type data: Union[data_readers.base_data.BaseData, pandas.DataFrame]
+        :return: integer sampling size
+        :rtype: int
+        """
+        len_data = len(data)
+        if self._samples_per_update:
+            return self._samples_per_update
+        if len_data <= self._min_sample_size:
+            return int(len_data)
+        return max(int(self._sampling_ratio * len_data), self._min_sample_size)
```
AnhTruong, line 492: did we have a similar function for `StructuredDataProfile`? Can we use only one?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -493,12 +493,64 @@ def test_null_count(self):
...
+
+        # test data size * 0.20 < min_sample_size < data size
+        profile = dp.Profiler(pd.DataFrame([]))
+        sample_size = profile._get_sample_size(data[:10000])
+        self.assertEqual(5000, sample_size)
+
+        # test min_sample_size > data size * 0.20
+        profile = dp.Profiler(pd.DataFrame([]))
+        sample_size = profile._get_sample_size(data)
+        self.assertEqual(10000, sample_size)
```
AnhTruong, line 519: imo, for these tests we should set the value of self._min_sample_size and self._sampling_ratio and then test. Otherwise, if we change the default values later, the test will fail.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -570,6 +625,7 @@ def test_correct_total_sample_size_and_counts_and_mutability(self):
         file_path = os.path.join(test_root_path, 'data', 'csv/empty_rows.txt')
         data = pd.read_csv(file_path)
         profiler_options = ProfilerOptions()
+        profiler_options.structured_options.multiprocess.is_enabled = False
```
AnhTruong, line 628: so, we never test if the multiprocessing running for any test?

AnhTruong (commented):
several comments

dataprofiler/profilers/profile_builder.py
```diff
@@ -421,27 +418,16 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
         self._min_true_samples = min_true_samples
         self._profile = dict()
 
+        # matches structured data profile
+        # TODO: allow set via options
+        self._sampling_ratio = 0.2
+        self._min_sample_size = 5000
```
lettergram, line 424: Isn't this set by default already when you create the class?

dataprofiler/profilers/profile_builder.py
lettergram: It's not clear to me why we need to call `self._get_sample_size` within the `_update_profile_from_chunk`, we already pass in `sample_size` here.

dataprofiler/profilers/profile_builder.py
```diff
@@ -642,14 +639,29 @@ def _update_profile_from_chunk(df, profile=None, sample_size=None,
...
+                data_labeler = DataLabeler(
+                    labeler_type='structured',
+                    dirpath=data_labeler_options.data_labeler_dirpath,
+                    load_options=None)
+                self.options.set(
+                    {'data_labeler.data_labeler_object': data_labeler})
+
+            except Exception as e:
+                utils.warn_on_profile('data_labeler', e)
+                self.options.set({'data_labeler.is_enabled': False})
```
lettergram, line 663: I don't think this belongs in here, we don't instantiate anything in this function for good reason. It's `preprocessing` -> `processing` -> `postprocessing`, BUT the `processing` component isn't managed in here. IMO this convolutes the already hard to follow function

dataprofiler/profilers/profile_builder.py
```diff
@@ -678,11 +693,11 @@ def tqdm(l):
         # Create structured profile objects
         new_cols = set()
         for col in df.columns:
-            if col not in profile:
+            if col not in self._profile:
```
lettergram, line 696: I don't think any of these `self` are necessary. Also, I don't think you'd need to return profile if you were just doing it this way.

dataprofiler/tests/profilers/test_profile_builder.py
lettergram: ^ I'm personally of the opinion multiprocessing should run on all tests, unless otherwise specified as it's the default behavior.

dataprofiler/profilers/profile_builder.py
```diff
@@ -658,6 +670,9 @@ def tqdm(l):
                     print("Processing Column {}/{}".format(i+1, len(l)))
                     yield e
 
+        if not sample_size:
+            sample_size = self._get_sample_size(df)
+
```
lettergram, line 675: I think this should be called outside the function and passed in

dataprofiler/profilers/profile_builder.py
JGSweets: currently we use it in both spots, however this function is slightly different as it also utilizes the samples_per_update which is unique to this class. I did not want to alter the get_sample_size in `StructuredDataProfile` as it would prolong this fix

dataprofiler/profilers/profile_builder.py
```diff
@@ -602,37 +604,32 @@ def update_profile(self, data, sample_size=None, min_true_samples=None):
         :type min_true_samples
         :return: None
         """
-        if not sample_size:
-            sample_size = self._samples_per_update
```
lettergram, line 606: I think we should first check the data type, then get samples

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: fair

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: it requires significantly more time to run the tests. If we want to revery we can.

dataprofiler/profilers/profile_builder.py
JGSweets: No, these are new to this class.

dataprofiler/profilers/profile_builder.py
JGSweets: Technically this allows a profile to be instantiated without creating the labeler until profile time making it faster. Not sure if this is desirable.

However, where it was before puts the `processing` just before preprocessing anyways. In an ideal state, the data_labeler wouldn't be here anywhere in profiler anyways.

dataprofiler/profilers/profile_builder.py
JGSweets: Could you explain further? this code no longer returns self._profile

dataprofiler/profilers/profile_builder.py
```diff
@@ -768,9 +784,8 @@ def tqdm(l):
         print(notification_str)
         
         for col in tqdm(df.columns):
-            profile[col].update_column_profilers(clean_sampled_dict[col], pool)
```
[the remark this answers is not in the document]
JGSweets: it is already changed?

dataprofiler/profilers/profile_builder.py
JGSweets: moved

dataprofiler/profilers/profile_builder.py
JGSweets: updated

gme5078 (approved):

lettergram (approved):

PR #173, opened 2021-04-21
Add safety when spinning up pool

lettergram:
Currently, the pool can generate and pass in enough data it'll lock the machine. This PR adds protections ensuring the max number of processes stays low enough to ensure the memory isn't 100% full.


dataprofiler/profilers/profile_builder.py
```diff
@@ -689,13 +689,18 @@ def tqdm(l):
...
-                        
-        # Format the data
-        notification_str = "Finding the Null values in the columns..."
+                
+        # Generate pool and estimate datasize
         pool = None
+        est_data_size = df[:50000].memory_usage(index=False, deep=True).sum()
+        est_data_size = (est_data_size / min(50000, len(df))) * len(df)
+        pool, pool_size = utils.generate_pool(
+            max_pool_size=None, data_size=est_data_size, cols=len(df.columns))
```
lettergram, line 698: Basically, sample and then estimate the data_size. It's accurate 2-10% based on what I have seen, which is more than accurate enough to select the number of processes to use.

dataprofiler/profilers/utils.py
```diff
@@ -4,6 +4,7 @@
 import random
 import math
 import warnings
+import psutil
```
lettergram, line 7: cross platform, the functions used should be pretty safe.

dataprofiler/profilers/utils.py
```diff
@@ -163,34 +164,76 @@ def partition(data, chunk_size):
...
+    :param cols: columns of the dataset
+    :type cols: int
+    :return suggested_pool_size: suggeseted pool size 
+    :rtype suggested_pool_size: int
+    """
+    
+    # Return if there's no df_size
+    # Only do multiprocessing is over 1Mb
+    if data_size is None and data_size < 1000000:
+        return None
```
lettergram, line 182: Single thread if 1Mb or less

dataprofiler/profilers/utils.py
```diff
@@ -163,34 +164,76 @@ def partition(data, chunk_size):
...
+    :type data_size: int
+    :param cols: columns of the dataset
+    :type cols: int
+    :return suggested_pool_size: suggeseted pool size 
+    :rtype suggested_pool_size: int
+    """
+    
+    # Return if there's no df_size
+    # Only do multiprocessing is over 1Mb
+    if data_size is None or data_size < 1000000:
```
lettergram, line 181: less than 1mb will use single thread

dataprofiler/profilers/profile_builder.py
```diff
@@ -689,13 +689,18 @@ def tqdm(l):
                     options=structured_options
                 )
                 new_cols.add(col)
-                        
-        # Format the data
-        notification_str = "Finding the Null values in the columns..."
+                
+        # Generate pool and estimate datasize
         pool = None
+        est_data_size = df[:50000].memory_usage(index=False, deep=True).sum()
```
AnhTruong, line 695: should we sample df first?

lettergram: This does just take the first 50k, but it's fast. Below I generate an estimate which basically identifies the avg bytes per row then multiplies by the number of rows. It's pretty accurate, and the estimated space requirements per-process doubles the actual estimates (meaning the margin for error is 50%).

dataprofiler/profilers/profile_builder.py
gme5078, line 698: I don't think it should be creating a pool if mulitprocessing is off

gme5078 (changes requested):

dataprofiler/profilers/profile_builder.py
```diff
@@ -704,14 +704,19 @@ def tqdm(l):
                     options=structured_options
                 )
                 new_cols.add(col)
-                        
-        # Format the data
-        notification_str = "Finding the Null values in the columns..."
+                
+        # Generate pool and estimate datasize
         pool = None
         if options.structured_options.multiprocess.is_enabled:
```
JGSweets, line 710: my concern with setting this here rather than determined in options is that if a user sets True in MP, it will force it to False.
Maybe default in options should be `None` and if `None` then it determines here or options does auto determination if is `None`

lettergram: This really doesn't change the functionality? I just moved something that was previously out of the for loop into it?

I agree it might need to move, but isn't that a seperate PR as it's changing functionality of options (as opposed to protecting threading)?

JGSweets: I would say the data size check is new functionality, this would potentially change the outcome of options as opposed to right now where a user sets the option and it uses it.

lettergram: I suggested an update to the readme:

> **multiprocess** - Option to enable multiprocessing. Automatically selects the optimal number of processes to utilize based on the dataset and system constraints.

I think this covers the concern. Personally, I think we need to modify this to allow users to set their pool size. In which case, then I'd suggest we utilize what they specified and if they say `None` then we utilize the proposed method.

lettergram: Going to move modifications to the pool based on dataset size to separate PR, which would also need to include user defined process counts

AnhTruong (approved):

ChrisWallace2020 (approved):

Issue #164, opened 2021-04-20
Create an unstructured Compiler which combines the TextProfiler and the Unstructured Data Labeling


JGSweets:
**Is your feature request related to a problem? Please describe.**
Need Unstructured profiling

**Describe the outcome you'd like:**
Compiler which combines: TextProfiler and UnstructuredDataLabelerProfile to return a profile.

**Additional context:**
starting code, may not be exact same in the end.
```python
class UnstructuredCompiler(BaseColumnProfileCompiler):
    
    # NOTE: these profilers are ordered. Test functionality if changed.
    _profilers = [
        TextProfiler,
        UnstructuredDataLabelerProfile
    ]
    
    @property
    def profile(self):
        profile = {}
        for profiler in self._profiles.values():
            profile[profiler.profile_name].update(profiler.profile)
        return profile
```

PR #183, opened 2021-04-23
add UnstructuredCompiler and tests

JGSweets:
Addresses: https://github.com/capitalone/DataProfiler/issues/164


dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -226,3 +231,23 @@ def profile(self):
             profile["data_label"] = profiler.data_label
             profile["statistics"].update(profiler.profile)
         return profile
+
+
+class UnstructuredCompiler(BaseCompiler):
```
JGSweets, line 236: The compiler addition

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,19 +71,71 @@ def test_add_profilers(self):
...
     def test_update_match_are_abstract(self):
         six.assertCountEqual(
             self,
             {'profile'},
-            col_pro_compilers.BaseColumnProfileCompiler.__abstractmethods__
+            col_pro_compilers.BaseCompiler.__abstractmethods__
         )
 
 
+class TestUnstructuredCompiler(unittest.TestCase):
```
JGSweets, line 96: compiler test

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,19 +71,71 @@ def test_add_profilers(self):
...
+                          '1234']}}
+
+        output_profile = compiler.profile
+
+        # because vocab uses a set, it will be random order every time, hence
+        # we need to sort to check exact match between profiles
+        if ('statistics' in output_profile
+                and 'vocab' in output_profile['statistics']):
+            output_profile['statistics']['vocab'] = \
+                sorted(output_profile['statistics']['vocab'])
```
ChrisWallace2020, line 134: I understand the order is random, but do we need the `if` statement here to check that `statistics` and `vocab` are present? Should this be somewhat known/deterministic since its a controlled test?

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,19 +71,71 @@ def test_add_profilers(self):
...
+
+        self.maxDiff = None
```
ChrisWallace2020, line 136: What is this for?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -226,3 +231,23 @@ def profile(self):
...
+
+    # NOTE: these profilers are ordered. Test functionality if changed.
+    _profilers = [
+        TextProfiler,
+        UnstructuredDataLabelerProfile,
```
gme5078, line 241: I think maybe we should have uniformity in terms of Unstructured Profile class names and type (like the variable "type"). Maybe both labeled with the Unstructured prefix?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -226,3 +231,23 @@ def profile(self):
...
+    ]
+    _option_class = BaseOption
```
gme5078, line 243: This is just a place holder?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -36,7 +36,7 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
     Abstract numerical column profile subclass of BaseColumnProfiler. Represents
     a column in the dataset which is a text column. Has Subclasses itself.
     """
-    col_type = None
+    type = None
```
gme5078, line 39: Just curious: does numerical stats even need type?

dataprofiler/tests/profilers/test_column_profile_compilers.py
JGSweets: If they are not there, i want it to fail in the self.assertDictEqual 
which should give decently detailed results over the whole profile rather than just the key missing

JGSweets: I'm not particular about it, but was the intent

dataprofiler/tests/profilers/test_column_profile_compilers.py
JGSweets: I can remove, but basically it sets it so there's no limit on output of the profile dict if there's a failed assertion. Will detail the entire difference rather than clip if too long

dataprofiler/profilers/column_profile_compilers.py
JGSweets: Possibly, could we address this after this PR? I'll create an issue

dataprofiler/profilers/column_profile_compilers.py
JGSweets: yes, will add a todo

dataprofiler/profilers/numerical_column_stats.py
JGSweets: I would need to validate, but I'm not positive it does.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -103,64 +105,64 @@ def __add__(self, other):
...
+            if self._profiles[profile_key].thread_safe:
+
+                try:  # Add update function to be applied on the pool
+                    multi_process_dict[profile_key] = pool.apply_async(
+                        self._profiles[profile_key].update, (df_series,))
+                except Exception as e:  # Attempt again as a single process
+                    self._profiles[profile_key].thread_safe = False
+
+            if not self._profiles[profile_key].thread_safe:
+                single_process_list.append(profile_key)
```
lettergram, line 142: I think `profile_key` even further abstracts the confusing nature of what a "profile" is, IMO. 

I think @JGSweets mentioned something such as `inspectors` (so, it'd be `int_inspector` or `categorical_inspector`). 

Also instead of `profile_key` maybe `profile_type` or something makes more sense.

ChrisWallace2020: I personally think `profile_type` is considerably more clear

JGSweets: done

ChrisWallace2020 (dismissed):

ChrisWallace2020 (approved):

lettergram (approved):

PR #188, opened 2021-04-26
0.4.4 docs updated

gme5078:
https://gme5078.github.io/data-profiler/


lettergram:
I think this needs to be expanded, something like:

Example:

If `sample_size = 2` and `sample_array = [1.0, NULL, 2.0]` we should see the following:
```python
# The first two samples (1.0 and NULL) are used for the statistical analysis
profile = dp.Profiler(sample_array, sample_size=2) # [1.0] would be returned
```

In contrast, if we also set `min_true_samples=2` then the Data Reader will continue to read until the minimum true samples were found for the given column.
```python
# Use all samples in the statistical analysis until the number of "true" (non-NULL) values are reached (1.0, NULL, 2.0)
profile = dp.Profiler(sample_array, sample_size=2) # [1.0, 2.0] would be returned
```

Screenshot from 2021-04-26 14-26-44

ChrisWallace2020:
Small comment, super minor, in the API sidebar the order (top to bottom) is Profilers, Labelers, Data Readers, Validators, but I think it would make more sense if it went Data Readers, Labelers, Profilers, Validators, to sort of reflect the work flow of reading data, it going through a labeler, and then a profile being made.

Again not a super big deal, I think it looks super clean and comprehensive.

lettergram (approved):

PR #193, opened 2021-04-27
Counts null-rows appropriately

lettergram:


JGSweets:
partial fix to the issue, but doesn't fix the issue when it occurs in files without a system:
```python
import pandas as pd
from io import StringIO
s = """
1


4"""
pd.read_csv(StringIO(s), header=None, skip_blank_lines=False)
```
However, this causes an error in pandas that we do not yet know how to handle.
A separate PR down the road will need to fix this issue.

JGSweets (approved):

ChrisWallace2020 (approved):

Issue #167, opened 2021-04-20
Updating or Adding profiles have strange behavior for merging indices 


lettergram:
Example:
```
df_series1 = [None, 1, 2, 3]
df_series2 = [0, None, 2, 3]
profile1 = StructuredDataProfile(df_series1)
profile2 = StructuredDataProfile(df_series2)
profile3 = profile1 + profile2
profile3.profile.null_types_index['None'] # Will be [0, 1]
```

This should be something such as [0, 5] (basically, starting the index location at the end of the first profile). This could be different if the index locations were provided and there was no overlap.

ChrisWallace2020:
I'll tackle this starting 4/26

lettergram:
To preface, I'm not confident which approach is better. Simply pointing out the implications of a couple of approaches:

### Example 1 (current):
Use Index locations, union dicts
```
data1 = Data(file1)
#  [index: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
#     [0, 1, 2, None, None, None, 6, 7, 8, 9]]

data2 = Data(file2) 
 #   [index: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
 #    [None, None, None, 13, 14, 15, 16, 17, 18, 19]]
 
profile1 = profiler(data1) # null_list [3, 4, 5]
profile2 = profiler(data2) # null_list [0, 1, 2]

profile3 = profile1 + profile2
profile3.null_list # [0, 1, 2, 3, 4, 5]
profile3.sample_size # 20
profile3.null_ratio # 0.25

data3 = Data(file3)
#    [index: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
#     [21, 22, 23, 24, 25, None, None, None, None]] 

profile3.update(data3) # null_list [6, 7, 8, 9]
profile3.null_list # [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] <---- Is this accurate?
profile3.sample_size # 30
profile3.null_ratio # 0.3333

data4 = Data(file4)
#    [index: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
#     [21, 22, 23, 24, 25, None, None, None, None]] 

profile3.update(data4) # initial null_list [6, 7, 8, 9]
profile3.null_list # [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] <---- Is this accurate?
profile3.sample_size # 40
[...]

PR #199, opened 2021-04-28
Handling index overlap when merging and updating profiles

ChrisWallace2020:
See #167 

Shifts indices in StructuredProfile.update_profile and StructuredProfile.__add__ to make null row index dictionary accurate.

Also fixes separate issue discovered where when updating a Profile due to the fix in a previous PR where the null rows where truncated only to consider ones that were fully sampled, at update this accidentally removed all former null row statistics and the return from update_row_statistics only reflected data given to a specific update not the entire data set.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -66,6 +66,7 @@ class StructuredDataProfile(object):
         self._min_id = None
         self._max_id = None
         self._index_shift = None
+        self._last_batch_size = None
         self.profiles = {}
                          
         if df_series is not None and len(df_series) > 0:
@@ -231,6 +232,7 @@ class StructuredDataProfile(object):
     
     def _update_base_stats(self, base_stats):
         self.sample_size += base_stats["sample_size"]
+        self._last_batch_size = base_stats["sample_size"]
         self.sample = base_stats["sample"]
         self.null_count += base_stats["null_count"]
         self.null_types = utils._combine_unique_sets(
@@ -593,6 +595,15 @@ class Profiler(object):
         return min([self._profile[col].sample_size
                     for col in self._profile], default=0)
 
+    @property
+    def _min_sampled_from_batch(self):
+        """
+        Calculates and returns the number of rows that were completely sampled
+        in the most previous batch
+        """
+        return min([self._profile[col]._last_batch_size
+                    for col in self._profile], default=0)
+
     def report(self, report_options=None):
         if not report_options:
             report_options = {
@@ -683,10 +694,10 @@ class Profiler(object):
                 shift = self._profile[column]._index_shift
                 if shift is None:
                     null_row_indices = null_row_indices.intersection(
-                        data.index[sample_ids[:self._min_col_samples_used]])
+                        data.index[sample_ids[:self._min_sampled_from_batch]])
                 else:
                     null_row_indices = null_row_indices.intersection(
-                        data.index[sample_ids[:self._min_col_samples_used]] +
+                        data.index[sample_ids[:self._min_sampled_from_batch]] +
                         shift)
 
             # Find the common null indices between the columns
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -383,10 +383,10 @@ class StructuredDataProfile(object):
             min_id = min(df_series.index)
             max_id = max(df_series.index)
         else:
-            warnings.warn("Unable to detect minimum and maximum index values. "
-                          "There may be future overlapping in indices when "
-                          "updating/merging which could result in inaccurate "
-                          "null row index reporting.")
+            warnings.warn("Unable to detect minimum and maximum index values "
+                          "for overlap detection. Updating/merging profiles "
+                          "may result in inaccurate null row index reporting "
+                          "due to unhandled overlapping indices.")
 
         # Select generator depending if sample_ids availability
         if sample_ids is None:
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -648,7 +648,7 @@ class Profiler(object):
     def _get_duplicate_row_count(self):
         return self.total_samples - len(self.hashed_row_dict)
 
-    def _update_row_statistics(self, data, sample_ids=None):
+    def _update_row_statistics(self, data, sample_ids):
         """
         Iterate over the provided dataset row by row and calculate
         the row statistics. Specifically, number of unique rows,
@@ -680,11 +680,8 @@ class Profiler(object):
             if null_type_dict:
                 null_row_indices = set.union(*null_type_dict.values())
 
-            if sample_ids is not None:
-                # If sample ids provided, only consider nulls in rows that
-                # were fully sampled
-                null_row_indices = null_row_indices.intersection(
-                    data.index[sample_ids[:self._min_col_samples_used]])
+            null_row_indices = null_row_indices.intersection(
+                data.index[sample_ids[:self._min_col_samples_used]])
 
             # Find the common null indices between the columns
             if first_col_flag:
@@ -885,12 +882,7 @@ class Profiler(object):
             pool.close()  # Close pool for new tasks
             pool.join()  # Wait for all workers to complete
 
-        # Only pass along sample ids if necessary
-        samples_for_row_stats = None
-        if min_true_samples not in [None, 0]:
-            samples_for_row_stats = np.concatenate(sample_ids)
-
-        self._update_row_statistics(df, samples_for_row_stats)
+        self._update_row_statistics(df, np.concatenate(sample_ids))
 
     def _remove_data_labelers(self):
         """
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -294,20 +294,6 @@ class StructuredDataProfile(object):
             sample_size = self._get_sample_size(df_series)
         if not min_true_samples:
             min_true_samples = self._min_true_samples
-
-        if utils.overlap(self.min_id, self.max_id,
-                         min(df_series.index), max(df_series.index)):
-            # Increment df_series index so that no overlap with current data
-            warnings.warn("Overlapping indices detected between data given "
-                          "to update_profile and profiled data, indices of "
-                          "provided data will be shifted to resolve this.")
-            if isinstance(df_series.index, pd.RangeIndex):
-                new_start = df_series.index.start + self.max_id + 1
-                new_stop = df_series.index.stop + self.max_id + 1
-                df_series.index = pd.RangeIndex(start=new_start, stop=new_stop,
-                                                step=df_series.index.step)
-            else:
-                df_series.index = [i + self.max_id + 1 for i in df_series.index]
         
         clean_sampled_df, base_stats = self.clean_data_and_get_base_stats(
             df_series=df_series, sample_size=sample_size,
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -232,14 +232,16 @@ class StructuredDataProfile(object):
             self.null_types, list(base_stats["null_types"].keys())
         )
 
-        if self.min_id is None:
-            self.min_id = base_stats["min_id"]
-        else:
-            self.min_id = min(self.min_id, base_stats["min_id"])
-        if self.max_id is None:
-            self.max_id = base_stats["max_id"]
-        else:
-            self.max_id = max(self.max_id, base_stats["max_id"])
+        if base_stats["min_id"] is not None:
+            if self.min_id is None:
+                self.min_id = base_stats["min_id"]
+            else:
+                self.min_id = min(self.min_id, base_stats["min_id"])
+        if base_stats["max_id"] is not None:
+            if self.max_id is None:
+                self.max_id = base_stats["max_id"]
+            else:
+                self.max_id = max(self.max_id, base_stats["max_id"])
 
         for null_type, null_rows in base_stats["null_types"].items():
             if type(null_rows) is list:
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -160,9 +160,9 @@ class StructuredDataProfile(object):
         other_max_id = other.max_id
         other_nti = copy.deepcopy(other.null_types_index)
         if utils.overlap(self.min_id, self.max_id, other.min_id, other.max_id):
-            warnings.warn(f"Attempted to merge profiles {self.name} and "
-                          f"{other.name} with overlapping indices. The indices "
-                          f"in {other.name} will be shifted to resolve this.")
+            warnings.warn(f"Attempted to merge profiles with overlapping "
+                          f"indices. The indices in {other.name} will be "
+                          f"shifted in the merged profile to resolve this.")
 
             other_min_id = other.min_id + self.max_id + 1
             other_max_id = other.max_id + self.max_id + 1
@@ -376,10 +376,8 @@ class StructuredDataProfile(object):
         # Record min and max index values if index is int
         min_id = None
         max_id = None
-        if isinstance(df_series.index, pd.RangeIndex):
-            min_id = df_series.index.start
-            max_id = df_series.index.stop
-        elif all([isinstance(i, int) for i in df_series.index]):
+        if isinstance(df_series.index, pd.RangeIndex) or \
+                all([isinstance(i, int) for i in df_series.index]):
             min_id = min(df_series.index)
             max_id = max(df_series.index)
         else:
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -236,13 +236,13 @@ class StructuredDataProfile(object):
             self.null_types, list(base_stats["null_types"].keys())
         )
 
-        if base_stats["min_id"] is not None:
-            if self.min_id is None:
+        if isinstance(base_stats["min_id"], int):
+            if isinstance(self.min_id, int):
                 self.min_id = base_stats["min_id"]
             else:
                 self.min_id = min(self.min_id, base_stats["min_id"])
-        if base_stats["max_id"] is not None:
-            if self.max_id is None:
+        if isinstance(base_stats["max_id"], int):
+            if isinstance(self.max_id, int):
                 self.max_id = base_stats["max_id"]
             else:
                 self.max_id = max(self.max_id, base_stats["max_id"])
@@ -284,7 +284,7 @@ class StructuredDataProfile(object):
                     (x1 <= y1 <= x2) or
                     (x1 <= y2 <= x2))
 
-        if (None not in [self.max_id, self.min_id] and
+        if (isinstance(self.min_id, int) and isinstance(self.max_id, int) and
                 all([isinstance(i, int) for i in df_series.index])):
             if overlap(self.min_id, self.max_id,
                        min(df_series.index), max(df_series.index)):
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -696,7 +696,7 @@ class Profiler(object):
                 null_in_row_count = null_in_row_count.union(null_row_indices)
 
         # If sample_ids provided, increment since that means only new data read
-        if sample_ids:
+        if sample_ids is not None:
             self.row_has_null_count += len(null_in_row_count)
             self.row_is_null_count += len(null_rows)
         else:
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -156,19 +156,23 @@ class StructuredDataProfile(object):
         )
 
         # Check if indices overlap, if they do, adjust attributes accordingly
+        other_min_id = other.min_id
+        other_max_id = other.max_id
+        other_nti = copy.deepcopy(other.null_types_index)
         if utils.overlap(self.min_id, self.max_id, other.min_id, other.max_id):
             warnings.warn(f"Attempted to merge profiles {self.name} and "
                           f"{other.name} with overlapping indices. The indices "
                           f"in {other.name} will be shifted to resolve this.")
-            other.min_id = other.min_id + self.max_id + 1
-            other.max_id = other.max_id + self.max_id + 1
+
+            other_min_id = other.min_id + self.max_id + 1
+            other_max_id = other.max_id + self.max_id + 1
 
             # Increment values (indices) in s by self.max_id + 1
             def increment(s):
                 return {x + self.max_id + 1 for x in s}
 
-            other.null_types_index = {k: increment(v) for k, v in
-                                      other.null_types_index.items()}
+            other_nti = {k: increment(v) for k, v in
+                         other.null_types_index.items()}
 
         merged_profile.name = self.name
         merged_profile._update_base_stats(
@@ -183,9 +187,9 @@ class StructuredDataProfile(object):
             {"sample": other.sample,
              "sample_size": other.sample_size,
              "null_count": other.null_count,
-             "null_types": copy.deepcopy(other.null_types_index),
-             "min_id": other.min_id,
-             "max_id": other.max_id}
+             "null_types": other_nti,
+             "min_id": other_min_id,
+             "max_id": other_max_id}
         )
         samples = list(dict.fromkeys(self.sample + other.sample))
         merged_profile.sample = random.sample(samples, min(len(samples), 5))
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -246,6 +246,7 @@ class StructuredDataProfile(object):
                           f"where null data present will be shifted forward "
                           f"when stored in profile: {self.name}")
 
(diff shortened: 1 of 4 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -322,12 +366,20 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+                "min_id": None, "max_id": None
             }
 
         # Pandas reads empty values in the csv files as nan
         df_series = df_series.apply(str)
 
+        # Record min and max index values if index is int
+        min_id = None
+        max_id = None
+        if all([isinstance(i, int) for i in df_series.index]):
```
JGSweets, line 379: should check if `pd.RangeIndex` first for speed

dataprofiler/profilers/profile_builder.py
```diff
@@ -153,18 +155,41 @@ def __add__(self, other):
...
         )
 
+        # Check if indices overlap, if they do, adjust attributes accordingly
+        other_min_id = other.min_id
+        other_max_id = other.max_id
+        other_nti = copy.deepcopy(other.null_types_index)
+        if utils.overlap(self.min_id, self.max_id, other.min_id, other.max_id):
+            warnings.warn(f"Attempted to merge profiles {self.name} and "
+                          f"{other.name} with overlapping indices. The indices "
+                          f"in {other.name} will be shifted to resolve this.")
```
JGSweets, line 165: self.name and other.name are the same

dataprofiler/profilers/profile_builder.py
```diff
@@ -153,18 +155,41 @@ def __add__(self, other):
...
+            warnings.warn(f"Attempted to merge profiles with overlapping "
+                          f"indices. The indices in {other.name} will be "
+                          f"shifted in the merged profile to resolve this.")
+
+            other_min_id = other.min_id + self.max_id + 1
+            other_max_id = other.max_id + self.max_id + 1
+
+            # Increment values (indices) in s by self.max_id + 1
+            def increment(s):
+                return {x + self.max_id + 1 for x in s}
```
JGSweets, line 172: maybe should be increment set?
Two things:
1. We should probably send in the value in which it is being incremented by
2. should take in input of value by which it is incrementing

dataprofiler/profilers/profile_builder.py
```diff
@@ -259,6 +295,14 @@ def update_profile(self, df_series, sample_size=None,
...
         if not min_true_samples:
             min_true_samples = self._min_true_samples
+
+        if utils.overlap(self.min_id, self.max_id,
+                         min(df_series.index), max(df_series.index)):
+            # Increment df_series index so that no overlap with current data
+            warnings.warn("Overlapping indices detected between data given "
+                          "to update_profile and profiled data, indices of "
+                          "provided data will be shifted to resolve this.")
+            df_series.index = [i + self.max_id + 1 for i in df_series.index]
```
JGSweets, line 305: should only alter the pd.RangeIndex if it is one as opposed to an entire list

dataprofiler/profilers/profile_builder.py
```diff
@@ -322,12 +366,25 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+        min_id = None
+        max_id = None
+        if isinstance(df_series.index, pd.RangeIndex) or \
+                all([isinstance(i, int) for i in df_series.index]):
+            min_id = min(df_series.index)
+            max_id = max(df_series.index)
+        else:
+            warnings.warn("Unable to detect minimum and maximum index values. "
+                          "As a result overlap when merging/updating cannot be "
+                          "detected and may result in inaccurate profile.")
```
JGSweets, line 386: an inaccurate

dataprofiler/profilers/profile_builder.py
```diff
@@ -153,18 +155,41 @@ def __add__(self, other):
             options=self.options,
         )
 
+        # Check if indices overlap, if they do, adjust attributes accordingly
+        other_min_id = other.min_id
+        other_max_id = other.max_id
+        other_nti = copy.deepcopy(other.null_types_index)
+        if utils.overlap(self.min_id, self.max_id, other.min_id, other.max_id):
```
JGSweets, line 162: should we have the warning here if can't detect overlap cause strings etc?

dataprofiler/profilers/profile_builder.py
```diff
@@ -153,18 +155,37 @@ def __add__(self, other):
...
+            warnings.warn(f"Attempted to merge profiles with overlapping "
+                          f"indices. The indices in {other.name} will be "
+                          f"shifted in the merged profile to resolve this.")
+
+            other_min_id = other.min_id + self.max_id + 1
+            other_max_id = other.max_id + self.max_id + 1
+
+            other_nti = {k: utils.increment(v, self.max_id + 1) for k, v in
+                         other.null_types_index.items()}
```
lettergram, line 171: double plus one?

ChrisWallace2020: The top 2 lines are incrementing the other.min_id and other.max_id, the bottom line increments every row index in other.null_type_index, so there is no double since these are for different attributes

dataprofiler/profilers/profile_builder.py
```diff
@@ -619,11 +680,8 @@ def _update_row_statistics(self, data, sample_ids=None):
             if null_type_dict:
                 null_row_indices = set.union(*null_type_dict.values())
 
-            if sample_ids is not None:
-                # If sample ids provided, only consider nulls in rows that
-                # were fully sampled
-                null_row_indices = null_row_indices.intersection(
-                    data.index[sample_ids[:self._min_col_samples_used]])
+            null_row_indices = null_row_indices.intersection(
+                data.index[sample_ids[:self._min_col_samples_used]])
```
JGSweets, line 684: i'm realizing now, that what we could do instead to not lose this speed up of the if statement, is make the add at the bottom be conditional as well.
```python
if sample_ids is not None:
    self.row_has_null_count += len(null_in_row_count)
    self.row_is_null_count += len(null_rows)
else:
```

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -661,6 +662,35 @@ def test_sample_size_passed_to_profile(self, *mocks):
         profiler._sampling_ratio = 0.2
         self.assertEqual(10000, update_mock.call_args[0][1])
 
+    def test_index_overlap_resolved_for_update_profile(self):
```
JGSweets, line 665: probably don't need to compute any of the compilers, can mock them out so we don't do data labeling, etc

ChrisWallace2020: Deja Vu, my bad

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -10,6 +10,7 @@
 
 import numpy as np
 import pandas as pd
+from io import BytesIO
```
JGSweets, line 13: builtin

JGSweets (approved):

dataprofiler/profilers/profile_builder.py
```diff
@@ -259,6 +291,20 @@ def update_profile(self, df_series, sample_size=None,
...
+        if utils.overlap(self.min_id, self.max_id,
+                         min(df_series.index), max(df_series.index)):
+            # Increment df_series index so that no overlap with current data
+            warnings.warn("Overlapping indices detected between data given "
+                          "to update_profile and profiled data, indices of "
+                          "provided data will be shifted to resolve this.")
+            if isinstance(df_series.index, pd.RangeIndex):
+                new_start = df_series.index.start + self.max_id + 1
+                new_stop = df_series.index.stop + self.max_id + 1
+                df_series.index = pd.RangeIndex(start=new_start, stop=new_stop,
```
JGSweets, line 304: realizing, will this change the external data passed in, unbeknownst to the user?

dataprofiler/profilers/profile_builder.py
```diff
@@ -322,12 +368,25 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+        min_id = None
+        max_id = None
+        if isinstance(df_series.index, pd.RangeIndex) or \
+                all([isinstance(i, int) for i in df_series.index]):
+            min_id = min(df_series.index)
+            max_id = max(df_series.index)
+        else:
+            warnings.warn("Unable to detect minimum and maximum index values. "
+                          "As a result overlap when merging/updating cannot be "
+                          "detected and may result in an inaccurate profile.")
```
lettergram, line 388: I think this can be refined such that the user understands what's going on. I don't think the phrase: "As a result overlap when merging/updating cannot be detected and may result in an inaccurate profile." is disconcerting and I'm not sure relays how they user can fix it and / or what they should be concerned about.

dataprofiler/profilers/profile_builder.py
```diff
@@ -322,12 +368,25 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
         # Pandas reads empty values in the csv files as nan
         df_series = df_series.apply(str)
 
+        # Record min and max index values if index is int
+        min_id = None
+        max_id = None
+        if isinstance(df_series.index, pd.RangeIndex) or \
+                all([isinstance(i, int) for i in df_series.index]):
+            min_id = min(df_series.index)
+            max_id = max(df_series.index)
```
lettergram, line 384: Couldn't index ranges also be strings?

dataprofiler/profilers/profile_builder.py
```diff
@@ -634,8 +695,13 @@ def _update_row_statistics(self, data, sample_ids=None):
...
 
-        self.row_has_null_count = len(null_in_row_count)
-        self.row_is_null_count = len(null_rows)
+        # If sample_ids provided, increment since that means only new data read
+        if sample_ids is not None:
+            self.row_has_null_count += len(null_in_row_count)
+            self.row_is_null_count += len(null_rows)
+        else:
+            self.row_has_null_count = len(null_in_row_count)
+            self.row_is_null_count = len(null_rows)
```
lettergram, line 704:
```
if sample_ids is None:
     self.row_has_null_count = 0
     self.row_is_null_count = 0
self.row_has_null_count += len(null_in_row_count)
self.row_is_null_count += len(null_rows)
```

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -804,6 +876,39 @@ def test_null_calculation_with_differently_sampled_cols(self):
...
+        opts = ProfilerOptions()
+        opts.structured_options.multiprocess.is_enabled = False
+        profile = dp.Profiler(data1, min_true_samples=2, samples_per_update=2,
+                              profiler_options=opts)
+        self.assertEqual(3, profile.row_has_null_count)
+        self.assertEqual(1, profile.row_is_null_count)
+        self.assertEqual(0.75, profile._get_row_has_null_ratio())
+        self.assertEqual(0.25, profile._get_row_is_null_ratio())
+
+        profile.update_profile(data2, min_true_samples=2, sample_size=2)
```
lettergram, line 906: Should have test with `min_true_samples` and without `min_true_samples`

lettergram (commented):
Update readme / docs to explain this

dataprofiler/profilers/profile_builder.py
lettergram: ^ I believe so, which is one of my concerns

dataprofiler/profilers/profile_builder.py
```diff
@@ -153,18 +155,37 @@ def __add__(self, other):
             options=self.options,
         )
 
+        # Check if indices overlap, if they do, adjust attributes accordingly
+        other_min_id = other.min_id
+        other_max_id = other.max_id
+        other_nti = copy.deepcopy(other.null_types_index)
+        if utils.overlap(self.min_id, self.max_id, other.min_id, other.max_id):
```
lettergram, line 162: I think this can be done in "update_base_stats"

dataprofiler/profilers/profile_builder.py
lettergram: I'm not following why this is necessary? This should be recalculated every time in the for loop above?

ChrisWallace2020: @JGSweets I struggled to explain the solution while you were away, please revisit this and clarify.

dataprofiler/profilers/profile_builder.py
```diff
@@ -230,7 +236,37 @@ def _update_base_stats(self, base_stats):
...
+            else:
+                self.min_id = min(self.min_id, base_min)
+        if isinstance(base_max, int):
+            if not isinstance(self.max_id, int):
+                self.max_id = base_max
+            else:
+                self.max_id = max(self.max_id, base_max)
+
+        # Update null row indices
+        for null_type, null_rows in base_nti.items():
```
ChrisWallace2020, line 269: @lettergram @JGSweets consolidated the overlap detection/index shift logic to be applied only to the null_type_index (not the index in data destructively) during _update_base_stats, so now this one solution covers both of the previous cases' scopes.

dataprofiler/profilers/profile_builder.py
JGSweets: Basically, the following code would cause a problem if we just "recalculated" it unless we alter `_update_base_stats` to take in `sample_ids`. This would require collating all base_stats prior to updating them.
However, in the current case, if sample_ids exist in this function (because min_true_samples > 0), the following code will only intersect the data that was calculated in `update_profile` (unless you maintained the list fo sample_ids between updates).
```python
            if sample_ids is not None:
                # If sample ids provided, only consider nulls in rows that
                # were fully sampled
                null_row_indices = null_row_indices.intersection(
                    data.index[sample_ids[:self._min_col_samples_used]])
```
It would not be ideal to maintain the sample_ids across all updates or between merges of data.

Example:
First profile (`profiler = dp.Profiler(data1, min_true_samples=4)`):
assume len(data1) == 10
sample_ids = 2,6,8,9
indexes = 2,6,8,9
Null indexes == 2,6
Update profile (profiler.update_profile(data2):
assume len(data2) == 10
sample_ids = 1,3,4,9
indexes = 1,3,4,9 => (updated bc overlap) => (11, 13, 14, 19)
Null indexes = 3,4 => (updated bc overlap) => 13, 14

During `update_row_statistics` the following would occur in the current scenario
{2,6,13,14}.intersect((11, 13, 14, 19)) => {13,14}

if we kept the original code, 
```python
self.row_has_null_count = 2
self.row_is_null_count = 2 
```
the new code instead does
```python
# after first profile self.row_has_null_count = self.row_is_null_count = 2
self.row_has_null_count += 2
self.row_is_null_count += 2
```
Hence why we now do:
```python
        if sample_ids is not None:
            self.row_has_null_count += len(null_in_row_count)
            self.row_is_null_count += len(null_rows)
        else:
```

JGSweets: > ```
> if sample_ids is None:
>      self.row_has_null_count = 0
>      self.row_is_null_count = 0
> self.row_has_null_count += len(null_in_row_count)
> self.row_is_null_count += len(null_rows)
> ```
we could do that, i find it a little strange that we would set something to 0 and then update, but i do see the logic.

dataprofiler/profilers/profile_builder.py
JGSweets: I believe I tested this and a `pd.RangeIndex.start` and `pd.RangeIndex.stop` have to be integers.
https://pandas.pydata.org/docs/reference/api/pandas.RangeIndex.html
confirmed above

dataprofiler/profilers/profile_builder.py
```diff
@@ -230,7 +236,37 @@ def _update_base_stats(self, base_stats):
...
+        if isinstance(base_min, int):
+            if not isinstance(self.min_id, int):
+                self.min_id = base_min
+            else:
+                self.min_id = min(self.min_id, base_min)
+        if isinstance(base_max, int):
+            if not isinstance(self.max_id, int):
+                self.max_id = base_max
+            else:
+                self.max_id = max(self.max_id, base_max)
```
JGSweets, line 266: I think this can be slightly more clear if we do the following assuming min_id is either `None` or an `int`:
```python
if self.min_id is None:
    self.min_id = base_min
elif base_min is not None and base_min < self.min_id:
    self.min_id = base_min
    
if self.max_id is None:
    self.max_id = base_max
elif base_max is not None and base_max > self.max_id:
    self.max_id = base_min
```
thoughts? not going to block the pr on this though.

JGSweets: or could do:
```python
if self.min_id is None:
    self.min_id = base_min
elif base_min is not None:
    self.min_id = min(self.min_id, base_min)
    
if self.max_id is None:
    self.max_id = base_max
elif base_max is not None:
    self.max_id = max(self.max_id, base_max)
```

dataprofiler/profilers/profile_builder.py
```diff
@@ -230,7 +236,37 @@ def _update_base_stats(self, base_stats):
...
+            warnings.warn(f"Overlapping indices detected. To resolve, indices "
+                          f"where null data present will be shifted forward "
+                          f"when stored in profile: {self.name}")
+
+            # Shift indices (min, max, and all indices in null types index
+            base_min = base_min + self.max_id + 1
+            base_max = base_max + self.max_id + 1
+
+            base_nti = {k: {x + self.max_id + 1 for x in v} for k, v in
+                        base_stats["null_types"].items()}
```
JGSweets, line 254: I was actually dreaming about this refactor earlier as I got back. perfect.

JGSweets (dismissed):

dataprofiler/profilers/profile_builder.py
JGSweets: If we do this refactor, we might want to make `self.min_id` -> `self._min_id` and `self.max_id` -> `self._max_id` bc user shouldn't really be interacting with these.

dataprofiler/profilers/profile_builder.py
lettergram: Index generally can be strings though: https://stackoverflow.com/questions/41536343/i-have-string-index-in-pandas-dataframe-how-can-i-select-by-startswith

Would this be an issue?

JGSweets: It is contained in an if statement which requires it to be an int

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -661,6 +663,78 @@ def test_sample_size_passed_to_profile(self, *mocks):
...
         self.assertEqual(10000, update_mock.call_args[0][1])
 
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnPrimitiveTypeProfileCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnStatsProfileCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnDataLabelerCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler')
+    def test_index_overlap_resolved_for_update_profile(self, *mocks):
```
AnhTruong, line 673: the function name is kind of strange. Can we remove `resolved`?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -661,6 +663,78 @@ def test_sample_size_passed_to_profile(self, *mocks):
...
+        data = pd.Series([0, None, 1, 2, None])
+        profile = StructuredDataProfile(data)
+        self.assertEqual(0, profile.min_id)
+        self.assertEqual(4, profile.max_id)
```
AnhTruong, line 677: should max_id be 3 (number 2) in this case?

AnhTruong: or you mean min_id as start_id and max_id as end_id?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -661,6 +663,78 @@ def test_sample_size_passed_to_profile(self, *mocks):
...
+        self.assertDictEqual(profile.null_types_index, {'nan': {1, 4}})
+        profile.update_profile(data)
+        # Now all indices will be shifted by max_id + 1 (5)
+        # So the 2 None will move from indices 1, 4 to 6, 9
```
AnhTruong, line 681: I can see this works for the cases with time-series data, and the processing is over time. But in case that the user chooses randomly some proportions of the same large dataset, then profile them and merge profiles, does that work?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -661,6 +663,78 @@ def test_sample_size_passed_to_profile(self, *mocks):
...
+        # Ensure merged profile included shifted indices
+        profile3 = profile1 + profile2
+        self.assertEqual(0, profile3.min_id)
+        self.assertEqual(9, profile3.max_id)
+        self.assertDictEqual(profile3.null_types_index, {'nan': {1, 4, 6, 9}})
+
+        # Ensure original profile not overwritten
+        self.assertEqual(0, profile2.min_id)
+        self.assertEqual(4, profile2.max_id)
+        self.assertDictEqual(profile2.null_types_index, {'nan': {1, 4}})
```
AnhTruong, line 707: can we add this check for profile1 as well?

AnhTruong (commented):
some initial thoughts with the unit tests

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: Yes it doesn't have to do with non-None or anything like that, its purely just to remember the range of the index.

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: I figured it was unnecessary since the test function before this one uses the same data and asserts it about "profile1" in that function, but I can add it back.

dataprofiler/profilers/profile_builder.py
lettergram: > Basically, the following code would cause a problem if we just "recalculated" it unless we alter _update_base_stats to take in sample_ids. This would require collating all base_stats prior to updating them.

I see the issue, but I don't know if this is the appropriate solution. 

> {2,6,13,14}.intersect((11, 13, 14, 19)) => {13,14} # only when sample_ids is not None

This is a different problem that needs to be addressed. Shouldn't each update be isolated, i.e. 

    First profile (profiler = dp.Profiler(data1, min_true_samples=4)):
    assume len(data1) == 10
    sample_ids = 2,6,8,9
    indexes = 2,6,8,9
    Null indexes == 2,6

    Update profile (profiler.update_profile(data2)):
    assume len(data2) == 10
    sample_ids = 1,3,4,9
    indexes = 1,3,4,9 => (updated bc overlap) => (11, 13, 14, 19)
    Null indexes = 3,4 => (updated bc overlap) => 13, 14

    {2, 3}.union({13,14}.intersect((11, 13, 14, 19))) => {2, 3, 13,14} # only when sample_ids is not None

This should be the union of all the values (which would be the correct calculation):
```
```

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: Done

dataprofiler/profilers/profile_builder.py
JGSweets: Ideally, `{13,14}.intersect((11, 13, 14, 19)) => {13, 14}` is calculated with base_stats

followed by a merge function which does:
```python
null_indexes = {2, 6}.union({13, 14})
self.row_has_null_count += len(updated_null_in_row_count)
self.row_is_null_count += len(updated_null_rows)
```
which would mirror what occurs in the lower level profilers of calculating all stats on the new dataset being added, followed by a merge function.

The problem is the redesign required to achieve this becomes complicated. We need to do the calc with `base_stats`, but this means compiling all base stats until all are processed bc of `sample_ids` and min_true_samples. Feasible, but would be what I see currently as the new technical requirement.

However, the current state is no different than the previous code despite an if statement. I would suggest that any such change should occur in a subsequent PR so that users can have the fix in now for the current errors in Nulls.

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Done

JGSweets (changes requested):
Actually, realizing now, since indexes moved to only in update_base_stats, these will be incorrect only using the first half.
```python
null_row_indices = null_row_indices.intersection(
                    data.index[sample_ids[:self._min_col_samples_used]])
```

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -804,6 +879,58 @@ def test_null_calculation_with_differently_sampled_cols(self):
...
+                'ColumnStatsProfileCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnDataLabelerCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler')
+    def test_null_row_stats_correct_after_updates(self, *mocks):
+        data1 = pd.DataFrame([[1, None],
+                             [1, 1],
+                             [None, None],
+                             [None, 1]])
+        data2 = pd.DataFrame([[None, None],
```
lettergram, line 894: Check: 0, 1, N

lettergram (approved):

PR #200, opened 2021-04-28
Improved Data Labeler Notebook

gme5078:
Added more sections and moved some around to be more cohesive.

This is a work in progress only because I want to change the datasets (to be smaller and more digestible) and the display of the printed info.


gme5078:
This is no longer a work in progress. Fire away with your reviews.

AnhTruong (commented):
can we clear all outputs first?

AnhTruong:
`Setting Parameters` section is not really necessary. If we need to keep it, we can some how integrate it to the above components, e.g., setting params for preprocessor, post processor, model, then plug them into the system.

Prediction of DataLabeler at the cell level may come after the prediction results at the column level. For example, we can say, even the DP provides results at the column level, DataLabeler also provides the results at the cell level when data is read row by row. But IMO, this’s somewhat overlap with the later part when we use DataLabeler to predict the results for text at different levels.

`df = data.data.rename({'comment': 'UNKNOWN'}, axis=1)`: we can set the default_label now. So, it’d be better to do it that way, the users don’t have to change the column name.

`Let's display the training results of the last epoch`: looks like the results are shown already in the previous cell?

`Saving and Loading a Data Labeler`: I think it’s covered in the above section when the labeler is saved after being trained on a structured dataset, and then loaded to predict

lettergram (commented):
* Should say `labeler` not `data_labeler`
* Clear output
* IMO select a different dataset for training, it's not a good one to show the low accuracy IMO. All of those samples are floats, every column. You'd be better off selecting a dataset with int, float, string or something to that effect.

AnhTruong (approved):

lettergram (approved):

ChrisWallace2020 (approved):

AnhTruong (approved):

Issue #175, opened 2021-04-21
TextProfiler Unstructured needs options for setting properties of the profiler


JGSweets:
**Is your feature request related to a problem? Please describe.**
Need the ability to set properties for unstructured's TextProfiler

For now, only create `TextProfilerOption`.

**Describe the outcome you'd like:**

LIkely final outcome after all options are setup.
```python
options = ProfilerOptions()
options.unstructured_options.text.is_case_sensitive = ...
options.unstructured_options.text.stop_words = [...]
options.unstructured_options.text.words.is_enabeld = ...
options.unstructured_options.text.vocab.is_enabled = ...
.
.
.
```

Desired outcome after this issue is solved:
```python
options = TextProfilerOption(...)
TextProfiler(name=..., options=options)
```

PR #209, opened 2021-05-04
Add options for TextProfilerOptions

AnhTruong:


dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,67 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+        :vartype is_case_sensitive: bool
+        :ivar stop_words: option set for stop words.
+        :vartype stop_words: Union[None, list(str)]
+        :ivar words: option set for word update.
+        :vartype words: BooleanOption
+        :ivar vocab: option set for vocab update.
+        :vartype vocab: BooleanOption
+        """
+        self.is_case_sensitive = True
+        self.stop_words = None
```
ChrisWallace2020, line 781: Do these add to our existing stop words or overwrite them (if the user specifies them via options)

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,67 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+        :rtype: list(str)
+        """
+        errors = []
+        if not isinstance(self.is_case_sensitive, bool):
+            errors.append("{}.is_case_sensitive must be a Boolean."
+                          .format(variable_path))
+
+        if self.stop_words is not None and \
+                (not isinstance(self.stop_words, list)
+                 or len(self.stop_words) < 1
```
ChrisWallace2020, line 812: Could the user ever want no stop words? This prohibits them from doing so but is there a use case for this?

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,118 @@ def test_case_sensitivity(self):
...
+        text_profile.update(sample)
+        profile = text_profile.profile
+
+        expected_word_count = {'sentence': 1, 'Test': 1}
+        expected_vocab = [' ', ',', '.', '!', 'T', 'h', 'i', 's', 'a',
+                          'e', 't', 'n', 'c']
+        self.assertDictEqual(expected_word_count, profile['word_count'])
+        self.assertCountEqual(expected_vocab, profile['vocab'])
+
+        # is_case_sensitive options
```
ChrisWallace2020, line 224: Can you add a test where the data is case sensitive, and there are 2 words that are the same, but one is capitalized, and ensure that it still treats them as 2 separate words?

AnhTruong: good point! I've added it

dataprofiler/profilers/profiler_options.py
AnhTruong: makes sense. I removed that check. The users can still set empty list if they want.

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,66 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
             variable_path=variable_path + '.structured_options')
 
         return errors
+
+
+class TextProfilerOptions(BaseOption):
```
JGSweets, line 766: Should be a boolean option by base because it should have `is_enabled` as such should be initialized properly in init

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,66 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+    def __init__(self):
```
JGSweets, line 767: properties should be settable on construction with defaults to enabled

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,66 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+        """
+        Constructs the TextProfilerOption object with default values.
+
+        :ivar is_case_sensitive: option set for case sensitivity.
```
JGSweets, line 771: need to have the `is_enabled`

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,66 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+        Checks to see if a property is enabled or not and returns boolean.
+
+        :param prop: The option to check if it is enabled
+        :type prop: String
+        :return: Whether or not the property is enabled
+        :rtype: Boolean
+        """
+        return BaseColumnOptions.is_prop_enabled(self, prop)
+
+    def _validate_helper(self, variable_path='TextProfilerOptions'):
```
JGSweets, line 796: instead of referencing `BaseColumnOptions`, lets just rename it so it is universal.
e.g. `BaseInspectorOptions` (the naming convention that will come to pass)
or for now could say `BaseProfilerOptions` if you don't want to say `inspector` yet

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -28,6 +28,9 @@ def __init__(self, name, options=None):
         # TODO: Add line length
         #self.line_length = {'max': None, 'min': None,...} #numeric stats mixin?
 
+        if options:
+            options.validate()
+
```
JGSweets, line 33: i dont believe we do this in any other profile. However, we should validate that the options coming in our in fact `TextProfilerOptions`

JGSweets: the validate is happening at the top level of profiler

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -92,11 +95,22 @@ def __init__(self, name, options=None):
...
+                self.__calculations.update(
+                    {"vocab": TextProfiler._update_vocab})
+            if options.words.is_enabled:
+                self.__calculations.update(
+                    {"words": TextProfiler._update_words})
+        else:
+            self.__calculations = {
+                "vocab": TextProfiler._update_vocab,
+                "words": TextProfiler._update_words,
+            }
```
JGSweets, line 113: this is covered by `BaseColumnProfiler._filter_properties_w_options(self.__calculations, options)`

JGSweets: please revert

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,129 @@ def test_case_sensitivity(self):
             expected_word_count = {'hello': 1, 'name': 1, 'grant': 2, 'bob': 1,
                                    'friends': 1}
             self.assertDictEqual(expected_word_count, profile['word_count'])
+
```
JGSweets, line 168: we need tests validating setting the options work, and they should be in their own tests like the other options
We can keep the integration tests which validate they options affect things how we would expect them here.

JGSweets: I would also suggest making these tests smaller as each test should limit to a specific scope

JGSweets (changes requested):
Generally good, but some tweaks should improve overall.

dataprofiler/profilers/profiler_options.py
JGSweets: actually after further inspection should be based off of `BaseColumnOptions` which we can rename

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,66 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+        """
+        Validates the options do not conflict and cause errors.
+
+        :param variable_path: current path to variable set.
+        :type variable_path: str
+        :return: list of errors (if raise_error is false)
+        :rtype: list(str)
+        """
+        errors = []
```
JGSweets, line 805: will need the same validate checks as the option validates:
e.g.
```python
        if not variable_path:
            variable_path = self.__class__.__name__

```

dataprofiler/profilers/profiler_options.py
AnhTruong: changed class name and inherits from that

dataprofiler/profilers/profiler_options.py
AnhTruong: I think if we want, we have to do it for every other classes. Currently, some of them have, some do not have. Not sure we want to do it in this pR.

dataprofiler/profilers/unstructured_text_profile.py
AnhTruong: true, changed it back

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -378,5 +378,16 @@ def test_warning_tf_multiple_dp_with_update(self):
...
+        test_root_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+        test_dir = os.path.join(test_root_path, 'data')
+        path = os.path.join(test_dir, 'csv/diamonds.csv')
+
+        data = dp.Data(path)
+        profile_options = dp.ProfilerOptions()
+        profile_options.set({"data_labeler.is_enabled": False,})
+        print('running dp1')
+        profile1 = dp.Profiler(data, profiler_options=profile_options)
+
```
JGSweets, line 391: is this unintentional?

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,61 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
             variable_path=variable_path + '.structured_options')
 
         return errors
+
+
+class TextProfilerOptions(BaseInspectorOptions):
+
+    def __init__(self, is_enabled=True):
```
JGSweets, line 768: should also include: `is_case_sensitive`, `stop_words`, not sure about dealing with `words` and `voacb`

AnhTruong: not sure it's needed now. At least some other classes don't have the default values now. And like you said, I don't know how to make it for words and vocab.

dataprofiler/profilers/unstructured_text_profile.py
AnhTruong: we can put it to UnstructuredProfiler later when we refactor all of this. For now, it's in the constructor.

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: removed

dataprofiler/profilers/profiler_options.py
JGSweets: `is_case_sensitive` and `stop_words` should still be able to be set through the constructor

dataprofiler/profilers/unstructured_text_profile.py
JGSweets: as this is in a feature branch, i don't think it is necessary to do this. I suggest removing.

JGSweets: should do the same options check the other classes do

dataprofiler/tests/profilers/profiler_options/test_base_column_options.py
```diff
@@ -1,11 +1,11 @@
-from dataprofiler.profilers.profiler_options import BaseColumnOptions
+from dataprofiler.profilers.profiler_options import BaseInspectorOptions
```
JGSweets, line 1: rename file

dataprofiler/profilers/unstructured_text_profile.py
JGSweets: e.g.
```python
if options and not isinstance(options, FloatOptions):
    raise ValueError("FloatColumn parameter 'options' must be of type"
                     " FloatOptions.")
```

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
...
+            dict(prop='stop_words',
+                 value_list=[2, 'a', [1, 2]],
+                 errors=[
+                     "TextProfilerOptions.stop_words must be None "
+                     "or list of strings."
+                 ]),
+        ]
+
+        # # this code can be abstracted to limit code everywhere else
+        # # AKA, for loop below could be abstracted to a utils func
```
ChrisWallace2020, line 101: Is this supposed to be here? If so, I would maybe move it directly above the loop a few lines below, rather than before the default configuration check

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
...
+            # non errors
+            dict(prop='is_enabled', value_list=[False, True], errors=[]),
+            dict(prop='is_case_sensitive', value_list=[False, True], errors=[]),
+            dict(prop='stop_words',
+                 value_list=[None, ['word1', 'word2'], []],
+                 errors=[]),
+            dict(prop='words', value_list=[False, True], errors=[]),
+            dict(prop='vocab', value_list=[False, True], errors=[]),
+
+            # errors
```
ChrisWallace2020, line 91: Only one attribute being tested with errors seems a little light compared to the non errors, could you add some error cases for is_enabled, is_case_sensitive, words, and vocab?

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,95 @@ def test_case_sensitivity(self):
...
+        options.stop_words = ['hello', 'sentence', 'is', 'a']
+
+        text_profile = TextProfiler("Name", options=options)
+        sample = pd.Series(["This is test, a Test sentence.!!!"])
+        text_profile.update(sample)
+        profile = text_profile.profile
+
+        expected_word_count = {'This': 1, 'Test': 1, 'test': 1}
+        expected_vocab = [' ', ',', '.', '!', 'T', 'h', 'i', 's', 'a',
+                          'e', 't', 'n', 'c']
```
ChrisWallace2020, line 216: What happened to the comma, period, and explanation point here, because they were not in the provided list of stop words, but also weren't present in the word_count dictionary?

dataprofiler/profilers/profiler_options.py
```diff
@@ -184,18 +184,18 @@ def _validate_helper(self, variable_path='HistogramOption'):
         return errors
 
 
-class BaseColumnOptions(BooleanOption):
+class BaseInspectorOptions(BooleanOption):
 
-    def __init__(self):
+    def __init__(self, is_enabled=True):
```
JGSweets, line 189: nice!

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,64 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+        :ivar words: option set for word update.
+        :vartype words: BooleanOption
+        :ivar vocab: option set for vocab update.
+        :vartype vocab: BooleanOption
+        """
+        self.is_case_sensitive = is_case_sensitive
+        self.stop_words = stop_words
+        self.words = BooleanOption(is_enabled=True)
+        self.vocab = BooleanOption(is_enabled=True)
+        super().__init__(is_enabled=is_enabled)
```
JGSweets, line 788: not gong to prevent on this but, but i would suggest putting this as the first call. There are other cases we have in which it is needed to call after, but when this is last, we open the option for something in the base's construction to override the the init. This could ultimately be dangerous for this case specifically, but not guaranteed to happen.
e.g. if a parent class decided to add `self.stop_words = False` in its `__init__` it would override the setting here since it is called last while one would assume that the child class's props in the `init` would be the final to be set.

dataprofiler/profilers/profiler_options.py
```diff
@@ -761,3 +761,64 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+
+        errors = super()._validate_helper(variable_path=variable_path)
+
+        if not isinstance(self.is_case_sensitive, bool):
+            errors.append("{}.is_case_sensitive must be a Boolean."
+                          .format(variable_path))
+
+        if self.stop_words is not None and \
+                (not isinstance(self.stop_words, list)
+                 or not all(
```
JGSweets, line 810: not needed change, but it might be more readable if instead of breaking at the `all` we break at the for:
e.g.
```python
if (self.stop_words is not None and
        (not isinstance(self.stop_words, list)
         or not all(isinstance(item, str)
                    for item in self.stop_words))):
```
thoughts?

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
+from dataprofiler.profilers.profiler_options import TextProfilerOptions, \
+    BooleanOption
+from dataprofiler.tests.profilers.profiler_options.test_base_inspector_options \
+     import TestBaseInspectorOptions
```
JGSweets, line 4: is this proper?

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
...
+        # validate, variable path being passed
+        expected_error = ("type object 'test.is_case_sensitive' has no "
+                          "attribute 'is_enabled'")
+        with self.assertRaisesRegex(AttributeError, expected_error):
+            option._set_helper({'is_case_sensitive.is_enabled': True}, 'test')
+
+        expected_error = ("type object 'test.stop_words' has no "
+                          "attribute 'is_enabled'")
+        with self.assertRaisesRegex(AttributeError, expected_error):
+            option._set_helper({'stop_words.is_enabled': True}, 'test')
```
JGSweets, line 32: I think we are missing the two tests for:
`vocab.is_enabled` and `option.words.is_enabled`

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
...
+        # Treat is_case_sensitive and stop_words as BooleanOption
+        expected_error = ("type object 'is_case_sensitive' has no attribute "
+                          "'is_enabled'")
+        with self.assertRaisesRegex(AttributeError, expected_error):
+            option.set({'is_case_sensitive.is_enabled': True})
+
+        expected_error = ("type object 'stop_words' has no attribute "
+                          "'is_enabled'")
+        with self.assertRaisesRegex(AttributeError, expected_error):
+            option.set({'stop_words.is_enabled': True})
```
JGSweets, line 72: should also check for `words` and `vocab`

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
...
+
+        params_to_check = [
+            # non errors
+            dict(prop='is_enabled', value_list=[False, True], errors=[]),
+            dict(prop='is_case_sensitive', value_list=[False, True], errors=[]),
+            dict(prop='stop_words',
+                 value_list=[None, ['word1', 'word2'], []],
+                 errors=[]),
+            dict(prop='words', value_list=[False, True], errors=[]),
+            dict(prop='vocab', value_list=[False, True], errors=[]),
```
JGSweets, line 89: shouldn't this error because words/vcab needs to be a `BooleanOption`?

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
JGSweets: agreed, need tests for the other 3 cases

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
JGSweets: agreed

JGSweets: reading the code below, this cannot be abstracted anymore because it specifically look sfor hte property values `words` and `vocab`

JGSweets: I believe the code below could be reverted back to its original form if above in params_to_check are updated to use
```python
dict(prop='words', value_list=[BooleanOption(is_enabled=False), 
                               BooleanOption(is_enabled=True)], errors=[]),
dict(prop='vocab', value_list=[BooleanOption(is_enabled=False), 
                               BooleanOption(is_enabled=True)], errors=[]),
```

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
JGSweets: e.g.
```python
dict(prop='words', value_list=[BooleanOption(is_enabled=False), 
                               BooleanOption(is_enabled=True)], errors=[]),
dict(prop='vocab', value_list=[BooleanOption(is_enabled=False), 
                               BooleanOption(is_enabled=True)], errors=[]),
```

if this is done, the code below could be agnostic to parameters again and not need to be altered from its original form

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,138 @@
...
+                        msg='Errored for prop: {}, value: {}.'.format(prop,
+                                                                      value))
+                else:
+                    self.assertIsNone(
+                        validate_errors,
+                        msg='Errored for prop: {}, value: {}.'.format(prop,
+                                                                      value))
+
+        # this time testing raising an error
+        option.stop_words = 'fake word'
```
JGSweets, line 133: this only succeeds because the last param_to_check is `stop_words`
I suggest adding: `option = self.get_options()` prior to this change which will prevent other issues from occurring.

dataprofiler/tests/profilers/test_unstructured_text_profile.py
JGSweets: also, it looks like `is` and `a` were included in the vocab. Group question, should `stop` words prevent vocab from being read?

Could be decided in subsequent PR.

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,95 @@ def test_case_sensitivity(self):
...
+        options = TextProfilerOptions()
+        options.stop_words = []
+
+        text_profile = TextProfiler("Name", options=options)
+        sample = pd.Series(["This is test"])
+        text_profile.update(sample)
+        profile = text_profile.profile
+
+        expected_word_count = {'This': 1, 'is': 1, 'test': 1}
+        self.assertDictEqual(expected_word_count, profile['word_count'])
```
JGSweets, line 230: good, next time though, can just stick with the same sentence so we have consistency and all that changes is the stop words

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,95 @@ def test_case_sensitivity(self):
...
+                " TextProfilerOptions."):
+            TextProfiler("Name", options="wrong_data_type")
+
+    def test_options_default(self):
+        options = TextProfilerOptions()
+
+        text_profile = TextProfiler("Name", options=options)
+        sample = pd.Series(["This is test, a Test sentence.!!!"])
+        text_profile.update(sample)
+        profile = text_profile.profile
```
JGSweets, line 181: no need to change rn, but i would suggest testing the property of the class instead rather than the profile which could potentially change which could causes these tests to fail

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,95 @@ def test_case_sensitivity(self):
...
+
+        expected_word_count = {'sentence': 1, 'Test': 1, 'test': 1}
+        expected_vocab = [' ', ',', '.', '!', 'T', 'h', 'i', 's', 'a',
+                          'e', 't', 'n', 'c']
+        self.assertDictEqual(expected_word_count, profile['word_count'])
+        self.assertCountEqual(expected_vocab, profile['vocab'])
+
+    def test_options_case_sensitive(self):
+        options = TextProfilerOptions()
+        options.is_case_sensitive = False
```
JGSweets, line 191: should add a comment saying we are assuming the default is tested in previous tests

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,95 @@ def test_case_sensitivity(self):
             expected_word_count = {'hello': 1, 'name': 1, 'grant': 2, 'bob': 1,
                                    'friends': 1}
             self.assertDictEqual(expected_word_count, profile['word_count'])
-        
\ No newline at end of file
+
+    def test_text_profile_with_wrong_options(self):
+        with self.assertRaisesRegex(ValueError,
```
JGSweets, line 170: good

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -164,4 +165,95 @@ def test_case_sensitivity(self):
...
+                "TextProfiler parameter 'options' must be of type"
+                " TextProfilerOptions."):
+            TextProfiler("Name", options="wrong_data_type")
+
+    def test_options_default(self):
```
JGSweets, line 175: My one concern with these tests is all of them are single samples. if we could break them up such that they are multiple samples, would be better,
i.e. just turn something like `pd.Series(["This is test, a Test sentence.!!!"])` -> `pd.Series(["This is test,", " a Test sentence.!!!"])`

JGSweets (changes requested):
Missing some tests and needs some slight test alterations for robustness. After added will approve!

dataprofiler/tests/profilers/test_unstructured_text_profile.py
AnhTruong: usually, word-processing doesn't care about punctuations. So, they are all excluded from word update.
About the consistency of word and vocab updates on the character, we can discuss further, but may be in the next PR.

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: added three more tests

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: moved the comments down, and revert the functions

dataprofiler/profilers/profiler_options.py
AnhTruong: not sure if it has any effect now, but moved to the first call to prevent potential issues

dataprofiler/profilers/profiler_options.py
AnhTruong: looks better, changed it

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: can you elaborate?

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: added

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: added

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: revert the function

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: added

dataprofiler/tests/profilers/test_unstructured_text_profile.py
AnhTruong: changed the sample to be consistent with other tests. The reason I changed it previously is to make the sample simple enough to read.

dataprofiler/tests/profilers/test_unstructured_text_profile.py
AnhTruong: added

dataprofiler/tests/profilers/test_unstructured_text_profile.py
AnhTruong: added the case with two samples

AnhTruong:
https://github.com/capitalone/DataProfiler/issues/175

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
```diff
@@ -0,0 +1,178 @@
...
+
+            # errors
+            dict(prop='is_case_sensitive',
+                 value_list=[2, 'string'],
+                 errors=[
+                     "TextProfilerOptions.is_case_sensitive must "
+                     "be a Boolean."
+                 ]),
+            dict(prop='stop_words',
+                 value_list=[2, 'a', [1, 2]],
```
JGSweets, line 125: we should also add a test for mixed list e.g. [`1, 'a']`

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
JGSweets: import being on the next line, I'm not sure what python standards suggest.

dataprofiler/tests/profilers/profiler_options/test_unstructured_text_profile_options.py
AnhTruong: added

JGSweets (approved):

ChrisWallace2020 (approved):

Issue #165, opened 2021-04-20
Top Level UnstructuredProfiler (User Interaction)


JGSweets:
**Is your feature request related to a problem? Please describe.**
Need unstructured profiling

**Describe the outcome you'd like:**
Should follow same format as original profiler to mimic structured, however do the same for unstructured.

**Additional context:**
Starting code (can change):
```python
class UnstructuredProfiler(BaseProfiler):
    
    def __init__(self, data, samples_per_update=None, min_true_samples=None, 
                 profiler_options=None):
        super().__init__(data,
                         samples_per_update,
                         min_true_samples,
                         profiler_options)
        raise NotImplementedError()
        
    def __add__(self, other):
        """
        Merges two Unstructured profilers together overriding the `+` operator.
        :param other: unstructured profile being add to this one.
        :type other: UnstructuredProfiler
        :return: merger of the two unstructured profilers
        """
        raise NotImplementedError()

    def _get_sample_size(self, data):
        """
        Determines the minimum sampling size for detecting column type.
        :param data: data to be profiled
        :type data: pandas.core.series.Series
        :return: integer sampling size
        :rtype: int
        """
        len_df = len(data)
        if len_df < self._min_sample_size:
            return int(len_df)
        return max(int(self._sampling_ratio * len_df), self._min_sample_size)
    
    def report(self, report_options=None):
        """Empty lines"""
        raise NotImplementedError()
        
    def _clean_and_sample_lines(self, data, min_true_samples=None, sample_ids=None):
        """
        Identify null lines and return them in a dictionary as well as
        remove any nulls data.
        
        :param df_series: a given column
        :type df_series: pandas.core.series.Series
        :param sample_size: Number of samples to use in generating the profile
        :type sample_size: int
        :param min_true_samples: Minimum number of samples required for the
            profiler
        :type min_true_samples: int
        :param sample_ids: Randomized list of sample indices
        :type sample_ids: list(list)
        :return: updated column with null removed and dictionary of null
            parameters
        :rtype: pd.Series, dict
        """
        raise NotImplementedError()
        
    def _update_base_stats(self, base_stats):
        raise NotImplementedError()

    def _update_line_statistics(self, data):
        """
        Iterate over the provided dataset line by line and calculate
        the line statistics. Specifically, number of unique lines,
        lines containing null values, and total lines reviewed. This
        function is safe to use in batches.
        :param data: a dataset
        :type data: pandas.DataFrame
        """
        raise NotImplementedError()
        
[...]

PR #211, opened 2021-05-06
Add unstructured profiler

JGSweets:
Will update when TextProfilerOptions is made. Addresses: https://github.com/capitalone/DataProfiler/issues/165

### DIFF
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -246,9 +246,13 @@ class UnstructuredCompiler(BaseCompiler):
     @property
     def profile(self):
         profile = {
-            "data_label":
-                self._profiles[UnstructuredLabelerProfile.type].profile,
-            "statistics":
-                self._profiles[TextProfiler.type].profile
+            "data_label": dict(),
+            "statistics": dict()
         }
+        if UnstructuredLabelerProfile.type in self._profiles:
+            profile["data_label"] = \
+                self._profiles[UnstructuredLabelerProfile.type].profile
+        if TextProfiler.type in self._profiles:
+            profile["statistics"] = \
+                self._profiles[TextProfiler.type].profile
         return profile
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -557,9 +557,10 @@ class UnstructuredProfiler(object):
 
     def _update_base_stats(self, base_stats):
         """
-        TBD
+        Updates the samples and line count of the class for the given dataset
+        batch.
 
-        :param base_stats: TBD
+        :param base_stats: dictionary of basic sampling / data stats
         :type base_stats: dict
         :return: None
         """
@@ -570,20 +571,31 @@ class UnstructuredProfiler(object):
     @property
     def profile(self):
         """
+        Returns the UnstructuredCompiler which compiles the low level
+        inspectors.
 
-        Returns:
-
+        :return: None
         """
         return self._profile
 
     def report(self, report_options=None):
         """
+        Returns the unstructured report based on all profiled data fed into the
+        profiler. User can specify the output_formats: (pretty, compact,
+        serializable, flat).
+            Pretty: floats are rounded to four decimal places, and lists are
+                shortened.
+            Compact: Similar to pretty, but removes detailed statistics such as
+                runtimes, label probabilities, index locations of null types,
+                etc.
+            Serializable: Output is json serializable and not prettified
+            Flat: Nested output is returned as a flattened dictionary
 
-        Args:
-            report_options:
-
-        Returns:
-
+        :var report_options: optional format changes to the report
+            `dict(output_format=<FORMAT>)`
+        :type report_options: dict
+        :return: dictionary report
+        :rtype: dict
         """
         if not report_options:
             report_options = {
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -603,7 +603,6 @@ class UnstructuredProfiler(object):
             }
 
         output_format = report_options.get("output_format", None)
-        omit_keys = report_options.get("omit_keys", [])
 
         report = OrderedDict([
             ("global_stats", {
@@ -616,7 +615,7 @@ class UnstructuredProfiler(object):
         ])
 
         report["data_stats"] = self._profile.profile
-        return _prepare_report(report, output_format, omit_keys)
+        return _prepare_report(report, output_format)
 
     @staticmethod
     def _clean_data_and_get_base_stats(data, sample_size,
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -22,7 +22,7 @@ import numpy as np
 from . import utils
 from .. import data_readers
 from .column_profile_compilers import ColumnPrimitiveTypeProfileCompiler, \
-    ColumnStatsProfileCompiler, ColumnDataLabelerCompiler
+    ColumnStatsProfileCompiler, ColumnDataLabelerCompiler, UnstructuredCompiler
 from ..labelers.data_labelers import DataLabeler
 from .helpers.report_helpers import calculate_quantiles, _prepare_report
 from .profiler_options import ProfilerOptions, StructuredOptions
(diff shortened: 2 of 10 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +444,474 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
 
 
+# TODO: remove when proper when TextProfilerOptions available
+from .profiler_options import BaseColumnOptions, BooleanOption
+class TextProfilerOptions(BaseColumnOptions):
+    def __init__(self):
+        super().__init__()
+        self.is_case_sensitive = False
+        self.vocab = BooleanOption()
+        self.words = BooleanOption()
```
JGSweets, line 454: temporary, will remove when options built

dataprofiler/profilers/profile_builder.py
```diff
@@ -719,45 +1187,6 @@ def _update_row_statistics(self, data, sample_ids=None):
             self.row_has_null_count = len(null_in_row_count)
             self.row_is_null_count = len(null_rows)
 
-    def update_profile(self, data, sample_size=None, min_true_samples=None):
```
JGSweets, line 722: just moved

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +444,474 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+    def profile(self):
+        """
+
+        Returns:
+
+        """
+        return self._profile
+
+    def report(self, report_options=None):
+        """
```
JGSweets, line 589: need to update

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +444,474 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+        :return: None
+        """
+        self.total_samples += base_stats["sample_size"]
+        self.sample = base_stats["sample"]
+        self._empty_line_count += base_stats["empty_line_count"]
+
+    @property
+    def profile(self):
+        """
+
```
JGSweets, line 582: need to update

dataprofiler/profilers/histogram_utils.py
```diff
@@ -2,329 +2,11 @@
 https://github.com/numpy/numpy/blob/v1.19.0/numpy/lib/histograms.py
 Histogram-related functions
 """
-import functools
```
JGSweets, line 5: exact same code as in numpy, we didn't need to copy and could just import instead as seen at bottom of the changes

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,476 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+        self.total_samples = 0
+        self._empty_line_count = 0
+
+        # TODO: allow set via options
+        self._sampling_ratio = 0.2
+        self._min_sample_size = 5000
+
+        # assign data labeler
+        # TODO: when available, update to use "unstructured_options"
+        self.options.structured_options.unstructured_text = TextProfilerOptions()
```
JGSweets, line 490: bc we don't have `ProfilerOptions.unstructured_options` yet

AnhTruong (commented):
I'd suggest remove the save+load part, and its tests, to be easier to review. Currently the PR is really large.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,402 @@ def test_min_max_id_properly_update(self, *mocks):
...
+        # mock out _profile
+        profiler1._profile = 1
+        profiler2._profile = 2
+
+        # merge profilers
+        merged_profile = profiler1 + profiler2
+        self.assertEqual(10, merged_profile.total_samples)
+        self.assertEqual(4, merged_profile._empty_line_count)
+        # note how sample doesn't include whitespace lines
+        self.assertCountEqual(['this', ' is', 'here\n', 'more data', 'is my'],
```
AnhTruong, line 812: should this include the string `test`?

AnhTruong (commented):
some comments on the initial review

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -246,9 +246,13 @@ class UnstructuredCompiler(BaseCompiler):
...
-                self._profiles[TextProfiler.type].profile
+            "data_label": dict(),
+            "statistics": dict()
         }
+        if UnstructuredLabelerProfile.type in self._profiles:
+            profile["data_label"] = \
+                self._profiles[UnstructuredLabelerProfile.type].profile
+        if TextProfiler.type in self._profiles:
+            profile["statistics"] = \
+                self._profiles[TextProfiler.type].profile
```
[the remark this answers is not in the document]
JGSweets: We should have done it in my previous PR... I missed that.  Just to not make this even longer, can we save it for another PR as i'm not technically editing the intention of the class itself but fixing a bug?

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -20,11 +20,20 @@ def __init__(self, data_labeler_dirpath=None, options=None):
...
+            self.data_labeler = options.data_labeler_object
+        if self.data_labeler is None:
+            data_labeler_dirpath = None
+            if options:
+                data_labeler_dirpath = options.data_labeler_dirpath
+
+            self.data_labeler = DataLabeler(
+                labeler_type='unstructured',
+                dirpath=data_labeler_dirpath,
+                load_options=None)
```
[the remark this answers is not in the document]
JGSweets: Right now the intent is to build both the unstructured and structured out separately and then refactor when we integrate. Brings unstructured out faster, but we can the clean it up soon after while it's live.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,402 @@ def test_min_max_id_properly_update(self, *mocks):
...
+        profiler = UnstructuredProfiler(None)
+        self.assertIsNone(profiler.encoding)
+        self.assertIsNone(profiler.file_type)
+        self.assertIsNone(profiler._profile)
+        self.assertIsNone(profiler._samples_per_update)
+        self.assertEqual(0, profiler._min_true_samples)
+        self.assertEqual(0, profiler.total_samples)
+        self.assertEqual(0, profiler._empty_line_count)
+        self.assertEqual(0.2, profiler._sampling_ratio)
+        self.assertEqual(5000, profiler._min_sample_size)
```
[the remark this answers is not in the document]
JGSweets: This is the top level for unstructured.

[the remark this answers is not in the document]
JGSweets: Also, this is testing the properties themselves of the class hence it makes sense to have them here

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,402 @@ def test_min_max_id_properly_update(self, *mocks):
...
+        self.assertEqual(profiler._min_true_samples, 5)
+
+        # can properties update correctly for data
+        data = pd.Series(['this', 'is my', '\n\r', 'test'])
+        profiler = UnstructuredProfiler(data)
+        self.assertEqual(4, profiler.total_samples)
+        self.assertEqual(1, profiler._empty_line_count)
+        self.assertEqual("<class 'pandas.core.series.Series'>",
+                         profiler.file_type)
+        self.assertIsNone(profiler.encoding)
```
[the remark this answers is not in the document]
JGSweets: added

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: > should this include the string test?
it won't have test because the merged samples is a random sampling between the two previous sets of samples, hence it maybe a mixture of both and not a concatenation

> can we add the case when is_case_sensitive is False, then the merging of 'Test' and 'test' will become only 'test'?
sample doesn't alter case, it takes the direct data

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,477 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+        """
+        Updates the samples and line count of the class for the given dataset
+        batch.
+
+        :param base_stats: dictionary of basic sampling / data stats
+        :type base_stats: dict
+        :return: None
+        """
+        self.total_samples += base_stats["sample_size"]
+        self.sample = base_stats["sample"]
```
ChrisWallace2020, line 568: I understand this is how sample is currently saved in `Profiler` as well, but if this is called at the end of `_update_profile_from_chunk`, which comes after `update_profile`, then that means the sample always comes from the newest data profiled, should it do a random sampling of the 2 samples concatenated, like how it does in `__add__`?

dataprofiler/profilers/column_profile_compilers.py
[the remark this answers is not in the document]
AnhTruong: that's fine for now. But we should do it later soon.

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,477 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+
+        :var report_options: optional format changes to the report
+            `dict(output_format=<FORMAT>)`
+        :type report_options: dict
+        :return: dictionary report
+        :rtype: dict
+        """
+        if not report_options:
+            report_options = {
+                "output_format": None,
```
ChrisWallace2020, line 602: Should you also set `"omit_keys"` to `[]` here?

dataprofiler/profilers/profile_builder.py
JGSweets: TBH i'm not sure, it kinda makes sense that it would only be from newest profile since that was the batch just processed via update, but could go either way. I think after these are merged we can make a decision which we apply to both cases.

dataprofiler/profilers/profile_builder.py
JGSweets: good call, just removing instead.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,417 @@ def test_min_max_id_properly_update(self, *mocks):
...
+            'other', 'behavior', 'case', 'win', 'always', 'shared', 'single',
+            'on', 'have', 'repeats', 'but', 'whereby', 'lead', 'Removing',
+            'sometimes', 'code', 'the', '6', 'trying', 'those', 'up', 'it',
+            'business', 'two', 'rigidly', 'handling', 'truth', 'not', 'for',
+            'totally', 'is', 'part', 'hours', 'application', 'new', 'into',
+            'handle', 'rule', 'need', 'or', 'to', 'pieces', 'when', 'needs',
+            'can', 'handles', 'generic', 'Report', 'coupled', 'this', 'of',
+            'usually', 'try', 'if', 'just', 'special', 'Share', 'potential',
+            'your', 'unrelated', 'you', 'abstraction', 'exact', 'by',
+        }
```
AnhTruong, line 1012: imo we should come up with less text, cause otherwise it's hard to manually inspect the results like this.

AnhTruong (commented):
some other comments

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Fair enough, it just seems weird because imagine the data changes over time with updates, the sample would only reflect the latest data, which may not be a good representation of the whole set.

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,361 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+
+        :param data: data to be profiled
+        :type data: Union[data_readers.base_data.BaseData, pandas.DataFrame]
+        :param sample_size: number of samples to profile from the data
+        :type sample_size: int
+        :param min_true_samples: minimum number of non-null samples to profile
+        :type min_true_samples
+        :return: None
+        """
+        if isinstance(data, data_readers.base_data.BaseData):
```
ChrisWallace2020, line 758: Can't this only handle `TextData` anyway, so can we make this `isinstance` check more rigorous?

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Why? Are they not needed here?

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,361 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+
+        samples = list(dict.fromkeys(self.sample + other.sample))
+        merged_profile.sample = random.sample(list(samples),
+                                              min(len(samples), 5))
+        merged_profile._profile = self._profile + other._profile
+        return merged_profile
+
+    def _get_sample_size(self, data):
+        """
+        Determines the minimum sampling size for detecting column type.
```
ChrisWallace2020, line 543: Should this be updated? Theres no columns here its just a single column, right?

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,361 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+            "sample": random.sample(list(data.values),
+                                    min(len(data), 5)),
+        }
+
+        return data, base_stats
+
+    def _update_profile_from_chunk(self, data, sample_size,
+                                   min_true_samples=None):
+        """
+        Iterate over the columns of a dataset and identify its parameters.
```
ChrisWallace2020, line 697: You aren't iterating over columns here

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,361 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+
+        :param data: a text dataset
+        :type data: Union[pd.Series, pd.DataFrame, list]
```
ChrisWallace2020, line 700: Can't `data` only be series or list here?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,394 @@ def test_min_max_id_properly_update(self, *mocks):
...
+        # test data size * 0.20 < min_sample_size < data size
+        sample_size = profiler._get_sample_size(data[:10000])
+        self.assertEqual(5000, sample_size)
+
+        # test min_sample_size > data size * 0.20
+        sample_size = profiler._get_sample_size(data)
+        self.assertEqual(10000, sample_size)
+
+        # test min_sample_size > data size * 0.10
+        profiler._sampling_ratio = 0.5
```
ChrisWallace2020, line 850: The comment doesn't match the ratio value

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,394 @@ def test_min_max_id_properly_update(self, *mocks):
...
+        sample_size = profiler._get_sample_size(data)
+        self.assertEqual(25000, sample_size)
+
+    def test_clean_data_and_get_base_stats(self, *mocks):
```
ChrisWallace2020, line 854: Can you leave comments about what got sampled in the cases

dataprofiler/profilers/profile_builder.py
AnhTruong: agree with Chris. Should do sampling on the merged samples.

JGSweets: If we decide to do this, should be in subsequent PR such that both are fixed together, there is more than just this that only looks at the recent for updates, e.g. file type etc

dataprofiler/profilers/profile_builder.py
JGSweets: o wait, i was thinking that they were for omitting columns, i can re-add

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,477 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+                            format(type(self).__name__, type(other).__name__))
+
+        merged_profile = UnstructuredProfiler(
+            data=None, samples_per_update=self._samples_per_update,
+            min_true_samples=self._min_true_samples, options=self.options
+        )
+        merged_profile.encoding = self.encoding \
+            if self.encoding == other.encoding else 'multiple files'
+        merged_profile.file_type = self.file_type \
+            if self.file_type == other.file_type else 'multiple files'
```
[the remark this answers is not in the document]
JGSweets: can be pandas dataframe as input and not a file. could be reading in csv technically but as text

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,477 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+            use_data_labeler = data_labeler_options.is_enabled
+        if use_data_labeler \
+                and data_labeler_options.data_labeler_object is not None:
+            data_labeler = data_labeler_options.data_labeler_object
+            data_labeler_options.data_labeler_object = None
+
+        # remove data labelers
+        # TODO: fix to correct options when unstructured options avail
+        if use_data_labeler:
+            if data_labeler is None:
```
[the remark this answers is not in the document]
JGSweets: this was removed with save / load, hence N/A

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,417 @@ def test_min_max_id_properly_update(self, *mocks):
...
+
+        # update with second dataset
+        profiler.update_profile(data2)
+        self.assertEqual(10, profiler.total_samples)
+        self.assertEqual(4, profiler._empty_line_count)
+        # note how sample doesn't include whitespace lines
+        self.assertCountEqual(['here\n', ' is', 'more data'], profiler.sample)
+
+
+class TestUnstructuredProfilerWData(unittest.TestCase):
```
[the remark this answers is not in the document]
JGSweets: was intending w = with

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: done

dataprofiler/profilers/profile_builder.py
JGSweets: We could limit it to be that, but technically it could be CSVData depending on the data format

JGSweets: i.e. `records` would work

dataprofiler/profilers/profile_builder.py
JGSweets: fixed

dataprofiler/profilers/profile_builder.py
JGSweets: fixed

dataprofiler/profilers/profile_builder.py
```diff
@@ -444,6 +445,361 @@ def clean_data_and_get_base_stats(df_series, sample_size,
...
+
+        if sample_size < len(data):
+            warnings.warn("The data will be profiled with a sample size of {}. "
+                          "All statistics will be based on this subsample and "
+                          "not the whole dataset.".format(sample_size))
+
+        # process the text data
+        notification_str = "Calculating the statistics... "
+        print(notification_str)
+        pool = None
```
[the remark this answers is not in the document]
JGSweets: yes, but leaving it abstracted for now

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: fixed

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: updated

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Wait why do they get set as `None` here but default to `[]` a few lines lower? Shouldn't this behavior be consistent?

ChrisWallace2020 (dismissed):

dataprofiler/profilers/profile_builder.py
AnhTruong: yeah, another PR to do that is fine.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: sample doesn't alter case, it takes the direct data

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,379 @@ def test_min_max_id_properly_update(self, *mocks):
...
+                          'v', 'r', 'o', 'a', '6', 'n', 'h', ' ', 'g', 'R', 't',
+                          'W', '.', 'm', 'c', 'l', 'e', 'p', 'w', 'S', 'd', 'D',
+                          ':']
+        self.assertCountEqual(
+            expected_vocab,
+            self.report['data_stats']['statistics'].pop('vocab'))
+
+        # vocab order doesn't matter, case insensitive, remove stop words
+        # assuming if words are correct, rest of TextProfiler is merged properly
+        # vocab order doesn't matter, case insensitive, remove stop words
```
AnhTruong, line 985: this line is the same with the above, remove?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -749,6 +752,379 @@ def test_min_max_id_properly_update(self, *mocks):
...
+            'Removing': 1, 'duplication': 1, 'that': 1, 'repeats': 1, 'the': 2,
+            'handling': 1, 'of': 1, 'exact': 1, 'same': 1, 'rule': 1, 'is': 1,
+            'also': 1, 'usually': 1, 'win': 2, 'Duplicate': 1, 'words': 1,
+            'Reply': 1, 'Share': 1, 'Report': 1}
+
+        # adapt to the stop words (brittle test)
+        stop_words = \
+            self.profiler._profile._profiles['unstructured_text']._stop_words
+        for key in list(expected_word_count.keys()):
+            if key.lower() in stop_words:
```
AnhTruong, line 1003: why we need to lower the key here? When we compare with stop_words we should keep the words as is?

JGSweets: stop words are lower, hence matching the key which is case sensitive needs matching

JGSweets: we don't actually alter the key in the dict

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: done

AnhTruong (approved):

ChrisWallace2020 (approved):

PR #216, opened 2021-05-11
Unstructured Options

ChrisWallace2020:
Implement `UnstructuredOptions` which contain `TextProfilerOptions` and `DataLabelerOptions`, and is stored within `ProfilerOptions`. Required some integration with saving and loading profiles since the labelers had to be removed and restored.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -26,7 +26,7 @@ from .column_profile_compilers import ColumnPrimitiveTypeProfileCompiler, \
 from ..labelers.data_labelers import DataLabeler
 from .helpers.report_helpers import calculate_quantiles, _prepare_report
 from .profiler_options import ProfilerOptions, StructuredOptions, \
-    UnstructuredOptions, TextProfilerOptions
+    UnstructuredOptions
 
 
 class StructuredDataProfile(object):
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -788,6 +788,7 @@ class UnstructuredProfiler(object):
         :rtype: DataLabeler
         """
         data_labeler = None
+        data_labeler_options = None
 
         # determine if the data labeler is enabled
         use_data_labeler = True
@@ -796,7 +797,7 @@ class UnstructuredProfiler(object):
             use_data_labeler = data_labeler_options.is_enabled
 
         # remove the data labeler from options
-        if use_data_labeler \
+        if use_data_labeler and data_labeler_options is not None \
                 and data_labeler_options.data_labeler_object is not None:
             data_labeler = data_labeler_options.data_labeler_object
             data_labeler_options.data_labeler_object = None
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -26,7 +26,7 @@ from .column_profile_compilers import ColumnPrimitiveTypeProfileCompiler, \
 from ..labelers.data_labelers import DataLabeler
 from .helpers.report_helpers import calculate_quantiles, _prepare_report
 from .profiler_options import ProfilerOptions, StructuredOptions, \
-    TextProfilerOptions
+    UnstructuredOptions, TextProfilerOptions
 
 
 class StructuredDataProfile(object):
@@ -464,9 +464,12 @@ class UnstructuredProfiler(object):
         :type options: ProfilerOptions Object
         :return: Profiler
         """
+
         if not options:
-            options = ProfilerOptions()
-        elif not isinstance(options, ProfilerOptions):
+            options = UnstructuredOptions()
+        elif isinstance(options, ProfilerOptions):
+            options = options.unstructured_options
+        elif not isinstance(options, UnstructuredOptions):
             raise ValueError("The profile options must be passed as a "
                              "ProfileOptions object.")
 
@@ -486,7 +489,7 @@ class UnstructuredProfiler(object):
         self._min_sample_size = 5000
 
         # assign data labeler
-        data_labeler_options = self.options.unstructured_options.data_labeler
+        data_labeler_options = self.options.data_labeler
         if data_labeler_options.is_enabled \
                 and data_labeler_options.data_labeler_object is None:
 
@@ -497,13 +500,11 @@ class UnstructuredProfiler(object):
                     dirpath=data_labeler_options.data_labeler_dirpath,
                     load_options=None)
                 self.options.set(
-                    {'unstructured_options.data_labeler.data_labeler_object':
-                         data_labeler})
+                    {'data_labeler.data_labeler_object': data_labeler})
 
             except Exception as e:
                 utils.warn_on_profile('data_labeler', e)
-                self.options.set({'unstructured_options.data_labeler.is_enabled':
-                                      False})
+                self.options.set({'data_labeler.is_enabled': False})
 
         if data is not None:
             self.update_profile(data)
@@ -737,8 +738,8 @@ class UnstructuredProfiler(object):
         print(notification_str)
         pool = None
         if self._profile is None:
-            self._profile = UnstructuredCompiler(
-                data, options=self.options.unstructured_options, pool=pool)
+            self._profile = UnstructuredCompiler(data, options=self.options,
+                                                 pool=pool)
         else:
             self._profile.update_profile(data, pool=pool)
 
@@ -783,16 +784,15 @@ class UnstructuredProfiler(object):
         """
         Helper method for removing all data labelers before saving to disk.
 
-        :return: data_labeler(s) used for unstructured labelling (if labeler
-        exists in structured options it will be returned in the 1 index)
-        :rtype: list(DataLabeler)
+        :return: data_labeler used for unstructured labelling
+        :rtype: DataLabeler
         """
         data_labeler = None
 
         # determine if the data labeler is enabled
         use_data_labeler = True
-        if self.options and isinstance(self.options, ProfilerOptions):
-            data_labeler_options = self.options.unstructured_options.data_labeler
+        if self.options and isinstance(self.options, UnstructuredOptions):
+            data_labeler_options = self.options.data_labeler
             use_data_labeler = data_labeler_options.is_enabled
 
         # remove the data labeler from options
@@ -801,20 +801,6 @@ class UnstructuredProfiler(object):
             data_labeler = data_labeler_options.data_labeler_object
             data_labeler_options.data_labeler_object = None
 
-        # determine if the structured data labeler is enabled
-        use_struct_data_labeler = True
-        if self.options and isinstance(self.options, ProfilerOptions):
-            struct_data_labeler_options = \
-                self.options.structured_options.data_labeler
-            use_struct_data_labeler = data_labeler_options.is_enabled
-
-        # remove the structured data labeler from options
-        struct_data_labeler = None
-        if use_struct_data_labeler \
-                and struct_data_labeler_options.data_labeler_object is not None:
-            struct_data_labeler = data_labeler_options.data_labeler_object
-            struct_data_labeler_options.data_labeler_object = None
-
         # remove the data labeler from the unstructured profiler
         if use_data_labeler:
             if data_labeler is None:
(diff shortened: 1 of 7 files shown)


dataprofiler/profilers/profiler_options.py
```diff
@@ -655,6 +656,67 @@ def _validate_helper(self, variable_path='DataLabelerOptions'):
         return errors
 
 
+class TextProfilerOptions(BaseInspectorOptions):
```
ChrisWallace2020, line 659: Simply moved this up

dataprofiler/profilers/profile_builder.py
```diff
@@ -840,17 +857,25 @@ def _restore_data_labelers(self, data_labeler=None):
...
                 self.options.set(
-                    {'data_labeler.data_labeler_object': data_labeler})
+                    {'unstructured_options.data_labeler.data_labeler_object':
+                         data_labeler})
 
             except Exception as e:
                 utils.warn_on_profile('data_labeler', e)
-                self.options.set({'data_labeler.is_enabled': False})
+                self.options.set({'unstructured_options.data_labeler.is_enabled':
+                                      False})
```
JGSweets, line 866: do we need to set the structured to be disabled too?

dataprofiler/profilers/profile_builder.py
```diff
@@ -883,7 +908,7 @@ def save(self, filepath=None):
             pickle.dump(data, outfile)
 
         # Restore all data labelers
-        self._restore_data_labelers(data_labeler)
+        self._restore_data_labelers(data_labelers)
```
JGSweets, line 911: could be `self._restore_data_labelers(unstruct_data_labeler, struct_data_labeler)`

dataprofiler/profilers/profile_builder.py
```diff
@@ -486,8 +486,7 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
         self._min_sample_size = 5000
 
         # assign data labeler
-        # TODO: when available, update to use "unstructured_options"
-        self.options.structured_options.unstructured_text = TextProfilerOptions()
+        self.options.unstructured_options.text = TextProfilerOptions()
```
gme5078, line 489: Old code was a placeholder, now that you're implementing this, we can get rid of this line :)

gme5078 (changes requested):

dataprofiler/profilers/profile_builder.py
```diff
@@ -802,29 +802,46 @@ def _remove_data_labelers(self):
...
+
         # remove the data labeler from the unstructured profiler
         if use_data_labeler:
             if data_labeler is None:
                 data_labeler = \
                     self._profile._profiles['data_labeler'].data_labeler
             self._profile._profiles['data_labeler'].data_labeler = None
 
-        return data_labeler
+        return [data_labeler, struct_data_labeler]
```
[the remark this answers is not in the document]
ChrisWallace2020: Good idea, will do

dataprofiler/profilers/profile_builder.py
```diff
@@ -1399,13 +1425,23 @@ def _remove_data_labelers(self):
...
         data_labelers = {}
 
-        # Delete data labeler for profiler
-        data_labeler_options = self.options.structured_options.data_labeler
-        if data_labeler_options.is_enabled \
-                and data_labeler_options.data_labeler_object is not None:
-            data_labelers["data_labeler"] = data_labeler_options \
-                                            .data_labeler_object
-            data_labeler_options.data_labeler_object = None
+        # Delete data labeler from structured options
```
JGSweets, line 1428: instead fo delete, lets use the word remove like above. also, lets have the same format for both, either both determine if enabled then remove like this or determine if enabled and then remove for struct, then for unstruct.
this could potentially be a util function that both could use.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1444,11 +1480,35 @@ def _restore_data_labelers(self, data_labelers={}):
...
+                        labeler_type='unstructured',
+                        dirpath=data_labeler_options.data_labeler_dirpath,
+                        load_options=None)
+                self.options.set(
+                    {'structured_options.data_labeler.data_labeler_object':
+                         data_labeler})
+
+            except Exception as e:
+                utils.warn_on_profile('data_labeler', e)
+                self.options.set({'structured_options.data_labeler.is_enabled':
```
JGSweets, line 1510: unstructured_options?

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: We should, yes

dataprofiler/profilers/profile_builder.py
```diff
@@ -1444,11 +1480,35 @@ def _restore_data_labelers(self, data_labelers={}):
...
                 self.options.set(
-                    {'data_labeler.data_labeler_object': data_labeler})
+                    {'structured_options.data_labeler.data_labeler_object':
+                         data_labeler})
                 
             except Exception as e:
                 utils.warn_on_profile('data_labeler', e)
-                self.options.set({'data_labeler.is_enabled': False})
+                self.options.set({'structured_options.data_labeler.is_enabled':
+                                      False})
```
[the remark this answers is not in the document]
ChrisWallace2020: Yes, will add

dataprofiler/profilers/profiler_options.py
```diff
@@ -753,41 +827,38 @@ def _validate_helper(self, variable_path='ProfilerOptions'):
...
+                    variable_path, prop, prop_check[prop].__name__))
+            else:
+                errors += self.properties[prop]._validate_helper(
+                    variable_path=(variable_path + '.' + prop
+                                   if variable_path else prop))
         return errors
 
 
-class TextProfilerOptions(BaseInspectorOptions):
+class ProfilerOptions(BaseOption):
```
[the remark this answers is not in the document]
JGSweets: yup, unfortunate that git did it like this

dataprofiler/profilers/profile_builder.py
```diff
@@ -1444,11 +1480,35 @@ def _restore_data_labelers(self, data_labelers={}):
...
+                        dirpath=data_labeler_options.data_labeler_dirpath,
+                        load_options=None)
+                self.options.set(
+                    {'structured_options.data_labeler.data_labeler_object':
+                         data_labeler})
+
+            except Exception as e:
+                utils.warn_on_profile('data_labeler', e)
+                self.options.set({'structured_options.data_labeler.is_enabled':
+                                      False})
```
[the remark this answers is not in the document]
ChrisWallace2020: Yes I do, at the end of `UnstructuredProfiler._restore_data_labelers` there is a comment `# Restore structured data labeler if there was one` with that logic

dataprofiler/profilers/profile_builder.py
```diff
@@ -840,17 +857,25 @@ def _restore_data_labelers(self, data_labeler=None):
...
         # Restore data labelers for unstructured data labeling
         if use_data_labeler:
             data_labeler_profile = self._profile._profiles['data_labeler']
             data_labeler_profile.data_labeler = data_labeler
 
+        # Restore structured data labeler if there was one
+        if data_labelers and data_labelers[1] and self.options and \
+                isinstance(self.options, ProfilerOptions):
+            self.options.structured_options.data_labeler.data_labeler_object \
+                = data_labelers[1]
```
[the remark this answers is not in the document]
ChrisWallace2020: This will be fixed shortly, but originally it was the unstructured option's labeler in `0`, and the structured one in `1`, this will be fixed to be made more clear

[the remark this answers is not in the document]
lettergram: I think you do that, you should add a comment -- sounds good though.

dataprofiler/profilers/profile_builder.py
```diff
@@ -486,8 +486,7 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
         self._min_sample_size = 5000
 
         # assign data labeler
-        # TODO: when available, update to use "unstructured_options"
-        self.options.structured_options.unstructured_text = TextProfilerOptions()
+        self.options.unstructured_options.text = TextProfilerOptions()
         data_labeler_options = self.options.structured_options.data_labeler
```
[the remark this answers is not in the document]
ChrisWallace2020: That is an oversight on my part, that should be unstructured on 490

dataprofiler/tests/profilers/profiler_options/test_unstructured_options.py
```diff
@@ -0,0 +1,127 @@
...
+                                    r"must be a\(n\) DataLabelerOptions\."):
+            option.validate()
+
+        # Both incorrect
+        option.text = BooleanOption()
+        self.assertCountEqual(option.validate(raise_error=False),
+                              ["UnstructuredOptions.text must be a(n) "
+                               "TextProfilerOptions.",
+                               "UnstructuredOptions.data_labeler must be a(n) "
+                               "DataLabelerOptions."])
```
JGSweets, line 114: similar to above with the set / set_helper

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -938,7 +938,8 @@ def setUpClass(cls):
 
         # turn off data labeler because if model changes, results also change
         profiler_options = ProfilerOptions()
-        profiler_options.set({'data_labeler.is_enabled': False})
+        profiler_options.structured_options.data_labeler.is_enabled = False
+        profiler_options.unstructured_options.data_labeler.is_enabled = False
```
gme5078, line 942: I think we need integration tests with both the labelers set since both profiles interact with both data labelers

gme5078 (changes requested):

ChrisWallace2020:
Big refactor, summary of changes:
- Now, `Profiler` and `UnstructuredProfiler` both can take in either `ProfilerOptions` or `StructuredOptions`(for `Profiler`)/`UnstructuredOptions`(for `UnstructuredProfiler`)
- `Profiler.options` is stored as type `StructuredOptions` and `UnstructuredProfiler.options` is stored as type `UnstructuredOptions`

This solves the need for both profilers to remove and restore labeler objects from both, and just concern themselves with the options class that applies to the profiler.

This required extensive rewriting of any calls to `self.options` in both classes since now there is no need to go one step further into `self.options.(un)structured_options.___`.

Unrelated, I also changed the signature of `Profiler._update_profile_from_chunk` to not pass options since they are stored within `Profiler` any way, and this way it matches the signature of how it is in `UnstructuredProfiler`.

Some tests were rewritten to accomodate the refactor, the last order of business from the first wave of comments is to rewrite the `UnstructuredOptions` tests to reduce the amount of redundancy that is common when testing options.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1387,8 +1381,7 @@ def update_profile(self, data, sample_size=None, min_true_samples=None):
         if not sample_size:
             sample_size = self._get_sample_size(data)
 
-        self._update_profile_from_chunk(
-            data, sample_size, min_true_samples, self.options)
+        self._update_profile_from_chunk(data, sample_size, min_true_samples)
```
ChrisWallace2020, line 1384: Don't need to pass options anymore

dataprofiler/profilers/profile_builder.py
```diff
@@ -1194,7 +1193,7 @@ def _update_row_statistics(self, data, sample_ids=None):
             self.row_is_null_count = len(null_rows)
 
     def _update_profile_from_chunk(self, df, sample_size=None,
-                                   min_true_samples=None, options=None):
+                                   min_true_samples=None):
```
ChrisWallace2020, line 1196: Don't need to take options anymore

dataprofiler/profilers/profile_builder.py
```diff
@@ -784,16 +784,15 @@ def _remove_data_labelers(self):
...
         # determine if the data labeler is enabled
-        # TODO: fix to correct options when unstructured options avail
         use_data_labeler = True
-        if self.options and isinstance(self.options, ProfilerOptions):
-            data_labeler_options = self.options.structured_options.data_labeler
+        if self.options and isinstance(self.options, UnstructuredOptions):
+            data_labeler_options = self.options.data_labeler
             use_data_labeler = data_labeler_options.is_enabled
 
         # remove the data labeler from options
```
gme5078, line 798: I think it would rarely ever fail, but its possible that data_labeler_options would never be set, but still be called on line 800, which could cause an error. I'm not going to request this change, but just pointing it out if you want to make it more robust

ChrisWallace2020: Thats a good point, it should be fine since self.options gets set even if one isn't provided, so unless the user messes around with it and makes it something else, we should be fine. Truth be told I'm not sure why we even check the instance here, it seems it should be somewhat obvious. But you make a good point.

gme5078 (approved):

lettergram (approved):

PR #217, opened 2021-05-12
bug: histogram index out of range

scottiegarcia:
### Details

Histogram occasionally fails when values are outside of the bin edges. I've ran into this with columns that have long number identifiers. Although the data technically shouldn't be have histogram computed for them, this change keeps it from breaking by putting indices greater than the last bin edge into the last bin.

This error is caught and fails safely on profiling, but does not do so when merging profiles. An alternative would be to implement the same safe error handling to skip histogram computation on errors.

### To reproduce
```
import pandas as pd
from dataprofiler import Profiler, ProfilerOptions

vals = ['4948484949555554544949495054485054', '4948484948485749515554495054485054', '4948484948505251545552524952485054', '4948484952485048485551524952485054', '4948484948515550575556535154485054', '4948484950545549485651495054485054', '4948484954565649505449524950485054', '49484849535456545155495054485054', '4948484954515651515451495054485054', '4948484957575651505156554954485054']

df = pd.DataFrame({"column_a": vals})
df_1 = df[:5]
df_2 = df[5:]

profile_options = ProfilerOptions()

profile_options.set({
    "histogram_and_quantiles.is_enabled": True
})

profile_1 = Profiler(data=df_1, profiler_options=profile_options)
profile_2 = Profiler(data=df_2, profiler_options=profile_options)

profiles = profile_1 + profile_2

report = profiles.report()

print(report)  
```


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -310,6 +310,8 @@ def _histogram_bin_error(self, input_array):
 
         inds = np.digitize(input_array, bin_edges)
 
+        inds = np.where(inds >= len(input_array), len(input_array) - 1, inds)
```
JGSweets, line 313: Maybe, should the line `bin_edges[-1] += 1e-3` be relative to the scaling of the numbers? This line is supposed to catch said issue, but clearly it is not. 1e-3 maybe too small for the large numbers bc of overflow. Or is it an issue with merge and the min/max of the bins aren't properly being updated.

I do agree this would fix the issue, but I'm concerned that after merging, this histogram may be incorrectly binned.

Thoughts?

JGSweets (commented):
Should add the code show to bug out as a test as well.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: The fix should work, I've double checked. @JGSweets : 1e-3 doesn't work for the large numbers as you said. And, this fix doesn't affect the merge/min/max.

JGSweets: Correct, but is the code not working correctly because the bin edges are incorrect at min/max?

JGSweets: The reason for using the +/- is because it is not expensive, but using the where operation would be on a large dataset, hence a fix not using where would be ideal. One that adjust the +/- relative to the bin.

But we first need to make sure the bins are correct after merge.

JGSweets: Maybe do a swap, set temp value to current, set bin value to no.inf, do digitize, and finally swap back

AnhTruong:
Agree with Jeremy, should add the aforementioned tests.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: actually the check should be for length of `bin_edges` instead of `input_array`, i.e., should be
```
inds = np.where(inds >= len(bin_edges), len(bin_edges) - 1, inds)
```

AnhTruong: then, we can remove `bin_edges[-1] += 1e-3` before it and `bin_edges[-1] -= 1e-3` after it

AnhTruong:
@scottiegarcia

AnhTruong:
As discussed, we believe this should speed up the code fix:

```
        temp_last_edge = bin_edges[-1]
        bin_edges[-1] = np.inf

        inds = np.digitize(input_array, bin_edges)
        if temp_last_edge == np.inf:
            inds = np.minimum(inds, len(bin_edges) - 1)

        # reset the edge
        bin_edges[-1] = temp_last_edge
```
So, we only need to reassign the index when dealing with the largest max number. Otherwise, the algorithm should be one-pass over the data.
@scottiegarcia

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -481,16 +478,20 @@ def _update_histogram(self, df_series):
             self._stored_histogram['histogram']['bin_edges'] = bin_edges
 
         # update loss for the stored bins
-        histogram_loss = self._histogram_bin_error(df_series)
+        try:
+            histogram_loss = self._histogram_bin_error(df_series)
+        except BaseException:
+            histogram_loss = None
```
AnhTruong, line 484: I'm not sure we should do this try here, as it may affect the later profile merging. Or, at least some warning message is needed. 
Thoughts @JGSweets @lettergram

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -420,6 +420,29 @@ def _clean_report(report):
...
+        profile_options = ProfilerOptions()
+
+        profile_options.set({
+            "histogram_and_quantiles.is_enabled": True
+        })
+
+        profile_1 = dp.Profiler(data=df_1, profiler_options=profile_options)
+        profile_2 = dp.Profiler(data=df_2, profiler_options=profile_options)
+
+        profile_1 + profile_2
```
AnhTruong, line 443: This test should be moved to `test_int_column_profile` and also a corresponding one for `test_float_column_profile`

AnhTruong: Also, please add test for the fixed function `_histogram_bin_error`

dataprofiler/profilers/numerical_column_stats.py
JGSweets: With the expected fix, is this still needed? Can other errors occur?

dataprofiler/tests/profilers/test_profile_builder.py
scottiegarcia: moved tests to those files and added the `_histogram_bin_error` test

dataprofiler/profilers/numerical_column_stats.py
scottiegarcia: Only other way would be values below the minimum index that would break `np.digitize`

scottiegarcia: Not able to replicate with large negatives. I've removed the try/except block.

scottiegarcia:
Thanks for the feedback, let me know if any other changes are necessary/what updates I'll need to make in regards to the failing tests pointed out [here](https://github.com/capitalone/DataProfiler/pull/218)

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -270,3 +270,19 @@ def test_timeit(self):
...
+            "histogram": {
+                "bin_edges": np.array([0.0, 4.0, 8.0, 12.0, 16.0])
+            }
+        }
+
+        input_array = [0, 3, 5, 9, 11, 17]
+
+        sum_error = num_profiler._histogram_bin_error(input_array)
+
+        assert sum_error == (2-0)**2 + (2-3)**2 + (6-5)**2 + (10-9)**2 + (10-11)**2 + (17-14)**2
```
AnhTruong, line 288: can we add a comment on the value of ids array, then this error calculation become more clear?

AnhTruong: aslo, would be complete to add two more cases with max negative number and max positive number `1.8e+308`

AnhTruong (commented):
Thanks. To make the test complete, I'd recommend adding two more cases. I'll approve then. The code looks good.

scottiegarcia:
In adding tests for when `sum_error` is infinite, I ran into issues in which profile reporting would break when trying to add histogram. I've editted `_get_best_histogram_for_profile` to still return the histogram if the only histogram method used had infinite loss, instead of returning None which was breaking the downstream report creation. Let me know what y'all think of going this way about it.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -270,3 +271,104 @@ def test_timeit(self):
...
+                "bin_edges": np.array([0.0, 4.0, 8.0, 12.0, 16.0])
+            }
+        }
+
+        input_array = [0, 3, 5, 9, 11, 17]
+
+        sum_error = num_profiler._histogram_bin_error(input_array)
+
+        # Sum of errors should be difference of each input value to midpoint of bin squared
+        # bin_midpoints = [2, 6, 10, 14]
```
AnhTruong, line 290: I meant, we should also add comment on `ids = [1, 1, 2, 3, 3, 4]` for the input array

AnhTruong (dismissed):
Not related to this PR, but we're having a tensorflow version issue which causes the tests failed. Once it's fixed, this PR can be merged (given another approval).

JGSweets (dismissed):

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1019,3 +1019,29 @@ def test_histogram_option_integration(self):
...
+
+
+    def test_profile_merge_bin_edges_indices(self):
+        vals = [4948484949555554544949495054485054.0, 4948484948485749515554495054485054.0,
+                4948484948505251545552524952485054.0, 4948484952485048485551524952485054.0,
+                4948484948515550575556535154485054.0, 4948484950545549485651495054485054.0,
+                4948484954565649505449524950485054.0, 49484849535456545155495054485054.0,
+                4948484954515651515451495054485054.0, 4948484957575651505156554954485054.0]
+
+        data = pd.Series(vals)
```
JGSweets, line 1031: the profile at this level takes in a series of strings, so I think we need to do this:
```python
data = pd.Series(vals).astype(str)
```

Should fix one error, the other errors seem to be install issues that will need further debugging.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1019,3 +1019,29 @@ def test_histogram_option_integration(self):
 
         histogram, _ = num_profiler._histogram_for_profile('custom')
         self.assertEqual(100, len(histogram['bin_counts']))
+
```
JGSweets, line 1022: remove extra space

JGSweets (changes requested):
Small bug fix, unsure yet about errors in other locations.

AnhTruong:
please update this branch, currently it's out of date.

JGSweets (approved):

PR #218, opened 2021-05-13
Modify ProfilerOptions.set to require specification when setting certain options

ChrisWallace2020:
If an option being set exists within both `StructuredOptions` and `UnstructuredOptions`, and the user is trying to set it at the `ProfilerOption` level, this may unintentionally set it for both structured and unstructured cases. This is fine in most cases, e.g. disabling all data labelling, but can be incorrect in others, e.g. setting the unstructured and structured data_labeler objects to be the same. This modification checks that options that require that extra specificity have it, and raises ValueErrors where necessary, and is made to be modular so that if we discover other options later that need this level of attention they can simply be added to the set of `overlap_options` in `ProfilerOptions.set`


ChrisWallace2020:
Test run failed that didn't fail when I tested locally, adding WIP tag until resolved.

lettergram:
Works:
```
Collecting tensorflow-gpu>=2.3.0
  Downloading tensorflow_gpu-2.4.1-cp37-cp37m-manylinux2010_x86_64.whl (394.3 MB)
```
Doesn't work:
```
Collecting tensorflow-gpu>=2.3.0
  Downloading tensorflow_gpu-2.5.0-cp37-cp37m-manylinux2010_x86_64.whl (454.3 MB)
```

JGSweets (dismissed):

JGSweets (approved):

gme5078 (approved):

Issue #168, opened 2021-04-20
`PAD` incorrectly being included in data_label_representation` and `avg_predictions` fields


scottiegarcia:
**General Information:**
 - OS: macOS Catalina 10.15.7
 - Python version: 3.7
 - Library version: 0.4.2

**Describe the bug:**
`PAD` incorrectly being included in data_label_representation` and `avg_predictions` fields

**To Reproduce:**
```
import pandas as pd
from dataprofiler import Profiler, ProfilerOptions

df = pd.DataFrame([1,2,3], columns=["a"])

profile_options = ProfilerOptions()

profile_options.set({"data_labeler.is_enabled": True})

profile = Profiler(data=df, profiler_options=profile_options)

report = profile.report()

assert "PAD" in report["data_stats"]["a"]["statistics"]["avg_predictions"]   
```

**Expected behavior:**
Field should be removed before returning to user.

**Screenshots:**
N/A

**Additional context:**
B/A

PR #226, opened 2021-05-20
Remove PAD from structured labeling output

JGSweets:
Addresses: https://github.com/capitalone/DataProfiler/issues/168

Removes PAD from postprocess output, removes pad in data labeling column profile

Since the model still can return PAD as a result as is passed to the postprocessor, label_mapping was left unchanged.

### DIFF
--- a/dataprofiler/tests/labelers/test_data_processing.py
+++ b/dataprofiler/tests/labelers/test_data_processing.py
@@ -1862,18 +1862,19 @@ class TestStructCharPostprocessor(unittest.TestCase):
         results['conf'] = confidences
 
         expected_confidence_output = np.array([
-            [0, 0, 1, 0, 0],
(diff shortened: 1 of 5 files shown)


dataprofiler/tests/labelers/test_data_processing.py
```diff
@@ -1862,18 +1862,19 @@ def test_process(self):
         results['conf'] = confidences
 
         expected_confidence_output = np.array([
-            [0, 0, 1, 0, 0],
```
JGSweets, line 1865: i only took out the first column. not sure why git did this

ChrisWallace2020 (dismissed):

ChrisWallace2020 (approved):

AnhTruong (approved):

PR #234, opened 2021-05-24
Added is_structured property to BaseData and CSVData

ChrisWallace2020:
Boolean property that returns if a data object represents structured data. Raises NotImplemented for BaseData and checks that the selected data format is data frame for the CSV. Will be added to all Data readers in subsequent PRs.


dataprofiler/data_readers/csv_data.py
```diff
@@ -106,6 +106,10 @@ def quotechar(self):
     def header(self):
         return self._header
 
+    @property
+    def is_structured(self):
+        return self.data_format == "dataframe"
```
gme5078, line 111: I think this should just always be true? Isnt record format also structured?

dataprofiler/data_readers/base_data.py
```diff
@@ -72,6 +72,10 @@ def data(self):
     def data_format(self):
         return self._selected_data_format
 
+    @property
+    def is_structured(self):
```
JGSweets, line 76: let's add a comment describing the property. I know others may not, but we should add those in another task and make sure to add in future.

dataprofiler/data_readers/csv_data.py
```diff
@@ -106,6 +106,10 @@ def quotechar(self):
     def header(self):
         return self._header
 
+    @property
+    def is_structured(self):
```
JGSweets, line 110: let's add a comment describing the property. I know others may not, but we should add those in another task and make sure to add in future.

dataprofiler/data_readers/base_data.py
lettergram: Is this necessary?

dataprofiler/data_readers/csv_data.py
lettergram: ^ Yes

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -376,6 +376,24 @@ def test_len_data(self):
...
+
+        # With option specifying dataframe as data_format
+        data = CSVData(input_file_path=filename,
+                       options={"data_format": "dataframe"})
+        self.assertTrue(data.is_structured)
+
+        # With option specifying records as data_format
+        data = CSVData(input_file_path=filename,
+                       options={"data_format": "records"})
+        self.assertFalse(data.is_structured)
```
lettergram, line 396: Wouldn't we need tests for other stuff?

dataprofiler/data_readers/csv_data.py
ChrisWallace2020: @JGSweets can you confirm? I thought I remember you saying it was dependent on data format and that records were a way to strip away structure?

dataprofiler/tests/data_readers/test_csv_data.py
ChrisWallace2020: Like what? If its just dependent on data format and I checked the 2 data formats what else can I confirm?

dataprofiler/data_readers/csv_data.py
JGSweets: @ChrisWallace2020 is correct. Records is a list of strings, basically treating it as a text file.

dataprofiler/data_readers/base_data.py
JGSweets: I believe it should clarify the intent, no?

dataprofiler/data_readers/csv_data.py
JGSweets: The intent is for the profiler to know how to profile the dataset. Records would not be able to be profiled by the structured profiler

dataprofiler/data_readers/base_data.py
JGSweets: In terms of if it should be in base? Yes, I believe all our readers should clarify hence it should be in the abstract so anyone deriving from the base would know to include it.

JGSweets (approved):

dataprofiler/tests/data_readers/test_csv_data.py
lettergram: I think you need to test all the expected data types: parquet, avro, csv, json

gme5078 (approved):

dataprofiler/tests/data_readers/test_csv_data.py
JGSweets: I broke out each data type into its own task

PR #236, opened 2021-05-24
Add Unstructured Profiler to main

JGSweets:
Adds the current state of the unstructured profiler to main

NOTE: this will change the api for setting options from `profiler_options` -> `options`

```python
profile = dp.Profiler(data, options=profiler_options)
```


JGSweets:
Do we want to change version in this PR or subsequent?

gme5078:
Shouldnt we add the unstructured profile support to the wrapper first? Once that is done, I think it would another release for sure

JGSweets:
> Shouldnt we add the unstructured profile support to the wrapper first? Once that is done, I think it would another release for sure

I talked with @lettergram about this. Basically we want unstructured available and then we will add in wrapper later once we can fully validate inference support for choosing.

lettergram:
@JGSweets this PR should have the update as it updates the API, should be 0.4.7. If we get everything in before the next release, then we can update to 0.5.0.

dataprofiler/profilers/profile_builder.py
```diff
@@ -22,19 +22,20 @@
...
+    ColumnStatsProfileCompiler, ColumnDataLabelerCompiler, UnstructuredCompiler
 from ..labelers.data_labelers import DataLabeler
 from .helpers.report_helpers import calculate_quantiles, _prepare_report
-from .profiler_options import ProfilerOptions, StructuredOptions
+from .profiler_options import ProfilerOptions, StructuredOptions, \
+    UnstructuredOptions
 
 
-class StructuredDataProfile(object):
+class StructuredColProfiler(object):
```
AnhTruong, line 32: why not just `StructuredProfiler`?

JGSweets: Because it exists in the class below.

JGSweets: https://github.com/capitalone/DataProfiler/pull/236/files#diff-c869332fe6f1bea096dac03badb7365173fa62459962c2e9c020cb53341b1ba6R1062

JGSweets: line 1062 in `profile_builder.py`

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -103,64 +105,64 @@ def __add__(self, other):
...
-        
-        # Single process thread to loop through
-        for col_profile in single_process_list:
-            self._profiles[col_profile].update(df_series)
+                    self._profiles[profile_type] = returned_profile
+            except Exception as e:  # Attempt again as a single process
+                self._profiles[profile_type].thread_safe = False
+                single_process_list.append(profile_type)
+
+                # Single process thread to loop through
```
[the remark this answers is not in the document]
ChrisWallace2020: Fair enough

ChrisWallace2020 (approved):

PR #238, opened 2021-05-25
Add is_structured to remaining data readers

ChrisWallace2020:
Added is_structured property to TextData, ParquetData, JSONData (and AVROData via inheritance from JSON) as well as tests. Basically the property is true if the data_format selected is a dataframe.


JGSweets (commented):
In cases where `is_structured` is not dependent on the data, let's not read it in if it is not needed (other dependencies).

JGSweets (approved):

lettergram (approved):

PR #239, opened 2021-05-25
Abstract update profile and Structured can now take in pd.Series

JGSweets:
When abstracting the update profile, to make both profiles take in the same inputs, the structured profile needed to be allowed to take in a pd.Series.


dataprofiler/profilers/profile_builder.py
```diff
@@ -575,21 +575,54 @@ def _update_profile_from_chunk(df, profile=None, options=None, **kwargs):
...
         """
-        raise NotImplementedError()
+        encoding = None
+        file_type = None
+
+        if isinstance(data, data_readers.base_data.BaseData):
+            encoding = data.file_encoding
+            file_type = data.data_type
+            data = data.data
+        elif isinstance(data, (pd.Series, pd.DataFrame)):
```
JGSweets, line 600: instead of this, the profilers could have a variable which stipulate what types of frames can be read or a function. this would then allow for these differences.

or the bottom half of the function would be the abstraction and the top would be done by the profilers differently.

AnhTruong (approved):

gme5078 (approved):

PR #240, opened 2021-05-25
Normalize update_profile_from_chunk function signature for Profilers

JGSweets:
Ensures both have same function signature for the Profilers.

Also moves encoding / file type save to after update profile in case update_profile cancels the update.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -589,7 +589,7 @@ class BaseProfiler(object):
         raise NotImplementedError()
 
     @staticmethod
-    def _update_profile_from_chunk(df, profile=None, options=None, **kwargs):
+    def _update_profile_from_chunk(data, sample_size, min_true_samples=None):
         """
         Iterate over the dataset and identify its parameters via profiles.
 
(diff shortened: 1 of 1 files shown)


ChrisWallace2020 (approved):

dataprofiler/profilers/profile_builder.py
```diff
@@ -639,12 +639,6 @@ def update_profile(self, data, sample_size=None, min_true_samples=None):
                           "profiled.")
             return
 
-        # set file properties since data will be processed
```
JGSweets, line 642: moved in case `_update_profile_from_chunk` says the data is not profile-able with the given profiler e.g. unstructured + multicol dataframe. Hence, doesn't save the file being processed when an error is given.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1317,24 +1317,24 @@ def _update_row_statistics(self, data, sample_ids=None):
             self.row_has_null_count = len(null_in_row_count)
             self.row_is_null_count = len(null_rows)
 
-    def _update_profile_from_chunk(self, df, sample_size=None,
+    def _update_profile_from_chunk(self, data, sample_size,
```
JGSweets, line 1320: the function which calls this always gives a sample_size

JGSweets: `data` normalizes between struct and unstruct to be the same

AnhTruong (approved):

PR #242, opened 2021-05-25
Refactor save / load for profilers via BaseProfiler abstraction

JGSweets:
Refactor save / load for profilers via BaseProfiler abstraction

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -712,22 +712,19 @@ class BaseProfiler(object):
         # Restore data labeler for options
         use_data_labeler = True
         data_labeler_options = None
+        data_labeler_dirpath = None
         if self.options and isinstance(self.options, (StructuredOptions,
                                                       UnstructuredOptions)):
             data_labeler_options = self.options.data_labeler
             use_data_labeler = data_labeler_options.is_enabled
+            data_labeler_path = data_labeler_options.data_labeler_dirpath
 
         if use_data_labeler:
             try:
-                if data_labeler is None and data_labeler_options is None:
+                if data_labeler is None:
                     data_labeler = DataLabeler(
                         labeler_type=self._default_labeler_type,
-                        dirpath=None,
-                        load_options=None)
-                elif data_labeler is None:
-                    data_labeler = DataLabeler(
-                        labeler_type=self._default_labeler_type,
-                        dirpath=data_labeler_options.data_labeler_dirpath,
+                        dirpath=data_labeler_dirpath,
                         load_options=None)
                 self.options.set(
                     {'data_labeler.data_labeler_object': data_labeler})
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1067,8 +1067,6 @@ class UnstructuredProfiler(BaseProfiler):
         :type filepath: String
         :return: None
         """
-        # Set Default filepath
-
         # Create dictionary for all metadata, options, and profile
         data_dict = {
             "total_samples": self.total_samples,
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -688,6 +688,9 @@ class BaseProfiler(object):
 
         # Remove data labelers for all columns
         for profiler in profilers:
+
+            # profiles stored differently in Struct/Unstruct, this unifies
+            # labeler extraction
             if isinstance(self, StructuredProfiler):
                 profiler = profiler.profiles['data_label_profile']
 
@@ -739,6 +742,8 @@ class BaseProfiler(object):
                 utils.warn_on_profile('data_labeler', e)
                 self.options.set({'data_labeler.is_enabled': False})
 
+        # get all profiles, unstructured is a single profile and hence needs to
+        # be in a list, whereas structured is a dict and needs to be a list
         profilers = [self._profile]
         if isinstance(self, StructuredProfiler):
             profilers = self._profile.values()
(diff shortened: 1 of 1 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1352,9 +1349,7 @@ def tqdm(l):
         
         # If there are no minimum true samples, you can sort to save time
         if min_true_samples in [None, 0]:
-            # If there's a sample size, truncate
-            if sample_size is not None:
-                sample_ids[0] = sample_ids[0][:sample_size]
+            sample_ids[0] = sample_ids[0][:sample_size]
```
AnhTruong, line 1352: does that mean when sample_size is None, we take all samples? Does that the meaning of sample_size?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1393,7 +1388,7 @@ def tqdm(l):
         clean_sampled_dict = {}
         multi_process_dict = {}
         single_process_list = set()
-        if not sample_size: sample_size = len(data)
+        
         if sample_size < len(data):
```
AnhTruong, line 1392: if you delete the above line, does that work for this if statement when sample_size is None?

AnhTruong (commented):
some comments

dataprofiler/profilers/profile_builder.py
JGSweets: sample_size cannot be `None`, sample_size is always set within this function

dataprofiler/profilers/profile_builder.py
JGSweets: sample_size cannot be `none`

JGSweets: If a user doesn't set sample_size in the func `update_profile`, it is autoset within that function before it is passed to this one.

AnhTruong: kk

AnhTruong (dismissed):

dataprofiler/profilers/profile_builder.py
```diff
@@ -670,7 +706,67 @@ def _restore_data_labelers(self, data_labeler=None):
...
+        # Restore data labeler for options
+        use_data_labeler = True
+        data_labeler_options = None
+        if self.options and isinstance(self.options, (StructuredOptions,
+                                                      UnstructuredOptions)):
+            data_labeler_options = self.options.data_labeler
+            use_data_labeler = data_labeler_options.is_enabled
+
+        if data_labeler_options.is_enabled \
+                and data_labeler_options.data_labeler_object is None:
```
ChrisWallace2020, line 718: This would error if the above if statement wasn't entered, which yes, should always happen since we set options by default, but if thats the case, is the above if statement even necessary? Because the only way it doesn't get entered is unconventional user behavior, which at that point would save one unforseen error for another. Thoughts?

dataprofiler/profilers/profile_builder.py
```diff
@@ -660,7 +664,39 @@ def _remove_data_labelers(self):
...
+        # get all profiles, unstructured is a single profile and hence needs to
+        # be in a list, whereas structured is a dict and needs to be a list
+        profilers = [self._profile]
+        if isinstance(self, StructuredProfiler):
+            profilers = self._profile.values()
+
+        # Remove data labelers for all columns
+        for profiler in profilers:
+            if isinstance(self, StructuredProfiler):
+                profiler = profiler.profiles['data_label_profile']
```
ChrisWallace2020, line 692: What is this doing? This could use a comment its a bit confusing?

dataprofiler/profilers/profile_builder.py
```diff
@@ -670,7 +706,67 @@ def _restore_data_labelers(self, data_labeler=None):
...
+
+        profilers = [self._profile]
+        if isinstance(self, StructuredProfiler):
+            profilers = self._profile.values()
+
+        # Restore data labelers for all columns
+        for profiler in profilers:
+
+            if isinstance(self, StructuredProfiler):
+                profiler = profiler.profiles['data_label_profile']
```
ChrisWallace2020, line 740: This could use a comment

dataprofiler/profilers/profile_builder.py
JGSweets: i realized it was sort of a copy error between the struct / unstruct version, new should fix i hope.

dataprofiler/profilers/profile_builder.py
JGSweets: added comment

dataprofiler/profilers/profile_builder.py
JGSweets: commented

dataprofiler/profilers/profile_builder.py
```diff
@@ -714,17 +714,27 @@ def _restore_data_labelers(self, data_labeler=None):
...
-                and data_labeler_options.data_labeler_object is None:
+        if use_data_labeler:
             try:
-                if data_labeler is None:
+                if data_labeler is None and data_labeler_options is None:
+                    data_labeler = DataLabeler(
+                        labeler_type=self._default_labeler_type,
+                        dirpath=None,
+                        load_options=None)
+                elif data_labeler is None:
```
ChrisWallace2020, line 724: This if could be just one DataLabeler construction where dirpath is a variable that is None if options is None and otherwise its pulled from options

JGSweets: done

ChrisWallace2020: i.e.
`dirpath = None if data_labeler_options is None else data_labeler_options.data_labeler_dirpath`

ChrisWallace2020 (dismissed):

ChrisWallace2020 (approved):

lettergram (approved):

PR #243, opened 2021-05-27
Profiler is now a factory class and can create both structured and unstructured

ChrisWallace2020:
`Profiler` now mimics the factory class structure of `Data` for creating profilers, and can create both structured and unstructured. As of now, the logic for choosing which to create is as follows:

- If user specifies via kwarg, return the one they specify
- If user doesn't specify, and data is a `Data` object, refer to `is_structured`
- If user doesn't specify, and data is a pandas object:
        - Empty df/series -> Structured
        - Multi dimensional df -> Structured
        - If single col of string data that is varied in length or very long -> Unstructured
        - All other single col cases (uniform strings or any other dtype) -> Structured

Warnings are given every time inference is made in case the user intended for a specific profiler but didn't specify.
Test function written to ensure inference behaves as expected (laid out above).

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1520,9 +1520,9 @@ class Profiler(object):
 
         :param data: Data to be profiled
         :type data: Data class object
-        :param samples_per_update: Number of samples to use in generating profile
+        :param samples_per_update: Number of samples to use to generate profile
         :type samples_per_update: int
-        :param min_true_samples: Minimum number of samples required for the profiler
+        :param min_true_samples: Min number of samples required for the profiler
         :type min_true_samples: int
         :param options: Options for the profiler.
         :type options: ProfilerOptions Object
@@ -1538,7 +1538,7 @@ class Profiler(object):
             return UnstructuredProfiler(data, samples_per_update,
                                         min_true_samples, options)
         elif profiler_type is not None:
-            raise ValueError("Must specify 'profiler_type' to be 'structured'"
+            raise ValueError("Must specify 'profiler_type' to be 'structured' "
                              "or 'unstructured'.")
 
         # If user doesn't specify, and data is Data object, use is_structured
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1572,7 +1572,7 @@ class Profiler(object):
                 len_mean = sample_lens.mean()
                 len_range = sample_lens.max() - sample_lens.min()
                 # If strings are very long or varied, use unstructured
-                if len_mean > 20 or len_range > 20:
+                if len_mean > 25 or len_range > 25:
                     warnings.warn("Singular column string data given to "
                                   "Profiler inferred to be unstructured, will "
                                   "create UnstructuredProfiler.")
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1516,16 +1516,14 @@ class Profiler(object):
         :type profiler_type: str
         :return: BaseProfiler
         """
-        if profiler_type and profiler_type not in ["structured", "unstructured"]:
-            raise ValueError("Must specify 'profiler_type' to be 'structured' "
-                             "or 'unstructured'.")
 
         if profiler_type is None:
-            # If type unspecified, and data is Data object, use is_structured
+            # Default to structured
+            profiler_type = "structured"
+
+            # Unstructured if data is Data object and is_structured is False
             if isinstance(data, data_readers.base_data.BaseData):
-                if data.is_structured:
-                    profiler_type = "structured"
-                else:
+                if not data.is_structured:
                     profiler_type = "unstructured"
 
             # If type unspecified, and data is Pandas object, need to infer
@@ -1536,29 +1534,20 @@ class Profiler(object):
                                   "data is structured, if this is not the "
                                   "case please specify via 'profiler_type'"
                                   "kwarg.")
-                    profiler_type = "structured"
 
-                # If multi column df, data is structured
-                elif isinstance(data, pd.DataFrame) and len(data.columns) > 1:
-                    profiler_type = "structured"
-
-                # Single column df or series
-                else:
+                # Data is structured unless it is 1 column of string data
+                # that is either very long or varied in length
+                elif isinstance(data, pd.Series) or len(data.columns) == 1:
                     # Keep a series version of 1 col df for consistency
                     data_ser = data if isinstance(data, pd.Series) else data[0]
                     if data_ser.dtype == "object":
                         sample = data_ser.sample(min(5, len(data_ser)))
-                        sample_lens = sample.apply(len)
+                        sample_lens = sample.astype(str).apply(len)
                         len_mean = sample_lens.mean()
                         len_range = sample_lens.max() - sample_lens.min()
                         # If strings are very long or varied, use unstructured
-                        if len_mean > 15 or len_range > 15:
+                        if len_mean > 20 or len_range > 15:
                             profiler_type = "unstructured"
-                        else:
-                            profiler_type = "structured"
-                    else:
-                        # Structured for 1 col with non-string data
-                        profiler_type = "structured"
 
                     warnings.warn(f"Singular column data given to Profiler, "
                                   f"inferred to be {profiler_type}, if this is "
@@ -1575,3 +1564,6 @@ class Profiler(object):
         elif profiler_type == "unstructured":
             return UnstructuredProfiler(data, samples_per_update,
                                         min_true_samples, options)
+        else:
+            raise ValueError("Must specify 'profiler_type' to be 'structured' "
+                             "or 'unstructured'.")
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1636,7 +1636,8 @@ class StructuredProfiler(BaseProfiler):
         return profile
 
 
-def Profiler(data, samples_per_update=None, min_true_samples=0, options=None):
+def Profiler(data, samples_per_update=None, min_true_samples=0, options=None,
+             profiler_type=None):
     """
     Wrapper function for instantiating Structured and Unstructured Profilers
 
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1511,23 +1511,80 @@ def save(self, filepath=None):
...
+
+            # If single column, look at string length and variance
+            data_ser = data if isinstance(data, pd.Series) else data[0]
+            if data_ser.dtype == "object":
+                sample = data_ser.sample(min(5, len(data_ser)))
+                sample_lens = sample.apply(len)
+                len_mean = sample_lens.mean()
+                len_range = sample_lens.max() - sample_lens.min()
+                # If strings are very long or varied, use unstructured
+                if len_mean > 25 or len_range > 25:
```
ChrisWallace2020, line 1575: Currently this is how I deem if a list of strings is "unstructured". I figured this catches long strings that probably couldn't be labelled all on their own or if the strings have varying lengths, since most structured labels are usually not too varied in length.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1444,5 +1444,82 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+                              UnstructuredProfiler)
+
+        # User gives data that has .is_structured == True
+        empty_csv_df = dp.Data(data=empty_df, data_type="csv")
+        self.assertIsInstance(Profiler(empty_csv_df), StructuredProfiler)
+
+        # User gives data that has .is_structured == False
+        empty_csv_rec = dp.Data(data=empty_df, data_type="csv",
+                                options={"data_format": "records"})
+        self.assertIsInstance(Profiler(empty_csv_rec), UnstructuredProfiler)
```
JGSweets, line 1488: this is fine the way it is, but if we wanted to not be dependent on readers, could do:
```python
empty_struct_data = mock.Mock(spec=dp.data_readers.base_data.BaseData)
empty_struct_data.is_structured = True
```

and likewise for unstructured

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1444,5 +1444,82 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+                                        "oh wow this one is really long, like, "
+                                        "who would give this much string data, "
+                                        "this can't possibly be structured.",
+                                        "haha its short again now",
+                                        "k bye"])
+        unstruct_str_ser = unstruct_str_df[0]
+        self.assertIsInstance(Profiler(unstruct_str_df), UnstructuredProfiler)
+        self.assertIsInstance(Profiler(unstruct_str_ser), UnstructuredProfiler)
+        pass
+
```
gme5078, line 1522: Are we ready for unstructured integration tests? Like making sure the report works, update works, saving works, etc. all from a high level

dataprofiler/profilers/profile_builder.py
gme5078: Im still thinking about if this determination is appropriate, but in either case, I think the bar should be a little lower, like 15 characters or something

dataprofiler/profilers/profile_builder.py
```diff
@@ -1511,23 +1511,80 @@ def save(self, filepath=None):
...
+        if isinstance(data, data_readers.base_data.BaseData):
+            if data.is_structured:
+                return StructuredProfiler(data, samples_per_update,
+                                          min_true_samples, options)
+            else:
+                return UnstructuredProfiler(data, samples_per_update,
+                                            min_true_samples, options)
+
+        # If user doesn't specify, and data is Pandas object, need to infer
+        if isinstance(data, pd.DataFrame) or isinstance(data, pd.Series):
```
gme5078, line 1554: I know, it doesn't affect much, but I think it should be an elif

dataprofiler/profilers/profile_builder.py
```diff
@@ -1511,23 +1511,80 @@ def save(self, filepath=None):
...
+                                          min_true_samples, options)
+
+            # If multi column df, data is structured
+            if isinstance(data, pd.DataFrame) and len(data.columns) > 1:
+                return StructuredProfiler(data, samples_per_update,
+                                          min_true_samples, options)
+
+            # If single column, look at string length and variance
+            data_ser = data if isinstance(data, pd.Series) else data[0]
+            if data_ser.dtype == "object":
```
AnhTruong, line 1569: are we sure this always detect the string type?

dataprofiler/profilers/profile_builder.py
JGSweets: What do you think the standard length for an address would be or full name? 
If it is small, I don't really see an issue as much for profiling it as structured. Thoughts?

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Actually, it will be "object" if the types are mixed as well, so it doesn't properly ensure that they are all strings, should I switch it to some sort of `all(isinstance(str))` check?

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: We do have top level integration tests in `TestUnstructuredProfilerWData`, no?

dataprofiler/profilers/profile_builder.py
JGSweets: in pandas, object is a catchall that includes strings

dataprofiler/profilers/profile_builder.py
```diff
@@ -1511,23 +1511,80 @@ def save(self, filepath=None):
...
+                sample = data_ser.sample(min(5, len(data_ser)))
+                sample_lens = sample.apply(len)
```
JGSweets, line 1571: before applying `len` we might want to do `astype(str)` as the objects may not be strings, they could be dicts

dataprofiler/profilers/profile_builder.py
AnhTruong: maybe just len_range. Not sure len_mean is needed.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1444,5 +1444,82 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnPrimitiveTypeProfileCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnStatsProfileCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'ColumnDataLabelerCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler')
+    @mock.patch('dataprofiler.profilers.profile_builder.UnstructuredCompiler')
+    def test_profiler_factory_class_creates_correct_profiler(self, *mocks):
```
JGSweets, line 1468: should add a test to see what happens if we try to profile a series of dictionaries:
e.g.
```python
data = pd.Series([{'a': 1}, {'b': 2}, {'c': 3}])
```

dataprofiler/profilers/profile_builder.py
AnhTruong: in general, if we have a single col with strings, it'd probably good to consider unstructured profile imo, unless users specify the type.

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Do the astype to `data` or `sample`?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1444,5 +1444,82 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+        self.assertIsInstance(Profiler(struct_str_ser), StructuredProfiler)
+
+        # User gives 1 col data with strings of varying lengths
+        unstruct_str_df = pd.DataFrame(["short",
+                                        "this one here is a tad bit longer",
+                                        "oh wow this one is really long, like, "
+                                        "who would give this much string data, "
+                                        "this can't possibly be structured.",
+                                        "haha its short again now",
+                                        "k bye"])
```
AnhTruong, line 1517: as per my comment above, it's pretty strange if we remove the first and last short samples, then it becomes structuredProfiler (expectedly we should have the same profiler types)

dataprofiler/profilers/profile_builder.py
AnhTruong: can we add some tests with mixture of data types?

dataprofiler/profilers/profile_builder.py
JGSweets: just the sample to save processing

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: What is even the correct behavior in this case?

dataprofiler/profilers/profile_builder.py
JGSweets: @gme5078 @lettergram thoughts?

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: To be fair this inference is never going to be perfect and this example is already contrived so I'm not sure how to improve the inference further. In my opinion I should just warn them when we infer and tell them how to specify for themselves.

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: What would the correct behavior be in that case?

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: I figured len_mean would cover the case where the range is small but all the strings are extremely long (like a bunch of same-length paragraphs or sentences)

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: But what if data had mixed types, what is the proper behavior then?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1497,23 +1497,81 @@ def save(self, filepath=None):
...
+        :param profiler_type: Type of Profiler ("structured"/"unstructured")
+        :type profiler_type: str
+        :return: BaseProfiler
+        """
+        if profiler_type and profiler_type not in ["structured", "unstructured"]:
+            raise ValueError("Must specify 'profiler_type' to be 'structured' "
+                             "or 'unstructured'.")
+
+        if profiler_type is None:
+            # If type unspecified, and data is Data object, use is_structured
```
JGSweets, line 1524: if you set here: `profiler_type = "structured"` then you can remove the checks for the other cases, warnings would still need to occur like in the `len(data)` case, but less to check then.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: it would be treated like the string case. Event hough i'd  prefer it to be unstructured, this case will likely be treated as structured.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1497,23 +1497,81 @@ def save(self, filepath=None):
...
+                    profiler_type = "structured"
+
+                # If multi column df, data is structured
+                elif isinstance(data, pd.DataFrame) and len(data.columns) > 1:
+                    profiler_type = "structured"
+
+                # Single column df or series
+                else:
+                    # Keep a series version of 1 col df for consistency
+                    data_ser = data if isinstance(data, pd.Series) else data[0]
```
JGSweets, line 1548: I'm not sure if by calling data[0] if it creates a copy of the dataset. concern if the dataset was large, but need to look into it.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1497,23 +1497,81 @@ def save(self, filepath=None):
...
+        :type samples_per_update: int
+        :param min_true_samples: Min number of samples required for the profiler
+        :type min_true_samples: int
+        :param options: Options for the profiler.
+        :type options: ProfilerOptions Object
+        :param profiler_type: Type of Profiler ("structured"/"unstructured")
+        :type profiler_type: str
+        :return: BaseProfiler
+        """
+        if profiler_type and profiler_type not in ["structured", "unstructured"]:
```
JGSweets, line 1519: can move this to the bottom with the:
```python
if profiler_type == "structured":
            return StructuredProfiler(data, samples_per_update,
                                      min_true_samples, options)
        elif profiler_type == "unstructured":
            return UnstructuredProfiler(data, samples_per_update,
                                        min_true_samples, options)
```

dataprofiler/profilers/profile_builder.py
JGSweets: the string case here is covering it IMO. not sure how else we should deal with besides what you have written.

dataprofiler/profilers/profile_builder.py
JGSweets: That's why we apply the astype(str) to unify it and just treat them as strings for now. If we want to get more complicated later we can.

gme5078 (approved):

JGSweets (approved):

PR #246, opened 2021-05-27
Allow users to load from the Profiler factory class

JGSweets:
With recent refactor of the Profiler, users could no longer call Profiler.load, instead they had to do `StructuredProfiler.load`, etc.

This fixes that problem and intuitively reads the file and loads the correct profiler.


dataprofiler/profilers/profile_builder.py
```diff
@@ -806,15 +809,35 @@ def load(cls, filepath):
...
+        # remove profiler class if it exists
+        profiler_class = data.pop('profiler_class', None)
+
+        # if the user didn't load from the a given profiler class, we need
+        # to determine which profiler is being loaded.
+        profiler_cls = cls
+        if cls is BaseProfiler:
+            if 'StructuredProfiler' == profiler_class:
+                profiler_cls = StructuredProfiler
+            elif 'UnstructuredProfiler' == profiler_class:
```
ChrisWallace2020, line 825: Could you flip these equality statements? Feels weird to see <value> == <var> rather than <var> == <value>

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1167,6 +1167,9 @@ def test_save_and_load(self):
             data = dp.Data(os.path.join(data_folder, test_file))
             save_profile = UnstructuredProfiler(data)
 
+            # address case where empty_line_count is 0
+            save_profile._empty_line_count = 1
```
ChrisWallace2020, line 1171: Is this comment incorrect? It says 0 but you set it to 1

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1167,6 +1167,9 @@ def test_save_and_load(self):
             data = dp.Data(os.path.join(data_folder, test_file))
             save_profile = UnstructuredProfiler(data)
 
+            # address case where data is read in a way that has no empty lines
+            save_profile._empty_line_count = 1
```
ChrisWallace2020, line 1171: Sorry I'm still confused, you say "no empty lines" but its set to 1, can you explain this a bit further? Also why this addition is necessary?

dataprofiler/profilers/profile_builder.py
JGSweets: done

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: updated comment

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: TextData currently the entire dataset as a single string. that means `_empty_line_count == 0`. since that is also the default value, this value could be incorrectly saved if we don't test when it is not the default value

JGSweets: Can you suggest a different comment?

ChrisWallace2020: Oh so this way you check later than the _empty_line_count is correctly 0? Maybe say like "change off of default value to ensure it is correctly calculated to be 0"?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1537,6 +1540,93 @@ def test_profiler_factory_class_creates_correct_profiler(self, *mocks):
...
+    def test_save_and_load_unstructured(self):
+        data_folder = "dataprofiler/tests/data/"
+        test_files = ["txt/code.txt", "txt/sentence-10x.txt"]
+
+        for test_file in test_files:
+            # Create Data and StructuredProfiler objects
+            data = dp.Data(os.path.join(data_folder, test_file))
+            save_profile = UnstructuredProfiler(data)
+
+            # address case where empty_line_count is 0
```
ChrisWallace2020, line 1588: Change comment here too

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1537,6 +1540,93 @@ def test_profiler_factory_class_creates_correct_profiler(self, *mocks):
...
+            save_profile._empty_line_count = 1
+
+            # store the expected data_labeler
+            data_labeler = save_profile.options.data_labeler.data_labeler_object
```
ChrisWallace2020, line 1592: Why don't you do these data_labeler checks in the structured test function?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1537,6 +1540,93 @@ def test_profiler_factory_class_creates_correct_profiler(self, *mocks):
...
+                data_labeler.__class__)
+
+            # Check that reports are equivalent
+            save_report = save_profile.report()
+            load_report = load_profile.report()
+            self.assertDictEqual(save_report, load_report)
+
+            # validate both are still usable after
+            save_profile.update_profile(pd.DataFrame(['test']))
+            load_profile.update_profile(pd.DataFrame(['test']))
```
ChrisWallace2020, line 1630: Why aren't these checks in the structured test function?

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: these are copies of other tests where all i did was swap out the single line
`load_profile = Profiler.load("mock.pkl")`

Written by two diff ppl and one was written without the extra case

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: good catch

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: added data_label checks

JGSweets: added these checks too

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -431,16 +431,33 @@ def _clean_report(report):
...
 
+                # validate loaded profile has same data labeler class
+                self.assertIsInstance(
+                    load_profile.options.data_labeler.data_labeler_object,
+                    data_labeler.__class__)
+
+                # only checks first columns
+                # get first column
+                first_column_profile = list(load_profile.profile.values())[
+                    0]
```
AnhTruong, line 455: can we bring this line up?

JGSweets: fixed

AnhTruong (approved):

ChrisWallace2020 (approved):

PR #248, opened 2021-06-01
Fixes mock in test to use proper tf.keras

JGSweets:
Allows TF2.5.0 with fix to test which was incorrectly calling `'keras....` instead of `tf.keras....`

Reverts the TF version lock from: https://github.com/capitalone/DataProfiler/pull/220

### DIFF
--- a/dataprofiler/tests/labelers/test_data_labelers.py
+++ b/dataprofiler/tests/labelers/test_data_labelers.py
@@ -308,7 +308,7 @@ def mock_open(filename, *args):
         return StringIO('{}')
 
 
-@mock.patch("keras.models.load_model")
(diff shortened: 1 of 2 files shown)


dataprofiler/tests/labelers/test_data_labelers.py
```diff
@@ -308,7 +308,7 @@ def mock_open(filename, *args):
         return StringIO('{}')
 
 
-@mock.patch("keras.models.load_model")
+@mock.patch("tensorflow.keras.models.load_model")
```
gme5078, line 311: nice fix 👍

gme5078 (approved):

AnhTruong (approved):

PR #249, opened 2021-06-01
Update readme with unstructured profile

AnhTruong:


README.md
```diff
@@ -112,6 +112,33 @@ The format for a profile is below:
 }
 ```
 
+The format for an unstructured profile is below:
```
JGSweets, line 115: Should we also note the format for structured above?

README.md
```diff
@@ -283,6 +312,33 @@ print(json.dumps(report, indent=4))
...
+print(json.dumps(report, indent=4))
+```
+
+Another example of unstructured profile with pd.Series of string is given as below
+```python
+import dataprofiler as dp
+import pandas as pd
+import json
+
+text_data = pd.Series(['first string', 'second string'])
```
JGSweets, line 335: this will not work as `Profiler` assumes all `pd.Series` is structured.

Can you test all of the examples as well to ensure success?

JGSweets (changes requested):
An example given for unstructured actually provides structured results.

README.md
```diff
@@ -112,6 +112,33 @@ The format for a profile is below:
...
+            "entity_counts": {
+                "word_level": dict(int),
+                "true_char_level": dict(int),
+                "postprocess_char_level": dict(int)
+            },
+            "times": dict(float)
+        },
+        "statistics": {
+            "vocab": list(char),
+            "words": list(string),
```
JGSweets, line 135: should note these are minus the stop words

README.md
JGSweets: nvm

JGSweets:
@AnhTruong there's a lot going on in terms of commits in this PR. Instead can we do a new PR which uses a branch from off the original or rebase this branch on master? Thanks!

gme5078:
I agree with Jeremy on creating a new branch without so many commits

README.md
AnhTruong: added and tested, forgot to add the option (I commented out for another example)

README.md
AnhTruong: do we need a comment for stopwords here?

PR #255, opened 2021-06-04
Adding __add__ to the unstructured data labeler profile

gme5078:


dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -47,15 +48,91 @@ def __init__(self, data_labeler_dirpath=None, options=None):
...
+        Merges two dictionaries together and adds values together
+
+        :param first_dict: dictionary to be merged
+        :type first_dict: dict
+        :param second_dict: dictionary to be merged
+        :type second_dict: dict
+        :return: merged dictionary
+        """
+        if isinstance(first_dict, defaultdict):
+            merged_dict = defaultdict(first_dict.default_factory)
```
JGSweets, line 107: I'm a little confused by this, but I also think we can remove the else if we set merge_doct by default prior to the if

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -47,15 +48,91 @@ def __init__(self, data_labeler_dirpath=None, options=None):
...
+        else:
+            merged_dict = {}
+            
+        for item in first_dict:
+            if item in second_dict:
+                if isinstance(first_dict[item], dict) or \
```
JGSweets, line 113: Can do it in a single isinstance using a tuple

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -47,15 +48,91 @@ def __init__(self, data_labeler_dirpath=None, options=None):
...
+                        isinstance(first_dict[item], defaultdict):
+                    merged_dict[item] = self.add_nested_dictionaries(
+                        first_dict[item],
+                        second_dict[item])
+                else:
+                    merged_dict[item] = first_dict[item] + second_dict[item]
+            else:
+                merged_dict[item] = first_dict[item]
```
JGSweets, line 121: Since the dict coming in is not deep copied, this needs to be here, otherwise it is mutable

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -47,15 +48,91 @@ def __init__(self, data_labeler_dirpath=None, options=None):
...
+
+        for item in second_dict:
+            if item not in first_dict:
+                merged_dict[item] = second_dict[item]
```
JGSweets, line 125: Since the dict coming in is not deep copied, this needs to be here, otherwise it maybe mutable

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -131,7 +131,25 @@ def test_profile(self, processor_class_mock, model_class_mock):
 
         # key and value populated correctly
         self.assertDictEqual(expected_profile, profile)
+
+
+    def test_unstructured_labeler_profile_add(self):
```
JGSweets, line 136: The test should probably be a bit more complect with more labels. Technically you could make a profile and just set the values to test the merge.

JGSweets: Also we should avoid actually running the tensorflow model as the results change if the model changes

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -47,15 +48,91 @@ def __init__(self, data_labeler_dirpath=None, options=None):
...
+        merged_profile._update_percentages()
+
+        return merged_profile
+
     @property
     def label_encoding(self):
         return self.data_labeler.labels
 
+
+    def add_nested_dictionaries(self, first_dict, second_dict):
```
JGSweets, line 96: Since many profilers could use this we should consider putting it in a base class or utility func

JGSweets: Also, I wonder if a cleaner solution would be to copy the first dict and then merge in the values from the second as oppose to looping over both. Not saying it would be, but just a thought

JGSweets (dismissed):
Some cleanup of mutability, tests, and abstraction.

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
AnhTruong: ^ Adding more labels to the test

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -131,7 +131,36 @@ def test_profile(self, processor_class_mock, model_class_mock):
...
+        
+        profile2 = UnstructuredLabelerProfile()
+        sample = pd.Series(["samples"])
+        profile2.update(sample)
+
+        merged_profile = profile1 + profile2
         
+        self.assertEqual(merged_profile.entity_counts["word_level"]["UNKNOWN"],
+                         profile1.entity_counts["word_level"]["UNKNOWN"] +
+                         profile2.entity_counts["word_level"]["UNKNOWN"])
```
AnhTruong, line 157: can we add `PAD` and `TEST` labels here as well?

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -131,7 +131,36 @@ def test_profile(self, processor_class_mock, model_class_mock):
 
         # key and value populated correctly
         self.assertDictEqual(expected_profile, profile)
+
+    @mock.patch('dataprofiler.profilers.'
+                'unstructured_labeler_profile.DataLabeler')
+    def test_unstructured_labeler_profile_add(self, model_class_mock):
```
JGSweets, line 137: I would suggest that each profile is setup where an entity exists in both and the cases where it exists in one but not the other. This can be done by manually setting entity counts and the other parameters. Otherwise we arent truly testing all cases.

JGSweets: We would not need to set a labeler mock values because we wouldn't need to update the profile, just merge.

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -131,7 +131,41 @@ def test_profile(self, processor_class_mock, model_class_mock):
...
+
+        merged_profile = profile1 + profile2
+        expected_word_level = {"UNKNOWN": 8, "TEST": 2, "PAD": 1}
+        expected_true_char = {"PAD": 6}
         
+        self.assertDictEqual(merged_profile.entity_counts["word_level"], 
+                             expected_word_level)
+        self.assertDictEqual(merged_profile.entity_counts["true_char_level"], 
+                             expected_true_char)
+        self.assertDictEqual(merged_profile.entity_counts["postprocess_char_level"],
```
JGSweets, line 160: any reason we aren't testing this or mocking above?

dataprofiler/profilers/utils.py
```diff
@@ -245,3 +245,33 @@ def overlap(x1, x2, y1, y2):
...
+    :type second_dict: dict
+    :return: merged dictionary
+    """
+    merged_dict = {}
+    if isinstance(first_dict, collections.defaultdict):
+        merged_dict = collections.defaultdict(first_dict.default_factory)
+
+    for item in first_dict:
+        if item in second_dict:
+            if isinstance(first_dict[item], (dict, collections.defaultdict)):
```
JGSweets, line 265: can this just be `isinstance(first_dict[item], dict)` ? cause isn't: `isinstance(collections.defaultdict(int), dict) == True`?

JGSweets (approved):

AnhTruong (approved):

Issue #262, opened 2021-06-10
Sum not being returned in profiles by default


scottiegarcia:
**General Information:**
 - OS:  MacOS Cataline 10.15.7
 - Python version: 3.7.10
 - Library version: 0.5.1

**Describe the bug:**
`sum` is not being returned by default for numerical columns. I'd understand if it were for performance reasons, but it appears that it is still being calculated since it has an entry in the `times` dictionary.

**To Reproduce:**
```
 import pandas as pd
import dataprofiler
from dataprofiler import Profiler, ProfilerOptions

df = pd.DataFrame([1,2,3], columns=["a"])

profile_options = ProfilerOptions()

profile_options.set({
    "data_labeler.is_enabled": False,
    "histogram_and_quantiles.is_enabled": False,
})

profile = Profiler(data=df, options=profile_options)

report = profile.report()

assert "sum" in report["data_stats"]["a"]["statistics"] 
```

**Expected behavior:**
Numeric columns should return `sum` by default in reports.

JGSweets:
PR merged.

PR #264, opened 2021-06-11
Adding Sum to the Results

gme5078:
Simply added the Sum to the profile.

### DIFF
--- a/dataprofiler/tests/profilers/test_int_column_profile.py
+++ b/dataprofiler/tests/profilers/test_int_column_profile.py
@@ -23,6 +23,7 @@ class TestIntColumn(unittest.TestCase):
         self.assertEqual(profiler.match_count, 0)
         self.assertEqual(profiler.min, None)
         self.assertEqual(profiler.max, None)
+        self.assertEqual(profiler.sum, 0)
         self.assertEqual(profiler.mean, 0)
         self.assertEqual(profiler.variance, 0)
         self.assertTrue(profiler.stddev is np.nan)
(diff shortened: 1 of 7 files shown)


dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -269,6 +272,7 @@ def test_profile(self):
         expected_profile = dict(
             min=2.0,
             max=6.0,
+            sum=8,
```
ChrisWallace2020, line 275: How did you get 8 here? Because the sum of the numbers is 20.5, and after they get converted to strings there seems like there should be more than 8 characters

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -366,6 +370,7 @@ def test_profile_merge(self):
         expected_profile = dict(
             min=2.0,
             max=15.0,
+            sum=33,
```
ChrisWallace2020, line 373: Once again  not sure how 33 is the sum here, not saying it isn't correct, I may just be ignorant to the calculation details

dataprofiler/tests/profilers/test_int_column_profile.py
JGSweets: Values which are not integers do not get included

dataprofiler/tests/profilers/test_int_column_profile.py
JGSweets: As above, the float doesn't count.

JGSweets (approved):

JGSweets:
Addresses https://github.com/capitalone/DataProfiler/issues/262#issue-917663351

ChrisWallace2020 (approved):

PR #266, opened 2021-06-14
Add skewness and kurtosis statistics to numerical stats

andrew-yin:
Added the bias-corrected skewness and kurtosis values as defined [here (skewness)](https://en.wikipedia.org/wiki/Skewness#Sample_skewness) and [here (kurtosis)](https://en.wikipedia.org/wiki/Kurtosis#Standard_unbiased_estimator)] for numerical statistics. Also added it for float precision statistics, but let me know if this isn't needed/useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,8 +104,6 @@ The format for a structured profile is below:
 		'mean': float,
 		'var': float,
 		'std': float,
-		'skewness': float,
-		'kurtosis': float,
 		'sample_size': int,
 		'margin_of_error': float,
 		'confidence_level': float		
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -37,6 +37,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             'mean': None,
             'var': None,
             'std': None,
+            # TODO: skew, kurt here
             'sum': None,
             'sample_size': None,
             'margin_of_error': None,
@@ -127,6 +128,8 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             mean=self.np_type_to_type(self.mean),
             variance=self.np_type_to_type(self.variance),
             stddev=self.np_type_to_type(self.stddev),
+            skewness=self.np_type_to_type(self.skewness),
+            # TODO: Kurtosis here
             histogram=self._get_best_histogram_for_profile(),
             quantiles=self.quantiles,
             times=self.times,
@@ -136,6 +139,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                 mean=self.np_type_to_type(self.precision['mean']),
                 var=self.np_type_to_type(self.precision['var']),
                 std=self.np_type_to_type(self.precision['std']),
+                # TODO: Add skewness, kurtosis here
                 sample_size=self.np_type_to_type(self.precision['sample_size']),
                 margin_of_error=self.np_type_to_type(self.precision['margin_of_error']),
                 confidence_level=self.np_type_to_type(self.precision['confidence_level'])
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -38,15 +38,11 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             'mean': None,
             'var': None,
             'std': None,
-            'skewness': None,
-            'kurtosis': None,
             'sum': None,
             'sample_size': None,
             'margin_of_error': None,
             'confidence_level': 0.999
         }
-        self._precision_biased_skewness = np.nan
-        self._precision_biased_kurtosis = np.nan
 
         # https://www.calculator.net/confidence-interval-calculator.html
         self.__z_value_precision = 3.291
@@ -110,38 +106,12 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                 merged_profile.precision['std'] = math.sqrt(
                     merged_profile.precision['var'])
 
-                merged_profile._precision_biased_skewness = self._merge_biased_skewness(
-                    self.precision['sample_size'], self._precision_biased_skewness,
-                    self.precision['var'], self.precision['mean'],
-                    other.precision['sample_size'],
-                    other._precision_biased_skewness, other.precision['var'],
-                    other.precision['mean'])
-
-                merged_profile.precision['skewness'] = self._correct_bias_skewness(
-                    merged_profile.precision['sample_size'],
-                    merged_profile._precision_biased_skewness)
-
-                merged_profile._precision_biased_kurtosis = self._merge_biased_kurtosis(
-                    self.precision['sample_size'],
-                    self._precision_biased_kurtosis,
-                    self._precision_biased_skewness, self.precision['var'],
-                    self.precision['mean'],
-                    other.precision['sample_size'],
-                    other._precision_biased_kurtosis,
-                    other._precision_biased_skewness, other.precision['var'],
-                    other.precision['mean'])
-
-                merged_profile.precision['kurtosis'] = self._correct_bias_kurtosis(
-                    merged_profile.precision['sample_size'],
-                    merged_profile._precision_biased_kurtosis)
-
                 # Margin of error, 99.9% confidence level
                 merged_profile.precision['margin_of_error'] = \
                     merged_profile.__z_value_precision * (
                         merged_profile.precision['std']
                         / math.sqrt(merged_profile.precision['sample_size'])
                 )
-
 
         return merged_profile
 
@@ -170,8 +140,6 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                 mean=self.np_type_to_type(self.precision['mean']),
                 var=self.np_type_to_type(self.precision['var']),
                 std=self.np_type_to_type(self.precision['std']),
-                skewness=self.np_type_to_type(self.precision['skewness']),
-                kurtosis=self.np_type_to_type(self.precision['kurtosis']),
                 sample_size=self.np_type_to_type(self.precision['sample_size']),
                 margin_of_error=self.np_type_to_type(self.precision['margin_of_error']),
                 confidence_level=self.np_type_to_type(self.precision['confidence_level'])
@@ -229,8 +197,6 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             'max': len_per_float.max(),
             'mean': precision_sum / sample_size,
             'var': float(len_per_float.var()),
-            'biased_skewness': float(skew(len_per_float)),
-            'biased_kurtosis': float(kurtosis(len_per_float)),
             'sum': precision_sum,
             'sample_size': sample_size
         }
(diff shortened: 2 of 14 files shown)


dataprofiler/profilers/float_column_profile.py
```diff
@@ -233,31 +270,52 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
+
+            self._precision_biased_skewness = self._merge_biased_skewness(
+                self.precision['sample_size'],
+                self._precision_biased_skewness, self.precision['var'],
+                self.precision['mean'],
+                subset_precision['sample_size'],
+                subset_precision['biased_skewness'],
+                subset_precision['var'],
+                subset_precision['mean'])
+
```
JGSweets, line 312: will be able to remove with removal from precision

dataprofiler/profilers/float_column_profile.py
```diff
@@ -233,31 +270,52 @@ def _update_precision(self, df_series, prev_dependent_properties,
...
             sample_ratio = self.__precision_sample_ratio
         
-        # (min, max, var, sum, sample_size)
+        # (min, max, var, skewness, kurtosis, sum, sample_size)
         subset_precision = self._get_float_precision(df_series, sample_ratio)
         if subset_precision is None:
             return
         elif self.precision['min'] is None:
+            self._precision_biased_kurtosis = subset_precision.pop('biased_kurtosis')
+            self._precision_biased_skewness = subset_precision.pop('biased_skewness')
```
JGSweets, line 283: will be able to remove with precision revert

JGSweets (dismissed):
Initial thoughts, will have to dig a little deeper after initial changes are made. Good initial commits, just need to remove the precision component form this PR for now.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -105,13 +110,39 @@ def __add__(self, other):
...
+
                 # Margin of error, 99.9% confidence level
                 merged_profile.precision['margin_of_error'] = \
                     merged_profile.__z_value_precision * (
                         merged_profile.precision['std']
                         / math.sqrt(merged_profile.precision['sample_size'])
                 )
-            
+
+
```
ChrisWallace2020, line 145: Should only have 1 blank line

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -204,6 +206,56 @@ def batch_variance(mean_a, var_a, count_a, mean_b, var_b, count_b):
...
+        data = np.linspace(-3, 2, 11).tolist()
+        df2 = pd.Series(data)
+
+        data = np.full((10,), 1)
+        df3 = pd.Series(data)
+
+        num_profiler = IntColumn(df1.name)
+        num_profiler.update(df1.apply(str))
+
+        self.assertAlmostEqual(df1.skew(), num_profiler.skewness)
```
JGSweets, line 222: instead of utilizing another library, can we specify the actual value here since we know the calculation? Same for the other tests.

JGSweets: The list could be more simple if that makes it easier.

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -204,6 +206,56 @@ def batch_variance(mean_a, var_a, count_a, mean_b, var_b, count_b):
...
+        df = pd.concat([df1, df2_ints])
+        self.assertAlmostEqual(df.skew(), num_profiler.skewness)
+
+        df3_ints = df3[df3 == df3.round()]
+        num_profiler.update(df3.apply(str))
+        df = pd.concat([df1, df2_ints, df3_ints])
+        self.assertAlmostEqual(df.skew(), num_profiler.skewness)
+
+    def test_profiled_kurtosis(self):
+        data = np.linspace(-5, 5, 11).tolist()
```
JGSweets, line 235: as above about the calculation

dataprofiler/tests/profilers/test_int_column_profile.py
JGSweets: or at least specify what the value will be as oppose ot calculating.

dataprofiler/profilers/profiler_options.py
```diff
@@ -259,11 +259,14 @@ def __init__(self):
         :ivar is_numeric_stats_enabled: boolean to enable/disable all numeric
             stats
         :vartype is_numeric_stats_enabled: bool
+        TODO: Add comments for skew/kurt here
```
ChrisWallace2020, line 262: Make sure to add this before merge

README.md
```diff
@@ -83,6 +83,8 @@ The format for a structured profile is below:
             "mean": float,
             "variance": float,
             "stddev": float,
+            "skewness": float,
```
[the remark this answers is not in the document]
gme5078: We will need to update the structured profile like this and the profiler options

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -265,6 +298,143 @@ def _merge_variance(match_count1, variance1, mean1,
...
+        """
+        Apply bias correction to skewness
+
+        :param match_count: number of samples
+        :param biased_skewness: skewness without bias correction
+        :return: unbiased estimator of skewness
+        :rtype: NaN if sample size is too small, float otherwise
+        """
+        if biased_skewness is None or np.isnan(biased_skewness) or match_count < 3:
+            return np.nan
```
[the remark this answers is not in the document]
JGSweets: let's move this into a subsequent PR

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -444,6 +513,8 @@ def test_profile_merge_edge_case(self):
         profiler = profiler3 + profiler4
         self.assertEqual(profiler.min, 2)
         self.assertEqual(profiler.max, 5)
+        self.assertEqual(profiler.skewness, 0)
```
AnhTruong, line 516: can we use np or pd to calculate the value `0`, and same as below

dataprofiler/profilers/profiler_options.py
```diff
@@ -323,19 +332,32 @@ def _validate_helper(self, variable_path='NumericalOptions'):
...
+        if sum_disabled and not var_disabled:
             errors.append("{}: The numeric stats must toggle on the sum "
                           "if the variance is toggled on."
                           .format(variable_path))
+        if (sum_disabled or var_disabled) and not skew_disabled:
+            errors.append("{}: The numeric stats must toggle on the "
+                          "sum and variance if skewness is toggled on."
+                          .format(variable_path))
+        if (sum_disabled or var_disabled or skew_disabled) \
+            and not kurt_disabled:
```
AnhTruong, line 357: can we still calculate kurtosis without the need of skewness?

andrew-yin: The [algorithm](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Higher-order_statistics) for computing kurtosis relies on knowing M3 (third central moment without dividing by N), which I found is extracted from skewness. Not sure how else it could be done, thoughts?

AnhTruong: What I meant is, then we can find kurtosis even we have `skew_disabled` is True, as we can find skewness and update the kurtosis like you mentioned

andrew-yin: I guess this is possible, I just added these error checks to keep consistent with the original error check for variance. Although you can still technically compute variance without sum enabled (just need to compute sum in the background but not report it), it still raises an error so I figured the same should be for skewness/kurtosis for consistency. What do you think?

JGSweets: If any of the features require something else, it should be dependent on it being enabled. We shouldn't calculate it externally.

JGSweets: @AnhTruong, hence if the dependencies are correct, it should be done as is.

dataprofiler/tests/profilers/test_int_column_profile.py
JGSweets: @AnhTruong using pd / np would not make sense in this case as skew / kurtosis would not be able to be calculated on such small basis. We would have to look into an online algorithm for single values and to adapt it as such.

```
def online_kurtosis(data):
    n = mean = M2 = M3 = M4 = 0

    for x in data:
        n1 = n
        n = n + 1
        delta = x - mean
        delta_n = delta / n
        delta_n2 = delta_n * delta_n
        term1 = delta * delta_n * n1
        mean = mean + delta_n
        M4 = M4 + term1 * delta_n2 * (n*n - 3*n + 3) + 6 * delta_n2 * M2 - 4 * delta_n * M3
        M3 = M3 + term1 * delta_n * (n - 2) - 3 * delta_n * M2
        M2 = M2 + term1

    # Note, you may also calculate variance using M2, and skewness using M3
    # Caution: If all the inputs are the same, M2 will be 0, resulting in a division by 0.
    kurtosis = (n * M4) / (M2 * M2) - 3
    return kurtosis
```
from https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Higher-order_statistics

dataprofiler/profilers/profiler_options.py
AnhTruong: I still see this is a potential use case, but as we're having the consistency across other stats, we can raise a warning to reset other stats if the dependent stat is needed. But for now, we haven't had a user's request for such a case, then we can leave it as is.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -704,12 +877,74 @@ def _get_variance(self, df_series, prev_dependent_properties,
...
+        :param df_series: incoming data
+        :type df_series: pandas series
+        :param prev_dependent_properties: pre-update values needed
+            for computation
+        :type prev_dependent_properties: dict
+        :param subset_properties: incoming data statistics
+        :type subset_properties: dict
+        :return None
+        """
+        batch_biased_skewness = skew(df_series)
```
JGSweets, line 903: typically we want to leave in the fact that it is from scipy, hence  we can instead at the top just `import scipy` and call `scipy.stats.skew` as we don't need the modularity of `skew`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -704,12 +877,74 @@ def _get_variance(self, df_series, prev_dependent_properties,
...
+        :param df_series: incoming data
+        :type df_series: pandas series
+        :param prev_dependent_properties: pre-update values needed
+            for computation
+        :type prev_dependent_properties: dict
+        :param subset_properties: incoming data statistics
+        :type subset_properties: dict
+        :return None
+        """
+        batch_biased_kurtosis = kurtosis(df_series)
```
JGSweets, line 932: as above with `skew`: typically we want to leave in the fact that it is from `scipy`, hence we can instead at the top just import `scipy` and call `scipy.stats.kurtosis` as we don't need the modularity of `kurtosis`

dataprofiler/profilers/numerical_column_stats.py
JGSweets: If two approvals aren't made by the time this can be updated, we can fix it in this PR, otherwise we will add it to the subsequent PR updating for allowing the user to specify bias correction or not.

dataprofiler/profilers/numerical_column_stats.py
JGSweets: If two approvals aren't made by the time this can be updated, we can fix it in this PR, otherwise we will add it to the subsequent PR updating for allowing the user to specify bias correction or not.

JGSweets (approved):

JGSweets (approved):

AnhTruong (approved):

PR #269, opened 2021-06-15
Add option for turning on/off bias correction

andrew-yin:
By default, the computations for variance/skewness/kurtosis are corrected for bias. However, this update allows the user to set "bias" to true or false in options to provide either biased or unbiased statistics, respectively. By default, bias is False (i.e. there will NOT be bias included in these calculations).

### DIFF
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -3,7 +3,6 @@ import copy
 import math
 import numpy as np
 
-from scipy.stats import skew, kurtosis
 from .numerical_column_stats import NumericStatsMixin
 from .base_column_profilers import BaseColumnProfiler, \
     BaseColumnPrimitiveTypeProfiler
@@ -43,6 +42,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             'margin_of_error': None,
             'confidence_level': 0.999
         }
+        self._biased_precision_var = np.nan
 
         # https://www.calculator.net/confidence-interval-calculator.html
         self.__z_value_precision = 3.291
@@ -94,14 +94,21 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                 merged_profile.precision['sample_size'] = \
                     self.precision['sample_size'] + other.precision['sample_size']
 
-                merged_profile.precision['var'] = self._merge_variance(
+                merged_profile._biased_precision_var = self._merge_biased_variance(
                     self.precision['sample_size'],
-                    self.precision['var'], self.precision['mean'],
-                    other.precision['sample_size'], other.precision['var'],
+                    self._biased_precision_var, self.precision['mean'],
+                    other.precision['sample_size'],
+                    other._biased_precision_var,
                     other.precision['mean'])
+
                 merged_profile.precision['mean'] = \
                     merged_profile.precision['sum'] \
                     / merged_profile.precision['sample_size']
+
+                merged_profile.precision['var'] = self._correct_bias_variance(
+                    self.precision['sample_size'],
+                    self._biased_precision_var
+                )
             
                 merged_profile.precision['std'] = math.sqrt(
                     merged_profile.precision['var'])
@@ -196,7 +203,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
             'min': len_per_float.min(),
             'max': len_per_float.max(),
             'mean': precision_sum / sample_size,
-            'var': float(len_per_float.var()),
+            'biased_var': float(np.var(len_per_float)),
             'sum': precision_sum,
             'sample_size': sample_size
         }
@@ -245,6 +252,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         if subset_precision is None:
             return
         elif self.precision['min'] is None:
+            self._biased_precision_var = subset_precision.pop('biased_var')
             self.precision.update(subset_precision)
         else:
             # Update the calculations as data is valid
@@ -254,10 +262,10 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                 self.precision['max'], subset_precision['max'])            
             self.precision['sum'] += subset_precision['sum']
 
-            self.precision['var'] = self._merge_variance(
-                self.precision['sample_size'], self.precision['var'],
+            self._biased_precision_var = self._merge_biased_variance(
+                self.precision['sample_size'], self._biased_precision_var,
                 self.precision['mean'],
-                subset_precision['sample_size'], subset_precision['var'],
+                subset_precision['sample_size'], subset_precision['biased_var'],
                 subset_precision['mean'])
 
             self.precision['sample_size'] += subset_precision['sample_size']            
@@ -265,6 +273,10 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
                 / self.precision['sample_size']
 
         # Calculated outside
+        self.precision['var'] = self._correct_bias_variance(
+            self.precision['sample_size'],
+            self._biased_precision_var)
+
         self.precision['std'] = math.sqrt(self.precision['var'])
 
         # Margin of error, 99.9% confidence level
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -144,8 +144,8 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         """
         Property reporting statistics on the significant figures of each
         element in the data.
-        :return
-        :rtype dict
+        :return: Precision statistics
+        :rtype: dict
         """
         # First add the stats that don't need to be re-calculated
         precision = dict(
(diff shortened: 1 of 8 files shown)


dataprofiler/profilers/profiler_options.py
```diff
@@ -271,6 +273,9 @@ def __init__(self):
         self.skewness = BooleanOption(is_enabled=True)
         self.kurtosis = BooleanOption(is_enabled=True)
         self.histogram_and_quantiles = HistogramOption()
+        # By default, we do NOT want bias, hence bias is false. This
+        # convention is made to maintain consistency with scipy.stats
+        self.bias = BooleanOption(is_enabled=False)
```
JGSweets, line 278: I think the naming might be a little confusing and it might be best to change it to `bias_correction`
as in `bias_correction.is_enabled = True` I like the idea of. it matching scipy, but since we utilize the is_enabled it is kinda weird to say we are enabling bias. Thoughts?

andrew-yin: Agreed, bias_correction.is_enabled is more semantically clear to me, will make those changes

dataprofiler/profilers/float_column_profile.py
```diff
@@ -148,6 +139,38 @@ def profile(self):
         
         return profile
 
+    @property
+    def precision(self):
```
JGSweets, line 143: add description to precision

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -216,28 +218,37 @@ def _add_helper(self, other1, other2):
         if "sum" in self.__calculations.keys():
             self.sum = other1.sum + other2.sum
         if "skewness" in self.__calculations.keys():
-            self._biased_skewness = self._merge_biased_skewness(other1.match_count,
```
JGSweets, line 219: need to add the merge of `self.bias_correction` if two different give a warning and set the bias_correction False

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -188,10 +188,9 @@ def _add_helper(self, other1, other2):
             other2._NumericStatsMixin__calculations)
 
         # Check and potentially override bias correction computation
-        if other1.bias_correction == False or other2.bias_correction == False:
+        self.bias_correction = True
```
JGSweets, line 191: once a test for this is added, looks good.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -371,6 +371,73 @@ def test_profiled_kurtosis(self):
...
+        df2 = pd.Series(data)
+
+        data = np.full((10,), 1)
+        df3 = pd.Series(data)
+
+        # Disable bias correction
+        options = FloatOptions(); options.bias_correction.is_enabled = False
+        num_profiler = FloatColumn(df1.name, options=options)
+        num_profiler.update(df1.apply(str))
+        self.assertAlmostEqual(10, num_profiler.variance)
```
AnhTruong, line 388: why the variance is 10? I used numpy and it returns 43.5 or did I miss something here?

AnhTruong: in general, I'd suggest adding some comments of how to get the value 10, or show formular to calculate it.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -371,6 +371,73 @@ def test_profiled_kurtosis(self):
...
+
+        data = np.linspace(-3, 2, 11).tolist()
+        df2 = pd.Series(data)
+
+        data = np.full((10,), 1)
+        df3 = pd.Series(data)
+
+        # Disable bias correction
+        options = FloatOptions();
+        options.bias_correction.is_enabled = False
```
AnhTruong, line 418: do we have the test for `options.bias_correction.is_enabled = True`?

AnhTruong (commented):
some comments

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: The data is defined by np.linspace(-5,5,11), which forms a list of 11 equally-spaced numbers from -5 to 5 (i.e. -5, -4, -3.,...,3,4,5). I think the variance of just {-5,5,11} is 43.5 which is why you're getting that.

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: By default, options.bias_correction.is_enabled is True, so all tests without explicitly disabling the option is testing for the bias-corrected values.

JGSweets: i think @AnhTruong is asking if we have a merge that ensures when you add two with True it stays True.

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -80,10 +80,8 @@ def test_numerical_stats_option(self, *mocks):
                 self.assertIsNone(profile_column["statistics"]["max"])
                 self.assertEqual(0, profile_column["statistics"]["variance"])
                 self.assertIsNone(profile_column["statistics"]["quantiles"][0])
-                self.assertTrue(
-                    profile_column["statistics"]["skewness"] is np.nan)
-                self.assertTrue(
-                    profile_column["statistics"]["kurtosis"] is np.nan)
+                self.assertEqual(profile_column["statistics"]["skewness"], 0)
+                self.assertEqual(profile_column["statistics"]["kurtosis"], 0)
```
JGSweets, line 84: the standard we try to commit to is we do the `assert(expected_value, actual_value)` hence we just need to flip these around.

We should have a convo with @AnhTruong @lettergram about the implication of this being 0 and not `nan`

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -103,6 +101,8 @@ def test_numerical_stats_option(self, *mocks):
                 self.assertNotEqual(0, profile_column["statistics"]["variance"])
                 self.assertIsNotNone(
                     profile_column["statistics"]["quantiles"][0])
+                self.assertTrue(profile_column["statistics"]["skewness"] is np.nan)
+                self.assertTrue(profile_column["statistics"]["kurtosis"] is np.nan)
```
JGSweets, line 105: instead of assert is nan, can we use `assertEqual`?

ChrisWallace2020: would `assertIsNone` catch this or no since its `np.nan`?

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: This merge is tested by default in test_profile_merge and other similar cases, but should an explicit assert be added to make sure that bias stays True?

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
andrew-yin: I've tried both assertEqual and assertIsNone, but these both don't behave as expected for np.nan

dataprofiler/profilers/float_column_profile.py
```diff
@@ -148,6 +139,44 @@ def profile(self):
...
+        """
+        Property reporting statistics on the significant figures of each
+        element in the data.
+        :return
+        :rtype dict
```
gme5078, line 148: For return and rtype you need to close with a colon as well (":return:" and ":rtype:") or else the autodocumentation will fail. You should also give a description of what youre returning when you are returning something

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: correct, we should validate that if two trues are merged, it stays true

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -766,7 +837,10 @@ def test_profile(self):
             expected_quantiles = expected_profile.pop('quantiles')
             skewness = profile.pop('skewness')
             expected_skewness = expected_profile.pop('skewness')
+            variance = profile.pop('variance')
+            expected_variance = expected_profile.pop('variance')
 
+            self.maxDiff = None
```
JGSweets, line 843: remove?

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: @andrew-yin is this handled?

andrew-yin: @JGSweets just pushed those changes

JGSweets (approved):

AnhTruong (approved):
Looks good for the scope of this PR. But it might be helpful for the case when we want to calculate biased variance, unbiased kurtosis and unbiased skewness, i.e., each stats has its own option. Potentially for another PR @JGSweets ?

JGSweets:
> Looks good for the scope of this PR. But it might be helpful for the case when we want to calculate biased variance, unbiased kurtosis and unbiased skewness, i.e., each stats has its own option. Potentially for another PR @JGSweets ?

Could you think of a use-case where you would want to separate bias correction for each stat? For now I suggest that until a user requests that as an issue, let's focus on other missing features / bugs.

PR #270, opened 2021-06-15
Fixes propagation error in options when setting generic values

JGSweets:
Previously, `is_enabled` would propagate to all `is_enabled` values which may not be intuitive when calling `options.set(dict(is_enabled=True))`.

Now, if the user wants it to propagate to all properties or a specific set they must either use a wildcard or specify the set of options e.g.
```python
options.set({"*.is_enabled": True})
# or
options.set({"int.is_enabled": True})
```

Also resolves issues with tests which used set and erroneously propagated checks.

### DIFF
--- a/dataprofiler/tests/profilers/profiler_options/test_boolean_option.py
+++ b/dataprofiler/tests/profilers/profiler_options/test_boolean_option.py
@@ -59,8 +59,6 @@ class TestBooleanOption(TestBaseOption):
         # Option is_enabled is not a boolean
         option = self.get_options(is_enabled="Hello World")
         expected_error = ["{}.is_enabled must be a Boolean.".format(optpth)]
-        expected_error += ["{}.{}.is_enabled must be a Boolean." 
-                           .format(optpth, key) for key in self.keys]
         self.assertSetEqual(set(expected_error), 
                             set(option._validate_helper()))
     
@@ -78,8 +76,6 @@ class TestBooleanOption(TestBaseOption):
             option.validate(raise_error=True)
 
         expected_error = [expected_error]
-        expected_error += ["{}.{}.is_enabled must be a Boolean." 
-                           .format(optpth, key) for key in self.keys]
         self.assertSetEqual(set(expected_error), 
                             set(option.validate(raise_error=False)))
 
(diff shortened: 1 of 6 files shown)


dataprofiler/tests/profilers/profiler_options/test_boolean_option.py
```diff
@@ -59,8 +59,6 @@ def test_validate_helper(self):
         # Option is_enabled is not a boolean
         option = self.get_options(is_enabled="Hello World")
         expected_error = ["{}.is_enabled must be a Boolean.".format(optpth)]
-        expected_error += ["{}.{}.is_enabled must be a Boolean." 
-                           .format(optpth, key) for key in self.keys]
```
ChrisWallace2020, line 63: Why did you delete this?

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -43,10 +43,25 @@ def test_default_profiler_options(self, *mocks):
         # Stored in Profiler as StructuredOptions
         self.assertEqual(profile2.options, options.structured_options)
 
+    def test_set_failures(self, *mocks):
+
+        options = ProfilerOptions()
+
+        # check if no '*.' it raises an error bc attribute not found
+        expected_error = "type object '' has no attribute 'is_enabled'"
```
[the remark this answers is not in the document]
JGSweets: done

dataprofiler/tests/profilers/profiler_options/test_boolean_option.py
JGSweets: bc is_enabled no longer propagates setting everything at every level unless it is specified as "*.is_enabled"

ChrisWallace2020: Ah I see

ChrisWallace2020 (approved):

ChrisWallace2020 (approved):

ChrisWallace2020 (approved):

gme5078 (approved):

PR #271, opened 2021-06-16
Added seed library variable

gautomdas:


dataprofiler/profilers/utils.py
```diff
@@ -7,6 +7,7 @@
 import psutil
 import numpy as np
 import multiprocessing as mp
+from dataprofiler import settings
```
JGSweets, line 10: this needs to be spaced from the top 1st block of imports. For imports the structure is:
1. built-ins
2. 3rd party
3. Own library

dataprofiler/profilers/utils.py
```diff
@@ -88,7 +89,15 @@ def shuffle_in_chunks(data_length, chunk_size):
...
     
     rng = np.random.default_rng()
-    if 'DATAPROFILER_SEED' in os.environ:
+
+    if settings._seed is not None:
+        try:
+            seed_value = int(settings._seed)
+            rng = np.random.default_rng(seed_value)
+        except ValueError as e:
+            warnings.warn("Seed should be an integer", RuntimeWarning)
```
JGSweets, line 98: should reference that it should be more than 0

dataprofiler/profilers/utils.py
```diff
@@ -88,7 +89,15 @@ def shuffle_in_chunks(data_length, chunk_size):
         return []
     
     rng = np.random.default_rng()
-    if 'DATAPROFILER_SEED' in os.environ:
+
+    if settings._seed is not None:
+        try:
+            seed_value = int(settings._seed)
+            rng = np.random.default_rng(seed_value)
```
JGSweets, line 96: in this case, probably don't need to set an intermediate variable.

dataprofiler/tests/profilers/utils.py
```diff
@@ -3,6 +3,7 @@
 import random
 
 import numpy as np
+from dataprofiler import settings
```
JGSweets, line 6: as above with the imports, should be a space between these imports since numpy is a 3rd party lib

dataprofiler/tests/profilers/utils.py
```diff
@@ -11,8 +12,7 @@ def set_seed(seed=None):
     :param seed:
     :return:
     """
-    np.random.seed(seed)
-    random.seed(seed)
+    settings._seed = seed
```
JGSweets, line 15: this should also set the seed for np.random and random.seed in addition to the settings

JGSweets: or maybe we should remove this and the set_seed func in the __init__ should set these as well.

In this case, we would likely move the set_seed into its own file and import set_seed from the file so that it could be accessed directly via `dp.set_seed`

JGSweets: @granteden thoughts?

JGSweets: e.g. a utils

JGSweets: also utils shouldn't directly alter the `_seed` but utilize the `dp.set_seed`

dataprofiler/__init__.py
```diff
@@ -22,3 +23,6 @@
         '\tsudo apt-get -y install libsnappy-dev`\n',
         ImportWarning
     )
+
+def set_seed(seed=None):
+    settings._seed = seed
```
ChrisWallace2020, line 28: Need a blank new line at the end of the file

dataprofiler/settings.py
```diff
@@ -0,0 +1 @@
+_seed = None
```
ChrisWallace2020, line 1: Need an empty new line at the end of the file

dataprofiler/profilers/utils.py
JGSweets: `seed_value`

JGSweets: also, we should assume that settings._seed is in its correct form. either `None` or int > 0

dataprofiler/__init__.py
```diff
@@ -22,3 +23,6 @@
         '\tsudo apt-get -y install libsnappy-dev`\n',
         ImportWarning
     )
+
+def set_seed(seed=None):
```
JGSweets, line 27: we should validate here that seed is appropriate. e.g. `None` or int >= 0

dataprofiler/tests/profilers/utils.py
ChrisWallace2020: I think it should be a solution in one place, rather than top level in init and here. Because if you are just setting `_seed` here then what is the use case for the one in `__init__`?

JGSweets: actually, now that i think about it, i think we leave it out of `__init__` reason being, the user can choose to specify those externally. If we alter it globally, we could be altering it for someone else.

dataprofiler/__init__.py
```diff
@@ -22,3 +23,10 @@
...
+    # also check it's an integer
+    if seed is not None and (seed < 0 or int(seed)!=float(seed)):
```
JGSweets, line 29: if seed were a `{}` it would not have a `<` operator. hence i think we need to alter this to check specifically that it is an `int`
e.g. `if seed is not None and (not isinstance(seed, int) or seed < 0):`

dataprofiler/__init__.py
```diff
@@ -22,3 +23,10 @@
...
+        raise ValueError("Seed should be a non-negative integer.")
+    settings._seed = seed
+
```
JGSweets, line 32: this line might have whitespaces in it

dataprofiler/profilers/utils.py
```diff
@@ -87,8 +88,9 @@ def shuffle_in_chunks(data_length, chunk_size):
     if not data_length or data_length == 0 \
        or not chunk_size or chunk_size == 0:
         return []
-    
-    rng = np.random.default_rng()
+
+    rng = np.random.default_rng(int(settings._seed))
```
JGSweets, line 92: `since we already check, remove the `int` from here

dataprofiler/profilers/utils.py
```diff
@@ -87,8 +88,9 @@ def shuffle_in_chunks(data_length, chunk_size):
...
+
     if 'DATAPROFILER_SEED' in os.environ:
```
JGSweets, line 94: for this if statement, we should only override with env var if `dp._seed is None` e.g. if a user calls `dp.set_seed` the env var shouldn't override it.

JGSweets: `if 'DATAPROFILER_SEED' in os.environ and settings._seed is None:`

gautomdas: Okay, will fix

dataprofiler/__init__.py
```diff
@@ -7,6 +7,7 @@
                                     UnstructuredDataLabeler
 from .validators.base_validators import Validator
 from .version import __version__
+from . import settings
```
JGSweets, line 11: two spaces between imports and code

dataprofiler/profilers/utils.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import multiprocessing as mp
```
JGSweets, line 10: two spaces between imports and code

JGSweets (dismissed):

dataprofiler/tests/labelers/utils.py
```diff
@@ -0,0 +1,18 @@
...
+import random
+
+import numpy as np
+
+import dataprofiler as dp
+
+
+def set_seed(seed=None):
+    """
+    Sets the see for all possible random state libraries
```
ChrisWallace2020, line 12: typo, I'm guessing this is supposed to be "sets the seed"

dataprofiler/tests/profilers/utils.py
```diff
@@ -13,6 +15,7 @@ def set_seed(seed=None):
     """
     np.random.seed(seed)
     random.seed(seed)
+    dp.set_seed(seed)
```
ChrisWallace2020, line 18: Why is this written for profilers and labelers when its the same function? Is there a reason this isn't more centrally located?

dataprofiler/profilers/utils.py
```diff
@@ -87,9 +89,10 @@ def shuffle_in_chunks(data_length, chunk_size):
     if not data_length or data_length == 0 \
        or not chunk_size or chunk_size == 0:
         return []
-    
-    rng = np.random.default_rng()
-    if 'DATAPROFILER_SEED' in os.environ:
+
+    rng = np.random.default_rng(settings._seed)
```
gme5078, line 93: Did we want to make this so that there is only want rng call instead of potentially 2?

dataprofiler/tests/labelers/utils.py
ChrisWallace2020: Its also replicated in the other `set_seed` function

dataprofiler/tests/labelers/utils.py
```diff
@@ -0,0 +1,18 @@
...
+    :param seed:
+    :return:
+    """
+    np.random.seed(seed)
+    random.seed(seed)
```
ChrisWallace2020, line 17: Is there a reason these things aren't also set in the top level `__init__` function?

ChrisWallace2020: Then these test utils just call that one? I understand this was a whole discussion previously, I am just unaware of the conclusion and the structure that was decided upon.

JGSweets: We don't want to globally set the seed for other libraries.

JGSweets (approved):

ChrisWallace2020 (approved):

PR #272, opened 2021-06-16
NumericalStatsMixin: Count zeros/negatives include in merge and profile of profilers

az85252:
In the Numerical Stats Mixin class, I have updated the merge function to include for num_zeros and num_negatives. I also added num_zeros / num_negatives to the property func `def profile` for the int and float columns. Finally, I set the default options for num_zeros and num_negatives to be True while the text options for those is set to False.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ The format for a structured profile is below:
             "stddev": float,
             "skewness": float,
             "kurtosis": float,
+            "num_zeros": int
+            "num_negatives": int
             "histogram": { 
                 "bin_counts": list(int),
 		"bin_edges": list(float),
--- a/README.md
+++ b/README.md
@@ -85,8 +85,8 @@ The format for a structured profile is below:
             "stddev": float,
             "skewness": float,
             "kurtosis": float,
-            "num_zeros": int
-            "num_negatives": int
+            "num_zeros": int,
+            "num_negatives": int,
             "histogram": { 
                 "bin_counts": list(int),
 		"bin_edges": list(float),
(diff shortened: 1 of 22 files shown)


dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -28,17 +28,20 @@ def test_default_profiler_options(self, *mocks):
...
             self.assertTrue(column.properties["histogram_and_quantiles"])
             self.assertTrue(column.properties["min"])
             self.assertTrue(column.properties["max"])
             self.assertTrue(column.properties["sum"])
             self.assertTrue(column.properties["variance"])
             self.assertTrue(column.properties["is_numeric_stats_enabled"])
-            self.assertFalse(column.properties["num_zeros"].is_enabled)
-            self.assertFalse(column.properties["num_negatives"].is_enabled)
-
+            if(column_type is not "text"):
```
ChrisWallace2020, line 39: Should do `column_type != "text"`

JGSweets: also can remove the `()` around the statements

ChrisWallace2020:
It looks good, the only thing is that I would try to find a `test_add`/`test_merge` or something of the sort and add a check that the num zeros and negative actually got added and are what you expected (when merging). Because right now the test cases you added changes to didn't have any, so they were always the default value, so the test coverage is a bit lacking at the moment.

ChrisWallace2020 (dismissed):

JGSweets (commented):
@ChrisWallace2020 should we also have int/float `test_merge_profile` updated as well?

ChrisWallace2020:
@JGSweets yes you're right it should be checked at merge level, not just `_add_helper`

gme5078:
Can you please update the readme to have num_zeroes and num_negatives in the report?

JGSweets (dismissed):

README.md
```diff
@@ -85,6 +85,8 @@ The format for a structured profile is below:
             "stddev": float,
             "skewness": float,
             "kurtosis": float,
+            "num_zeros": int
```
gme5078, line 88: Can you add a comma at the end to follow the convention?

gme5078 (approved):

JGSweets (approved):

PR #273, opened 2021-06-17
Create independent skew/kurtosis calculation without bias correction

andrew-yin:
Removed dependence on scipy for obtaining biased calculations of skewness and kurtosis.

### DIFF
--- a/dataprofiler/profilers/utils.py
+++ b/dataprofiler/profilers/utils.py
@@ -297,7 +297,9 @@ def biased_skew(df_series):
     cubed_diffs = squared_diffs * diffs
     M2 = sum(squared_diffs)
     M3 = sum(cubed_diffs)
-    # Zero out if possible to prevent floating point errors
+    # This correction comes from the pandas implementation of
+    # skewness, which zeroes these values out before computation
+    # due to possible floating point errors that can occur.
     M2 = 0 if np.abs(M2) < 1e-14 else M2
     M3 = 0 if np.abs(M3) < 1e-14 else M3
 
@@ -327,6 +329,9 @@ def biased_kurt(df_series):
     fourth_diffs = squared_diffs * squared_diffs
     M2 = sum(squared_diffs)
     M4 = sum(fourth_diffs)
+    # This correction comes from the pandas implementation of
+    # kurtosis, which zeroes these values out before computation
+    # due to possible floating point errors that can occur.
     M2 = 0 if np.abs(M2) < 1e-14 else M2
     M4 = 0 if np.abs(M4) < 1e-14 else M4
 
--- a/dataprofiler/profilers/utils.py
+++ b/dataprofiler/profilers/utils.py
@@ -276,3 +276,62 @@ def add_nested_dictionaries(first_dict, second_dict):
             merged_dict[item] = copy.deepcopy(second_dict[item])
 
     return merged_dict
+
+def biased_skew(df_series):
+    """
+    Calculates the biased estimator for skewness of the given data.
+    The definition is formalized as g_1 here:
+        https://en.wikipedia.org/wiki/Skewness#Sample_skewness
+    :param df_series: data to get skewness of, assuming floats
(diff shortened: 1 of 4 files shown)


dataprofiler/profilers/utils.py
```diff
@@ -275,4 +275,63 @@ def add_nested_dictionaries(first_dict, second_dict):
...
+    squared_diffs = diffs ** 2
+    cubed_diffs = squared_diffs * diffs
+    M2 = sum(squared_diffs)
+    M3 = sum(cubed_diffs)
+    # Zero out if possible to prevent floating point errors
+    M2 = 0 if np.abs(M2) < 1e-14 else M2
+    M3 = 0 if np.abs(M3) < 1e-14 else M3
+
+    if (M2 == 0):
+        return 0.0
```
JGSweets, line 305: in this case, should we return np.nan as does scipy?

dataprofiler/profilers/utils.py
```diff
@@ -275,4 +275,63 @@ def add_nested_dictionaries(first_dict, second_dict):
...
+    mean = sum(df_series) / n
+    diffs = df_series - mean
+    squared_diffs = diffs ** 2
+    fourth_diffs = squared_diffs * squared_diffs
+    M2 = sum(squared_diffs)
+    M4 = sum(fourth_diffs)
+    M2 = 0 if np.abs(M2) < 1e-14 else M2
+    M4 = 0 if np.abs(M4) < 1e-14 else M4
+
+    if (M2 == 0):
```
JGSweets, line 333: same as above

dataprofiler/profilers/utils.py
JGSweets: or does scipy return 0 in this case?

andrew-yin: If M2 is zero, then the variance of the data must be zero, implying all elements must be the same value. This doesn't mean that the skewness/kurtosis is undefined, but rather zero as well. Scipy/pandas return zero as well.

JGSweets: what about the float point issue for small numbers as pointed out by pandas in their calc?

andrew-yin: Pandas applies the function [_zero_out_fperr](https://github.com/pandas-dev/pandas/blob/b4dcf3b535d06dd3d597e37be46a4f178fe6b542/pandas/core/nanops.py#L1496) to M2/M3/M4 after computing them. The function essentially returns 0 if the given value is less than 1e-14, otherwise it returns the original value, which is what I also correct for.

dataprofiler/profilers/utils.py
```diff
@@ -275,4 +275,63 @@ def add_nested_dictionaries(first_dict, second_dict):
...
+        return 0.0
```
JGSweets, line 334: should the value be -3 for this?

dataprofiler/profilers/utils.py
JGSweets: let's add a comment referencing where the idea comes form, it's good we commented it, but we should mention the idea is derived bc pandas experiences with issues

dataprofiler/profilers/utils.py
```diff
@@ -275,4 +275,63 @@ def add_nested_dictionaries(first_dict, second_dict):
...
+        return np.nan
+
+    mean = sum(df_series) / n
+    diffs = df_series - mean
+    squared_diffs = diffs ** 2
+    fourth_diffs = squared_diffs * squared_diffs
+    M2 = sum(squared_diffs)
+    M4 = sum(fourth_diffs)
+    M2 = 0 if np.abs(M2) < 1e-14 else M2
+    M4 = 0 if np.abs(M4) < 1e-14 else M4
```
JGSweets, line 331: needs same comment as above

gme5078 (approved):

JGSweets (approved):

PR #275, opened 2021-06-21
Update default value for variance/skew/kurtosis

andrew-yin:
Previously, variance/skewness/kurtosis have a default value of zero, even if their value is incomputable based on the sample size. This changes the default value to NaN when these statistics cannot be computed on the data.

### DIFF
--- a/dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
+++ b/dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
@@ -81,10 +81,10 @@ class TestProfilerOptions(unittest.TestCase):
                     profile_column["statistics"]["histogram"]["bin_edges"])
                 self.assertIsNone(profile_column["statistics"]["min"])
                 self.assertIsNone(profile_column["statistics"]["max"])
-                self.assertEqual(0, profile_column["statistics"]["variance"])
+                self.assertTrue(np.isnan(profile_column["statistics"]["variance"]))
                 self.assertIsNone(profile_column["statistics"]["quantiles"][0])
-                self.assertEqual(0, profile_column["statistics"]["skewness"])
-                self.assertEqual(0, profile_column["statistics"]["kurtosis"])
+                self.assertTrue(np.isnan(profile_column["statistics"]["skewness"]))
+                self.assertTrue(np.isnan(profile_column["statistics"]["kurtosis"]))
 
         # Assert that the stats are enabled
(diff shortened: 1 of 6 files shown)


JGSweets (commented):
In `_merge_biased_skewness` and `_merge_biased_kurtosis`, if `variance` is nan, wouldn't this also cause an error? Is this even possible to occur?

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -101,7 +101,8 @@ def test_numerical_stats_option(self, *mocks):
                     profile_column["statistics"]["histogram"]["bin_edges"])
                 self.assertIsNotNone(profile_column["statistics"]["min"])
                 self.assertIsNotNone(profile_column["statistics"]["max"])
-                self.assertNotEqual(0, profile_column["statistics"]["variance"])
+                print(profile_column["statistics"]["variance"])
```
ChrisWallace2020, line 104: Should delete this

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -101,7 +101,8 @@ def test_numerical_stats_option(self, *mocks):
...
+                self.assertEqual(0.5, profile_column["statistics"]["variance"])
```
ChrisWallace2020, line 105: You could also just do `self.assertFalse(np.isnan(profile_column["statistics"]["variance"]))` if all you care about is whether or not it is nan, although its not a big deal.

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
JGSweets: Should assert the value still.

andrew-yin:
When variance, skewness, and kurtosis are np.nan due to indeterminate values in the dataset, future updates on these statistics should be ignored. This will be included in a separate PR.

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
JGSweets: will approve as is, but i think the original intent is we just care in options that it isn't the default value anymore, but this is fine too.

JGSweets (approved):

ChrisWallace2020 (approved):

PR #276, opened 2021-06-22
TextOptions for num_zeros and num_negatives disabled by default

az85252:
When num_zeros and num_negatives are declared as TextOptions, those should be disabled and if something changes those to be is_enabled = True, throw errors saying that they shouldn't. Test cases are added and modified to fit these.

### DIFF
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -311,7 +311,7 @@ class NumericalOptions(BaseInspectorOptions):
         """
         if self.min.is_enabled or self.max.is_enabled or self.sum.is_enabled \
                 or self.variance.is_enabled or self.skewness.is_enabled \
-                or self.kurtosis.is_enabled or self.bias_correction.is_enabled \
+                or self.kurtosis.is_enabled \
                 or self.histogram_and_quantiles.is_enabled \
                 or self.num_zeros.is_enabled or self.num_negatives.is_enabled:
             return True
@@ -334,7 +334,6 @@ class NumericalOptions(BaseInspectorOptions):
         self.variance.is_enabled = value
         self.skewness.is_enabled = value
         self.kurtosis.is_enabled = value
-        self.bias_correction.is_enabled = value
         self.num_zeros.is_enabled = value
         self.num_negatives.is_enabled = value
         self.histogram_and_quantiles.is_enabled = value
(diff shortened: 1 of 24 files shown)


requirements.txt
```diff
@@ -2,6 +2,7 @@ h5py>=2.10.0
 wheel>=0.33.1
 future>=0.18.2
 numpy>=1.18.5
+scipy>=1.4.1
```
JGSweets, line 5: please remove

dataprofiler/tests/profilers/profiler_options/test_numerical_options.py
```diff
@@ -99,13 +99,13 @@ def test_validate(self):
...
             expected_error = "{}.{}.is_enabled must be a Boolean."\
                 .format(optpth, key)
-            
+            default_bool = options.properties[key].is_enabled
             options.set({skey: "Hello World"})
             with self.assertRaisesRegex(ValueError, expected_error):
                 options.validate(raise_error=True)    
-            self.assertEqual([expected_error], 
+            self.assertIn(expected_error,
                              options.validate(raise_error=False))
```
ChrisWallace2020, line 107: This line should be tabbed back to line up with `expected_error`

dataprofiler/profilers/profiler_options.py
```diff
@@ -599,8 +599,55 @@ def _validate_helper(self, variable_path='TextOptions'):
...
+        will return False.
+
+        :return: true if any numeric stats property is enabled, otherwise false
+        :rtype bool:
+        """
+        if self.min.is_enabled or self.max.is_enabled or self.sum.is_enabled \
+                or self.variance.is_enabled or self.skewness.is_enabled \
+                or self.kurtosis.is_enabled or self.bias_correction.is_enabled \
+                or self.histogram_and_quantiles.is_enabled \
+                or self.num_zeros.is_enabled or self.num_negatives.is_enabled:
```
ChrisWallace2020, line 628: If these 2 are always disabled should they even be present here? Also why can't we just use the one in `NumericOptions` here? I understand the different setter since you want to set differently in this case, but this property seems identical to the `NumericOptions` one.

dataprofiler/profilers/profiler_options.py
```diff
@@ -599,8 +599,49 @@ def _validate_helper(self, variable_path='TextOptions'):
...
+    def is_numeric_stats_enabled(self):
+        """
+        Returns the state of numeric stats being enabled / disabled. If any
+        numeric stats property is enabled it will return True, otherwise it
+        will return False.
+
+        :return: true if any numeric stats property is enabled, otherwise false
+        :rtype bool:
+        """
+        return super().is_numeric_stats_enabled
```
ChrisWallace2020, line 624: This isn't necessary since this property is already inherited from the parent, you don't need to make a version of it in the child class.

az85252: In order for the below function, is_numeric_stats_enabled(self, value) to work properly, we need to have a override function is_numeric_stats_enabled(self) inside TextOptions class.

ChrisWallace2020: Then please add a comment explaining as much since this may seem unnecessary to anyone reading this function.

JGSweets: something still needs off here as `super().is_numeric_stats_enabled` itself shouldn't work bc it should not include num_zeros and num_negatives (should be false by default) in which case do we have a test validating that this property works?

JGSweets (approved):

JGSweets (approved):

ChrisWallace2020 (approved):

PR #277, opened 2021-06-22
Numerical Stats Differences

gme5078:
Okay, so, some comments:

Firstly, this PR gets the differences of only a few numerical statistics because it is strict to the AC, histograms/quantiles are off the table for this PR.

Secondly, how each datatype difference is treated:
- Int/Float (Interchangeable): Diff = Stat1 - Stat2
- Set/List (Interchangeable, duplicates are stripped from list): Diff = [Unique values in Stat1, Shared Values, Unique values in Stat2]
- Strings: Diff = [stat1, stat2]

If there is no difference, Diff = "unchanged"
I think there will one day be an option to not return unchanged values, if you would like that added to this PR, I can do so.


dataprofiler/profilers/utils.py
```diff
@@ -343,3 +343,39 @@ def biased_kurt(df_series):
 
     kurt = n * M4 / M2 ** 2 - 3
     return kurt
+
+def find_diff(stat1, stat2):
```
JGSweets, line 347: I suggest making individual functions for the differences and allow the creator of the report to set what they want where for each difference.

gme5078: So like allowing the user to overall set chi squared for all number differences or do you mean like the user can set chi squared for specific number differences like if they wanted max diff to be chi squared but still have mean be subtraction?

gme5078: in either case, I'm not sure I want to add options when the options dont exist yet?

dataprofiler/profilers/utils.py
```diff
@@ -343,3 +343,63 @@ def biased_kurt(df_series):
...
+    returns [unique values of stat1, shared values, unique values of stat2].
+
+    :param stat1: the first statistical input
+    :type stat1: Union[int, float, str, list, set]
+    :param stat2: the second statistical input
+    :type stat2: Union[int, float, str, list, set]
+    :return: the difference of the stats
+    """
+    diff = "unchanged"
+    if stat1 is None or stat2 is None:
```
ChrisWallace2020, line 361: If both are None this will return `[None, None]` but shouldn't we want "unchanged" in that case?

dataprofiler/profilers/utils.py
```diff
@@ -343,3 +343,63 @@ def biased_kurt(df_series):
...
+    Finds the difference between two stats. If there is no difference, returns
+    "unchanged". For ints/floats, returns stat1 - stat2. For strings, returns
+    list containing [stat1, stat2]. For lists/sets, removes duplicates and
+    returns [unique values of stat1, shared values, unique values of stat2].
+
+    :param stat1: the first statistical input
+    :type stat1: Union[int, float, str, list, set]
+    :param stat2: the second statistical input
+    :type stat2: Union[int, float, str, list, set]
+    :return: the difference of the stats
```
ChrisWallace2020, line 358: This docstring lays out all the cases but this is just the int/float case

dataprofiler/profilers/utils.py
```diff
@@ -343,3 +343,63 @@ def biased_kurt(df_series):
...
+
+def find_diff_of_strings(stat1, stat2):
+    """
+    Finds the difference between two stats. If there is no difference, returns
+    "unchanged". For strings, returns list containing [stat1, stat2]. 
+
+    :param stat1: the first statistical input
+    :type stat1: Union[str]
+    :param stat2: the second statistical input
+    :type stat2: Union[str]
```
ChrisWallace2020, line 376: They types should just be str

dataprofiler/profilers/utils.py
```diff
@@ -343,3 +343,63 @@ def biased_kurt(df_series):
...
+    :return: the difference of the stats
+    """
+    diff = "unchanged"
+    if stat1 is None or stat2 is None:
+        diff = [stat1, stat2]
+    elif set(stat1) != set(stat2):
+        unique1 = [element for element in stat1 if element not in stat2]
+        shared = [element for element in stat1 if element in stat2]
+        unique2 = [element for element in stat2 if element not in stat1]
+        diff = [unique1, shared, unique2]
```
ChrisWallace2020, line 404: Should `unique1`, `shared`, `unique2` have types that reflect the types of `stat1` and `stat2`. i.e. if they are lists we return a list of lists, if they are sets we return a list of sets?

dataprofiler/tests/profilers/test_utils.py
```diff
@@ -44,3 +44,42 @@ def test_uneven_chunk_sample(self):
...
+        
+    def test_find_diff(self):
+        """
+        Checks to see if the find difference function is operating
+        appropriately.
+        """
+
+        # Ensure lists and sets are handled appropriately
+        self.assertEqual("unchanged",
+                         utils.find_diff_of_lists_and_sets([3, 2], [2, 3, 2]))
```
ChrisWallace2020, line 56: This feels like it this could be considered a "change", I don't know what customers said their use cases were, but maybe lists take count into consideration? Idk I am not dead set on this there can be a conversation but I think this should have the difference as `[2]`.

dataprofiler/tests/profilers/test_utils.py
```diff
@@ -44,3 +44,42 @@ def test_uneven_chunk_sample(self):
...
+        self.assertEqual([[1], [2, 3], [4]],
+                         utils.find_diff_of_lists_and_sets([1, 2, 3], [2, 3, 4]))
+        self.assertEqual("unchanged",
+                         utils.find_diff_of_lists_and_sets({3, 2}, {2, 3}))
+        self.assertEqual([[1], [2, 3], [4]],
+                         utils.find_diff_of_lists_and_sets({1, 2, 3}, {2, 3, 4}))
+        self.assertEqual("unchanged",
+                         utils.find_diff_of_lists_and_sets({2, 3}, [2, 3]))
+        self.assertEqual([[1], [2, 3], [4]],
+                         utils.find_diff_of_lists_and_sets([1, 2, 3], {2, 3, 4}))
```
ChrisWallace2020, line 66: Is there ever a case where these functions would get called with 2 things of different types?

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -502,3 +502,30 @@ def test_merge_num_zeros_and_negatives(self):
...
+        other1.match_count = 10
+        
+        other2.min = 3
+        other2.max = None
+        other2._biased_variance = 9
+        other2.sum = 6
+        other2.match_count = 20
+        
+        expected_diff = {
+            'min': 'unchanged',
```
AnhTruong, line 524: instead of `unchanged`, I'd suggest we make a change ratio for every case, so unchanged corresponds to 0%

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -502,3 +502,30 @@ def test_merge_num_zeros_and_negatives(self):
...
+        """
+        other1, other2 = TestColumn(), TestColumn()
+        other1.min = 3
+        other1.max = 4
+        other1._biased_variance = 1
+        other1.sum = 6
+        other1.match_count = 10
+        
+        other2.min = 3
+        other2.max = None
```
AnhTruong, line 518: does it make sense to have min=3 and max=None?

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -502,3 +502,30 @@ def test_merge_num_zeros_and_negatives(self):
...
+        other2._biased_variance = 9
+        other2.sum = 6
+        other2.match_count = 20
+        
+        expected_diff = {
+            'min': 'unchanged',
+            'max': [4, None],
```
AnhTruong, line 525: as above, imo this should be `nan` or some similar representation

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -502,3 +502,30 @@ def test_merge_num_zeros_and_negatives(self):
...
+            'sum': 'unchanged',
+            'mean': 0.3,
+            'variance': -8.362573099415204,
+            'stddev': -2.0238425028660023
```
AnhTruong, line 529: where do we get mean, var and std from other1 and other2? Should we only calculate the difference based on the given stats from other1 and other2?

dataprofiler/tests/profilers/test_utils.py
AnhTruong: I think the things you mentioned can be covered from sample_size and word_counts/category_counts @ChrisWallace2020

dataprofiler/tests/profilers/test_utils.py
```diff
@@ -44,3 +44,42 @@ def test_uneven_chunk_sample(self):
...
+        self.assertEqual(1.0, utils.find_diff_of_numbers(5.0, 4))
+        self.assertEqual("unchanged", utils.find_diff_of_numbers(5.0, 5.0))
+        self.assertEqual("unchanged", utils.find_diff_of_numbers(5, 5.0))
+        self.assertEqual([4, None],
+                         utils.find_diff_of_numbers(4, None))
+
+        # Ensure strings are handled appropriately
+        self.assertEqual("unchanged",
+                         utils.find_diff_of_strings("Hello", "Hello"))
+        self.assertEqual(["Hello", "team"],
```
AnhTruong, line 82: should the output include three parts `[['Hello'], [], ['team']]`?

AnhTruong (commented):
some first comments on the unit tests. Once we discuss on these, I'll review next the code implementation.

dataprofiler/profilers/utils.py
gme5078: good catch, fixing

dataprofiler/profilers/utils.py
gme5078: I don't think it would add anything imo. I chose to make them interchangeable since I think we already have some lists/sets that interchange

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
gme5078: This was just an example to show you what happens when None are involved. it might make sense if the user disables Max from options

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
gme5078: I think we plan on adding percents and Chi Squared and T Test in the future

dataprofiler/tests/profilers/test_utils.py
gme5078: @ChrisWallace2020 , if you can find me an example in our report, I would change it but otherwise, I think sets and lists should be treated the same.

dataprofiler/tests/profilers/test_utils.py
gme5078: Don't know but I was treating sets and lists the same so I thought they ought to be interchangeable too. I think there are some parts of the repo that go from lists to sets, so maybe its possible

dataprofiler/tests/profilers/test_utils.py
gme5078: I dont think so. I'm imagining using this on Data Label for example so, I think the user would want something like:
["PHONE_NUMBER", "BAN"] if the data labels were different

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
gme5078: I've been just subtracting the difference from other1 and other2. I thought about recalculating it based on the intermediate statistics, but then I thought what would be the point? Like, I think a data scientist wants to know how much variance has shifted, but I don't think recalculating it particularly makes sense. I'm open to ideas tho

ChrisWallace2020 (approved):

JGSweets (approved):

PR #279, opened 2021-06-23
IntColumn Stats Diff

gme5078:
The only stats int has are numerical stats, so the difference function is pretty small

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -247,7 +247,7 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         if "num_negatives" in self.__calculations.keys():
             self.num_negatives = other1.num_negatives + other2.num_negatives
 
-    def _diff_helper(self, other_profile):
+    def diff(self, other_profile):
         """
         Finds the differences for several numerical stats.
 
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -247,7 +247,7 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         if "num_negatives" in self.__calculations.keys():
             self.num_negatives = other1.num_negatives + other2.num_negatives
 
-    def diff(self, other_profile):
+    def diff(self, other_profile, options=None):
         """
         Finds the differences for several numerical stats.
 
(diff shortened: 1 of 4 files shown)


dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -788,3 +788,35 @@ def test_profile_merge_bin_edges_indices(self):
...
+
+        # Assert the difference report is correct
+        expected_diff = {
+            'max': 3.0,
+            'mean': 2.0,
+            'min': 1.0,
+            'stddev': 0.5857864376269049,
+            'sum': 8.0,
+            'variance': 2.0
+        }
```
JGSweets, line 814: should have both a negative and positive value in case we decide to utilize absolutes at some point or protect against.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -247,7 +247,7 @@ def _add_helper(self, other1, other2):
         if "num_negatives" in self.__calculations.keys():
             self.num_negatives = other1.num_negatives + other2.num_negatives
 
-    def _diff_helper(self, other_profile):
+    def diff(self, other_profile, options=None):
```
gme5078, line 250: ultimately renamed so the int column can just inherit this function instead of the int column creating its own. No reason for it to be private either.

gme5078: And allowing options for future use.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -528,4 +528,4 @@ def test_diff_helper(self):
             'variance': -8.362573099415204,
             'stddev': -2.0238425028660023
         }
-        self.assertDictEqual(expected_diff, other1._diff_helper(other2))
+        self.assertDictEqual(expected_diff, other1.diff(other2))
```
ChrisWallace2020, line 531: You should rename this test function `test_diff` now since its no longer called `_diff_helper`

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -788,3 +788,35 @@ def test_profile_merge_bin_edges_indices(self):
...
+            'variance': -94.0
+        }
+        self.assertDictEqual(expected_diff, profiler1.diff(profiler2))
+        
+        # Assert type error is properly called
+        with self.assertRaises(TypeError) as exc:
+            profiler1.diff("Inproper input")
+        self.assertEqual(str(exc.exception),
+                         "Unsupported operand type(s) for diff: 'IntColumn' and"
+                         " 'str'")
```
ChrisWallace2020, line 822: Should you also add this test in `test_numeric_stats_mixin_profile`?

JGSweets: Ultimately, I agree this is an issue bc we have to repeat the tests for anything inheriting. However, I think this needs to be addressed in a test refactor for the mixin all at once. Additionally, if we ultimately merge the int / float profiles similar to how text can be string / text, this would also have an impact on this change.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
JGSweets: agree with the recommended change

JGSweets (approved):

ChrisWallace2020 (approved):

PR #280, opened 2021-06-24
Add warnings for insufficient match count and invalid data for skew/kurt calculations

andrew-yin:
There are a couple of cases where the skewness and kurtosis values cannot be computed. Warnings are added for when skewness/kurtosis cannot be corrected for bias despite it being enabled in options. Furthermore, when invalid data is detected (e.g. infinite values), warnings are raised and any future updates of skewness/kurtosis are no longer computed.

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -413,6 +413,9 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         :rtype: NaN if sample size is too small, float otherwise
         """
         if np.isnan(biased_skewness) or match_count < 3:
+            warnings.warn("Insufficient match count to correct bias in skewness. Bias correction"
+                          "can be manually disabled by setting bias_correction.is_enabled to"
+                          "False in ProfilerOptions.", RuntimeWarning)
             return np.nan
 
         skewness = np.sqrt(match_count * (match_count - 1)) \
@@ -485,6 +488,9 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         :rtype: NaN if sample size is too small, float otherwise
         """
         if np.isnan(biased_kurtosis) or match_count < 4:
+            warnings.warn("Insufficient match count to correct bias in kurtosis. Bias correction"
+                          "can be manually disabled by setting bias_correction.is_enabled to"
+                          "False in ProfilerOptions.", RuntimeWarning)
             return np.nan
 
         kurtosis = (match_count - 1) / ((match_count - 2) *
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -958,6 +958,12 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         :type subset_properties: dict
         :return None
         """
+        # If skewness is still NaN but has a valid match count, this
+        # must mean that there were previous invalid values in
+        # the dataset.
+        if np.isnan(self._biased_skewness) and self.match_count > 0:
+            return
+
         batch_biased_skewness = utils.biased_skew(df_series)
         subset_properties["biased_skewness"] = batch_biased_skewness
         batch_count = subset_properties["match_count"]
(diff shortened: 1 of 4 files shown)


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -439,14 +442,14 @@ def _merge_biased_kurtosis(match_count1, biased_kurtosis1, biased_skewness1,
...
-        if np.isnan(biased_kurtosis2):
-            biased_kurtosis1 = 0
         if match_count1 < 1:
             return biased_kurtosis2
         elif match_count2 < 1:
             return biased_kurtosis1
+        if np.isnan(biased_kurtosis1):
+            biased_kurtosis1 = 0
+        if np.isnan(biased_kurtosis2):
+            biased_kurtosis1 = 0
```
JGSweets, line 452: if any value nan that has match_count > 1, should we instead just return nan? In that case could this be an elif part of the above?

andrew-yin: Agreed, changes made

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -335,14 +335,12 @@ def _merge_biased_variance(match_count1, biased_variance1, mean1,
...
-        if np.isnan(biased_variance1):
-            biased_variance1 = 0
-        if np.isnan(biased_variance2):
-            biased_variance2 = 0
         if match_count1 < 1:
             return biased_variance2
         elif match_count2 < 1:
             return biased_variance1
+        elif np.isnan(biased_variance1) or np.isnan(biased_variance2):
+            return np.nan
```
JGSweets, line 343: nice

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1231,3 +1231,71 @@ def test_profile_merge_bin_edges_indices(self):
         profile_2.update(data_2)
 
         profile_1 + profile_2
+
+    def test_invalid_values_skew_kurt(self):
+        data = pd.Series(['-inf', 'inf'])
+        profiler = FloatColumn(data.name)
+
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
```
JGSweets, line 1240: instead of using the warnings catch, if we want to simplify, we can utilize the `self.assertWarnsRegex`

JGSweets: However, this seems to check for two warnings. we might have to do one at a time to validate, unsure.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1231,3 +1231,71 @@ def test_profile_merge_bin_edges_indices(self):
...
+
+            profiler.update(data)
+            # Verify values are NaN
+            self.assertTrue(np.isnan(profiler._biased_skewness))
+            self.assertTrue(np.isnan(profiler._biased_kurtosis))
+            # Verify warning was raised properly
+            self.assertEqual(2, len(w))
```
JGSweets, line 1247: should variance also have a warning considering it also would return nan, yes?

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: I'm curious if there's an obvious way to make it only return 1 warning for all 3 when it is an incoming data issue

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: I agree, however this wasn't in the AC of the original task. I was going to bring this up with you, just never had the chance.

andrew-yin: It's also a bit different, since variance relies on np.var, not one of our independently written functions

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: An idea can be that rather than checking var/skew/kurt individually, you can check if the sum is indeterminate/infinite on the values and warn there. Thoughts @JGSweets

JGSweets: Makes sense to me.

dataprofiler/tests/profilers/test_float_column_profile.py
JGSweets: True, but we would expect the same occurrence of the issue, no?

JGSweets (approved):

AnhTruong (approved):

PR #281, opened 2021-06-24
Add correlation update feature to DP

AnhTruong:


dataprofiler/profilers/profile_builder.py
```diff
@@ -1201,8 +1200,12 @@ def __add__(self, other):
         # merge profiles
         for profile_name in self._profile:
             merged_profile._profile[profile_name] = (
-                self._profile[profile_name] + other._profile[profile_name]
+                    self._profile[profile_name] + other._profile[profile_name]
```
JGSweets, line 1203: the previous pep8 changes make sense, but I'm not sure this should have two tabs.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1285,6 +1289,9 @@ def _get_row_has_null_ratio(self):
     def _get_duplicate_row_count(self):
         return self.total_samples - len(self.hashed_row_dict)
 
+    def _get_correlation(self):
+        return self.correlation_matrix
+
```
JGSweets, line 1294: what's the intent here? If get correlation just returns an existing property?

An intent I can see for having `_get_correlation` would be to separate `_get_correlation` from merging the correlation where `_get_correlation` would only calculate the correlation for the new series before merging or if this is the first update of the correlation would provide the correlation for the profile.

Such workflow would allow the `_get_correlation` and merging of correlations to be timed separately without overlap of function timing

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1362,118 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        Update correlation matrix for cleaned data.
+
+        :param clean_samples: the input cleaned dataset
+        :type clean_samples: dict()
+        """
+        columns = self.options.correlation.columns
+        if columns is None:
+            for col in self._profile:
+                if self._profile[col].profile['data_type'] not in ['int', 'float'] \
+                or self._profile[col].profile['statistics']['null_count'] > 0:
```
JGSweets, line 1376: this needs to be tabbed over twice, i also suggest if you put `()` around the entire statement  you won't need the `\`

JGSweets: Also, I'm concerned that this is using profile. Maybe it would be better to open up the API for the compiler to allow it to return the `data_type` based on current state as a functional property

JGSweets: potentially a future PR revision though

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1362,118 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        if len(clean_samples) <= 1:
+            return
+
+        data = pd.DataFrame(clean_samples)
+        data = data.apply(pd.to_numeric, errors='coerce')
+        correlation_matrix = data.corr()
+
+        if self.correlation_matrix is None:
+            self.correlation_matrix = correlation_matrix
+            return
```
JGSweets, line 1387: i would suggest putting this at the top and put the warning inside the `if`. That way, instead of running through  all the calculations, we skip it which would otherwise be a waste of time.

JGSweets: we can leave the return in the if, but no need for it at end of function

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1362,118 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+
+        # currently return None with the exising correlation
+        # TODO: implement the update with the exising correlation
+        warnings.warn("Currently the update with existing correlation "
+                      "doesnt work.")
```
JGSweets, line 1392: I'll think about this, but I think we should polish this text a little more. Something along the lines of merging correlations will be available in a future update, as a result the correlations have been turned off for this profile.

dataprofiler/profilers/profile_builder.py
JGSweets: Also, if we skip correlations because they exist, should we set it to something besides `None`, but also not the same as the previous corr matrix? i'm not sure it is safe for them to assume the correlations are correct after an update.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1362,118 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        corr_mat2 = other.correlation_matrix
+        mean2 = np.array(
+            [other._profile[profile_name].profile['statistics']['mean']
+             for profile_name in other._profile])
+        std2 = np.array(
+            [other._profile[profile_name].profile['statistics']['stddev']
+             for profile_name in other._profile])
+        n2 = other.total_samples
+        return self._merge_correlation_helper(corr_mat1, mean1, std1, n1,
+                                              corr_mat2, mean2, std2, n2)
```
JGSweets, line 1420: I'm confused,  thought this doesn't work from our previous talks, does it work for merging two separate profiles?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1201,8 +1200,12 @@ def __add__(self, other):
         # merge profiles
         for profile_name in self._profile:
             merged_profile._profile[profile_name] = (
-                self._profile[profile_name] + other._profile[profile_name]
+                    self._profile[profile_name] + other._profile[profile_name]
             )
+
+        # merge correlation
+        merged_profile.correlation_matrix = self._merge_correlation(other)
```
JGSweets, line 1207: is this feature not being turned off till we solve the issues we discussed?

dataprofiler/profilers/profiler_options.py
```diff
@@ -578,7 +612,43 @@ def _validate_helper(self, variable_path='CategoricalOptions'):
         :return: list of errors (if raise_error is false)
         :rtype: list(str)
         """
-        return super()._validate_helper(variable_path) 
+        return super()._validate_helper(variable_path)
+
+
+class ColumnsCorrelationOptions(BaseInspectorOptions):
```
JGSweets, line 618: Maybe just CorrelationOption?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -456,16 +523,22 @@ def test_save_and_load(self):
         test_files = ["csv/guns.csv", "csv/iris.csv"]
 
         def _clean_report(report):
+            global_stats = report["global_stats"]
+            if "correlation_matrix" in global_stats and \
+                    report["global_stats"]["correlation_matrix"] is not None:
+                report["global_stats"]["correlation_matrix"] = np.concatenate(
+                    report["global_stats"]["correlation_matrix"].values).tolist()
```
JGSweets, line 530: were they not already concatenated? I'll have to inspect this in more detail later.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1685,6 +1772,12 @@ def test_save_and_load_structured(self):
...
+
```
JGSweets, line 1780: maybe we should make this a util since we reuse it a lot.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: It's a pd.DF, yes? so can we just use the `.tolist()` from pandas?

JGSweets (commented):
Initial thoughts, We will have to think about restructuring the functions so they can be timed properly. Additionally, we should talk about the merging since the code is here and if it works for the situation provided.

dataprofiler/profilers/profile_builder.py
AnhTruong: the intent is to get top correlation pairs. Let's say, we have many columns, and we only want to show the correlations of the top 10.

dataprofiler/profilers/profile_builder.py
AnhTruong: agree, and some statistics as well

dataprofiler/profilers/profile_builder.py
AnhTruong: restructured the if. 
For the 'updated' correlation, I don't know what's the appropriate value to set. Currently, I keep it as the previous value, with the warning, then the users cannot use it anyway. But I'm open to a discussion on this. Plus, we'll have to deal with this shortly, so, it'd be temporally fine to leave it that way imo.

dataprofiler/profilers/profile_builder.py
AnhTruong: modified the message. Does that look better?

dataprofiler/profilers/profile_builder.py
AnhTruong: as we discussed, this merge works for two profiles without nan values. I think you mentioned about the `update_correlation`, not the merge.

dataprofiler/profilers/profile_builder.py
AnhTruong: as above

dataprofiler/profilers/profiler_options.py
AnhTruong: i changed it already

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: it's a dataframe. So, I have to convert to array, concatenate them and convert to list. df doesn't have .tolist()

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: you meant the clean_report function? I can do it if we want.

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: it'll become list of lists

dataprofiler/profilers/profile_builder.py
JGSweets: Currently, this just returns the whole matrix though right? Based on the other functions this might be a bit confusing as the other `_get_*` appear to do a calculation of the statistic referenced. Also, I think the separation as mentioned above is important so we can start adding time tracking to the top level profilers as well.

dataprofiler/profilers/profile_builder.py
JGSweets: does the null count need to be acquired from the profile or can we get it directly from the class property?

dataprofiler/profilers/profile_builder.py
JGSweets: Either way we leave it, I think the if should go above the calculations as we don't need to waste compute if we aren't even going to use the values anyways.

dataprofiler/profilers/profile_builder.py
JGSweets: I'm not a fan of the phrase doesn't work as oppose to coming in a future release.

JGSweets: There also needs to be description of what that means, e.g. how does that affect the user's profile.

dataprofiler/profilers/profiler_options.py
JGSweets: What does that mean?

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Isn't that what is desired?

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: indeed

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: changed. Initially I thought the assertequal doesn't work with list of lists, so I have to do np.concatenate

dataprofiler/profilers/profiler_options.py
AnhTruong: to CorrelationOption

dataprofiler/profilers/profile_builder.py
AnhTruong: fixed

dataprofiler/profilers/profile_builder.py
AnhTruong: I'm not completely follow what you mentioned above, but my understanding is to separate calculate corr from scratch vs update corr from an existing one? It maybe good, but currently we don't have that for any other stat calculation. Should be a separate PR for all of them?

About the timing, do we have the timing update for get_row_statistics? If yes, we can just do the same as they are at the same level?

dataprofiler/profilers/profile_builder.py
AnhTruong: any suggestion of how to get it from the class?

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: brought to utils

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,125 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+
+        :param clean_samples: the input cleaned dataset
+        :type clean_samples: dict()
+        """
+        columns = self.options.correlation.columns
+        if columns is None:
+            for col in self._profile:
+                if (self._profile[col].profile['data_type'] not in ['int', 'float']
+                        or self._profile[col].profile['statistics']['null_count'] > 0):
+                    clean_samples.pop(col)
```
JGSweets, line 1374: this basically means any column that has a single nan will no longer be part of corr? That's dangerous, no?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,125 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        :type corr_mat2: pd.DataFrame
+        :param mean2: mean of columns of profile2
+        :type mean2: np.array
+        :param std2: standard deviation of columns of profile2
+        :type std2: np.array
+        :return: merged correlation matrix
+        """
+        if len(corr_mat1.columns) != len(corr_mat2.columns) or \
+                len(corr_mat1.columns) <= 1:
+            return None
```
JGSweets, line 1448: is `None` appropriate? cause then if someone updated after merge, wouldn't it try to update again?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,125 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        self.correlation_matrix = data.corr()
+
+    def _update_correlation(self, clean_samples):
+        """
+        Update correlation matrix for cleaned data.
+
+        :param clean_samples: the input cleaned dataset
+        :type clean_samples: dict()
+        """
+        if self.correlation_matrix is not None:
```
JGSweets, line 1389: maybe it should be based on samples processed as oppose to `None` this would fix the other issue and allow it to be set to `None` still.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,125 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        if len(corr_mat1.columns) != len(corr_mat2.columns) or \
+                len(corr_mat1.columns) <= 1:
+            return None
+        if set(corr_mat1.columns).difference(set(corr_mat2.columns)):
+            return None
+
+        std_mat1 = np.outer(std1, std1)
+        std_mat2 = np.outer(std2, std2)
+        mean_diff_vector = mean1 - mean2
+        mean_diff_mat = np.outer(mean_diff_vector, mean_diff_vector)
```
JGSweets, line 1455: what if the column order changed? Wouldn't this be an issue? Not saying it did, but what if? Do we want to map or have an auto block on even schema order change?

dataprofiler/profilers/profile_builder.py
AnhTruong: My view about this is that we're gonna have several options: exclude columns with nan, keep those columns but with nan results, keep those columns but only calculate results based on not-nan values, keep those columns but calculate results with replaced values for nan values. Later we can think about how to deal with other cases. For now, I think we can just add some warning messages at the very least.

dataprofiler/profilers/profile_builder.py
AnhTruong: not sure if we need an update after merge, but even that's the case it only requires checking the columns, right?

dataprofiler/profilers/profile_builder.py
AnhTruong: imo None here means either we don't have any data yet, or the current data doesn't have enough numerical data to calculate the correlation. So, I don't really see issue with this.

JGSweets: The problem becomes if you merge profiles and the value gets set as None bc they don't overlap in sets, if you then update the merged profile, it will then set the correlation because it is None.

Really, it should only update if the match count has not yet been updated since we are disabling updates after the first profile.

dataprofiler/profilers/profile_builder.py
JGSweets: looks like they have a property called `null_count`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,125 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+
+        :param clean_samples: the input cleaned dataset
+        :type clean_samples: dict()
+        """
+        if self.correlation_matrix is not None:
+            # currently return None with the exising correlation
+            # TODO: implement the update with the exising correlation
+            warnings.warn("Currently, the updated correlation is kept the same "
+                          "as the existing value. Updating correlations with "
+                          "existing value will be available in a future update")
```
JGSweets, line 1394: def better. maybe:
> Currently, the updating correlations is disabled and will remain unchanged from its current value. Updating correlations will be available in a future update.

JGSweets: thoughts?

dataprofiler/profilers/profile_builder.py
AnhTruong: good point, i've added a fix to that, please look at the merge_correlation function

dataprofiler/profilers/profile_builder.py
AnhTruong: changed

dataprofiler/profilers/profile_builder.py
AnhTruong: changed

dataprofiler/profilers/profile_builder.py
AnhTruong: changed to use with class property

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -323,7 +323,10 @@ def test_warning_tf_run_dp_multiple_times(self):
                 columns.append(col)
                 predictions.append(results['data_stats'][col]['data_label'])
 
-    def test_warning_tf_run_dp_merge(self):
+    from unittest import mock
```
JGSweets, line 326: move mock to top of file.

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -323,7 +323,10 @@ def test_warning_tf_run_dp_multiple_times(self):
...
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'StructuredProfiler._merge_correlation')
+    def test_warning_tf_run_dp_merge(self, *mocks):
```
JGSweets, line 329: Not that this is yours, but what is the intent of this test? it doesn't seem to assert anything. @granteden @ChrisWallace2020 ?

@AnhTruong why is the mock necessary?

JGSweets: Maybe instead the correlations should be disabled like the other props below?

JGSweets: similar statement to the tests below these in: `test_warning_tf_multiple_dp_with_update`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -207,6 +220,7 @@ def test_stream_profilers(self, *mocks):
         self.assertEqual(0, profiler.row_is_null_count)
         self.assertEqual(3, profiler.total_samples)
         self.assertEqual(2, len(profiler.hashed_row_dict))
+        self.assertIsNone(profiler.correlation_matrix)
```
JGSweets, line 223: considering the update is mocked out, not sure theres a reason to validate the `correlation_matrix`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -218,6 +232,7 @@ def test_stream_profilers(self, *mocks):
         self.assertEqual(2, profiler.row_is_null_count)
         self.assertEqual(8, profiler.total_samples)
         self.assertEqual(5, len(profiler.hashed_row_dict))
+        self.assertIsNone(profiler.correlation_matrix)
```
JGSweets, line 235: as above

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -239,6 +254,53 @@ def test_correct_duplicate_row_count_test(self):
         self.assertEqual(2999, self.trained_schema.total_samples)
         self.assertEqual(0.0, self.trained_schema._get_duplicate_row_count())
 
+    def test_correlation(self):
```
JGSweets, line 257: should mock out labeler if it isn't auto mocked out.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -239,6 +254,53 @@ def test_correct_duplicate_row_count_test(self):
...
+        # data with one column, correlation not updated
+        data = pd.DataFrame([1.0, None, 1.0, None, 5.0])
+        profiler = dp.StructuredProfiler(data)
+        self.assertIsNone(profiler.correlation_matrix)
```
JGSweets, line 261: good

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -239,6 +254,53 @@ def test_correct_duplicate_row_count_test(self):
...
+        expected_corr_mat = data.corr()
+        pd.util.testing.assert_frame_equal(expected_corr_mat,
+                                           profiler.correlation_matrix)
+
+        # data with multiple numerical columns, with nan values
+        data = pd.DataFrame({'a': [np.nan, np.nan, 1, 7, 5, 9, 4, 10, 7, 2],
+                             'b': [10, 11, np.nan, 4, 2, 5, 6, 3, 9, 8],
+                             'c': [1, 5, 3, 5, 7, 2, 6, 8, np.nan, np.nan]})
+        profiler = dp.StructuredProfiler(data)
+        self.assertIsNone(profiler.correlation_matrix)
```
JGSweets, line 287: So current state cannot take any nan.

Can you actually make one column not have nans? In this case, the value should still be `None`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -239,6 +254,53 @@ def test_correct_duplicate_row_count_test(self):
...
+        data2 = data[int(len(data) / 2):]
+
+        profile1 = dp.StructuredProfiler(data1)
+        profile2 = dp.StructuredProfiler(data2)
+        merged_profile = profile1 + profile2
+
+        expected_corr_mat = data.corr()
+        pd.util.testing.assert_frame_equal(expected_corr_mat,
+                                           merged_profile.correlation_matrix)
+
```
JGSweets, line 303: I think this needs further discussion. I think it would be a silent confusing if the user receives a report that when merged only reflects one of the profiles. 

IMO, the correlations should be set to `None` in this case as we cannot compute the merge and the new correlations might actually be drastically different that the chosen values.

@ChrisWallace2020 @granteden @lettergram thoughts?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -413,11 +476,13 @@ def test_text_data_raises_error(self):
     @mock.patch('dataprofiler.profilers.profile_builder.StructuredProfiler.'
                 '_update_row_statistics')
     @mock.patch('dataprofiler.profilers.profile_builder.StructuredColProfiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'StructuredProfiler._update_correlation')
```
JGSweets, line 480: if this is moved to the top, would not need to change the array value below

dataprofiler/tests/profilers/utils.py
```diff
@@ -105,3 +105,29 @@ def get_depth(od):
     if isinstance(od, dict):
         return 1 + (max(map(get_depth, od.values())) if od else 0)
     return 0
+
+
+def clean_report(report):
```
JGSweets, line 110: ty

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: done

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
AnhTruong: changed, correlation option is disabled now

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: agreed, removed it. Originally I kept it without the mock

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: removed

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: done

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: good point, added it

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: I don't really follow your comment. Why the correlation should be set to None in this case. Two halves of the data both have correlations, then we merge from them?

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: i didn't change any array value below this. Which changes you mentioned specifically?

AnhTruong: i see, changed it now

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -190,6 +201,8 @@ def test_add_profilers(self, *mocks):
     @mock.patch('dataprofiler.profilers.profile_builder.'
                 'ColumnDataLabelerCompiler')
     @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler')
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'StructuredProfiler._update_correlation')
```
JGSweets, line 205: why do we need this here? shouldn't we enforce correlation in the stream to be `None` or what we think it should be?

JGSweets: Or is this being done elsewhere?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -239,6 +252,67 @@ def test_correct_duplicate_row_count_test(self):
...
+        data = pd.DataFrame({'a': [np.nan, np.nan, 1, 7, 5, 9, 4, 10, 7, 2],
+                             'b': [10, 11, np.nan, 4, 2, 5, 6, 3, 9, 8],
+                             'c': [1, 5, 3, 5, 7, 2, 6, 8, np.nan, np.nan]})
+        profiler = dp.StructuredProfiler(data)
+        self.assertIsNone(profiler.correlation_matrix)
+
+        # data with only one numerical columns without nan values
+        data = pd.DataFrame({'a': [np.nan, np.nan, 1, 7, 5, 9, 4, 10, 7, 2],
+                             'b': [10, 11, np.nan, 4, 2, 5, 6, 3, 9, 8],
+                             'c': [1, 5, 3, 5, 7, 2, 6, 8, np.nan, np.nan]})
```
JGSweets, line 293: all 3 still have nans here, or am i reading it wrong?

Also, maybe add a case where only 1 column has NaN?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -239,6 +252,67 @@ def test_correct_duplicate_row_count_test(self):
...
+    def test_merge_correlation(self, *mock):
+        data = pd.DataFrame({'a': [3, 2, 1, 7, 5, 9, 4, 10, 7, 2],
+                             'b': [10, 11, 1, 4, 2, 5, 6, 3, 9, 8],
+                             'c': [1, 5, 3, 5, 7, 2, 6, 8, 1, 2]})
+        data1 = data[:int(len(data) / 2)]
+        data2 = data[int(len(data) / 2):]
+
+        profile1 = dp.StructuredProfiler(data1)
+        profile2 = dp.StructuredProfiler(data2)
+        merged_profile = profile1 + profile2
```
JGSweets, line 310: this seems dangerous to test with the exact same data. I would instead suggest having data1 and data2 be different to ensure it is correct.

JGSweets: Probably simplify the data significantly too so that you can manually calculate the expected correlations as oppose to getting the result and just putting that as the test. (e.g. if code was actually incorrect and the result was just copied the test would pass but be wrong).

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: this test uses mock, then 'statistics' is not updated, therefore I had to use mock for update_correlation. The merge is tested in test_merge_correlation

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: added

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: The concern is that merge != stream. We need to test the condition when we stream data the the correlation gets set to `None`

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: data1 and data2 are different
```
data1 = data[:int(len(data) / 2)]
data2 = data[int(len(data) / 2):]
```

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Also, maybe instead of mocking out update, you mock out get? and set the correlation to a specific value?

Then you can test the stream functionality:
```
if self.total_samples > 0:
            # currently return None with the existing correlation
            # TODO: implement the update with the existing correlation
            warnings.warn("Currently, the updating correlations is disabled "
                          "and will remain unchanged from its current value. "
                          "Updating correlations will be available in a "
                          "future update.")
            return
```
However, I still am of the opinion that in a stream this should set the correlation matrix to `None`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,131 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+    def _update_correlation(self, clean_samples):
+        """
+        Update correlation matrix for cleaned data.
+
+        :param clean_samples: the input cleaned dataset
+        :type clean_samples: dict()
+        """
+        if self.total_samples > 0:
+            # currently return None with the exising correlation
+            # TODO: implement the update with the exising correlation
```
JGSweets, line 1391: two of these: `exising` -> `existing`

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: for merge makes sense.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,131 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+            for col in self._profile:
+                if (self._profile[col].profile['data_type'] not in ['int', 'float']
+                        or self._profile[col].null_count > 0):
+                    clean_samples.pop(col)
+        if len(clean_samples) <= 1:
+            return
+
+        data = pd.DataFrame(clean_samples)
+        data = data.apply(pd.to_numeric, errors='coerce')
+        self.correlation_matrix = data.corr()
```
JGSweets, line 1380: instead of setting, we should return the corr and allow `_update_correlation` to set the value.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,131 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        :type clean_samples: dict()
+        """
+        if self.total_samples > 0:
+            # currently return None with the exising correlation
+            # TODO: implement the update with the exising correlation
+            warnings.warn("Currently, the updating correlations is disabled "
+                          "and will remain unchanged from its current value. "
+                          "Updating correlations will be available in a "
+                          "future update.")
+            return
```
JGSweets, line 1396: I believe this is still dangerous to set the value to the previous when we update as it gives a false expectation of what the correlations could be.

i would prefer to set the value to None if it gets updated.

JGSweets: This also means the warning message needs to change.

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: added the test for stream

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: added the test for stream

dataprofiler/profilers/profile_builder.py
AnhTruong: fixed

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,132 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+            # currently return None with the existing correlation
+            # TODO: implement the update with the existing correlation
+            warnings.warn("Currently, the correlation update is disabled "
+                          "with existing correlation, which will be reset "
+                          "to None. Updating correlations will be available "
+                          "in a future update.")
+            self.correlation_matrix = None
+            return
+
+        self._get_correlation(clean_samples)
```
JGSweets, line 1399: when we return the correlation via _get_correlation we can then do 
`self.correlation_matrix = self._get_correlation(clean_samples)`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,132 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+            for col in self._profile:
+                if (self._profile[col].profile['data_type'] not in ['int', 'float']
+                        or self._profile[col].null_count > 0):
+                    clean_samples.pop(col)
+        if len(clean_samples) <= 1:
+            return None
+
+        data = pd.DataFrame(clean_samples)
+        data = data.apply(pd.to_numeric, errors='coerce')
+        return data.corr()
```
ChrisWallace2020, line 1380: Not in this PR, but maybe in a subsequent, we could allow user to specify correlation method via options, because I know thats an argument to `.corr()`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1355,11 +1359,132 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+
+        :param other: the other profile that needs to be merged
+        :return:
+        """
+        corr_mat1 = self.correlation_matrix
+        corr_mat2 = other.correlation_matrix
+        if corr_mat1 is None:
+            return corr_mat2
+        if corr_mat2 is None:
+            return corr_mat1
```
ChrisWallace2020, line 1413: This indicates case where correlation is disabled in one profiler but not the other, should we error in this case then? @JGSweets @gme5078 ?

JGSweets: Not error, but I see the point. If either is None, should we return None since it cannot merge?

dataprofiler/profilers/profile_builder.py
AnhTruong: yeah, we thought about this. Currently only Pearson's correlation is feasible for merging. Spearman or Chi square are tough for now

dataprofiler/profilers/profile_builder.py
```diff
@@ -1257,7 +1260,8 @@ def report(self, report_options=None):
                 "unique_row_ratio": self._get_unique_row_ratio(),
                 "duplicate_row_count": self._get_duplicate_row_count(),
                 "file_type": self.file_type,
-                "encoding": self.encoding
+                "encoding": self.encoding,
+                "correlation_matrix": self.correlation_matrix
```
ChrisWallace2020, line 1264: Should either leave out of report for now or serialize it somehow (json or something of the sort)

AnhTruong: convert to json now

JGSweets (dismissed):
Lets follow-up this PR with the standardization for duplicate columns

ChrisWallace2020 (approved):

JGSweets (approved):

PR #282, opened 2021-06-24
Add diff to DateTime Column

andrew-yin:
DateTimeColumns can now be compared in terms of maximum, minimum, and the formats.

### DIFF
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -128,6 +128,13 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
         return None
 
     def diff(self, other_profile, options=None):
+        """
+        Generates the differences between max, min, and formats of two DateTime columns
+
+        :return: Dict containing the differences between max, min, and format in their
+        appropriate output formats
+        :rtype: dict
+        """
         cls = self.__class__
         if not isinstance(other_profile, cls):
             raise TypeError("Unsupported operand type(s) for diff: '{}' "
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -141,14 +141,9 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
                             "and '{}'".format(cls.__name__,
                                               other_profile.__class__.__name__))
 
-        # We can use find_diff_of_numbers for max/min to get a timedelta object,
-        # since datetime objects can be compared and subtracted naturally
-        min_diff = utils.find_diff_of_numbers(self._dt_obj_min, other_profile._dt_obj_min)
-        max_diff = utils.find_diff_of_numbers(self._dt_obj_max, other_profile._dt_obj_max)
-
         differences = {
-            "min": utils.timedelta_to_readable_output(min_diff),
-            "max": utils.timedelta_to_readable_output(max_diff),
+            "min": utils.find_diff_of_dates(self._dt_obj_min, other_profile._dt_obj_min),
+            "max": utils.find_diff_of_dates(self._dt_obj_max, other_profile._dt_obj_max),
             "format": utils.find_diff_of_lists_and_sets(
                 self.date_formats, other_profile.date_formats)
         }
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -143,9 +143,12 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
 
         # We can use find_diff_of_numbers for max/min to get a timedelta object,
         # since datetime objects can be compared and subtracted naturally
+        min_diff = utils.find_diff_of_numbers(self._dt_obj_min, other_profile._dt_obj_min)
+        max_diff = utils.find_diff_of_numbers(self._dt_obj_max, other_profile._dt_obj_max)
+
         differences = {
-            "min": str(utils.find_diff_of_numbers(self._dt_obj_min, other_profile._dt_obj_min)),
-            "max": str(utils.find_diff_of_numbers(self._dt_obj_max, other_profile._dt_obj_max)),
+            "min": utils.timedelta_to_readable_output(min_diff),
+            "max": utils.timedelta_to_readable_output(max_diff),
             "format": utils.find_diff_of_lists_and_sets(
                 self.date_formats, other_profile.date_formats)
         }
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -127,6 +127,23 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
             return float(self.match_count) / self.sample_size
         return None
 
+    def diff(self, other_profile, options=None):
+        cls = self.__class__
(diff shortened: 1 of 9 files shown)


dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -127,6 +127,23 @@ def data_type_ratio(self):
...
+        cls = self.__class__
+        if not isinstance(other_profile, cls):
+            raise TypeError("Unsupported operand type(s) for diff: '{}' "
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+
+        # We can use find_diff_of_numbers for max/min to get a timedelta object,
+        # since datetime objects can be compared and subtracted naturally
+        differences = {
+            "min": str(utils.find_diff_of_numbers(self._dt_obj_min, other_profile._dt_obj_min)),
```
gme5078, line 140: What does the result look like? Can I see an example?

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -127,6 +127,23 @@ def data_type_ratio(self):
             return float(self.match_count) / self.sample_size
         return None
 
+    def diff(self, other_profile, options=None):
```
gme5078, line 130: Dont forget to right a docstring :)

dataprofiler/profilers/datetime_column_profile.py
gme5078: Oh nvm, I see in the test. Was there any options for the output format of the timedelta object?

JGSweets: Ya, we should prob use one of the existing formats from the profile

dataprofiler/tests/profilers/test_datetime_column_profile.py
```diff
@@ -371,3 +371,45 @@ def test_datetime_column_with_wrong_options(self):
...
+            2.5, 12.5, '2013-03-10 15:43:30', 5, '03/10/14 15:43',
+            'Mar 11, 2013'
+        ]
+        df2 = pd.Series(data2).apply(str)
+        profiler2 = DateTimeColumn(df2.name)
+        profiler2.update(df2)
+
+        expected_diff = {
+            'min': "1 days 08:16:30",
+            'max': "-10 days +08:17:00",
```
JGSweets, line 392: i like that it does this

dataprofiler/profilers/datetime_column_profile.py
andrew-yin: There's not a way to directly change the output format as far as I can tell besides writing a custom function. If we were to use one of the existing formats, the only one that seems to be suitable for a difference in times is "%H:%M:%S.%f", but I also don't know how "negative" time differences could be represented in this format and if it's necessarily a better practice if it ends up becoming more unreadable. Thoughts? @gme5078 @JGSweets

gme5078: The only thing I dont like about the current format is that if the data scientists want to set a threshold, it feels like they might have to parse the output. I think we can leave it like this for now and change if our customers ask for it.
We cant really use any of the existing formats in the datetimecolumn since those are mostly dates and dates dont make sense as a metric of difference in time.

andrew-yin: @gme5078 I guess it depends on the use case. For serialization purposes, the default output for a timedelta object cast to str is the most readable and understandable format to me. However, if diff() is being used in actual code, then maybe we can change this to returning the actual timedelta object. I'm just not sure what the intended use cases of diff() are.

dataprofiler/profilers/utils.py
```diff
@@ -411,3 +411,21 @@ def find_diff_of_lists_and_sets(stat1, stat2):
...
+
+    :param diff: The difference in time to get the readable output
+    :type diff: datetime.timedelta object
+    :return: Readable output
+    :rtype: str
+    """
+    if diff.days >= 0:
+        return str(diff)
+
+    return "-"+str(abs(diff))
```
JGSweets, line 431: add spaces around the `+`

dataprofiler/profilers/utils.py
```diff
@@ -411,3 +411,21 @@ def find_diff_of_lists_and_sets(stat1, stat2):
...
+    output. This returns a readable output for timedelta that
+    accounts for potential negative differences.
+
+    :param diff: The difference in time to get the readable output
+    :type diff: datetime.timedelta object
+    :return: Readable output
+    :rtype: str
+    """
+    if diff.days >= 0:
+        return str(diff)
```
JGSweets, line 429: since we do the `-` below, i think we should do the `+` here to standardize it, thoughts? other time formats also indicate `+`s

dataprofiler/profilers/utils.py
```diff
@@ -411,3 +411,21 @@ def find_diff_of_lists_and_sets(stat1, stat2):
         unique2 = [element for element in stat2 if element not in stat1]
         diff = [unique1, shared, unique2]
     return diff
+
```
JGSweets, line 414: should be two spaces between functions  not within a class

dataprofiler/profilers/utils.py
```diff
@@ -411,3 +411,21 @@ def find_diff_of_lists_and_sets(stat1, stat2):
...
+def timedelta_to_readable_output(diff):
```
JGSweets, line 415: I think it might be better to just have a util function which is `find_diff_of_datetime` and to just have one call. Thoughts?

JGSweets: Also, we will need a test specifically for this function or the new function described above.

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -127,6 +127,33 @@ def data_type_ratio(self):
...
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+
+        # We can use find_diff_of_numbers for max/min to get a timedelta object,
+        # since datetime objects can be compared and subtracted naturally
+        min_diff = utils.find_diff_of_numbers(self._dt_obj_min, other_profile._dt_obj_min)
+        max_diff = utils.find_diff_of_numbers(self._dt_obj_max, other_profile._dt_obj_max)
+
+        differences = {
+            "min": utils.timedelta_to_readable_output(min_diff),
```
JGSweets, line 150: if we update for the new function as described below, can then just call it here and don't need the diff funcs above.

dataprofiler/profilers/utils.py
andrew-yin: I feel like it's a bit redundant to find the differences between the dates within the function, because that would essentially be rewriting find_diff_of_numbers() inside the function.

dataprofiler/profilers/utils.py
```diff
@@ -411,3 +411,36 @@ def find_diff_of_lists_and_sets(stat1, stat2):
...
+    :type stat2: datetime.datetime object
+    :return: Difference in stats
+    :rtype: str
+    """
+    # We can use find_diff_of_numbers since datetime objects
+    # can be compared and subtracted naturally
+    diff = find_diff_of_numbers(stat1, stat2)
+    if isinstance(diff, str):
+        return diff
+    if isinstance(diff, list):
```
JGSweets, line 439: just add a comment for why this case occurs and will approve

JGSweets (approved):

gme5078 (approved):

PR #288, opened 2021-06-25
Updating the documentation to v0.5.2

gme5078:
Should note addition to statistics, setting the seed, and contributions.

Here is how it looks:
https://gme5078.github.io/data-profiler/


JGSweets (commented):
In future revision we need to give access or an API view of top level __init__ such that ppl understand the access to the `dataprofiler.Profiler`,`dataprofiler.set_seed`, etc

JGSweets (approved):

PR #291, opened 2021-06-28
StructuredProfiler _profile dict -> list refactor (part 1/3)

ChrisWallace2020:
This is the 1st part of a 3 part refactor, as follows:
1) Refactor `StructuredProfiler._profile` to be a list instead of a dict, and store a mapping that maps column names to a list of indices (in `_profile`) that correspond to data that was in a column of the given name. Refactor tests accordingly.
2) Refactor `StructuredProfiler` `report['data_stats']` to be a list that mirrors `StructuredProfiler._profile`, and stores a column name to index mapping in `global_stats`. Refactor tests accordingly.
3) Update documentation (README as well as notebooks)

Note that after this refactor (1/3) a lot of tests will fail due to `report` not being compatible with the new `_profile`, but to limit PR scope and streamline reviewing, this will have to be ignored until the 2nd refactor, at which point all tests (old and new) will pass.

To see the plan for the larger refactor (steps 1 and 2), it is here: #278
This PR is just a subset of changes made in #278

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1258,19 +1258,20 @@ class StructuredProfiler(BaseProfiler):
         :rtype: Dict[int, int]
         """
 
-        len_schema_1 = len(schema1)
-        len_schema_2 = len(schema2)
+        len_schema1 = len(schema1)
+        len_schema2 = len(schema2)
 
-        if strict and len_schema_1 != len_schema_2:
+        # If both non-empty, must be same length
+        if 0 < len_schema1 != len_schema2 > 0:
             raise ValueError("Attempted to merge profiles with different "
                              "numbers of columns")
 
         # In the case of __add__ with one of the schemas not initialized
-        if strict and (len_schema_1 == 0 or len_schema_2 == 0):
+        if strict and (len_schema1 == 0 or len_schema2 == 0):
             raise ValueError("Cannot merge empty profiles.")
 
         # In the case of _update_from_chunk with uninitialized schema
-        if not strict and len_schema_2 == 0:
+        if not strict and len_schema2 == 0:
             return {col_ind: col_ind for col_ind_list in schema1.values()
                     for col_ind in col_ind_list}
 
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1274,14 +1274,11 @@ class StructuredProfiler(BaseProfiler):
         :rtype: Dict[int, int]
         """
 
-        one_schema_empty = (len(schema1) == 0 and len(schema2) > 0) \
-                           or (len(schema1) > 0 and len(schema2) == 0)
-        # In the case of __add__ with one of the schemas not initialized
-        if strict and one_schema_empty:
-            raise ValueError("Cannot merge profiles due to schema mismatch")
+        if strict and (len(schema1) == 0 or len(schema2) == 0):
+            raise ValueError("Cannot merge empty profiles.")
 
         # In the case of _update_from_chunk with uninitialized schema
-        if len(schema2) == 0:
+        if not strict and len(schema2) == 0:
             return {col_ind: col_ind for col_ind_list in schema1.values()
                     for col_ind in col_ind_list}
 
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1173,7 +1173,19 @@ class StructuredProfiler(BaseProfiler):
         StructuredProfiler specific checks to ensure two profiles can be added
         together.
         """
-        if self._col_name_to_idx != other._col_name_to_idx:
+        schemas_differ = True
+        # Schemas must completely match if both contain duplicates
+        if self._duplicate_cols_present and other._duplicate_cols_present \
+                and self._col_name_to_idx == other._col_name_to_idx:
+            schemas_differ = False
+        # Can allow permutation of unique columns
+        elif not self._duplicate_cols_present and not \
+                other._duplicate_cols_present \
+                and self._col_name_to_idx.keys() == \
+                other._col_name_to_idx.keys():
+            schemas_differ = False
+
+        if schemas_differ:
             raise ValueError('Profiles do not have the same schema.')
         elif not all([isinstance(other._profile[idx],
                                  type(self._profile[idx]))
@@ -1203,8 +1215,14 @@ class StructuredProfiler(BaseProfiler):
 
         # merge profiles
         for idx in range(len(self._profile)):
+            other_idx = idx
+            # If columns unique, order may be permutated
+            if not self._duplicate_cols_present:
+                # Determine which index in other._profile corresponds to idx
+                col_name = self._get_col_name_from_idx(idx)
+                other_idx = other._col_name_to_idx[col_name][0]
             merged_profile._profile.append(self._profile[idx] +
-                                           other._profile[idx])
+                                           other._profile[other_idx])
         merged_profile._col_name_to_idx = copy.deepcopy(self._col_name_to_idx)
         return merged_profile
 
@@ -1235,24 +1253,13 @@ class StructuredProfiler(BaseProfiler):
         """
         return min([col._last_batch_size for col in self._profile], default=0)
 
-    def _profile_idx_to_data(self, data, prof_idx):
+    def _get_col_name_from_idx(self, idx):
         """
-        Go from _profile list index to the column in data that corresponds to
-        the index in _profile
+        Determine which column name corresponds to a _profile index
         """
-        # Find column name that corresponds to index in _profile
-        for col in self._col_name_to_idx:
-            col_idxs = self._col_name_to_idx[col]
-            if prof_idx in col_idxs:
-                # Case where col isn't duplicated, no need to look through ids
-                if len(col_idxs) == 1:
-                    return data[col]
-                # If there are duplicate columns under the same name, need
-                # to figure place in list of indexes, as this corresponds
-                # to place in list of series returned by data[col]
-                for i in range(len(col_idxs)):
-                    if col_idxs[i] == prof_idx:
-                        return data[col][i]
+        for col_name in self._col_name_to_idx:
+            if idx in self._col_name_to_idx[col_name]:
+                return col_name
 
     def report(self, report_options=None):
         if not report_options:
@@ -1395,35 +1402,43 @@ class StructuredProfiler(BaseProfiler):
 
         duplicate_cols_given = len(data.columns) != len(data.columns.unique())
 
-        # Error out if initialized with unique columns but given duplicates
-        if self._initialized and duplicate_cols_given \
-                and not self._duplicate_cols_present:
-            raise ValueError("Attempted to update data with duplicate "
-                             "column names that weren't present before "
-                             "update. Schema must be identical when "
-                             "profiling data with duplicate column names.")
-
-        # Error out if duplicate columns present and given schema doesn't match
-        if self._duplicate_cols_present:
+        # Error out if schema given does not match schema present in data
+        if self._initialized:
             mapping_given = dict()
             for i in range(len(data.columns)):
                 col = data.columns[i]
+                if isinstance(col, str):
+                    col = col.lower()
                 mapping_given.setdefault(col, []).append(i)
-            if mapping_given != self._col_name_to_idx:
-                raise ValueError("Schema of data with duplicate column names "
-                                 "not respected when updating profile.")
+
+            # Determine if incoming schema is compatible with existing
+            schemas_differ = True
+            # If duplicates present, must completely match
+            if self._duplicate_cols_present and \
+                    mapping_given == self._col_name_to_idx:
+                schemas_differ = False
+            # If no duplicates present, can allow permutation of col order
+            # But cannot allow introduction of duplicates
+            elif not self._duplicate_cols_present and not duplicate_cols_given \
+                    and mapping_given.keys() == self._col_name_to_idx.keys():
+                schemas_differ = False
+
+            if schemas_differ:
+                raise ValueError("Schema of data given to "
+                                 "StructuredProfiler.update does not match "
+                                 "schema calculated at initialization.")
 
         try:
             from tqdm import tqdm
         except ImportError:
             def tqdm(l):
                 for i, e in enumerate(l):
-                    print("Processing Column {}/{}".format(i+1, len(l)))
+                    print("Processing Column {}/{}".format(i + 1, len(l)))
                     yield e
 
         # Shuffle indices once and share with columns
         sample_ids = [*utils.shuffle_in_chunks(len(data), len(data))]
-        
+
         # If there are no minimum true samples, you can sort to save time
         if min_true_samples in [None, 0]:
             sample_ids[0] = sample_ids[0][:sample_size]
@@ -1436,47 +1451,63 @@ class StructuredProfiler(BaseProfiler):
         # Newly introduced features (python3.8) improves the situation
         sample_ids = np.array(sample_ids)
 
-        # Create StructuredColProfilers (must be either first initialization
-        # or unique column names)
+        # Initialize data column index -> _profile index mapping
+        col_idx_to_prof_idx = dict()
+
+        # Create StructuredColProfilers upon initialization
+        # Record correlation between columns in data and index in _profile
         new_cols = False
-        if not self._duplicate_cols_present:
-            # Either initializing _profile for the first time or updating
-            # _profile that doesn't contain duplicate column names
-            for col in data.columns:
-                # If initializing for the first time, must fill mapping
-                # If initialized, with no duplicate columns, only add to mapping
-                # if not already there, since this means we are updating
-                if not self._initialized or col not in self._col_name_to_idx:
-                    # Append StructuredColProfiler to list of profiles
-                    # and record index where it was appended to in list
-                    self._col_name_to_idx.setdefault(col, []).append(
-                        len(self._profile))
-                    self._profile.append(StructuredColProfiler(
-                        sample_size=sample_size,
-                        min_true_samples=min_true_samples,
-                        sample_ids=sample_ids,
-                        options=self.options
-                    ))
-                    new_cols = True
-                
+        if not self._initialized:
+            for col_idx in range(data.shape[1]):
+                col_name = data.columns[col_idx]
+                if isinstance(col_name, str):
+                    col_name = col_name.lower()
+                # Record index in _profile corresponding to current col profile
+                self._col_name_to_idx.setdefault(col_name, []).append(
+                    len(self._profile))
+                # Record where this column of data corresponds to _profile id
+                col_idx_to_prof_idx[col_idx] = len(self._profile)
+                # Add blank StructuredColProfiler to _profile
+                self._profile.append(StructuredColProfiler(
+                    sample_size=sample_size,
+                    min_true_samples=min_true_samples,
+                    sample_ids=sample_ids,
+                    options=self.options
+                ))
+                new_cols = True
+        # Determine data column index -> profile index for unique col case
+        else:
+            # If duplicates present, schema enforced, so col_idx = prof_idx
+            col_idx_to_prof_idx = {i: i for i in range(len(self._profile))}
+            if not self._duplicate_cols_present:
+                for col_idx in range(data.shape[1]):
+                    col_name = data.columns[col_idx]
+                    if isinstance(col_name, str):
+                        col_name = col_name.lower()
+                    col_idx_to_prof_idx[col_idx] = \
+                    self._col_name_to_idx[col_name][0]
+
         # Generate pool and estimate datasize
         pool = None
         if self.options.multiprocess.is_enabled:
-            est_data_size = data[:50000].memory_usage(index=False, deep=True).sum()
+            est_data_size = data[:50000].memory_usage(index=False,
+                                                      deep=True).sum()
             est_data_size = (est_data_size / min(50000, len(data))) * len(data)
             pool, pool_size = utils.generate_pool(
                 max_pool_size=None, data_size=est_data_size,
                 cols=len(data.columns))
 
         # Format the data
-        notification_str = "Finding the Null values in the columns..."        
+        notification_str = "Finding the Null values in the columns..."
         if pool and new_cols:
             notification_str += " (with " + str(pool_size) + " processes)"
-        
+
+        # Keys are _profile indices
         clean_sampled_dict = {}
+        # Keys are column indices in data
         multi_process_dict = {}
         single_process_list = set()
-        
+
         if sample_size < len(data):
             warnings.warn("The data will be profiled with a sample size of {}. "
                           "All statistics will be based on this subsample and "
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1199,10 +1200,10 @@ def __add__(self, other):
...
 
         # merge profiles
-        for profile_name in self._profile:
-            merged_profile._profile[profile_name] = (
-                self._profile[profile_name] + other._profile[profile_name]
-            )
+        for idx in range(len(self._profile)):
+            merged_profile._profile.append(self._profile[idx] +
+                                           other._profile[idx])
+        merged_profile._col_name_to_idx = copy.deepcopy(self._col_name_to_idx)
```
JGSweets, line 1206: prior to this, how have we validated that the two schemas are the same to allow for merging?

JGSweets: the above only occurs if there are duplicate keys

JGSweets: If schemas are slightly different, and no duplicate keys, I believe this would error because indexes could be different or non-existent for profiles that partially profiled different columns in a datasets.
maybe best to loop through the  `_col_name_to_idx` based on keys

JGSweets: also in the case where we merge but not the same mapping, just copying the dict would not be an issue and instead would need to merge

JGSweets: Or do we only allow merging of exact schema matches.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1374,10 +1391,10 @@ def _update_profile_from_chunk(self, data, sample_size,
         elif isinstance(data, list):
             data = pd.DataFrame(data)
 
-        if len(data.columns) != len(data.columns.unique()):
-            raise ValueError('`StructuredProfiler` does not currently support '
-                             'data which contains columns with duplicate '
-                             'names.')
+        duplicate_cols_given = len(data.columns) != len(data.columns.unique())
+        num_cols = [len(idxs) for idxs in self._col_name_to_idx.values()]
```
JGSweets, line 1395: maybe we should store this value so we don't have to recalc all the time?

dataprofiler/profilers/profile_builder.py
gme5078: ^was thinking same thing, that schema could error when slightly different with this implementation

dataprofiler/profilers/profile_builder.py
JGSweets: or should we also store a list

dataprofiler/profilers/profile_builder.py
```diff
@@ -1374,10 +1391,10 @@ def _update_profile_from_chunk(self, data, sample_size,
...
+        duplicate_cols_present = any([num > 1 for num in num_cols])
```
JGSweets, line 1396: same as above, should we just have this stored as known info private to the profiler?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1402,17 +1419,46 @@ def tqdm(l):
...
+        # or unique column names)
+        new_cols = False
+        if not duplicate_cols_present:
+            # Given duplicate columns despite duplicate columns not being
+            # present in nonempty _profile
+            if initialized and duplicate_cols_given:
+                raise ValueError("Attempted to update data with duplicate "
+                                 "column names that weren't present before "
+                                 "update. Schema must be identical when "
+                                 "profiling data with duplicate column names.")
```
JGSweets, line 1432: this should error / be raised prior to any calculations being done, including sample_ids

dataprofiler/profilers/profile_builder.py
gme5078, line 1396: Whats the difference between duplicate_cols_given and duplicate_cols_present?

gme5078: oh nvm i see

dataprofiler/profilers/profile_builder.py
JGSweets: I'm slightly confused by these conditionals. The case where duplicate columns should error is if duplicate columns are present and new data is added which has the same name as the duplicate column.

JGSweets: However, I'm wondering if we should just require the schema to always match and not allow a profile to be updated a col at a time as this causes some headaches.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1402,17 +1419,46 @@ def tqdm(l):
...
+        else:
+            # Ensure same schema if initialized with duplicate columns
+            # No adding columns to _profile after initialization in this case
+            mapping_given = dict()
+            for i in range(len(data.columns)):
+                col = data.columns[i]
+                mapping_given.setdefault(col, []).append(i)
+            if mapping_given != self._col_name_to_idx:
+                raise ValueError("Schema of data with duplicate column names "
+                                 "not respected when updating profile.")
```
JGSweets, line 1461: this should be checked above prior to any calculation. 

However, is it no adding to columns period or no adding to a schema which has duplicate keys and you are adding the same key name?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -570,6 +564,93 @@ def test_dict_in_data_no_error(self, *mocks):
...
+            self.assertEqual(col_min, profiler._profile[col_idx].
+                             profile["statistics"]["min"])
+            self.assertEqual(col_max, profiler._profile[col_idx].
+                             profile["statistics"]["max"])
+            self.assertEqual(col_sum, profiler._profile[col_idx].
+                             profile["statistics"]["sum"])
+
+        # Check that update works as expected
+        new_data = pd.DataFrame([[100, 200, 300, 400, 500, 600]],
+                                columns=["a", "b", "a", "b", "c", "d"])
```
gme5078, line 597: what happens when the column order is switch around?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1439,54 +1485,108 @@ def tqdm(l):
...
-                if min_true_samples is None:
-                    min_true_samples = self._profile[col]._min_true_samples
-                try:
-                    multi_process_dict[col] = pool.apply_async(
-                        self._profile[col].clean_data_and_get_base_stats,
-                        (data[col], sample_size, min_true_samples, sample_ids))
-                except Exception as e:
-                    print(e)
-                    single_process_list.add(col)
+            for col in data.columns.unique():
```
JGSweets, line 1488: instead of looping by column name, just loop through by column index in the df?

JGSweets: I think the if statements maybe able to removed if this is the case and / or the code can be condensed.

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: In `_add_error_checks` the 2 profiles are required to have the same `_col_name_to_idx` mapping. I recall a conversation we had where you told me I had freedom to make schema requirements more strict in order to add this capability, is this no longer the case?

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: The logic is as follows:
If unitinitialized:
- Add a blank SCP for each column, and add each column to mapping (can be duplicate)

If initialized:
- If no duplicate cols in initial data, can add columns that weren't previously present, can update columns already present, but cannot give duplicate columns in data given to update
- If duplicate cols in initial data, no adding of cols at update, update data must look identical to data given at initialization

dataprofiler/profilers/profile_builder.py
```diff
@@ -1160,7 +1160,10 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
         self.row_has_null_count = 0
         self.row_is_null_count = 0
         self.hashed_row_dict = dict()
-        self._profile = dict()
+        self._profile = []
+        self._col_name_to_idx = dict()
+        self._initialized = False
+        self._duplicate_cols_present = False
```
JGSweets, line 1166: we can make these static values, but the operations themselves could just be reactive properties to information we already store so we don't have to store more info but rather have it as a property function.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1374,10 +1393,25 @@ def _update_profile_from_chunk(self, data, sample_size,
         elif isinstance(data, list):
             data = pd.DataFrame(data)
 
-        if len(data.columns) != len(data.columns.unique()):
-            raise ValueError('`StructuredProfiler` does not currently support '
-                             'data which contains columns with duplicate '
-                             'names.')
+        duplicate_cols_given = len(data.columns) != len(data.columns.unique())
```
JGSweets, line 1396: this does not account for fuzzy lower, correct? We should maybe make a check as a separate function for the case of allowing user to set case sensitivity

dataprofiler/profilers/profile_builder.py
```diff
@@ -1374,22 +1400,45 @@ def _update_profile_from_chunk(self, data, sample_size,
...
+
+        # Error out if schema given does not match schema present in data
+        if self._initialized:
+            mapping_given = dict()
+            for i in range(len(data.columns)):
+                col = data.columns[i]
+                if isinstance(col, str):
```
JGSweets, line 1410: maybe mention a comment above this suggest that pandas by default will give DFs an integer value for the name if columns aren't specified hence why this is needed

dataprofiler/profilers/profile_builder.py
```diff
@@ -1374,22 +1400,45 @@ def _update_profile_from_chunk(self, data, sample_size,
...
+                    col = col.lower()
+                mapping_given.setdefault(col, []).append(i)
```
JGSweets, line 1412: if `mapping_given` is initialized to a `defaultdict(list)`, can just do `mapping_given[col].append(i)`

JGSweets: not sure which is more greedy

dataprofiler/profilers/profile_builder.py
```diff
@@ -1170,11 +1173,23 @@ def _add_error_checks(self, other):
...
+            schemas_differ = False
+
+        if schemas_differ:
             raise ValueError('Profiles do not have the same schema.')
-        elif not all([isinstance(other._profile[p_name],
-                                 type(self._profile[p_name]))
-                      for p_name in self._profile]):  # options check
+        elif not all([isinstance(other._profile[idx],
+                                 type(self._profile[idx]))
+                      for idx in range(len(self._profile))]):  # options check
```
JGSweets, line 1192: How redundant is this with the checks in the update chunk? e.g. can we reduce these both to use the same func ultimately?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1223,17 +1243,23 @@ def _min_col_samples_used(self):
...
     def _min_sampled_from_batch(self):
         """
         Calculates and returns the number of rows that were completely sampled
         in the most previous batch
         """
-        return min([self._profile[col]._last_batch_size
-                    for col in self._profile], default=0)
+        return min([col._last_batch_size for col in self._profile], default=0)
+
+    def _get_col_name_from_idx(self, idx):
```
JGSweets, line 1256: should we use a func or just store this and its reverse mapping?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1402,104 +1451,143 @@ def tqdm(l):
...
+            col_idx_to_prof_idx = {i: i for i in range(len(self._profile))}
+            if not self._duplicate_cols_present:
+                # If columns unique, can allow permutation
+                for col_idx in range(data.shape[1]):
+                    col_name = data.columns[col_idx]
+                    if isinstance(col_name, str):
+                        col_name = col_name.lower()
+                    col_idx_to_prof_idx[col_idx] = \
+                        self._col_name_to_idx[col_name][0]
+
```
JGSweets, line 1490: I think if we utilize the mapping_given prior we may not need this at all.
e.g.
  1. Get mapping of new dataset (in case of add get mapping of other)
  2. Send to func to get dataset (other) to profile (self) mapping (raise error if schemas differ)
  3. If not initialized, create self._profile col profilers (not in add)
  4. Loop through using the mapping for updating etc.  (in add, just add based on mapping)

dataprofiler/profilers/profile_builder.py
```diff
@@ -1223,17 +1231,85 @@ def _min_col_samples_used(self):
...
+        :type strict: bool
+        :return: a mapping of indices in schema1 to indices in schema2
+        :rtype: Dict[int, int]
+        """
+
+        one_schema_empty = (len(schema1) == 0 and len(schema2) > 0) \
+                           or (len(schema1) > 0 and len(schema2) == 0)
+        # In the case of __add__ with one of the schemas not initialized
+        if strict and one_schema_empty:
+            raise ValueError("Cannot merge profiles due to schema mismatch")
```
JGSweets, line 1281: What is the case for both being empty?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1223,17 +1231,85 @@ def _min_col_samples_used(self):
...
+    def _get_col_name_from_idx(self, idx):
+        """
+        Determine which column name corresponds to a _profile index
+        """
+        for col_name in self._col_name_to_idx:
+            if idx in self._col_name_to_idx[col_name]:
+                return col_name
+
+    @staticmethod
+    def _get_and_validate_schema(schema1, schema2, strict=False):
```
JGSweets, line 1260: in this title, not completely clear that get means a mapping

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: There was a case where merging an initialized schema with unitinitialized should have raised an error, but didn't due to the logic for the uninitialized profiler in update_from_chunk case. So this ensures that in the merge case, this is no longer a blind spot. And the case for both being empty is 2 uninitialized profiles being merged, which I am unsure on what the correct behavior is, but if it is incorrect then I can make this 1 or statement, I just didn't know if both being empty was acceptable or not for merge.

ChrisWallace2020: This issue will be saved for a later refactor. We will want to allow merging empty profiles in the future.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1512,6 +1608,12 @@ def tqdm(l):
 
         self._update_row_statistics(data, samples_for_row_stats)
 
+        # Update personal attributes about state of profile
+        if not initialized:
+            num_cols = [len(self._col_name_to_idx[col])
+                        for col in self._col_name_to_idx]
+            self._duplicate_cols_present = any(num > 1 for num in num_cols)
+
```
JGSweets, line 1616: is this still needed?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1531,7 +1633,9 @@ def save(self, filepath=None):
                 "_samples_per_update": self._samples_per_update,
                 "_min_true_samples": self._min_true_samples,
                 "options": self.options,
-                "_profile": self.profile
+                "_profile": self.profile,
+                "_col_name_to_idx": self._col_name_to_idx,
+                "_duplicate_cols_present": self._duplicate_cols_present
```
JGSweets, line 1638: are these needed?

JGSweets, line 1638: is this still needed?

JGSweets, line 1638: is this line still needed?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1160,7 +1160,9 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
         self.row_has_null_count = 0
         self.row_is_null_count = 0
         self.hashed_row_dict = dict()
-        self._profile = dict()
+        self._profile = []
+        self._col_name_to_idx = defaultdict(list)
+        self._duplicate_cols_present = False
```
JGSweets, line 1165: do we still need `_duplicate_cols_present`?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1402,18 +1477,26 @@ def tqdm(l):
...
+        # Record correlation between columns in data and index in _profile
+        new_cols = False
+        if not initialized:
+            for col_idx in range(data.shape[1]):
+                col_name = data.columns[col_idx]
+                # Pandas cols are int by default, but need to fuzzy match strs
+                if isinstance(col_name, str):
+                    col_name = col_name.lower()
+                # Record index in _profile corresponding to current col profile
+                self._col_name_to_idx[col_name].append(len(self._profile))
```
JGSweets, line 1490: do we need this to be calculated here? if this is not initialized is 
`self._col_name_to_idx = mapping_given`?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1402,18 +1477,26 @@ def tqdm(l):
...
+                # Add blank StructuredColProfiler to _profile
+                self._profile.append(StructuredColProfiler(
```
JGSweets, line 1492: should we ensure that the name -> ind association is intact by using the mapping given as opposed to just appending and assuming it is matched?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1402,18 +1477,26 @@ def tqdm(l):
...
 
-        # Create structured profile objects
-        new_cols = set()
-        for col in data.columns:
-            if col not in self._profile:
-                self._profile[col] = StructuredColProfiler(
+        # Create StructuredColProfilers upon initialization
+        # Record correlation between columns in data and index in _profile
+        new_cols = False
+        if not initialized:
```
JGSweets, line 1483: since this is really the only place needed, can we just check the len of the profile here instead of using a var?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1223,17 +1231,74 @@ def _min_col_samples_used(self):
...
+
+            is_duplicate_col = len(schema1[key]) > 1
+            for schema1_col_ind, schema2_col_ind in zip(schema1[key],
+                                                        schema2[key]):
+                if is_duplicate_col and (schema1_col_ind != schema2_col_ind):
+                    raise ValueError(f"Different column indices under "
+                                     f"duplicate name '{key}', cannot update "
+                                     f"or merge unless schema is identical.")
+                schema_mapping[schema1_col_ind] = schema2_col_ind
+
```
gme5078, line 1300: I believe that Schema2 could have more columns than schema1 and no error would be thrown. That would be inappropriate right?

JGSweets: good catch, need to do a `len(schema1) == len(schema2)` check at beginning

dataprofiler/profilers/profile_builder.py
```diff
@@ -1223,17 +1231,74 @@ def _min_col_samples_used(self):
...
+        :rtype: Dict[int, int]
+        """
+
+        if strict and (len(schema1) == 0 or len(schema2) == 0):
+            raise ValueError("Cannot merge empty profiles.")
+
+        # In the case of _update_from_chunk with uninitialized schema
+        if not strict and len(schema2) == 0:
+            return {col_ind: col_ind for col_ind_list in schema1.values()
+                    for col_ind in col_ind_list}
```
JGSweets, line 1275: could assign lens at beginning to var so don't do multiple times

dataprofiler/profilers/profile_builder.py
```diff
@@ -1424,82 +1507,95 @@ def tqdm(l):
                 cols=len(data.columns))
 
         # Format the data
-        notification_str = "Finding the Null values in the columns..."        
-        if pool and len(new_cols) > 0:
+        notification_str = "Finding the Null values in the columns..."
+        if pool and new_cols:
```
gme5078, line 1511: I'm not sure why new_cols exists or why the notification addendum is dependent on new cols. Can we get rid of this?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -570,6 +564,139 @@ def test_dict_in_data_no_error(self, *mocks):
...
+            self.assertEqual(col_sum, profiler._profile[col_idx].
+                             profile["statistics"]["sum"])
+
+    def test_unique_col_permutation(self, *mocks):
+        data = pd.DataFrame([[1, 2, 3, 4],
+                             [5, 6, 7, 8]],
+                            columns=["a", "b", "c", "d"])
+        perm_data = pd.DataFrame([[4, 3, 2, 1],
+                                  [8, 7, 6, 5]],
+                                 columns=["d", "c", "b", "a"])
```
gme5078, line 623: Can we also test where they have different amounts of columns?

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Mapping given in the uninitialized case is a bunch of `{1:1, 2:2, 3:3, ...}` mappings, since its about column index to profile index, not column name to a list of profile indices.

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Not anymore, will remove

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Thats a good point, I never understood the need for this condition, will remove

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Removed

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Removed

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Implemented

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Implemented

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Implemented

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Implemented

dataprofiler/profilers/profile_builder.py
ChrisWallace2020: Implemented

dataprofiler/tests/profilers/test_profile_builder.py
ChrisWallace2020: Added

dataprofiler/profilers/profile_builder.py
```diff
@@ -1223,17 +1230,75 @@ def _min_col_samples_used(self):
...
+        :param strict: whether or not to strictly match (__add__ case)
+        :type strict: bool
+        :return: a mapping of indices in schema1 to indices in schema2
+        :rtype: Dict[int, int]
+        """
+
+        len_schema_1 = len(schema1)
+        len_schema_2 = len(schema2)
+
+        if strict and len_schema_1 != len_schema_2:
```
JGSweets, line 1264: this will allow profiles through when lens are different and strict isn't on which is not ideal.

JGSweets (approved):

gme5078 (approved):

PR #292, opened 2021-06-28
Added context manager

gautomdas:


dataprofiler/data_readers/data_utils.py
```diff
@@ -564,3 +564,22 @@ def load_as_str_from_file(file_path, file_encoding, max_lines=10,
...
+    :param seek: position to start in buffer
+    :type seek: int
+    :return: true if string is a buffer or false if string is a filepath
+    :rtype: boolean
+    """
+
+    if (type(filepath_or_buffer) == StringIO or type(filepath_or_buffer) == BytesIO):
+        return True
+    else:
+        return False
```
JGSweets, line 585: I think we can simplify this. a little bit:
```python
if isinstance(filepath_or_buffer, (StringIO, BytesIO):
    return True
return False
```

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
+from io import open, StringIO, BytesIO
+
+from . import data_utils
+
```
JGSweets, line 4: two spaces between imports and class / functions

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
...
+class FileOrBufferHandler():
+    def __init__(self, stream_or_buffer, open_method='r', options=dict()):
+        self._stream_or_buffer = stream_or_buffer
+        self.open_method = open_method
+        self.options = options
+
+    def __enter__(self):
+        if (data_utils.is_stream_buffer(self._stream_or_buffer)):
+            print(self.options)
```
JGSweets, line 13: remove the print

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
+from io import open, StringIO, BytesIO
+
+from . import data_utils
+
+class FileOrBufferHandler():
+    def __init__(self, stream_or_buffer, open_method='r', options=dict()):
```
JGSweets, line 6: instead of having options, lets just make seek an arg here, e.g. 
`def __init__(self, stream_or_buffer, open_method='r', seek=None):`

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
...
+        self._stream_or_buffer = stream_or_buffer
+        self.open_method = open_method
+        self.options = options
+
+    def __enter__(self):
+        if (data_utils.is_stream_buffer(self._stream_or_buffer)):
+            print(self.options)
+            if 'seek' in self.options:
+                self._stream_or_buffer.seek(self.options['seek'],0)
+                return self._stream_or_buffer
```
JGSweets, line 16: with the seek change, this can become `if seek is not None`

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
...
+            else:
+                return self._stream_or_buffer
+        else:
+            self._file = open(self._stream_or_buffer, self.open_method)
```
JGSweets, line 20: seek technically can be applied here, hence we could move the seek `if` out of the previous and to the end of this function

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, stream_or_buffer, open_method='r', options=dict()):
+        self._stream_or_buffer = stream_or_buffer
+        self.open_method = open_method
+        self.options = options
+
+    def __enter__(self):
+        if (data_utils.is_stream_buffer(self._stream_or_buffer)):
+            print(self.options)
+            if 'seek' in self.options:
+                self._stream_or_buffer.seek(self.options['seek'],0)
```
JGSweets, line 15: utilize the above suggestion for the input args and make sure there's a space after the comma

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
...
+            self._file = open(self._stream_or_buffer, self.open_method)
+            return self._file
+
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        if (data_utils.is_stream_buffer(self._stream_or_buffer)):
+            if 'seek' in self.options:
+                self._stream_or_buffer.seek(self.options['seek'],0)
+            else:
+                self._stream_or_buffer.seek(0,0)
```
JGSweets, line 29: I think this should reflect the same as the file where the seek just gets set to 0.
e.g. just need `self._stream_or_buffer.seek(0)` the `whence` value is not needed as it has a default.

dataprofiler/data_readers/filepath_or_buffer.py
JGSweets, line 6: instead of `stream_or_buffer`, lets stick closer to convention and use `filepath_or_buffer`

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
+from io import open, StringIO, BytesIO
+
+from . import data_utils
+
+class FileOrBufferHandler():
+    def __init__(self, stream_or_buffer, open_method='r', options=dict()):
+        self._stream_or_buffer = stream_or_buffer
+        self.open_method = open_method
+        self.options = options
```
JGSweets, line 9: if we are going to have `_file` below, we should add it to the init here as `None` to start

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
...
+                return self._stream_or_buffer
+            else:
+                return self._stream_or_buffer
+        else:
+            self._file = open(self._stream_or_buffer, self.open_method)
+            return self._file
+
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        if (data_utils.is_stream_buffer(self._stream_or_buffer)):
```
JGSweets, line 25: on `exit` we can just check `if _file is None`

JGSweets (commented):
We should also add tests to validate this works.

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,31 @@
+from io import open, StringIO, BytesIO
+
+from . import data_utils
+
+class FileOrBufferHandler():
```
JGSweets, line 5: add a description to this as well.

dataprofiler/data_readers/data_utils.py
```diff
@@ -564,3 +564,22 @@ def load_as_str_from_file(file_path, file_encoding, max_lines=10,
                 break
             
     return data_as_str
+
+def is_stream_buffer(filepath_or_buffer, encoding=None, seek=None):
```
AnhTruong, line 568: do we need encoding and seek?

AnhTruong (commented):
what does context manager actually mean in this PR?

dataprofiler/data_readers/data_utils.py
JGSweets: agreed, can remove those from the function input here since not being used

JGSweets:
@AnhTruong to explain it better than I can: https://www.geeksforgeeks.org/context-manager-in-python/

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
+from io import open, StringIO, BytesIO
+
+
```
JGSweets, line 3: no need for two spaces here, just below before the class and import

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
...
+    def __enter__(self):
+        if not data_utils.is_stream_buffer(self._filepath_or_buffer):
+            self._filepath_or_buffer = open(self._filepath_or_buffer, self.open_method)
+
+        if self.seek_offset is not None:
+            self._filepath_or_buffer.seek(self.seek_offset, self.seek_whence)
+            return self._filepath_or_buffer
+        else:
+            return self._filepath_or_buffer
```
JGSweets, line 37: instead of having the same return in both, remove these and just return outside the if

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        if (self._file is None):
```
JGSweets, line 41: can remove the `()` around the conditions

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
...
+        :type open_method: string
+        :param seek_offset: offset from start of the stream
+        :type seek_offset: int
+        :return: TextIOBase or BufferedIOBase class/subclass
+        """
+        self._filepath_or_buffer = filepath_or_buffer
+        self.open_method = open_method
+        self.seek_offset = seek_offset
+        self.seek_whence = seek_whence
+        self._file = None
```
JGSweets, line 27: Currently, the code is not using `_file`. However, I don't think we need it. As below, we can just set the `_filepath_or_buffer` to the input or to the `open` which I think works fine. 

Although `_filepath_or_buffer` no longer is an appropriate name as it won't be a filepath.
maybe rename it to `io_obj` not sure of best name.

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
...
+        else:
+            return self._filepath_or_buffer
+
+
+    def __exit__(self, exc_type, exc_value, exc_traceback):
+        if (self._file is None):
+            self._filepath_or_buffer.seek(0)
+            self._filepath_or_buffer.close()
+        else:
+            self._file.close()
```
JGSweets, line 45: We only want to close a file read from disk and not a StringIO/BytesIO.

Hence, I think we need to check that condition instead.
`if isinstance(filepath_or_buffer, (StringIO, BytesIO)):`

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        self._filepath_or_buffer = filepath_or_buffer
+        self.open_method = open_method
+        self.seek_offset = seek_offset
+        self.seek_whence = seek_whence
+        self._file = None
+
+    def __enter__(self):
+        if not data_utils.is_stream_buffer(self._filepath_or_buffer):
+            self._filepath_or_buffer = open(self._filepath_or_buffer, self.open_method)
```
JGSweets, line 31: I think we may want to make this more robust in the sense, what happens if someone inputs something that is not a filepath or is not a StringIO/BytesIO.

Maybe we can break this into 2 conditionals:
```python
self._file_obj = filepath_or_buffer
if isinstance(filepath_or_buffer, str):
    self._filepath_or_buffer = open(self._filepath_or_buffer, self.open_method)
elif not isinstance(filepath_or_buffer, (StringIO, BytesIO)):
   raise AttributeError(....)
```

Thoughts?

JGSweets: We would need to add a test for the `AttributeError`

dataprofiler/data_readers/data_utils.py
```diff
@@ -564,3 +564,21 @@ def load_as_str_from_file(file_path, file_encoding, max_lines=10,
...
+    :type encoding: str
+    :param seek: position to start in buffer
+    :type seek: int
+    :return: true if string is a buffer or false if string is a filepath
+    :rtype: boolean
+    """
+
+    if isinstance(filepath_or_buffer, (StringIO, BytesIO)):
+        return True
+    return False
```
JGSweets, line 584: I think once we added the context manager we can just incorporate the simple true / false in the class below as described in other comments

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
+from io import open, StringIO, BytesIO
```
JGSweets, line 1: for now, i suggest moving it to utils or maybe we should rename as I'm not sure this should be its own file. I'll wait for consensus from others...
@granteden @ChrisWallace2020 @AnhTruong ?

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,123 @@
...
+
+import pandas as pd
+
+from dataprofiler.data_readers.filepath_or_buffer import FileOrBufferHandler
+
+
+MODULE_PATH = os.path.dirname(os.path.abspath(__file__))
+test_root_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+
+class TestFilepathOrBuffer(unittest.TestCase):
```
JGSweets, line 13: Just remember to limit file character width to < 80

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def test_make_buffer_from_filepath(self):
+        """
+        Make sure FileOrBufferHandler can input a file and read it similarly to open()
+        """
+        for input_file in self.input_file_names:
+            with FileOrBufferHandler(input_file['path'], 'r') as filepath_or_buffer, open(input_file['path'], 'r') as input_file_check:
+                # check every 10th line for the first 100 lines
+                for i in range(0, 10):    
+                    for i in range(0, 9):
```
JGSweets, line 40: i assume the files are small enough that just doing the entire file wouldn't be expensive?

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,123 @@
...
+                        filepath_or_buffer.readline()
+                        input_file_check.readline()
+
+                    self.assertEqual(filepath_or_buffer.readline(), input_file_check.readline())
+                    
+            # check that file was properly closed
+            self.assertEqual(filepath_or_buffer.closed, input_file_check.closed)
+
```
JGSweets, line 121: 2 spaces between the end of the class and the if statement

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,123 @@
...
+        cls.input_file_names = [
+            dict(path=os.path.join(test_dir, 'csv/diamonds.csv'),
+                 count=53940, delimiter=',', has_header=[0],
+                 num_columns=10, encoding='utf-8'),
+            dict(path=os.path.join(test_dir, 'csv/iris.csv'),
+                 count=150, delimiter=',', has_header=[0],
+                 num_columns=6, encoding='utf-8'),  
+            dict(path=os.path.join(test_dir, 'json/iris-utf-8.json'), encoding='utf-8', count=150),
+            dict(path=os.path.join(test_dir, 'txt/discussion_hn.txt'), count=10000),
+            dict(path=os.path.join(test_dir, 'txt/discussion_reddit.txt'), count=10000),
```
JGSweets, line 28: Can remove all the extra info. Probably just need to test at most 2 files.

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,123 @@
+import unittest
+import os
+from io import open, StringIO, BytesIO
+
+import pandas as pd
```
JGSweets, line 5: can be removed

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,45 @@
...
+        self.seek_whence = seek_whence
+        self._encoding = encoding
+
+    def __enter__(self):
+        if isinstance(self._io_obj, str):
+            self._io_obj = open(self._io_obj, self.open_method, encoding=self._encoding)
+
+        elif not isinstance(self._io_obj, (StringIO, BytesIO)):
+            # Raise AttributeError if attribute value not found.
+            raise AttributeError(f'Type {type(self._io_obj)} is invalid.')
```
JGSweets, line 34: We should specifically say what is wrong as in why, e.g. it was not a string or a stream buffer

dataprofiler/data_readers/filepath_or_buffer.py
JGSweets: I'm inclined to merge this and fix in following PR if needed.

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -0,0 +1,49 @@
...
+        :type open_method: string
+        :param seek_offset: offset from start of the stream
+        :type seek_offset: int
+        :return: TextIOBase or BufferedIOBase class/subclass
+        """
+        self._filepath_or_buffer = filepath_or_buffer
+        self.open_method = open_method
+        self.seek_offset = seek_offset
+        self.seek_whence = seek_whence
+        self._encoding = encoding
```
JGSweets, line 28: if this is > 80 char, lets fix this in subsequent to keep the PR moving.

JGSweets (approved):

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,142 @@
...
+                self.assertEqual(filepath_or_buffer.readline(),
+                                    input_file_check.readline())
+
+        # check that file was properly closed
+        self.assertEqual(filepath_or_buffer.closed,
+                            input_file_check.closed)
+
+    def test_make_buffer_error_message(self):
+        """
+        Make sure FileOrBufferHandler can input a file and read it similarly to open()
```
ChrisWallace2020, line 129: This docstring doesn't really describe this test

dataprofiler/tests/data_readers/test_filepath_or_buffer.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        file_name = b"not a valid input"
```
ChrisWallace2020, line 131: What is the b in front of the string for?

JGSweets (approved):

ChrisWallace2020 (approved):

PR #293, opened 2021-06-28
Category counts

az85252:
I have allowed the system to track the counts of each category rather than just a list. Like, instead of just listing the categories like [a,b,c], we can now internally have a way to count each category. The categories will return as before like [a,b,c], but inside, _categories variable is now something like [a:1,b:2,c:3].

### DIFF
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -164,3 +164,10 @@ class CategoricalColumn(BaseColumnProfiler):
         self._update_helper(df_series, profile)
 
         return self
+
+    def counts(self):
+        """
+        Returns the dict of categories with the number of occurrences
+        for each category.
+        """
+        return dict(self._categories.items())
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -48,10 +48,8 @@ class CategoricalColumn(BaseColumnProfiler):
             raise TypeError("Unsupported operand type(s) for +: "
                             "'CategoricalColumn' and '{}'".format(
                                 other.__class__.__name__))
-        print(self._categories)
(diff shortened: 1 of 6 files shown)


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -1,7 +1,8 @@
 from . import BaseColumnProfiler
 from .profiler_options import CategoricalOptions
 from . import utils
-
+from collections import Counter
+from collections import defaultdict
```
ChrisWallace2020, line 5: Can make this 1 import line with `from collections import Counter, defaultdict`, also, this is a standard library, so its import should go above the others

ChrisWallace2020: Also where are you using `Counter`? I don't see it anywhere in the code

requirements.txt
```diff
@@ -2,6 +2,7 @@ h5py>=2.10.0
 wheel>=0.33.1
 future>=0.18.2
 numpy>=1.18.5
+scipy>=1.4.1
```
ChrisWallace2020, line 5: Why is this here?

ChrisWallace2020:
Need to add some tests to ensure categorical data was counted as expected at initialization, update, and merge.

dataprofiler/profilers/categorical_column_profile.py
JGSweets: I believe counter is not being used so we can remove it.

JGSweets: Also, need two spaces between imports and beginning of class.

requirements.txt
JGSweets: remove this req

az85252:
Refer to Ability to internally store counts of each category in a CategoricalColumn in #296

PR #296, opened 2021-06-28
Ability to internally store counts of each category in a CategoricalColumn

az85252:
I have allowed the system to track the counts of each category rather than just a list. Like, instead of just listing the categories like [a,b,c], we can now internally have a way to count each category. The categories will return as before like [a,b,c], but inside, _categories variable is now something like [a:1,b:2,c:3].

### DIFF
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -1,3 +1,4 @@
+from collections import defaultdict
 from . import BaseColumnProfiler
 from .profiler_options import CategoricalOptions
 from . import utils
@@ -29,7 +30,7 @@ class CategoricalColumn(BaseColumnProfiler):
             raise ValueError("CategoricalColumn parameter 'options' must be of"
                              " type CategoricalOptions.")
         super(CategoricalColumn, self).__init__(name)
-        self._categories = list()
+        self._categories = defaultdict(int)
         self.__calculations = {}
         self._filter_properties_w_options(self.__calculations, options)
 
@@ -49,8 +50,8 @@ class CategoricalColumn(BaseColumnProfiler):
                                 other.__class__.__name__))
 
         merged_profile = CategoricalColumn(None)
-        merged_profile._categories = self._categories.copy()
-        merged_profile._update_categories(other._categories)
+        merged_profile._categories = \
+            utils.add_nested_dictionaries(self._categories, other._categories)
         BaseColumnProfiler._add_helper(merged_profile, self, other)
         self._merge_calculations(merged_profile.__calculations,
                                  self.__calculations,
@@ -82,7 +83,7 @@ class CategoricalColumn(BaseColumnProfiler):
         """
         Property for categories.
         """
-        return self._categories
+        return list(self._categories.keys())
 
     @property
     def unique_ratio(self):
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -126,11 +127,8 @@ def _update_categories(self, df_series, prev_dependent_properties=None,
         :type df_series: pandas.DataFrame
         :return: None
         """
-        if hasattr(df_series, 'tolist'):
-            df_series = df_series.tolist()
 
-        self._categories = utils._combine_unique_sets(
-            self._categories, df_series)
+        self._categories.update(df_series.value_counts(dropna = False).to_dict())
```
JGSweets, line 131: no spaces around the = here `dropna=False`

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -126,11 +127,8 @@ def _update_categories(self, df_series, prev_dependent_properties=None,
         :type df_series: pandas.DataFrame
         :return: None
         """
-        if hasattr(df_series, 'tolist'):
-            df_series = df_series.tolist()
 
-        self._categories = utils._combine_unique_sets(
-            self._categories, df_series)
+        self._categories.update(df_series.value_counts(dropna=False).to_dict())
```
JGSweets, line 131: This like will replace the known values as opposed to updating counts:
I suggest utilizing the same functionality as in `__add__`, e.g.:
```python
category_count = df_series.value_counts(dropna=False).to_dict()
self._categories = utils.add_nested_dictionaries(self._categories, category_count)
```

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -167,3 +165,10 @@ def update(self, df_series):
         self._update_helper(df_series, profile)
 
         return self
+
+    def dict_of_categories(self):
+        """
+        Returns the dict of categories with the number of occurrences
+        for each category.
+        """
+        return dict(self._categories.items())
```
JGSweets, line 174: Instead of dict_of_categories, we could do a functional property `category_counts` and return a copy of the dict.
However, I suggest adding that to the subsequent PR which adds it to the profile.

JGSweets: For now we can just remove this function.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -335,3 +335,29 @@ def test_categorical_column_with_wrong_options(self):
...
+            "a", "a", "d", "d"
+        ])
+        profile2 = CategoricalColumn(df_categorical2.name)
+        profile2.update(df_categorical2)
+        profile3 = profile + profile2
+        # Check that _categories dict_of_categories is correct after merge/add function
+        self.assertEqual(profile3.dict_of_categories()['a'], 5)
+        self.assertEqual(profile3.dict_of_categories()['b'], 4)
+        self.assertEqual(profile3.dict_of_categories()['c'], 5)
+        self.assertTrue(profile3.dict_of_categories()['d'], 2)
```
JGSweets, line 363: If we end up updating the other tests, this test may become redundant.

gme5078 (approved):

JGSweets (approved):

PR #297, opened 2021-06-28
Add diff to DataLabeler column

andrew-yin:
diff() implemented for DataLabeler column to compare the predicted labels.


dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -301,3 +301,43 @@ def test_profile_merge_with_different_options(self, mock_instance):
...
+                [1, 0, 0, 0],
+                [1, 0, 0, 0],
+                [1, 0, 0, 0],
+                [0, 1, 0, 0],  # 2 repeated
+                [0, 1, 0, 0],
+                [0, 0, 1, 0],  # 3 repeated
+                [0, 0, 1, 0],
+                [0, 0, 1, 0],
+                [0, 0, 0, 1],  # 1 repeated
+            ])}  # counts [4, 2, 3, 1] => [a, b, c, d]
```
AnhTruong, line 331: should be [b, c, d, e] ?

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -301,3 +301,43 @@ def test_profile_merge_with_different_options(self, mock_instance):
...
+        mock_instance.return_value.predict.side_effect = mock_low_predict
+
+        data2 = pd.Series(['1'] * 10)
+        profiler2 = DataLabelerColumn(data2.name)
+        profiler2.update(data2)
+
+        diff = profiler.diff(profiler2)
+
+        expected_diff = {
+            "data_label": [["a"], ["b"], ["d", "c"]]
```
AnhTruong, line 341: can you add a comment on how to get this?

AnhTruong: do we need to update label changes, number of labels, entity counts, or model parameters?

AnhTruong (commented):
some comments

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
JGSweets: I feel like I would expect the order to be the opposite of this, e.g. `[profiler has but profiler2 doesn't , shared, profiler2 has but profiler doesn't]` Thoughts? 
@AnhTruong profiler has [b,c,d,e] as labels whereas profiler2 has [a,b] as labels.

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
JGSweets: agreed

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
gme5078: Jeremy  that order is the same as current state, I think?

gme5078 (approved):

gme5078 (dismissed):

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -301,3 +301,45 @@ def test_profile_merge_with_different_options(self, mock_instance):
...
+                [1, 0, 0, 0],
+                [1, 0, 0, 0],
+                [0, 1, 0, 0],  # 2 repeated
+                [0, 1, 0, 0],
+                [0, 0, 1, 0],  # 3 repeated
+                [0, 0, 1, 0],
+                [0, 0, 1, 0],
+                [0, 0, 0, 1],  # 1 repeated
+            ])}  # counts [4, 2, 3, 1] => [b, c, d, e]
+        mock_instance.return_value.predict.side_effect = mock_low_predict
```
ChrisWallace2020, line 332: Instead of mocking out DataLabeler to produce the desired `profiler.data_label`, could you instead just mock out `profiler.data_label` to be the string you want ("a|b|c")? Because the diff only needs `data_label`, it doesn't need to know where or how it was calculated.

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -238,6 +239,28 @@ def profile(self):
         }
         return profile
 
+    def diff(self, other_profile, options=None):
+        """
+        Generates the differences between the orders of two OrderColumns
```
JGSweets, line 244: `OrderColumns`

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -238,6 +239,28 @@ def profile(self):
...
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+        
+        labels = self.data_label.split('|')
+        other_labels = other_profile.data_label.split('|')
+
+        differences = {
+            "data_label": utils.find_diff_of_lists_and_sets(labels, other_labels)
+        }
+        return differences
```
JGSweets, line 262: should include the `avg_predictions` and `data_label_representation` diffs

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -238,6 +239,44 @@ def profile(self):
...
+        other_label_rep = other_profile.label_representation
+
+        avg_predictions_diff = dict()
+        label_rep_diff = dict()
+        for key, value in avg_preds.items():
+            avg_predictions_diff[key] = utils.find_diff_of_numbers(
+                value, other_avg_preds[key])
+        for key, value in label_rep.items():
+            label_rep_diff[key] = utils.find_diff_of_numbers(
+                value, other_label_rep[key])
```
gme5078, line 271: I don't think we can assume that both will have the same keys can we? People can use the their own data labeler to profile with

gme5078: I think you should iterate through self._possible_data_labels

JGSweets: That's a fair point, the two dicts could have varying keys. example:
e.g. 
{'a': 0.4, 'b': 0.2, 'c': 0.4}
{'b': 0.3, 'c': 0.4, 'd': 0.3} 
becomes
{'a': [0.4, None], 'b': 0.1, 'c': 'unchanged', 'd': [None, '0.3]} 
Thoughts?
Can loop through all keys as above and just use None when key is missing.

JGSweets: e.g. `self.other_avg_preds.get(key, None)`

JGSweets (approved):

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -301,3 +302,60 @@ def test_profile_merge_with_different_options(self, mock_instance):
...
+                    "b": -0.70,
+                    "c": 0.70
+                },
+                "label_representation": {
+                    "a": -0.84,
+                    "b": "unchanged",
+                    "c": 0.84
+                }
+            }
+            self.maxDiff = None
```
JGSweets, line 360: remove?

dataprofiler/profilers/utils.py
```diff
@@ -445,3 +445,37 @@ def find_diff_of_dates(stat1, stat2):
         return "+" + str(diff)
 
     return "-" + str(abs(diff))
+
+def find_diff_of_dicts(dict1, dict2):
```
JGSweets, line 449: we can start with this and refactor later if we have different dicts

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -238,6 +239,34 @@ def profile(self):
...
+
+        :return: Dict containing the differences between orders in their
+        appropriate output formats
+        :rtype: dict
+        """
+        cls = self.__class__
+        if not isinstance(other_profile, cls):
+            raise TypeError("Unsupported operand type(s) for diff: '{}' "
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
```
gme5078, line 254: Can you call the base class for the error check? You can use the returned empty dict for the differences dict

dataprofiler/profilers/utils.py
```diff
@@ -445,3 +445,37 @@ def find_diff_of_dates(stat1, stat2):
...
+    for key, value1 in dict1.items():
+        value2 = dict2.get(key, None)
+        if isinstance(value1, list):
+            diff[key] = find_diff_of_lists_and_sets(value1, value2)
+        elif isinstance(value1, datetime.datetime):
+            diff[key] = find_diff_of_dates(value1, value2)
+        elif isinstance(value1, str):
+            diff[key] = find_diff_of_strings(value1, value2)
+        else:
+            diff[key] = find_diff_of_numbers(value1, value2)
```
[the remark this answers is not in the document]
JGSweets: TBF, I was originally avoiding this intuitive nature due to potentially unintended consequences of future diffs. Intentionally being more explicit now and refactoring down the line as it grows. Unless we think all ints, strings, etc should diff this way.

JGSweets (approved):

gme5078 (approved):

PR #299, opened 2021-06-29
Categorical counts to be in the profile of CategoricalColumn

az85252:
Added feature that stores top n counted categories into profile statistics if is match is true

### DIFF
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -1,4 +1,6 @@
 from collections import defaultdict
+from operator import itemgetter
+
 from . import BaseColumnProfiler
 from .profiler_options import CategoricalOptions
 from . import utils
@@ -63,6 +65,7 @@ class CategoricalColumn(BaseColumnProfiler):
         """
         Property for profile. Returns the profile of the column.
         """
+        num_top_categories = 5
 
         profile = dict(
             categorical=self.is_match,
@@ -76,6 +79,9 @@ class CategoricalColumn(BaseColumnProfiler):
             profile["statistics"].update(
                 dict(categories=self.categories)
             )
+            profile["statistics"]['categorical_count'] = dict(
+                sorted(self._categories.items(), key=itemgetter(1),
+                       reverse=True)[:num_top_categories])
         return profile
 
     @property
@@ -166,3 +172,10 @@ class CategoricalColumn(BaseColumnProfiler):
         self._update_helper(df_series, profile)
 
         return self
+
+    def counts(self):
+        """
+        Returns the dict of categories with the number of occurrences
+        for each category.
+        """
+        return dict(self._categories.items())
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -64,8 +64,10 @@ class CategoricalColumn(BaseColumnProfiler):
     def profile(self):
         """
         Property for profile. Returns the profile of the column.
+        For categorical_count, it will display the top k categories most
+        frequently occurred in descending order.
         """
-        num_top_categories = 5
+        top_k_categories = 5
 
         profile = dict(
             categorical=self.is_match,
@@ -81,7 +83,7 @@ class CategoricalColumn(BaseColumnProfiler):
             )
             profile["statistics"]['categorical_count'] = dict(
                 sorted(self._categories.items(), key=itemgetter(1),
-                       reverse=True)[:num_top_categories])
+                       reverse=True)[:top_k_categories])
         return profile
 
     @property
@@ -172,10 +174,3 @@ class CategoricalColumn(BaseColumnProfiler):
         self._update_helper(df_series, profile)
 
         return self
-
-    def counts(self):
-        """
-        Returns the dict of categories with the number of occurrences
-        for each category.
-        """
-        return dict(self._categories.items())
--- a/dataprofiler/tests/profilers/test_categorical_column_profile.py
+++ b/dataprofiler/tests/profilers/test_categorical_column_profile.py
@@ -155,8 +155,7 @@ class TestCategoricalColumn(unittest.TestCase):
         profile = CategoricalColumn(df_categorical.name)
         profile.update(df_categorical)
         report = profile.profile
-        self.assertIsNotNone(report["times"])
-        report.pop("times")
+        self.assertIsNotNone(report.pop("times", None))
         expected_profile = dict(
             categorical=True,
             statistics=dict([
@@ -180,7 +179,6 @@ class TestCategoricalColumn(unittest.TestCase):
 
         report = profile.profile
         self.assertIsNotNone(report["times"])
-        report.pop("times")
         expected_profile = dict(
             categorical=False,
             statistics=dict([
@@ -225,7 +223,6 @@ class TestCategoricalColumn(unittest.TestCase):
 
         report = profile3.profile
         self.assertIsNotNone(report["times"])
-        report.pop("times")
         expected_profile = dict(
             categorical=False,
             statistics=dict([
@@ -246,7 +243,6 @@ class TestCategoricalColumn(unittest.TestCase):
 
         report = profile3.profile
         self.assertIsNotNone(report["times"])
-        report.pop("times")
         expected_profile = dict(
             categorical=False,
             statistics=dict([
@@ -263,7 +259,6 @@ class TestCategoricalColumn(unittest.TestCase):
 
         report = profile3.profile
         self.assertIsNotNone(report["times"])
-        report.pop("times")
         report_categories = report['statistics'].pop('categories')
         report_count = report['statistics'].pop('categorical_count')
         expected_profile = dict(
--- a/dataprofiler/tests/profilers/test_categorical_column_profile.py
+++ b/dataprofiler/tests/profilers/test_categorical_column_profile.py
@@ -155,21 +155,23 @@ class TestCategoricalColumn(unittest.TestCase):
         profile = CategoricalColumn(df_categorical.name)
         profile.update(df_categorical)
         report = profile.profile
-        six.assertCountEqual(
-            self, ['categorical', 'statistics', 'times'], report)
-        self.assertTrue(report["categorical"])
-        six.assertCountEqual(
-            self,
-            ['unique_count', 'unique_ratio', 'categories', 'categorical_count'], report['statistics']
+        self.assertIsNotNone(report["times"])
+        report.pop("times")
+        expected_profile = dict(
+            categorical=True,
+            statistics=dict([
+                ('unique_count', 3),
+                ('unique_ratio', .25),
+                ('categories', ["a", "b", "c"]),
+                ('categorical_count', {"a": 3, "b": 4, "c": 5})
+            ]),
         )
-        self.assertEqual(3, report["statistics"]["unique_count"])
-        self.assertEqual(0.25, report["statistics"]["unique_ratio"])
-        self.assertCountEqual(
-            ["a", "b", "c"], report["statistics"]["categories"]
-        )
-        expected_dict = {"a": 3, "b": 4, "c": 5}
-        self.assertEqual(expected_dict,
-                             report["statistics"]["categorical_count"])
+        # We have to pop these values because sometimes the order changes
+        self.assertCountEqual(expected_profile['statistics'].pop('categories'),
+                              report["statistics"].pop('categories'))
+        self.assertCountEqual(expected_profile['statistics'].pop(
+            'categorical_count'), report["statistics"].pop('categorical_count'))
+        self.assertEqual(report, expected_profile)
 
     def test_false_categorical_report(self):
         df_non_categorical = pd.Series(list(map(str, range(0, 20))))
@@ -177,17 +179,8 @@ class TestCategoricalColumn(unittest.TestCase):
         profile.update(df_non_categorical)
 
         report = profile.profile
+        self.assertIsNotNone(report["times"])
         report.pop("times")
-        six.assertCountEqual(
-            self, ['categorical', 'statistics'], report)
-        self.assertFalse(report["categorical"])
-        six.assertCountEqual(
-            self, ['unique_count', 'unique_ratio'], report['statistics']
-        )
-        self.assertEqual(20, report["statistics"]["unique_count"])
-        self.assertEqual(1.0, report["statistics"]["unique_ratio"])
-        self.assertNotIn("categorical_count", report["statistics"])
-
         expected_profile = dict(
             categorical=False,
             statistics=dict([
@@ -231,6 +224,7 @@ class TestCategoricalColumn(unittest.TestCase):
         self.assertDictEqual(expected_dict, profile3._categories)
 
         report = profile3.profile
+        self.assertIsNotNone(report["times"])
         report.pop("times")
         expected_profile = dict(
             categorical=False,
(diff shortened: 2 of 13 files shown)


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -166,3 +172,10 @@ def update(self, df_series):
         self._update_helper(df_series, profile)
 
         return self
+
+    def counts(self):
+        """
+        Returns the dict of categories with the number of occurrences
+        for each category.
+        """
+        return dict(self._categories.items())
```
JGSweets, line 181: is this needed? if we wanted this, i would suggest using the code in the profile
```python
 dict(
    sorted(self._categories.items(), key=itemgetter(1),
                reverse=True)[:num_top_categories])
```
but have an input arg for the num_top categories and then the profile could call:
`self.category_counts(top_k_categories=5)`

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,6 +65,7 @@ def profile(self):
         """
         Property for profile. Returns the profile of the column.
         """
+        num_top_categories = 5
```
JGSweets, line 68: maybe lets call it `top_k_categories` or does this not make sense?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -76,6 +79,9 @@ def profile(self):
             profile["statistics"].update(
                 dict(categories=self.categories)
             )
+            profile["statistics"]['categorical_count'] = dict(
+                sorted(self._categories.items(), key=itemgetter(1),
+                       reverse=True)[:num_top_categories])
```
JGSweets, line 84: just to confirm this is descending? e.g. largest number is on top? put a comment mentioning that maybe?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,13 +160,17 @@ def test_true_categorical_report(self):
...
+            ['unique_count', 'unique_ratio', 'categories', 'categorical_count'], report['statistics']
         )
         self.assertEqual(3, report["statistics"]["unique_count"])
         self.assertEqual(0.25, report["statistics"]["unique_ratio"])
         self.assertCountEqual(
             ["a", "b", "c"], report["statistics"]["categories"]
         )
+        expected_dict = {"a": 3, "b": 4, "c": 5}
+        self.assertDictEqual(expected_dict, report["statistics"]
+        ["categorical_count"])
```
JGSweets, line 172: instead of moving just the key to a new line:
```python
self.assertDictEqual(expected_dict, 
                 report["statistics"]["categorical_count"])
```

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,13 +160,17 @@ def test_true_categorical_report(self):
...
+
```
JGSweets, line 173: below in `test_false_categorical_report` we should probably change the test to validate the exact dict returned by the profile. In the current situation, it is not validating that `categorical_counts` is not in the dict. While we can do that, that would not future proof it like this this scenario. Instead if we specify the exact dict, it will prevent a potential future issue.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -348,3 +352,29 @@ def test_categorical_column_with_wrong_options(self):
                                    "CategoricalColumn parameter 'options' must"
                                    " be of type CategoricalOptions."):
             profiler = CategoricalColumn("Categorical", options="wrong_data_type")
+
+    def test_categorical_column_is_counted(self):
```
JGSweets, line 356: If we make this function be able to get the top k counts, then we can have this test validate that changing the k value works, however, these should be validated using the existing test, no?

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets: below that, test_categorical_merge also needs updating like `test_false_categorical_report` such that it tests `categorical_counts` as well

JGSweets (commented):
Looks like a test from `test_profile_builder.py` will need to be update:
function: `recursive_test_helper` within `test_profile_key_name_without_space` ~ line 362
```python
if prev_key not in ['data_stats', 'avg_predictions',
                                    'data_label_representation',
                                    'null_types_index', 'categorical_count']:
```

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,28 +160,43 @@ def test_true_categorical_report(self):
...
-            ['unique_count', 'unique_ratio', 'categories'], report['statistics']
+            ['unique_count', 'unique_ratio', 'categories', 'categorical_count'], report['statistics']
         )
         self.assertEqual(3, report["statistics"]["unique_count"])
         self.assertEqual(0.25, report["statistics"]["unique_ratio"])
         self.assertCountEqual(
             ["a", "b", "c"], report["statistics"]["categories"]
         )
+        expected_dict = {"a": 3, "b": 4, "c": 5}
+        self.assertTrue(report["statistics"]["categories"])
```
AnhTruong, line 171: i think we can remove this as this is subsumed by the one in line 167-169

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,28 +160,43 @@ def test_true_categorical_report(self):
...
+        self.assertFalse("categorical_count" in report["statistics"])
+
+        expected_profile = dict(
+            categorical=False,
+            statistics=dict([
+                ('unique_count', 20),
+                ('unique_ratio', 1),
+            ]),
+        )
+        self.assertEqual(report, expected_profile)
```
AnhTruong, line 199: did we have the tests for each of key in this dict already above?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -225,11 +251,80 @@ def test_categorical_merge(self):
...
+       # Check that _categories counts is correct after update function
+       self.assertEqual(counts['a'], 3)
+       self.assertEqual(counts['b'], 4)
+       self.assertEqual(counts['c'], 5)
+
+       df_categorical2 = pd.Series([
+           "a", "a", "d", "d"
+       ])
+       profile2 = CategoricalColumn(df_categorical2.name)
+       profile2.update(df_categorical2)
```
AnhTruong, line 317: can we add test for categorical count for profile2?

AnhTruong (commented):
some comments

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -76,6 +81,9 @@ def profile(self):
             profile["statistics"].update(
                 dict(categories=self.categories)
             )
+            profile["statistics"]['categorical_count'] = dict(
```
ChrisWallace2020, line 84: Should the name "categorical_count" be elaborated on to specify that not all categories will be present in this statistic, only the top 5. e.g. "top_categorical_counts" or "top_categories" or something of the sort?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,28 +160,43 @@ def test_true_categorical_report(self):
...
     def test_false_categorical_report(self):
         df_non_categorical = pd.Series(list(map(str, range(0, 20))))
         profile = CategoricalColumn(df_non_categorical.name)
         profile.update(df_non_categorical)
 
         report = profile.profile
+        report.pop("times")
         six.assertCountEqual(
-            self, ['categorical', 'statistics', 'times'], report)
+            self, ['categorical', 'statistics'], report)
```
ChrisWallace2020, line 183: Why did you pop "times" out instead of just adding to list of keys in the `assertCountEqual`. Also, what is `six`? Why isn't is `self`?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,28 +160,43 @@ def test_true_categorical_report(self):
...
         self.assertFalse(report["categorical"])
         six.assertCountEqual(
             self, ['unique_count', 'unique_ratio'], report['statistics']
         )
         self.assertEqual(20, report["statistics"]["unique_count"])
         self.assertEqual(1.0, report["statistics"]["unique_ratio"])
+        self.assertFalse("categorical_count" in report["statistics"])
```
ChrisWallace2020, line 190: This should be `self.assertNotIn`

dataprofiler/tests/profilers/test_categorical_column_profile.py
az85252: Jeremy told me to take times off for this specific test as the times can fluctuate and does not hold meaning in this specific test.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -216,6 +231,17 @@ def test_categorical_merge(self):
                          "NaN": 1, "None": 1, "nan": 1, "null": 1}
         self.assertDictEqual(expected_dict, profile3._categories)
 
+        report = profile3.profile
+        report.pop("times")
+        expected_profile = dict(
+            categorical=False,
```
ChrisWallace2020, line 237: I'm confused, a lot of these test_categorical functions have categorical as false in the expected dict. Is that intended? It seems like either is_match isn't accurate or these test cases aren't representative.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -225,11 +251,80 @@ def test_categorical_merge(self):
...
+        self.assertCountEqual(report_categories, ['abcd', 'aa', '2', np.nan,
+                                                  '4', 'b', '3', 'dfd', 'ee',
+                                                  'ff', 'nan', 'None', '1',
+                                                  'gg', 'null', 'NaN'])
+        expected_dict = dict({'aa': 5, '2': 4, 'abcd': 4, 'b': 3, np.nan: 2})
+        self.assertEqual(len(report_count), len(expected_dict))
+        self.assertEqual(report_count['aa'], expected_dict['aa'])
+        self.assertEqual(report_count['2'], expected_dict['2'])
+        self.assertEqual(report_count['abcd'], expected_dict['abcd'])
+        self.assertEqual(report_count['b'], expected_dict['b'])
```
ChrisWallace2020, line 291: You could just do `assertDictEqual` instead here.

dataprofiler/tests/profilers/test_categorical_column_profile.py
az85252: Yes, but I was told that we also need to test the exact profile object.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -160,28 +160,43 @@ def test_true_categorical_report(self):
...
+        self.assertEqual(expected_dict,
+                             report["statistics"]["categorical_count"])
 
     def test_false_categorical_report(self):
         df_non_categorical = pd.Series(list(map(str, range(0, 20))))
         profile = CategoricalColumn(df_non_categorical.name)
         profile.update(df_non_categorical)
 
         report = profile.profile
+        report.pop("times")
```
JGSweets, line 181: lets still add a check to make sure not None, e.g. `self.assertIsNotNone(report.pop("times, None))`

JGSweets: will validate something at least exists and that the key exists

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets, line 199: Since we have this profile we can remove the following asserts since they become redundant:
```python
six.assertCountEqual(self, ['categorical', 'statistics'], report)
self.assertFalse(report["categorical"])
six.assertCountEqual(self, ['unique_count', 'unique_ratio'], report['statistics'])
self.assertEqual(20, report["statistics"]["unique_count"])
self.assertEqual(1.0, report["statistics"]["unique_ratio"])
```

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -225,11 +250,85 @@ def test_categorical_merge(self):
...
+    def test_categorical_column_is_counted(self):
+       df_categorical = pd.Series([
+           "a", "a", "a", "b", "b", "b", "b", "c", "c", "c", "c", "c",
+       ])
+       profile = CategoricalColumn(df_categorical.name)
+
+       # Check that _categories counts in profile is empty after init.
+       report = profile.profile
+       counts = report["statistics"]["categorical_count"]
+       self.assertEqual(0, len(counts))
```
JGSweets, line 301: `self.assertDictEqual({}, counts)`

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -225,11 +250,85 @@ def test_categorical_merge(self):
...
+                                                  'ff', 'nan', 'None', '1',
+                                                  'gg', 'null', 'NaN'])
+        expected_dict = dict({'aa': 5, '2': 4, 'abcd': 4, 'b': 3, np.nan: 2})
+        self.assertEqual(len(report_count), len(expected_dict))
+        self.assertEqual(report_count['aa'], expected_dict['aa'])
+        self.assertEqual(report_count['2'], expected_dict['2'])
+        self.assertEqual(report_count['abcd'], expected_dict['abcd'])
+        self.assertEqual(report_count['b'], expected_dict['b'])
+
+    def test_categorical_column_is_counted(self):
```
JGSweets, line 292: I think this test is redundant and can be removed

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets: I think the intent is that we are merging and to prove that even if a categorical profiler doesn't start as categorical, it can update to one as it gets more data and the report back is updating as it should based on whether it is categorical or not.

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets: test being removed.

dataprofiler/profilers/categorical_column_profile.py
JGSweets: interesting thought, maybe refactor after this PR?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -155,39 +155,32 @@ def test_true_categorical_report(self):
...
-        six.assertCountEqual(
-            self,
-            ['unique_count', 'unique_ratio', 'categories', 'categorical_count'], report['statistics']
-        )
-        self.assertEqual(3, report["statistics"]["unique_count"])
-        self.assertEqual(0.25, report["statistics"]["unique_ratio"])
-        self.assertCountEqual(
-            ["a", "b", "c"], report["statistics"]["categories"]
+        self.assertIsNotNone(report["times"])
+        report.pop("times")
```
JGSweets, line 159: this could be a bit dangerous as while the test would fail, it would fail due to a code exception and not an assert.
This would happen in the case where times was not a key in the dict.

Hence, why I suggest reducing the two lines to: `self.assertIsNotNone(report.pop("times", None))`

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets: same as above

JGSweets (approved):

ChrisWallace2020 (approved):

PR #300, opened 2021-06-29
Add diff for FloatColumn

andrew-yin:
FloatColumn inherits diff() from NumericalStatsMixin, so all there is to do is test to ensure this inheritance behaves properly.

### DIFF
--- a/dataprofiler/profilers/float_column_profile.py
+++ b/dataprofiler/profilers/float_column_profile.py
@@ -115,7 +115,7 @@ class FloatColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :return: the FloatColumn differences
         :rtype: dict
         """
-        differences = super().diff(other_profile, options=None)
+        differences = NumericStatsMixin.diff(self, other_profile, options=None)
         other_precision = other_profile.profile['precision']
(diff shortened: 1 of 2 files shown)


JGSweets (commented):
We should also add the diff for the precision dict. No need to include confidence interval

dataprofiler/profilers/float_column_profile.py
```diff
@@ -105,6 +106,25 @@ def __add__(self, other):
...
+    def diff(self, other_profile, options=None):
+        """
+        Finds the differences for FloatColumnss.
+
+        :param other_profile: profile to find the difference with
+        :type other_profile: FloatColumn
+        :return: the FloatColumn differences
+        :rtype: dict
+        """
+        differences = super().diff(other_profile, options=None)
```
gme5078, line 118: I'm adding a general diff in the base column class so I think you will need to be specific that this is the Numerical Stats super.

gme5078: otherwise looks good to me

gme5078 (dismissed):

gme5078 (approved):

AnhTruong (approved):

PR #301, opened 2021-06-29
Add diff to TextColumn

andrew-yin:
TextColumn inherits most diff properties from NumericalStatsMixin, so all that's needed to do is add vocab to it.

### DIFF
--- a/dataprofiler/profilers/text_column_profile.py
+++ b/dataprofiler/profilers/text_column_profile.py
@@ -90,9 +90,9 @@ class TextColumn(NumericStatsMixin, BaseColumnPrimitiveTypeProfiler):
         :return: the text columns differences
         :rtype: dict
         """
+        differences = NumericStatsMixin.diff(self, other_profile, options)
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/text_column_profile.py
```diff
@@ -81,6 +81,21 @@ def profile(self):
...
+        Finds the differences for text columns
+
+        :param other_profile: profile to find the difference with
+        :type other_profile: TextColumn Profile
+        :return: the text columns differences
+        :rtype: dict
+        """
+        vocab_diff = utils.find_diff_of_lists_and_sets(
+            self.vocab, other_profile.vocab)
+        differences = super().diff(other_profile, options)
```
gme5078, line 95: Call the super first because it will do the instance check first

gme5078 (dismissed):

gme5078 (approved):

gme5078 (approved):

AnhTruong (approved):

PR #302, opened 2021-06-29
Adding Primitive Compiler Diff 

gme5078:
Okay so I had to make a lot of design choices here and I am open to alternate ideas if you have any!

One thing to note, there will be another PR later to allow Float and Int stats to be interchangeable, but that seemed outside the scope of this PR. I can add if you think otherwise tho! :)

### DIFF
--- a/dataprofiler/profilers/base_column_profilers.py
+++ b/dataprofiler/profilers/base_column_profilers.py
@@ -186,6 +186,22 @@ class BaseColumnProfiler(with_metaclass(abc.ABCMeta, object)):
                                                    other2.times)
 
         self.sample_size = other1.sample_size + other2.sample_size
+        
+    def diff(self, other_profile, options=None):
+        """
+        Finds the differences for several numerical stats.
+
+        :param other_profile: profile to find the difference with
+        :type other_profile: NumericStatsMixin Profile
+        :return: the numerical stats differences
+        :rtype: dict
+        """
+        cls = self.__class__
+        if not isinstance(other_profile, cls):
+            raise TypeError("Unsupported operand type(s) for diff: '{}' "
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+        return {}
 
     def _update_column_base_properties(self, profile):
         """
--- a/dataprofiler/profilers/base_column_profilers.py
+++ b/dataprofiler/profilers/base_column_profilers.py
@@ -189,11 +189,11 @@ class BaseColumnProfiler(with_metaclass(abc.ABCMeta, object)):
         
     def diff(self, other_profile, options=None):
         """
-        Finds the differences for several numerical stats.
+        Finds the differences for columns.
 
         :param other_profile: profile to find the difference with
-        :type other_profile: NumericStatsMixin Profile
-        :return: the numerical stats differences
+        :type other_profile: BaseColumnProfiler
+        :return: the stat differences
         :rtype: dict
         """
         cls = self.__class__
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -259,11 +259,10 @@ class ColumnPrimitiveTypeProfileCompiler(BaseCompiler):
         if data_type1 is not None or data_type2 is not None:
             diff_profile["data_type"] = utils.find_diff_of_strings(data_type1,
                                                                    data_type2)
-        
-        # Find diff of matching profile statistics
-        if "data_type" in diff_profile and diff_profile["data_type"] == "unchanged":
-            diff_profile["statistics"] = self._profiles[data_type1]\
-                .diff(other._profiles[data_type2], options)
+            # Find diff of matching profile statistics
+            if diff_profile["data_type"] == "unchanged":
+                diff_profile["statistics"] = self._profiles[data_type1]\
+                    .diff(other._profiles[data_type2], options)
 
         # If there is no data, pop the data
         if not diff_profile["data_type_representation"]:
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -209,6 +209,19 @@ class ColumnPrimitiveTypeProfileCompiler(BaseCompiler):
             )
         return profile
 
+    def find_data_type(self):
+        """
+        Finds the selected data_type in a primitive compiler
+        
+        :return: name of the selected data type
+        :rtype: str
+        """
+        matched_profile = None
+        for key, profiler in self._profiles.items():
+            if matched_profile is None and profiler.data_type_ratio == 1.0:
+                matched_profile = key
+        return matched_profile
+
     def diff(self, other, options=None):
         """
         Finds the difference between 2 compilers and returns the report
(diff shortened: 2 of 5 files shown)


dataprofiler/profilers/base_column_profilers.py
```diff
@@ -186,6 +186,22 @@ def _add_helper(self, other1, other2):
...
+        :type other_profile: BaseColumnProfiler
+        :return: the stat differences
+        :rtype: dict
+        """
+        cls = self.__class__
+        if not isinstance(other_profile, cls):
+            raise TypeError("Unsupported operand type(s) for diff: '{}' "
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+        return {}
```
JGSweets, line 204: technically, all these have `sample_size`, is sample_size part of the diff? if so we can do that here for all

JGSweets: I assume the intent of this for the error check, otherwise i'd suggest raise not implemented such that everyone does it

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -101,6 +101,21 @@ def __add__(self, other):
...
+        :param other: profile compiler finding the difference with this one.
+        :type other: BaseCompiler
+        :return: difference of the profiles
+        :rtype: dict
+        """
+        if type(other) is not type(self):
+            raise TypeError('`{}` and `{}` are not of the same profile compiler'
+                            ' type.'.format(type(self).__name__,
+                                            type(other).__name__))
+        return {}
```
JGSweets, line 118: I find it strange to return an empty dict, but we don't use it.

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -194,6 +209,62 @@ def profile(self):
...
+        """
+        Finds the difference between 2 compilers and returns the report
+        
+        :param other: profile compiler finding the difference with this one.
+        :type other: ColumnPrimitiveTypeProfileCompiler
+        :return: difference of the profiles
+        :rtype: dict
+        """
+        # Call super for compiler instance check
+        super().diff(other)
```
JGSweets, line 222: I find it strange to call it but not utilize it's return even if it is an empty dict

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -194,6 +209,62 @@ def profile(self):
...
+
+        # Initialize profile diff dict
+        profile = {
```
JGSweets, line 225: thoughts on changing the name to `diff_profile`?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -194,6 +209,62 @@ def profile(self):
...
+            "data_type_representation": dict(),
+        }
+
+        # Iterate through self for primitive profiles
+        matched_profile1 = None
+        for key, profiler in self._profiles.items():
+            # Find data_type
+            if matched_profile1 is None and profiler.data_type_ratio == 1.0:
+                matched_profile1 = key
```
JGSweets, line 234: Do you think we should have something in the base which gets this auto? I imagine `def profile` would use it too

JGSweets: e.g. `self.profile_type` or something?

dataprofiler/profilers/base_column_profilers.py
gme5078: honestly, I'm not sure where sample_size is propagated from in the profile yet. Intent is two-fold, one is to reduce duplicate code with the error check, and the other is to prevent the compilers from failing if a column doesnt have diff implemented yet

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -101,6 +101,21 @@ def __add__(self, other):
...
+    def diff(self, other, options=None):
+        """
+        Finds the difference between 2 compilers and returns the report
+
+        :param other: profile compiler finding the difference with this one.
+        :type other: BaseCompiler
+        :return: difference of the profiles
+        :rtype: dict
+        """
+        if type(other) is not type(self):
```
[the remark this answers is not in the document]
gme5078: This is the error check we do when we __add__ compilers, but I don't know actually know why. 

After googling, seems like isinstance allows for subclasses to be considered as the same where type() ensures it is the specific class the same

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,6 +71,132 @@ def test_add_profilers(self):
...
+    def test_diff_primitive_compilers(self):
+        # Test different data types
+        data1 = pd.Series(['-2', '-1', '1', '2'])
+        data2 = pd.Series(["YO YO YO", "HELLO"])
+        compiler1 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data1)
+        compiler2 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data2)
+
+        expected_diff = {
+            'data_type_representation': {
+                'datetime': 'unchanged',
```
AnhTruong, line 84: can we add a test for the datetime change as well?

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,6 +71,132 @@ def test_add_profilers(self):
...
+        options.int.is_enabled = False
+        compiler1 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data1,
+                                                                         options)
+        compiler2 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data2)
+        expected_diff = {
+            'data_type_representation': {
+                'datetime': 'unchanged', 
+                'float': 'unchanged', 
+                'string': 'unchanged', 
+                'int': [None, 1.0]},
```
AnhTruong, line 149: should we keep other results consistent with this format? 0 for not detected, 1 for detected, and None for non-existing

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,6 +71,132 @@ def test_add_profilers(self):
...
+        self.assertDictEqual(expected_diff, compiler1.diff(compiler2))
+
+        # Test disabled column in both compilers
+        compiler2 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data2,
+                                                                         options)
+        expected_diff = {
+            'data_type_representation': {
+                'datetime': 'unchanged', 
+                'float': 'unchanged', 
+                'string': 'unchanged'
```
AnhTruong, line 161: as above, `int` should be [None, None]

AnhTruong (commented):
some first comments about the tests

dataprofiler/profilers/column_profile_compilers.py
gme5078: Good point, I use it now

dataprofiler/profilers/column_profile_compilers.py
gme5078: utilized now

dataprofiler/profilers/column_profile_compilers.py
gme5078: done

dataprofiler/profilers/column_profile_compilers.py
gme5078: Separated out into its own function, but this is the only compiler that will use it

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -70,6 +71,132 @@ def test_add_profilers(self):
...
+        # Test different data types
+        data1 = pd.Series(['-2', '-1', '1', '2'])
+        data2 = pd.Series(["YO YO YO", "HELLO"])
+        compiler1 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data1)
+        compiler2 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data2)
+
+        expected_diff = {
+            'data_type_representation': {
+                'datetime': 'unchanged',
+                'int': 1.0,
```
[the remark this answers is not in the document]
gme5078: We could but that is not in this PR, that can be changed in the utils in a different PR

dataprofiler/tests/profilers/test_column_profile_compilers.py
gme5078: Thats not what this is showing. This is showing that one profile had None for int ratio and the other profile had 1.0 for int ratio. That change could be made in the utils file in a diff PR tho.

dataprofiler/tests/profilers/test_column_profile_compilers.py
gme5078: If both values are none than nothing has changed exactly, no need to include it

dataprofiler/tests/profilers/test_column_profile_compilers.py
gme5078: added!

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -194,6 +209,67 @@ def profile(self):
...
+        data_type1 = self.selected_data_type
+        data_type2 = other.selected_data_type
+        if data_type1 is not None or data_type2 is not None:
+            diff_profile["data_type"] = utils.find_diff_of_strings(data_type1,
+                                                                   data_type2)
+        
+        # Find diff of matching profile statistics
+        if "data_type" in diff_profile and diff_profile["data_type"] == "unchanged":
+            diff_profile["statistics"] = self._profiles[data_type1]\
+                .diff(other._profiles[data_type2], options)
```
JGSweets, line 266: indent over remove first condition?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -194,6 +209,67 @@ def profile(self):
             )
         return profile
 
+    @property
+    def selected_data_type(self):
```
JGSweets, line 213: do you want to use this in the func above?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -194,6 +209,67 @@ def profile(self):
...
+        Finds the selected data_type in a primitive compiler
+        
+        :return: name of the selected data type
+        :rtype: str
+        """
+        matched_profile = None
+        if self._profiles:
+            for key, profiler in self._profiles.items():
+                if matched_profile is None and profiler.data_type_ratio == 1.0:
+                    matched_profile = key
```
JGSweets, line 224: Can return immediately here

dataprofiler/profilers/column_profile_compilers.py
JGSweets: nvm, for loop needs ratios

JGSweets (approved):

dataprofiler/tests/profilers/test_column_profile_compilers.py
AnhTruong: my concern is about the consistency of the results. Above we have `'int': 1` or `'int': -1`. Here we have `'int': [None, 1]`. Should we make a common output format?

dataprofiler/tests/profilers/test_column_profile_compilers.py
AnhTruong: imo `unchanged` is different from `[None, None]`. Unchanged is when we have type but the type is unchanged, but [None, None] means none of profiles has the int option enabled. [None, None] should be consistent with the format that I mentioned above.

dataprofiler/tests/profilers/test_column_profile_compilers.py
AnhTruong: @JGSweets @ChrisWallace2020 : thoughts?. Also for my comment below.

dataprofiler/tests/profilers/test_column_profile_compilers.py
JGSweets: If the options suggest to leave it out of the profiler in both cases, I think it is fair to not have it in the diff.

Example, user turns off everything but the labeling in two profiles. We shouldn't give them information on int / float in the diff

JGSweets: At best, I could see `None` to maintain consistency. but not a list

dataprofiler/tests/profilers/test_column_profile_compilers.py
ChrisWallace2020: I think this is fine because if we just made it 1, that would imply `None` is synonymous with `0`, and I think that distinction should be kept intact

ChrisWallace2020 (approved):

PR #303, opened 2021-07-01
StructuredProfiler report dict -> list refactor (part 2/3)

ChrisWallace2020:
This is the 2nd part of a 3 part refactor, as follows:

1. Refactor StructuredProfiler._profile to be a list instead of a dict, and store a mapping that maps column names to a list of indices (in _profile) that correspond to data that was in a column of the given name. Refactor tests accordingly.
2. Refactor StructuredProfiler report['data_stats'] to be a list that mirrors StructuredProfiler._profile, and stores a column name to index mapping in global_stats. Refactor tests accordingly.
3. Update documentation (README as well as notebooks)

### DIFF
--- a/.github/workflows/test-python-package.yml
+++ b/.github/workflows/test-python-package.yml
@@ -5,7 +5,7 @@ name: Test Python Package
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, duplicate_column_fix ]
 
 jobs:
   build:
--- a/dataprofiler/profilers/helpers/report_helpers.py
+++ b/dataprofiler/profilers/helpers/report_helpers.py
@@ -87,24 +87,6 @@ def _prepare_report(report, output_format=None, omit_keys=None):
             "data_stats.*.statistics.histogram"
         ])
         output_format = "pretty"
-
-    # Modify omit_keys to account for data_stats being a list and not a dict
-    # With column names for keys
-    new_omit_keys = []
-    for omit_key in omit_keys:
-        key_list = omit_key.split(".")
-        if len(key_list) > 1 and key_list[0] == "data_stats" \
-                and key_list[1] != "*":
-            idxs = report["global_stats"]["profile_schema"][key_list[1]]
-            for i in idxs:
-                swapped_list = key_list
-                swapped_list[1] = str(i)
-                swapped_str = ".".join(swapped_list)
-                new_omit_keys.append(swapped_str)
-        else:
-            new_omit_keys.append(omit_key)
-
-    omit_keys = new_omit_keys
     
     for key in report:
         
@@ -118,47 +100,65 @@ def _prepare_report(report, output_format=None, omit_keys=None):
         if isinstance(value, set):
             value = sorted(list(value))
 
+        if "*" in omit_keys:
+            continue
+
         # For data_stats (in structured case), need to recurse through a list
         # As well as account for omit_keys containing indices instead of keys
-        if key == "data_stats" and isinstance(value, list):
-            fmt_report["data_stats"] = []
-            # split off any remaining keys for the recursion
-            # i.e. [test0, test1.test2] -> omit_keys => [test1.test2]
-            # This will filter out all the data_stats omit keys
-            data_stat_omit_keys = []
-            for omit_key in omit_keys:
-                omit_key_split = omit_key.split('.', 1)
+        elif key == "data_stats" and isinstance(value, list) \
+                and "data_stats" not in omit_keys:
 
-                # Must have more keys left for recursion
-                if len(omit_key_split) > 1:
-                    next_key_layer = omit_key_split[-1]
-                    prior_key_layer = omit_key_split[0]
-                    if len(next_key_layer) > 0:
-                        if prior_key_layer == '*' or prior_key_layer == key:
-                            data_stat_omit_keys.append(next_key_layer)
+            # Map cols to keep in report with their layer's omit keys
+            kept_cols_oks = dict()
+            # Cols that will have None in report since they were omitted
+            none_cols = []
+            for col_ind in range(len(value)):
+                # column is being omitted
+                if f"data_stats.{str(value[col_ind].get('column_name'))}" \
+                        in omit_keys:
+                    none_cols.append(col_ind)
+                    continue
 
-            # Recurse through indices in data_stats and call report_helper
-            for i in range(len(value)):
-                i_index_omit_keys = []
-                # Filter off data_stats omit keys depending on index
-                for nl_key in data_stat_omit_keys:
-                    key_split = nl_key.split(".", 1)
+                # update omit keys
+                next_layer_omit_keys = []
+                # Track omit_keys that pertain to data_stats
+                ds_omit_keys = [omit_key[11:] for omit_key in omit_keys
+                                if omit_key[:11] == "data_stats."]
+                for omit_key in ds_omit_keys:
+                    omit_key_split = omit_key.split('.', 1)
 
-                    # Must have more keys left for recursion
-                    if len(key_split) > 1:
-                        next_key_layer = key_split[-1]
-                        index = key_split[0]
-                        if len(next_key_layer) > 0:
-                            if (index.isdigit() and int(index) == i) \
-                                    or index == "*":
-                                i_index_omit_keys.append(next_key_layer)
+                    # this omit occurs only at this level, don't include
+                    if len(omit_key_split) == 1:
+                        continue
 
-                # Recursively prepare each column in data_stats list
-                fmt_report["data_stats"].append(
-                    _prepare_report(value[i], output_format, i_index_omit_keys))
+                    # this omit applies to everything below
+                    elif omit_key_split[0] == "*":
+                        next_layer_omit_keys.append(omit_key_split[1])
+
+                    # this omit applies to the name of this indexed column
+                    elif omit_key_split[0] == \
+                            str(value[col_ind].get("column_name")):
+                        next_layer_omit_keys.append(omit_key_split[1])
+
+                # update report and list for column we are keeping
+                kept_cols_oks[col_ind] = next_layer_omit_keys
+
+            # Helper to create reports based on whether ind is kept or omitted
+            def ds_helper(ind):
+                # Return None if col is to be omitted
+                if ind in none_cols:
+                    return None
+                # Return prepared report with this columns next level omit keys
+                # Index into report["data_stats"][ind] (value[ind])
+                return _prepare_report(value[ind], output_format,
+                                       kept_cols_oks[ind])
+
+            # Get subset of columns to keep
+            fmt_report["data_stats"] = [ds_helper(ind)
+                                        for ind in range(len(value))]
 
         # Do not recurse or modify profile_schema
-        elif key == "profile_schema":
+        elif key == "profile_schema" and "profile_schema" not in omit_keys:
             fmt_report[key] = value
 
         elif isinstance(value, dict):
--- a/dataprofiler/profilers/helpers/report_helpers.py
+++ b/dataprofiler/profilers/helpers/report_helpers.py
@@ -87,6 +87,24 @@ def _prepare_report(report, output_format=None, omit_keys=None):
             "data_stats.*.statistics.histogram"
         ])
         output_format = "pretty"
+
+    # Modify omit_keys to account for data_stats being a list and not a dict
+    # With column names for keys
+    new_omit_keys = []
+    for omit_key in omit_keys:
+        key_list = omit_key.split(".")
+        if len(key_list) > 1 and key_list[0] == "data_stats" \
+                and key_list[1] != "*":
+            idxs = report["global_stats"]["profile_schema"][key_list[1]]
+            for i in idxs:
+                swapped_list = key_list
+                swapped_list[1] = str(i)
+                swapped_str = ".".join(swapped_list)
+                new_omit_keys.append(swapped_str)
+        else:
+            new_omit_keys.append(omit_key)
+
+    omit_keys = new_omit_keys
     
     for key in report:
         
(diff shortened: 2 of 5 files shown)


.github/workflows/test-python-package.yml
```diff
@@ -5,7 +5,7 @@ name: Test Python Package
 
 on:
   pull_request:
-    branches: [ main ]
+    branches: [ main, duplicate_column_fix ]
```
ChrisWallace2020, line 8: Now tests run on this branch as well

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +118,50 @@ def _prepare_report(report, output_format=None, omit_keys=None):
         if isinstance(value, set):
             value = sorted(list(value))
 
-        if isinstance(value, dict):
+        # For data_stats (in structured case), need to recurse through a list
+        # As well as account for omit_keys containing indices instead of keys
+        if key == "data_stats" and isinstance(value, list):
```
JGSweets, line 123: So the reason for both conditions is unstructured has the key but is not a list?

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +118,50 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+            fmt_report["data_stats"] = []
```
JGSweets, line 124: any reason we cannot format in place? why do we need to create a new list and replace?

JGSweets: I see bc we want to remove from. format. Another option is to keep a list of indexes to keep and subset after (only fix those keeping)

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -87,6 +87,24 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+            idxs = report["global_stats"]["profile_schema"][key_list[1]]
+            for i in idxs:
+                swapped_list = key_list
+                swapped_list[1] = str(i)
+                swapped_str = ".".join(swapped_list)
+                new_omit_keys.append(swapped_str)
+        else:
+            new_omit_keys.append(omit_key)
+
+    omit_keys = new_omit_keys
```
JGSweets, line 107: do we need to restructured this? can we keep it as a list of strings?

Instead when we are at data_stats, can we knowingly if structured match what could be * or otherwise to the column_name?

JGSweets: Personally, I wonder if _prepare report would be better if it kept a path and utilized regex of the omits.

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +118,50 @@ def _prepare_report(report, output_format=None, omit_keys=None):
         if isinstance(value, set):
             value = sorted(list(value))
 
-        if isinstance(value, dict):
```
JGSweets, line 103: above at `if key in omit_keys:` if we kept original structure, could either utilize regex or for string vs int column name can convert key to `str` for comparison.

Also, i wonder if this should also count any place that has a `*` e.g. `data_stats.*` wouldn't work here would it?

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +118,50 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+                        next_key_layer = key_split[-1]
+                        index = key_split[0]
+                        if len(next_key_layer) > 0:
+                            if (index.isdigit() and int(index) == i) \
+                                    or index == "*":
+                                i_index_omit_keys.append(next_key_layer)
+
+                # Recursively prepare each column in data_stats list
+                fmt_report["data_stats"].append(
+                    _prepare_report(value[i], output_format, i_index_omit_keys))
```
JGSweets, line 158: if the above string for omits is unchanged, at this point we could have something like:
```python
"data_stats.TEST1.statistics.times",
"data_stats.TEST2.statistics.avg_predictions",
"data_stats.*.statistics.data_label_representation",
```

In this case would something like this work?
```python
# if * then nothing else should exist
if '*' in omit_keys:
    continue

# recurse through indices calling prepare
for col_ind in range(value):
    # column is being omitted
    if str(report[key][col_ind].get("column_name")) in omit_keys:
        continue
    
    # update omit keys
    next_layer_omit_keys = []
    for omit_key in omit_keys:
       omit_key_split = omit_key.split('.', 1)
       
       # this omit occurs only at this level, don't include
       if len(omit_key_split) == 1:
           continue
           
        # this omit applies to everything below where more matching can occur
        elif omit_key[:2] == '*.':
            next_layer_omit_keys.append(omit_key[2:]
            
        # this omit validates that the column name matches to get included
        elif omit_key_split[0] == str(report[key][col_ind].get("column_name")):
            next_layer_omit_keys.append(omit_key_split[1]
    
    report[key][col_ind]._prepare_report(report[key][col_ind], output_format, next_layer_omit_keys))
```

JGSweets: Just trying to avoid the omit key change additions and reverts

JGSweets: Also if we didn't want to do a columns to keep or a list at the end, can reverse the for loop and pop it from the list
e.g. `report[key].pop()` followed by a `continue`

JGSweets (commented):
Do our tests go to the depth of checking the column name for validation our reports are correct?

dataprofiler/profilers/helpers/report_helpers.py
ChrisWallace2020: Correct

ChrisWallace2020:
The old ones not so much but the ones I wrote assert that the right names are where you expect them in the column and I check a few stats (min, max, sum) to make sure the correct calculations are there too.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1312,28 +1312,33 @@ def report(self, report_options=None):
         omit_keys = report_options.get("omit_keys", [])
         num_quantile_groups = report_options.get("num_quantile_groups", 4)
 
-        columns = list(self._profile.values())
+        report_schema = dict()
+        for i in range(len(self._profile)):
+            col = self._profile[i].name
+            report_schema.setdefault(col, []).append(i)
+
```
JGSweets, line 1319: since we already loop through profile below, should we just calculate the `profile_schema` in that loop?

ChrisWallace2020: Good idea, will add

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,68 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+
+            # Map cols to keep in report with their layer's omit keys
+            kept_cols_oks = dict()
+            # Cols that will have None in report since they were omitted
+            none_cols = []
+            for col_ind in range(len(value)):
+                col_name = str(value[col_ind].get('column_name'))
+                # column is being omitted
+                if f"data_stats.{col_name}" \
+                        in omit_keys:
```
JGSweets, line 118: Any reason why this is not on one line?

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,68 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+        if "*" in omit_keys:
+            continue
+
+        # For data_stats (in structured case), need to recurse through a list
+        elif key == "data_stats" and isinstance(value, list) \
+                and "data_stats" not in omit_keys:
+
+            # Map cols to keep in report with their layer's omit keys
+            kept_cols_oks = dict()
+            # Cols that will have None in report since they were omitted
```
JGSweets, line 112: i suggest having a space between the line and the next comment, below too

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,68 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+
+            # Helper to create reports based on whether ind is kept or omitted
+            def ds_helper(ind):
+                # Return None if col is to be omitted
+                if ind in none_cols:
+                    return None
+                # Return prepared report with this columns next level omit keys
+                # Index into report["data_stats"][ind] (value[ind])
+                return _prepare_report(value[ind], output_format,
+                                       kept_cols_oks[ind])
```
JGSweets, line 154: i find this more confusing than just computing the report while in the loop above. Instead of appending to another col, maybe just set the value as none for the key before each continue?

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,68 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+                        in omit_keys:
+                    none_cols.append(col_ind)
+                    continue
+
+                # update omit keys
+                next_layer_omit_keys = []
+                # Track omit_keys that pertain to data_stats
+                ds_omit_keys = [omit_key[11:] for omit_key in omit_keys
+                                if omit_key[:11] == "data_stats."]
+                for omit_key in ds_omit_keys:
```
JGSweets, line 127: this is like doing the same loop twice. I think it would be better to just have an if statement in the loop below and utilize a continue maybe.

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,68 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+                    omit_key_split = omit_key.split('.', 1)
+
+                    # this omit occurs only at this level, don't include
+                    if len(omit_key_split) == 1 and omit_key in {"*", col_name}:
+                        none_cols.append(col_ind)
+                        continue
```
JGSweets, line 133: instead of continue, should we actually break here since we identified it? 

Since the second half already requires it to be an exact match, I don't think we need the `len(omit_key_split) == 1`, no?

dataprofiler/profilers/helpers/report_helpers.py
ChrisWallace2020: Whoops, at one point this was too long but I made some changes and never made it back to 1 line.

dataprofiler/profilers/helpers/report_helpers.py
ChrisWallace2020: Fixed, short circuiting in omit case is consolidated and report creation is done column by column

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,54 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+                        # Pull out omit_key information given its for data_stats
+                        omit_key = omit_key[11:]
+
+                    omit_key_split = omit_key.split('.', 1)
+
+                    if omit_key_split[0] in {"*", col_name}:
+                        # Already omitted entire column in case above
+                        # Therefore guaranteed to have further keys in this case
+                        if len(omit_key_split) < 2:
+                            raise ValueError("Invalid omit keys given to report")
```
JGSweets, line 136: not sure we should error if it technically overlaps

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,54 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+
+            fmt_report["data_stats"] = []
+
+            for col_ind in range(len(value)):
+                col_name = str(value[col_ind].get('column_name'))
+                # column is being omitted
+                if f"data_stats.{col_name}" in omit_keys \
+                        or "data_stats.*" in omit_keys:
+                    fmt_report["data_stats"].append(None)
+                    continue
```
JGSweets, line 118: the issue becomes, these are technically already looping through omit keys. Here we have 2 `in` conditions. and then below we have a third loop through `in` with the `for`.  We also then have to remember that these conditions exist in combination with what we are doing below. Maybe it is best to move these into the for loop below for clarity:

```python
next_layer_omit_keys = []
for omit_key in omit_keys:
    
    # case where omit_key removes this column entirely
    if omit_key in {"*", col_name, f"data_stats.{col_name}":
        fmt_report["data_stats"].append(None)
        break
    
    # case where the omit extends into this column
    omit_key_split = omit_key.split('.', 1)
    if omit_key_split[0] in {"*", col_name}:
        next_layer_omit_keys.append(omit_key_split[1])
    
    fmt_report["data_stats"].append(
        _prepare_report(value[col_ind], output_format,
                        next_layer_omit_keys))
```

ChrisWallace2020: Resolved, now we check this condition within the omit_key loop

dataprofiler/profilers/helpers/report_helpers.py
```diff
@@ -100,7 +100,51 @@ def _prepare_report(report, output_format=None, omit_keys=None):
...
+                        next_layer_omit_keys.append(next_key_split[1])
+
+                # update report and list for column we are keeping
+                if not is_omitted_col:
+                    fmt_report["data_stats"].append(
+                        _prepare_report(value[col_ind], output_format,
+                                        next_layer_omit_keys))
+
+        # Do not recurse or modify profile_schema
+        elif key == "profile_schema" and "profile_schema" not in omit_keys:
```
JGSweets, line 144: this is stored as a list, correct?

ChrisWallace2020: It is a dictionary mapping strings to list[int]. The reason this is necessary is because in the "pretty" format, those lists get turned into strings with ... in the middle if they are too long, but this schema is necessary for navigating data_stats, so we can't modify or alter it at all, hence the necessity for this.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -303,30 +303,122 @@ def test_report(self):
...
+                "encoding", "profile_schema"
             ]
         )
         flat_report = self.trained_schema.report(report_options={"output_format":"flat"})
         self.assertEqual(test_utils.get_depth(flat_report), 1)
         with mock.patch('dataprofiler.profilers.helpers.report_helpers._prepare_report') as pr_mock:
             self.trained_schema.report(report_options={"output_format":'pretty'})
-            self.assertEqual(pr_mock.call_count, 2)
+            # Once for global_stats, once for each of 16 columns
+            self.assertEqual(pr_mock.call_count, 17)
```
JGSweets, line 315: previously it was just a dict which passed into the func, but now we loop through columns, makes sense.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -303,30 +303,122 @@ def test_report(self):
...
+    def test_pretty_report_doesnt_cast_schema(self):
+        report = self.trained_schema.report(
+            report_options={"output_format": "pretty"})
+        self.assertIsInstance(report["global_stats"]["profile_schema"], dict)
+        self.assertIsInstance(report["data_stats"], list)
+        for idx_list in report["global_stats"]["profile_schema"].values():
+            self.assertIsInstance(idx_list, list)
+            for idx in idx_list:
+                self.assertIsInstance(idx, int)
+                self.assertTrue(idx >= 0)
```
JGSweets, line 348: instead would it be better to just assert the dict?

ChrisWallace2020: Fixed

JGSweets (approved):

AnhTruong (approved):
lgtm

PR #305, opened 2021-07-02
Add entity percentages to unstructured labeler profile

andrew-yin:
Pretty simple and short PR because entity percentages are already calculated. This just adds it to the profile.

### DIFF
--- a/dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
+++ b/dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
@@ -139,6 +139,39 @@ class TestUnstructuredLabelerProfile(unittest.TestCase):
 
     @mock.patch('dataprofiler.profilers.'
                 'unstructured_labeler_profile.DataLabeler')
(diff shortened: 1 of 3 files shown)


dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -121,6 +121,11 @@ def test_profile(self, processor_class_mock, model_class_mock):
                 'true_char_level': defaultdict(int, {'UNKNOWN': 1}),
                 'word_level': defaultdict(int)
             },
+            entity_percentages={
+                'postprocess_char_level': defaultdict(int, {'UNKNOWN': 1.0}),
+                'true_char_level': defaultdict(int, {'UNKNOWN': 1.0}),
+                'word_level': defaultdict(int)
```
gme5078, line 127: Can we add more entities to this test so that there are several values for each dict?

JGSweets (approved):

AnhTruong (approved):

PR #307, opened 2021-07-06
Make correlation matrix as the full matrix

AnhTruong:
The resulted correlation matrix now has the size n x n, where n is the total number of columns in the input data.
Columns with nan value(s) has nan correlation with all other columns in the matrix.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1261,7 +1261,7 @@ class StructuredProfiler(BaseProfiler):
                 "duplicate_row_count": self._get_duplicate_row_count(),
                 "file_type": self.file_type,
                 "encoding": self.encoding,
-                "correlation_matrix": self._get_serializable_corr_mat()
+                "correlation_matrix": self.correlation_matrix
             }),
             ("data_stats", OrderedDict()),
         ])
@@ -1359,11 +1359,6 @@ class StructuredProfiler(BaseProfiler):
             self.row_has_null_count = len(null_in_row_count)
             self.row_is_null_count = len(null_rows)
 
-    def _get_serializable_corr_mat(self):
-        if self.correlation_matrix is not None:
-            return self.correlation_matrix.to_json(orient='split')
-        return self.correlation_matrix
-
     def _get_correlation(self, clean_samples):
         """
         Calculate correlation matrix on the cleaned data.
@@ -1372,17 +1367,26 @@ class StructuredProfiler(BaseProfiler):
         :type clean_samples: dict()
         """
         columns = self.options.correlation.columns
+        clean_column_ids = []
         if columns is None:
-            for col in self._profile:
+            for col_id, col in enumerate(self._profile.keys()):
                 if (self._profile[col].profile['data_type'] not in ['int', 'float']
                         or self._profile[col].null_count > 0):
                     clean_samples.pop(col)
+                else:
+                    clean_column_ids.append(col_id)
         if len(clean_samples) <= 1:
             return None
 
         data = pd.DataFrame(clean_samples)
         data = data.apply(pd.to_numeric, errors='coerce')
-        return data.corr()
+
+        n_cols = len(self._profile)
+        corr_mat = np.empty((n_cols, n_cols))
+        corr_mat.fill(np.nan)
+        rows = [[id] for id in clean_column_ids]
+        corr_mat[rows, clean_column_ids] = np.corrcoef(data, rowvar=False)
+        return corr_mat
 
     def _update_correlation(self, clean_samples):
         """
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1372,17 +1367,26 @@ def _get_correlation(self, clean_samples):
...
         if len(clean_samples) <= 1:
             return None
 
         data = pd.DataFrame(clean_samples)
         data = data.apply(pd.to_numeric, errors='coerce')
-        return data.corr()
+
+        n_cols = len(self._profile)
+        corr_mat = np.empty((n_cols, n_cols))
+        corr_mat.fill(np.nan)
```
JGSweets, line 1386: `np.full((n_cols, n_cols), np.nan)`

AnhTruong: `fill` is actually faster: https://stackoverflow.com/questions/1704823/create-numpy-matrix-filled-with-nans

dataprofiler/profilers/profile_builder.py
```diff
@@ -1372,17 +1367,26 @@ def _get_correlation(self, clean_samples):
...
+        rows = [[id] for id in clean_column_ids]
+        corr_mat[rows, clean_column_ids] = np.corrcoef(data, rowvar=False)
```
JGSweets, line 1388: i believe we can do
`corr_mat[np.ix_(clean_column_ids, clean_column_ids)] = np.corrcoef(data, rowvar=False)`

dataprofiler/profilers/profile_builder.py
JGSweets: I'm not sure if speed is what is needed here as much as cleanliness of code as `n` should be relatively small.  I'll leave the choice up to you.

dataprofiler/profilers/profile_builder.py
JGSweets: However, not sure that it is cleaner.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1420,26 +1424,32 @@ def _merge_correlation(self, other):
         if corr_mat1 is None or corr_mat2 is None:
             return None
 
-        if len(corr_mat1.columns) != len(corr_mat2.columns) or \
-                len(corr_mat1.columns) <= 1:
+        col_ids1 = np.where(~np.isnan(corr_mat1).all(axis=0))[0]
+        col_ids2 = np.where(~np.isnan(corr_mat2).all(axis=0))[0]
+
+        if len(col_ids1) != len(col_ids2) or len(col_ids1) <= 1:
```
JGSweets, line 1430: I'm not sure this is correct. Just because one profile has a `NaN` in one column doesn't mean it should not be mergeable for the other columns, right?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1420,26 +1424,32 @@ def _merge_correlation(self, other):
...
-             for profile_name in corr_mat1.columns])
+             for i, profile_name in enumerate(self._profile.keys())
+             if i in col_ids1])
         n1 = self.total_samples
 
         mean2 = np.array(
             [other._profile[profile_name].profile['statistics']['mean']
-             for profile_name in corr_mat1.columns])
+             for i, profile_name in enumerate(self._profile.keys())
+             if i in col_ids1])
```
ChrisWallace2020, line 1448: Should this be col_ids2?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1420,26 +1424,32 @@ def _merge_correlation(self, other):
...
         std2 = np.array(
             [other._profile[profile_name].profile['statistics']['stddev']
-             for profile_name in corr_mat1.columns])
+             for i, profile_name in enumerate(self._profile.keys())
+             if i in col_ids1])
```
ChrisWallace2020, line 1452: Should this be col_ids2?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1420,26 +1424,32 @@ def _merge_correlation(self, other):
         if corr_mat1 is None or corr_mat2 is None:
             return None
 
-        if len(corr_mat1.columns) != len(corr_mat2.columns) or \
-                len(corr_mat1.columns) <= 1:
+        col_ids1 = np.where(~np.isnan(corr_mat1).all(axis=0))[0]
+        col_ids2 = np.where(~np.isnan(corr_mat2).all(axis=0))[0]
```
ChrisWallace2020, line 1428: Can you leave a comment as this is a few operations nested together.

dataprofiler/profilers/profile_builder.py
JGSweets: good catch

dataprofiler/profilers/profile_builder.py
AnhTruong: at this part of code, col_ids1 should be the same as col_ids2, but I changed it for readability

dataprofiler/profilers/profile_builder.py
AnhTruong: added

dataprofiler/profilers/profile_builder.py
AnhTruong: changed

dataprofiler/profilers/profile_builder.py
AnhTruong: changed

dataprofiler/profilers/profile_builder.py
AnhTruong: agree. We can discuss further on that later.

dataprofiler/profilers/profile_builder.py
AnhTruong: I keep it as is for now, at least it's a bit faster than np.ix

dataprofiler/profilers/profile_builder.py
AnhTruong: i've done some tests, it's not so much difference when n is large, so I changed to np.full for readability

JGSweets (approved):

ChrisWallace2020 (approved):

PR #313, opened 2021-07-08
Bug fix to make sure samples being stored by UnstructuredProfiler save

ChrisWallace2020:
Previous bug had issue where when loading UnstructuredProfiler, `.sample` was not saved (and therefore not loaded) so when trying to do any operations with it you received an AttributeError saying UnstructuredProfiler does not have a 'sample' attribute. This fixes that bug and includes a test inclusion that failed before the inclusion of the bug fix.


JGSweets (commented):
I believe, `TestUnstructuredProfiler.test_base` needs to update as well.

JGSweets (approved):

AnhTruong (approved):

PR #317, opened 2021-07-08
Add update functionality to correlation 

andrew-yin:
~~Currently, some tests fail because some compiler profiles are mocked out in certain test cases. I also haven't figured out a way to update the correlation matrix if a new column is added to the data, but I'm not sure if this is even possible mathematically.~~

Previously, the correlation matrix was set to `None` when the profile was updated. This update enables the correlation matrix to be updated with new data.


dataprofiler/profilers/profile_builder.py
```diff
@@ -1392,24 +1392,39 @@ def _get_correlation(self, clean_samples):
...
+            if (len(batch_means) != len(prev_dependent_properties["mean"])):
+                warnings.warn("A new column was added when updating. Correlations cannot be calculated"
+                              " for new columns.", RuntimeWarning)
+                return
+
+            self.correlation_matrix = self._merge_correlation_helper(
+                 self.correlation_matrix, prev_dependent_properties["mean"],
+                 prev_dependent_properties["std"], self.total_samples,
+                 batch_corr, batch_means, batch_stds, len(data.index))
             return
```
andrew-yin, line 1425: Not sure if the best practice here is to do nothing, thoughts?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1550,6 +1565,18 @@ def tqdm(l):
...
 
+        # Record the previous mean/std values of columns that need to
+        # have correlation updated. Non-numeric columns have NaN
+        # so that they stay NaN when updating.
+        prev_dependent_properties = {
+            'mean': np.full(len(self._profile.keys()), np.nan),
+            'std': np.full(len(self._profile.keys()), np.nan)
+        }
+        for id, col in enumerate(self._profile.keys()):
+            if self._profile[col].profile['data_type'] in ['int', 'float']:
```
andrew-yin, line 1576: This line causes some test cases to fail with a KeyError on "statistics". Some test cases mock out the compilers, so the profile property can't be called without an error.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1392,24 +1392,39 @@ def _get_correlation(self, clean_samples):
...
+                        and col in data.columns:
+                    batch_means[id] = data[col].mean()
+                    batch_stds[id] = data[col].std()
+
+            # If these lengths are different, then some columns were added to to the
+            # profile.
+            if (len(batch_means) != len(prev_dependent_properties["mean"])):
+                warnings.warn("A new column was added when updating. Correlations cannot be calculated"
+                              " for new columns.", RuntimeWarning)
+                return
```
JGSweets, line 1419: Add a note to remove when @ChrisWallace2020 pr gets merged as it won't allow it.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -86,6 +86,7 @@ def __init__(self, options=None):
                     'bin_edges': None
                 }
             }
+        self._batch_history = []
```
andrew-yin, line 89: This could be changed to a queue of some fixed size if we don't want to store every single update history.

dataprofiler/profilers/profile_builder.py
JGSweets: I would move the ` self._merge_correlation_helper(` outside the if to below and `_merge_correlation_helper` should be able to handle the both cases.

Ultimately the structure in _update should be: 
1. get corr
2. merge corr
    a. can handle both null case
    b. can handle one set of values are null
    c. can handle both sets are non-null

dataprofiler/profilers/profile_builder.py
```diff
@@ -1550,6 +1572,22 @@ def tqdm(l):
...
         sample_ids = np.array(sample_ids)
 
+        # Record the previous mean/std values of columns that need to
+        # have correlation updated. Non-numeric columns have NaN
+        # so that they stay NaN when updating.
+        prev_dependent_properties = {
+            'mean': np.full(len(self._profile.keys()), np.nan),
+            'std': np.full(len(self._profile.keys()), np.nan)
+        }
+        for id, col in enumerate(self._profile.keys()):
```
JGSweets, line 1582: since this code and code above are very similar, can we refactor such that we don't repeat?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1372,7 +1372,7 @@ def _get_correlation(self, clean_samples):
         clean_column_ids = []
         if columns is None:
             for col_id, col in enumerate(self._profile.keys()):
-                if (self._profile[col].profile['data_type'] not in ['int', 'float']
+                if (self ._profile[col].profile['data_type'] not in ['int', 'float']
```
gme5078, line 1375: Interesting, does this compile with that space? Should remove anyway

dataprofiler/profilers/profile_builder.py
AnhTruong: agree

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -360,22 +359,84 @@ def test_merge_correlation(self, *mocks):
...
+            [0.26594894270403086, -0.4907239, 1.0]
         ])
-        np.testing.assert_array_equal(expected_corr_mat,
+        np.testing.assert_array_almost_equal(expected_corr_mat,
                                       merged_profile.correlation_matrix)
 
-        # merge between an existing corr and None correlation (with data)
+    def test_correlation_update(self):
+        profile_options = dp.ProfilerOptions()
+        profile_options.set({"correlation.is_enabled": True})
```
AnhTruong, line 370: should we set correlation option to True as the update works now?
@JGSweets thoughts?

AnhTruong (commented):
some comments

dataprofiler/profilers/profile_builder.py
```diff
@@ -1392,24 +1392,46 @@ def _get_correlation(self, clean_samples):
...
+                        # Force bias correction for stddev
+                        n = batch_properties['match_count']
+                        batch_stds[id] = np.sqrt(batch_properties['biased_variance'] * n / (n - 1))
+
+            # If these lengths are different, then some columns were added to to the
+            # profile.
+            if (len(batch_means) != len(prev_dependent_properties["mean"])):
+                # TODO: Remove this when adding new columns isn't supported.
+                warnings.warn("A new column was added when updating. Correlations cannot be calculated"
+                              " for new columns.", RuntimeWarning)
```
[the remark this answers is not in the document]
andrew-yin: This will be removed soon as I didn't realize adding new columns would not be allowed after Chris's update merged.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -911,6 +912,9 @@ def _update_helper(self, df_series_clean, profile):
                                      df_series=df_series_clean,
                                      prev_dependent_properties=prev_dependent_properties,
                                      subset_properties=subset_properties)
+        if len(self._batch_history) == 5:
```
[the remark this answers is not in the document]
JGSweets: This will be tunable in the future, but I think this is fine for now.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Let's get a profiled time first before we enable by default.

JGSweets (approved):

dataprofiler/profilers/profile_builder.py
```diff
@@ -1545,6 +1582,11 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
         :type std2: np.array
         :return: merged correlation matrix
         """
+        if len(mean1) == 0:
+           return corr_mat2
+        elif len(mean2) == 0:
+           return corr_mat1
+
```
AnhTruong, line 1589: should we add the case when either corr_mat1 or corr_mat2 is None?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1545,6 +1580,15 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
         :type std2: np.array
         :return: merged correlation matrix
         """
+        if corr_mat1 is None:
+            return corr_mat2
+        elif corr_mat2 is None:
+            return corr_mat1
```
AnhTruong, line 1586: should we return None in this case and above?

AnhTruong (commented):
should we add tests for the case with only one numeric col and multiple columns but with only one numeric col

JGSweets (dismissed):

dataprofiler/profilers/profile_builder.py
AnhTruong:
```
        if self.total_samples == 0:
            return corr_mat2
        if other.total_samples == 0:
            return corr_mat1

            return None
```

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -271,41 +271,53 @@ def test_correlation(self, *mock):
         profile_options = dp.ProfilerOptions()
         profile_options.set({"correlation.is_enabled": True})
 
-        # data with one column, correlation not updated
+        # data with one column
         data = pd.DataFrame([1.0, None, 1.0, None, 5.0])
         profiler = dp.StructuredProfiler(data, options=profile_options)
-        self.assertIsNone(profiler.correlation_matrix)
+        expected_corr_mat = np.array([[np.nan]])
```
AnhTruong, line 277: imo, this case should return either `None` or `np.array([[1.0]])` ?

JGSweets: Interesting catch, yes it looks like it should be `1.0`

andrew-yin: Since the data has non-numeric values, should it be np.nan?

JGSweets: However:
```
In [25]: np.corrcoef(pd.DataFrame([1, 2]), rowvar=False)
Out[25]: 1.0

In [26]: np.corrcoef(pd.DataFrame([[1], [2]]), rowvar=False)
Out[26]: 1.0
```

AnhTruong: actually, this case has None value, then the returning value as nan makes sense. But we need to add the case without nan, e.g. 
```
>>> np.corrcoef(np.array([1, 2, 3, 4, 5]), rowvar=False)
1.0
```

AnhTruong (approved):

JGSweets (approved):

PR #318, opened 2021-07-08
Add diff to Unstructured Labeler

andrew-yin:


dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -219,7 +220,70 @@ def test_unstructured_labeler_profile_add(self, mock):
...
+            'entity_percentages': {
+                'postprocess_char_level' : utils.find_diff_of_dicts(
+                    entity_percentages1['postprocess_char_level'],
+                    entity_percentages2['postprocess_char_level']),
+                'true_char_level': utils.find_diff_of_dicts(
+                    entity_percentages1['true_char_level'],
+                    entity_percentages2['true_char_level']),
+                'word_level': utils.find_diff_of_dicts(
+                    entity_percentages1['word_level'],
+                    entity_percentages2['word_level'])
```
JGSweets, line 281: We should manually set these as I'm not sure using the function in the test which is within the function is appropriate.

andrew-yin: I thought this was the method we decided on since we have some floating point precision errors and assertDictEquals fails because on that. The past diff PRs use the util functions in the tests as well, thoughts?

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -88,6 +88,28 @@ def __add__(self, other):
...
+    def diff(self, other_profile, options=None):
+        cls = self.__class__
+        if not isinstance(other_profile, cls):
+            raise TypeError("Unsupported operand type(s) for diff: '{}' "
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+
+        entity_counts_diff = {}
+        entity_percentages_diff = {}
+        for i in ['word_level', 'true_char_level', 'postprocess_char_level']:
```
ChrisWallace2020, line 100: Kinda nitpicky, but I usually associate `i` with index, whereas these are different strings, so maybe use `key` or `level` or something to show that they are dictionary keys.

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
ChrisWallace2020: Shouldn't floating point errors not be a concern though since the values you chose would make predictable, calculatable percentages (6/20 = 0.3, 14/20 = 0.7, ...)?

andrew-yin: This was discussed last time, but 0.7 - 0.4 became 0.2999..., which failed checks. Typing that into the Python console results in that as well.

ChrisWallace2020: This is a bit of an ask but it may be worth it in the future, we could write our own util function for dict equality that uses almost equal on floats? Cuz this seems like it could be a recurring problem down the line that would be helpful for other tests too. @JGSweets thoughts?

dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -88,6 +88,28 @@ def __add__(self, other):
 
         return merged_profile
 
+    def diff(self, other_profile, options=None):
+        cls = self.__class__
```
gme5078, line 92: Dont forget a docstring

dataprofiler/tests/profilers/test_unstructured_labeler_profile.py
```diff
@@ -219,7 +220,70 @@ def test_unstructured_labeler_profile_add(self, mock):
...
+                'unstructured_labeler_profile.'
+                'CharPostprocessor')
+    def test_diff(self, mock1, mock2):
+        """
+        Tests to see that entity percentages match the counts given
+        """
+        profiler1 = UnstructuredLabelerProfile()
+        profiler1.char_sample_size = 20
+        profiler1.word_sample_size = 10
+        profiler1.entity_counts["postprocess_char_level"]["UNKNOWN"] = 6
```
gme5078, line 236: Can we try a profile that has different entity labels and also no entity labels (empty profile)?

dataprofiler/profilers/unstructured_labeler_profile.py
gme5078: options should be a part of the docstring too

gme5078 (approved):

ChrisWallace2020 (approved):

PR #319, opened 2021-07-09
Categorical Profile Differences

gme5078:
Couple things:
Some differences are only calculated if they are categorical. This is the same as the profile and this is because it would be too intensive to get these statistics when they are not categorical. 
Also, String_diff doubles as bool diff so I renamed it, let me know if you think that is inappropriate.


dataprofiler/profilers/float_column_profile.py
```diff
@@ -108,7 +108,7 @@ def __add__(self, other):
 
     def diff(self, other_profile, options=None):
         """
-        Finds the differences for FloatColumnss.
```
gme5078, line 111: typo lol

dataprofiler/profilers/utils.py
```diff
@@ -478,4 +478,8 @@ def find_diff_of_dicts(dict1, dict2):
         if key not in diff:
             diff[key] = [None, value]
 
+    # If both dicts have no keys, it is unchanged
+    if diff == {}:
+        diff = "unchanged"
```
gme5078, line 483: just some edge case protection

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -60,6 +60,47 @@ def __add__(self, other):
...
+
+        differences["categorical"] = \
+            utils.find_diff_of_strings_and_bools(self.is_match, other_profile.is_match)
+
+        differences["statistics"] = dict([
+                ('unique_count', utils.find_diff_of_numbers(len(self.categories),
+                                                            len(other_profile.categories))),
+                ('unique_ratio', utils.find_diff_of_numbers(self.unique_ratio, 
+                                                            other_profile.unique_ratio)),
+            ])
```
ChrisWallace2020, line 82: Should these all also only be calculated below if the column is categorical?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -60,6 +60,47 @@ def __add__(self, other):
...
+                                           other_profile.gini_impurity)
+            cat_count1 = dict(sorted(self._categories.items(),
+                                     key=itemgetter(1),
+                                     reverse=True))
+            cat_count2 = dict(sorted(other_profile._categories.items(),
+                                     key=itemgetter(1),
+                                     reverse=True))
+
+            differences["statistics"]['categorical_count'] = \
+                utils.find_diff_of_dicts(cat_count1, cat_count2)
```
ChrisWallace2020, line 100: Just noting this here, depending on which PR goes in first, will need to have someone add unalikability here once #320 is in.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -308,6 +308,47 @@ def test_gini_impurity(self):
...
+                    'n': 1,
+                    'maybe': [None, 2]
+                }
+            }
+        }
+
+        self.assertDictEqual(expected_diff, profile.diff(profile2))
+
+        # Test with one categorical column matching
+        df_categorical = pd.Series(["THIS", "is", "not", "a", "categorical",
```
ChrisWallace2020, line 339: If this is supposed to not be categorical, naming it `df_categorical` is a bit confusing.

gme5078: haha good catch

dataprofiler/profilers/categorical_column_profile.py
gme5078: Yea, lets get alberts in first

dataprofiler/profilers/categorical_column_profile.py
gme5078: No, these ones arent as intensive and also follow suit with the profile

ChrisWallace2020 (approved):

JGSweets (approved):

ChrisWallace2020 (approved):

PR #322, opened 2021-07-12
Add new report structure to gh pages docs

ChrisWallace2020:
Updates to example notebooks will be handled by the next large documentation update, this just handles README-level fixes. like the new report layout.


gme5078 (approved):
perfect

PR #324, opened 2021-07-13
Added diff for the stats compiler

gme5078:
Very simple compiler diff.

For the sake of speed, I could add the other compiler diffs to this PR, but I'm not sure its appropriate to make a larger PR even if its mostly straightforward.

### DIFF
--- a/dataprofiler/profilers/column_profile_compilers.py
+++ b/dataprofiler/profilers/column_profile_compilers.py
@@ -301,12 +301,11 @@ class ColumnStatsProfileCompiler(BaseCompiler):
 
         # Iterate through profiles
         all_profiles = set(self._profiles.keys()) | set(other._profiles.keys())
-        if all_profiles:
-            for key in all_profiles:
-                if key in self._profiles and key in other._profiles:
-                    diff = self._profiles[key].diff(other._profiles[key], 
-                                                    options)
-                    diff_profile.update(diff)
+        for key in all_profiles:
+            if key in self._profiles and key in other._profiles:
+                diff = self._profiles[key].diff(other._profiles[key], 
+                                                options)
+                diff_profile.update(diff)
 
         return diff_profile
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -287,6 +287,28 @@ def profile(self):
...
+        :type other: ColumnStatsProfileCompiler
+        :return: difference of the profiles
+        :rtype: dict
+        """
+        # Call super for compiler instance check
+        diff_profile = super().diff(other, options)
+
+        # Iterate through profiles
+        all_profiles = set(self._profiles.keys()) | set(other._profiles.keys())
+        if all_profiles:
```
ChrisWallace2020, line 304: You shouldn't actually need this if, because if its just empty, then the for loop has nothing to loop through, and hence it won't happen anyway.

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -223,6 +223,51 @@ def test_disabling_columns_during_primitive_diff(self):
         expected_diff = {}
         self.assertDictEqual(expected_diff, compiler1.diff(compiler2))
 
+    def test_compiler_stats_diff(self):
+        data1 = pd.Series(['1', '9', '9'])
+        data2 = pd.Series(['10', '9', '9', '9'])
+        options = StructuredOptions()
+
+        # Test normal diff
+        compiler1 = col_pro_compilers.ColumnStatsProfileCompiler(data1,options)
```
ChrisWallace2020, line 232: You don't need to pass options here since nothing is different from default

JGSweets (dismissed):

ChrisWallace2020 (approved):

JGSweets (approved):

PR #325, opened 2021-07-13
Top k categories

az85252:
I have added a feature that by default, the category count in statistics of profile will return all counts be default, but we can set the top k categories it will display. (set_categorical_count_size(k))


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -222,6 +222,17 @@ def update(self, df_series):
...
+        """
+        Sets the number of top categories to be displayed when we call profile
+        and categorical_count in the statistics dictionary
+
+        :param k: the number we are changing top_k_categories to
+        :type k: int
+        :return: None
+        """
+        self._top_k_categories = k
+
```
JGSweets, line 235: since the user won't have access to the function, lets use the `options` class to allow the set of this value.

e.g. https://github.com/capitalone/DataProfiler/blob/43c65ecad973d81bb6cc453b89bdaec2b46336c2/dataprofiler/profilers/profiler_options.py#L867
in TextProfilerOptions for Unstructured

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, name, options=None):
         self._categories = defaultdict(int)
         self.__calculations = {}
         self._filter_properties_w_options(self.__calculations, options)
+        self._top_k_categories = None
```
ChrisWallace2020, line 38: Shouldn't this default to the value in options?

ChrisWallace2020: Is there a value in options for this?

dataprofiler/profilers/profiler_options.py
```diff
@@ -695,14 +695,15 @@ def _validate_helper(self, variable_path='OrderOptions'):
 
 class CategoricalOptions(BaseInspectorOptions):
 
-    def __init__(self):
+    def __init__(self, is_enabled=True, top_k_categories=None):
```
gme5078, line 698: Hmmm, is this going to be the new pattern going forward? I see a lot of option inits with and without parameters so im unsure

ChrisWallace2020: We should probably discuss this because there is currently no consistency (e.g. `DataLabelerOptions` sets everything to `None`, doesn't take parameters, `TextProfilerOptions` takes a bunch of optional parameters)

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -375,6 +373,32 @@ def test_unalikeability(self):
...
+            ["a", "a", "b", "c", "d", "e", "e", "e", "f", "g"])
+        profile = CategoricalColumn(df_series.name, options)
+        profile.update(df_series)
+        self.assertEqual(
+            len(profile.profile['statistics']['categorical_count']), 7)
+        profile._top_k_categories = 6
+        self.assertEqual(
+            len(profile.profile['statistics']['categorical_count']), 6)
+
+        options = CategoricalOptions(top_k_categories=6)
```
ChrisWallace2020, line 388: I would change these to:
`options = CategoricalOptions()` just once at the top, and then for each case you can just do
`options.top_k_categories = ___` so that you only create 1 options object and can keep reusing it. This also won't have to change if we decide to take away the parameter from the option init.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -375,6 +373,32 @@ def test_unalikeability(self):
         profile.update(df_categorical)
         self.assertEqual(profile.unalikeability, 2*(10 + 15 + 6)/90)
 
+    def test_top_k_categories_change(self):
+        options = CategoricalOptions()
```
ChrisWallace2020, line 377: See comment below on line 388, can just keep this `options` and manually set the `top_k_categories` for each case

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -130,7 +132,7 @@ def profile(self):
             profile["statistics"]['unalikeability'] = self.unalikeability
             profile["statistics"]['categorical_count'] = dict(
                 sorted(self._categories.items(), key=itemgetter(1),
-                       reverse=True)[:top_k_categories])
+                       reverse=True)[:self._top_k_categories])
```
JGSweets, line 135: if this value is `None` does it fail?

JGSweets: It does not, interesting..

dataprofiler/profilers/profiler_options.py
JGSweets: The pattern started without. However, I believe I mentioned before it should be the pattern moving forward to include. As such we should in the future hopefully be able to init options via construction in one command rather than two.

dataprofiler/profilers/profiler_options.py
```diff
@@ -695,14 +695,15 @@ def _validate_helper(self, variable_path='OrderOptions'):
...
         """
         Options for the Categorical Column
 
         :ivar is_enabled: boolean option to enable/disable the column.
         :vartype is_enabled: bool
         """
-        BaseInspectorOptions.__init__(self)
+        BaseInspectorOptions.__init__(self, is_enabled=True)
+        self.top_k_categories = top_k_categories
```
JGSweets, line 706: We will need to update the `_validate below to ensure `top_k_categories` can be either `None` or a positive integer.

dataprofiler/profilers/profiler_options.py
```diff
@@ -713,7 +714,13 @@ def _validate_helper(self, variable_path='CategoricalOptions'):
         :return: list of errors (if raise_error is false)
         :rtype: list(str)
         """
-        return super()._validate_helper(variable_path)
+        errors = super()._validate_helper(variable_path)
+        if not (self.top_k_categories is None or (
+                isinstance(self.top_k_categories, int) and self.top_k_categories >= 0)):
+            errors.append("top_k_categories must be either None"
+                          " or a positive integer".format(variable_path))
```
JGSweets, line 721: I think this would fail would not catch the case where a user could set it to a value which is not an integer.

JGSweets: whereas:
```python
if self.top_k_categories is not None and (
                isinstance(self.top_k_categories, int) 
                or self.top_k_categories < 1)):
```
would cover the case when it is not none and it isn't a positive integer

JGSweets: Also, we need to utilize the variable_path in the error to that the user knows where this option exists:
```python
errors.append("{}.top_k_categories must be either None"
                          " or a positive integer".format(variable_path))
```
```

File: dataprofiler/tests/profilers/profiler_options/test_categorical_options.py [hunk 9]
```diff
@@ -18,6 +23,20 @@ def test_set(self):
     
     def test_validate_helper(self):
         super().test_validate_helper()
+        # These ones should throw an error
+        expected_error = "top_k_categories must be either None" \
```
JGSweets, line 27: when the error above is fixed, this will also need to be updated

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -8,7 +8,12 @@ class TestCategoricalOptions(TestBaseInspectorOptions):
     option_class = CategoricalOptions
     
     def test_init(self):
-        super().test_init()
+        option = self.get_options()
+        self.assertDictEqual({"is_enabled": True, "top_k_categories": None}, option.properties)
+        option = self.get_options(is_enabled=False)
+        self.assertDictEqual({"is_enabled": False, "top_k_categories": None}, option.properties)
+        option = self.get_options(top_k_categories=2)
+        self.assertDictEqual({"is_enabled": True, "top_k_categories": 2}, option.properties)
```
JGSweets, line 16: `test_set` need to also be updated for the new variable.

we can do this by ensuring the new variable can be properly set similar to `TestTextProfilerOptions.test_set`

dataprofiler/profilers/profiler_options.py
```diff
@@ -695,14 +695,15 @@ def _validate_helper(self, variable_path='OrderOptions'):
...
-    def __init__(self):
+    def __init__(self, is_enabled=True, top_k_categories=None):
         """
         Options for the Categorical Column
 
         :ivar is_enabled: boolean option to enable/disable the column.
         :vartype is_enabled: bool
         """
-        BaseInspectorOptions.__init__(self)
+        BaseInspectorOptions.__init__(self, is_enabled=True)
```
ChrisWallace2020, line 705: You take `is_enabled` in as a parameter with a default of `True`, but then just pass `True` anyway, meaning that the user could never have it be `False` via construction, make sure to pass the `is_enabled` given by the user (optionally) here instead.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -11,7 +11,7 @@
 
 from dataprofiler.profilers import CategoricalColumn
 from dataprofiler.profilers.profile_builder import StructuredColProfiler
-
+from ...profilers.profiler_options import CategoricalOptions
```
ChrisWallace2020, line 14: Don't do relative imports in test modules, use absolute here. Relative imports need to be used in the code itself though, but tests should exist separate of the library, not as if they live within the library, if that makes sense.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -375,6 +373,32 @@ def test_unalikeability(self):
...
+        self.assertEqual(
+            len(profile.profile['statistics']['categorical_count']), 4)
+
+        options.top_k_categories = 7
+        df_series = pd.Series(["a", "c", "b", "c", "d"])
+        profile = CategoricalColumn(df_series.name, options)
+        profile.update(df_series)
+        self.assertEqual(
+            len(profile.profile['statistics']['categorical_count']), 4)
+
```
JGSweets, line 401: ideally these should have comments explaining what each assert is testing so a user knows the intent:
1. `None`
2. top_k < num cat
3. top_k > num_cat

Is the 4th a duplicate of 3? if so we can remove, no?

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -283,12 +283,10 @@ def test_categorical_merge(self):
...
-        expected_dict = dict({'aa': 5, '2': 4, 'abcd': 4, 'b': 3, np.nan: 2})
-        self.assertEqual(len(report_count), len(expected_dict))
-        self.assertEqual(report_count['aa'], expected_dict['aa'])
-        self.assertEqual(report_count['2'], expected_dict['2'])
-        self.assertEqual(report_count['abcd'], expected_dict['abcd'])
-        self.assertEqual(report_count['b'], expected_dict['b'])
+        expected_dict = {'aa': 5, '2': 4, 'abcd': 4, 'b': 3, np.nan: 2,
+                         'dfd': 2, '3': 2, '4': 2, 'ee': 2, 'null': 1, 'ff': 1,
+                         'NaN': 1, '1': 1, 'nan': 1, 'gg': 1, 'None': 1}
+        self.assertCountEqual(report_count, expected_dict)
```
JGSweets, line 289: will approve, but for future, the tests should be unless the assert specifically says otherwise in the docs `self.assert...(expected, actual)`

JGSweets (approved):

ChrisWallace2020 (approved):

PR #327, opened 2021-07-13
Added CSVData tests for io streams

gautomdas:
Added the tests to compare CSVData functionality with streams


dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -309,6 +429,60 @@ def test_set_header(self):
...
+
+            # set bad header setting
+            options = dict(header='abcdef')
+            with self.assertRaisesRegex(ValueError,
+                                        '`header` must be one of following: auto, '
+                                        'none for no header, or a non-negative '
+                                        'integer for the row that represents the '
+                                        'header \(0 based index\)'):
+                csv_data = CSVData(filename, options=options)
+                first_value = csv_data.data.loc[0][0]
```
JGSweets, line 456: We can remove this as it is covered in the other tests. Additionally, if we did want these tests, I would suggest not putting them in the for loop as we would only need to test it once.

dataprofiler/tests/data_readers/test_csv_data.py
```diff
@@ -208,6 +208,31 @@ def test_specifying_data_type(self):
...
+                byte_string = BytesIO(fp.read())
+                input_data_obj = Data(byte_string, data_type='csv')
+                self.assertEqual(input_data_obj.data_type, 'csv')
+                self.assertEqual(input_data_obj.data_type, 'csv', input_file["path"])
+                self.assertEqual(input_data_obj.delimiter,
+                                input_file['delimiter'],
+                                input_file["path"])
+
+            with open(input_file['path'], 'r', encoding=input_file['encoding']) as fp:
+                byte_string = StringIO(fp.read())
```
JGSweets, line 227: not a blocker for approval, but `byte_string` maybe should be `buffer` or `stream`

JGSweets (commented):
I think this is fine for now, but I think we should refactor this immediately after to clean up tests and set the future precedent.

Essentially, wether it is read from the file or from the buffer, they should be identical meaning the tests can be used the same.

Something like this:
```python
def setUpClass(cls):
    .
    .
    .

      
    cls.buffer_list = []
    for input_file in cls.input_file_names:
        #
        # Create stream here
        # buffer = ...
        buffer_info = input_file.copy()
        buffer_info['path'] = buffer
        cls.buffer_list.append(buffer_info)
        )
        
.
.
.
# then in tests also loop through the buffer list with existing tests not made specifically for streams
```

Again, not saying it should be this PR. We can refactor directly after.

JGSweets (approved):

ChrisWallace2020 (approved):

PR #336, opened 2021-07-14
Data Labeler Compiler Diff

gme5078:
Pretty straightforward.
Might be strange that we have a for loop for one profile, but in the future that may change and it is consistent with the profile property.
Took a page out of Andrew's book for mocking.


dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -328,6 +328,35 @@ def profile(self):
...
+            if key in self._profiles and key in other._profiles:
+                diff_data_label = utils.find_diff_of_strings_and_bools(
+                    self._profiles[key].data_label, 
+                    other._profiles[key].data_label)
+                diff_profile["data_label"] = diff_data_label
+                diff = self._profiles[key].diff(other._profiles[key], options)
+                diff_profile["statistics"].update(diff)
+        
+        if diff_profile["statistics"] == {}:
+            diff_profile.pop("statistics")
```
JGSweets, line 357: not needed change for approval, but if we were to match the style above, `if not diff_profile["statistics"]` as oppose tot he `== {}`

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -269,6 +269,80 @@ def test_compiler_stats_diff(self):
...
+                        'b': -0.7,
+                        'c': 0.7
+                    },
+                    'label_representation': {
+                        'a': -0.84, 
+                        'b': 'unchanged',
+                        'c': 0.84
+                    }
+                },
+                'data_label': ['a', 'b']
```
JGSweets, line 330: why is this `data_label` different than the one in statistics? Also, in profile do we have `data_label` at both levels? Should this instead pop it from the statistics level and reinsert it at the top level?

gme5078: good catch, diff in the datalabeler profie is not supposed to return data label

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -233,6 +233,7 @@ def profile(self):
         Property for profile. Returns the profile of the column.
         """
         profile = {
+            "data_label": self.data_label,
```
gme5078, line 236: @JGSweets @AnhTruong I've made a small refactor to put the data_label in the profile instead of having it extracted from the compiler level. Was there a specific reason the data_label was left out of the profile? Is this refactor inappropriate?

JGSweets: I think this is appropriate.

The update: `profile["data_label"] = col_profile.pop("data_label")` completes the refactor.

JGSweets (approved):

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -325,9 +325,36 @@ def profile(self):
...
+            profile["data_label"] = col_profile.pop("data_label")
+            profile["statistics"].update(col_profile)
         return profile
+    
+    def diff(self, other, options=None):
+        """
+        Finds the difference between 2 compilers and returns the report
+
+        :param other: profile compiler finding the difference with this one.
+        :type other: ColumnDataLabelerCompiler
```
AnhTruong, line 338: can you add the descriptions for the `options`?

dataprofiler/profilers/column_profile_compilers.py
```diff
@@ -325,9 +325,36 @@ def profile(self):
...
+        :return: difference of the profiles
+        :rtype: dict
+        """
+        # Call super for compiler instance check
+        diff_profile = super().diff(other, options)
+        diff_profile["statistics"] = dict()
+
+        # Iterate through profile(s)
+        all_profiles = set(self._profiles.keys()) | set(other._profiles.keys())
```
AnhTruong, line 347: can we just do `all_profiles = set(self._profiles.keys()) & set(other._profiles.keys())`, then we don't need the if statement down below inside the for loop?

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -269,6 +269,79 @@ def test_compiler_stats_diff(self):
...
+        options = StructuredOptions()
+        options.data_labeler.is_enabled = False
+        compiler1 = col_pro_compilers.ColumnDataLabelerCompiler(data, options)
+        expected_diff = {}
+        self.assertDictEqual(expected_diff, compiler1.diff(compiler2))
+
+        # Test disabling both datalabeler profiles for compiler diff
+        compiler2 = col_pro_compilers.ColumnDataLabelerCompiler(data, options)
+        expected_diff = {}
+        self.assertDictEqual(expected_diff, compiler1.diff(compiler2))
```
AnhTruong, line 343: can we add a test with one profile is empty, then the result should be the other profile?

AnhTruong (commented):
some comments

dataprofiler/tests/profilers/test_column_profile_compilers.py
gme5078: The result wont be the other profile, it would be empty, since you cant have a difference unless you have 2 profiles to compare

JGSweets (approved):

AnhTruong (approved):

PR #337, opened 2021-07-15
Move profile to numeric stats mixin

AnhTruong:
Others inherited class will call the profile from numeric stat mixin:
- float
- int
- text


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -247,6 +247,28 @@ def _add_helper(self, other1, other2):
...
             self.num_negatives = other1.num_negatives + other2.num_negatives
 
+    def profile(self):
+        """
+        Property for profile. Returns the profile of the column.
+        :return:
+        """
+
+        profile = dict(
+            min=self.np_type_to_type(self.min),
```
JGSweets, line 257: since `num_zeros` and `num_negatives` are calculated here in numericstats, I think they should also be. part of the profile here too. Text which doesn't have it can simply pop them.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -131,21 +131,10 @@ def profile(self):
...
-            min=self.np_type_to_type(self.min),
-            max=self.np_type_to_type(self.max),
-            sum=self.np_type_to_type(self.sum),
-            mean=self.np_type_to_type(self.mean),
-            variance=self.np_type_to_type(self.variance),
-            stddev=self.np_type_to_type(self.stddev),
-            skewness=self.np_type_to_type(self.skewness),
-            kurtosis=self.np_type_to_type(self.kurtosis),
+        profile = NumericStatsMixin.profile(self)
+        additive_profile = dict(
```
JGSweets, line 135: i would suggest a naming change or not having an intermediary variable. Instead, just put the dict directly inside of `update`:
e.g. `profile.update({...})` 
or for var name change `float_specific_profile`.

JGSweets: same for below vars

JGSweets: i can see the naming intent for abstraction, but not sure the name itself makes sense to me.
The other way to do it would be to reverse the roles and do:
```
profile = {...}
profile.update(NumericStatsMixin.profile(self))
```

AnhTruong: I've changed it to 
- update from numericstatsmixin
- update with the specific stats using profile.udpate

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: we can discuss about this. IMO, we don't want to update first and pop out later. That's why I don't update these two stats in numeric, but only update them for float and int

dataprofiler/profilers/text_column_profile.py
```diff
@@ -64,21 +64,10 @@ def profile(self):
...
-            skewness=self.skewness,
-            kurtosis=self.kurtosis,
-            histogram=self._get_best_histogram_for_profile(),
-            quantiles=self.quantiles,
-            vocab=self.vocab,
-            times=self.times
-        )
+        profile = NumericStatsMixin.profile(self)
+        profile.pop('num_zeros')
+        profile.pop('num_negatives')
```
JGSweets, line 69: should add a comment here

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -503,6 +503,64 @@ def test_merge_num_zeros_and_negatives(self):
...
+
+        num_profiler.match_count = 0
+        num_profiler.min = mock_profile['min']
+        num_profiler.max = mock_profile['max']
+        num_profiler.sum = mock_profile['sum']
+        num_profiler.histogram_selection = 'auto'
+        num_profiler.histogram_methods['auto']['histogram'] = \
+            mock_profile['histogram']
+        num_profiler.quantiles = mock_profile['quantiles']
+        num_profiler.times = mock_profile['times']
```
JGSweets, line 540: this seems strange to me that we are manually setting the values and just checking if the profile returns correctly.

I would rather send actual data into the profiler and ensure the calcs / results out are correct.

AnhTruong: To test if the profile works, I think assigning properties and checking them later is not bad. At least, we did that for other tests for numeric profile as well, using TestColumn(). If we want to send the data, we have to run it under int/float/text.

gme5078 (approved):

JGSweets (approved):

PR #338, opened 2021-07-15
Add correlation update functionality with NaNs - "Complete" row removal

andrew-yin:
Previously, any column that contains NaNs will not have correlation computed on it. This update enables correlations to be calculated and updated even if numeric columns contain NaNs.

The method of handling NaNs is a "complete" removal of the entire row that contains NaN values. This guarantees a positive semi-definite correlation matrix. Let me know what you think about this implementation.


dataprofiler/profilers/profile_builder.py
```diff
@@ -1431,53 +1436,61 @@ def _update_row_statistics(self, data, sample_ids=None):
...
-                else:
+                if self._profile[idx].profile['data_type'] in ['int', 'float']:
                     clean_column_ids.append(idx)
 
-        data = pd.DataFrame(clean_samples)
-        data = data.apply(pd.to_numeric, errors='coerce')
+        # Exclude all rows with NaN in numeric columns
+        # and add dropped-NaN stats to properties for
+        # future updating
+        data_dropped = data.dropna(subset=data.columns[clean_column_ids])
```
JGSweets, line 1456: NA values dropped should have already been determined via `clean_data_and_get_base_stats` which is called in the update chunk method prior to this function. Can we pass that info as well so that we don't recalculate?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1431,53 +1436,61 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+        properties = {
+            'mean': np.full(len(self._profile), np.nan),
+            'std': np.full(len(self._profile), np.nan)
+        }
+        for id in clean_column_ids:
+            properties['mean'][id] = data_dropped.iloc[:, id].mean()
+            properties['std'][id] = data_dropped.iloc[:, id].std()
+        properties['count'] = len(data_dropped.index)
```
JGSweets, line 1464: Instead of recalculating mean / std we should utilize the pre-calculated values and do an update to them specifically for these alterations due to NaNs. We want to reduce as many repeat calculations as possible. 

Additionally, we need to decide whether the NaNs apply across the dataset or within col to col interactions.

dataprofiler/profilers/profile_builder.py
andrew-yin: @JGSweets I thought that does it on a per-column basis. In this case, if one column has a NaN, the rest of the columns should have their corresponding values dropped as well even if there's non-null values. For example, if `col1=[NaN, 3, 5]` and `col2=[1,2,3]` , we still want to drop the `1` from `col2` since there's a `NaN` in `col1`.

dataprofiler/profilers/profile_builder.py
andrew-yin: The concern I saw with updating the given values is that the values that need to be removed from the mean calculation are dependent on the other columns containing NaN values. Searching for which values need to be removed and then updating the mean seemed like an equally expensive operation to me. 

In regards to applying NaN removal to the whole dataset vs pairwise, it becomes quite complex of a process to update correlations pairwise with NaNs, because then each column could potentially have different "means/stddevs" depending on what indices are NaN in one column vs another, and would require needing to find a way to store multiple different values for means and standard deviations for different columns. Thoughts? @JGSweets

dataprofiler/profilers/profile_builder.py
AnhTruong: Agree with Andrew. with this option of nan processing for correlation, I don't think we can use the clean samples obtained from the previous step, as the columns' indices are non-overlap

AnhTruong (commented):
Can we figure out a way to reuse the calculated mean/var/std? Also, we should add an option on how we want to calculate the correlation matrix with nan values. IMO, we should cover at least three options: 
- pairwise correlations where we remove any row with nan for any pair of correlations (pandas provides this)
- removing any row with nan values (this PR)
- imputation nan using mean/median/mode

AnhTruong:
Another option is ignoring the columns with nan values as we're having now.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1163,6 +1163,11 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0,
         self._profile = []
         self._col_name_to_idx = defaultdict(list)
         self.correlation_matrix = None
+        self._corr_prev_properties = {
```
AnhTruong, line 1166: if we decide to use this, should it be saved with the profiler so that later on we will load it and potentially merge with other profile?

andrew-yin:
Converting to a draft for the time being as a result of more discussion about how we should deal with NaNs

lettergram:
@andrew-yin Where are we with this?

andrew-yin:
@lettergram An alternative implementation of this (#342) has already been merged. I've left this up as we decided some more discussion is needed on whether the update method should be an option that the user can choose, but additional testing of the overall speed of the correlation calculation is needed as well. Should I close this PR for the time being?

JGSweets:
Closing for now until we can re-review.

PR #339, opened 2021-07-15
Unstructured Text Diff

gme5078:
Okay, so I'm calling this a work in progress because I would like to discuss several things:

1. When getting the difference in word count, do we look at comparing exclusively the dict of the top k words or the entire dict (same for vocab)?
2. When we decide the top_k_values, do we choose self.top_k_values (for the same rationale that we have an order of profile1 - profile2, where only one profile can be subtracted) or the max between the two profiles?
3. Should the difference of dicts return actually a list of 3 dicts. [profile1_key:value, shared_key:diff, profile2_key:value] or should we stick with the earlier status quo?

This PR has my suggestions, but I'd like input.

### DIFF
--- a/dataprofiler/profilers/unstructured_text_profile.py
+++ b/dataprofiler/profilers/unstructured_text_profile.py
@@ -216,6 +216,44 @@ class TextProfiler(object):
         merged_profile.sample_size = self.sample_size + other.sample_size
 
         return merged_profile
+    
+    def diff(self, other_profile, options=None):
+        """
+        Finds the differences for two unstructured text profiles
+
+        :param other_profile: profile to find the difference with
+        :type other_profile: TextProfiler
+        :param options: options for diff output
+        :type options: dict
+        :return: the difference between profiles
+        :rtype: dict
+        """
+        cls = self.__class__
+        if not isinstance(other_profile, cls):
+            raise TypeError("Unsupported operand type(s) for diff: '{}' "
+                            "and '{}'".format(cls.__name__,
+                                              other_profile.__class__.__name__))
+
+        diff = {}
+
+        diff["vocab"] = utils.find_diff_of_lists_and_sets(
+            list(self.vocab_count.keys()), list(other_profile.vocab_count.keys())
+        )
+
+        diff["vocab_count"] = utils.find_diff_of_dicts_with_different_keys(
+            dict(self.vocab_count.most_common(self._top_k_chars)),
+            dict(other_profile.vocab_count.most_common(self._top_k_chars)))
+
+        diff["words"] = utils.find_diff_of_lists_and_sets(
+            list(self.word_count.keys()),
+            list(other_profile.word_count.keys())
+        )
+        
+        diff["word_count"] = utils.find_diff_of_dicts_with_different_keys(
+            dict(self.word_count.most_common(self._top_k_words)),
+            dict(other_profile.word_count.most_common(self._top_k_words)))
+
+        return diff
 
     @property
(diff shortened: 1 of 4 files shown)


JGSweets:
To me, if the user specified a Top k, we should only show the Top k and it should be the`[profile1 not in proifle2, shared, profile2 not in profile1]`

JGSweets (approved):

dataprofiler/tests/profilers/test_utils.py
```diff
@@ -124,3 +124,19 @@ def test_find_diff(self):
             "g": [None, 15]
         }
         self.assertDictEqual(expected_diff, utils.find_diff_of_dicts(dict1, dict2))
+
+    def test_diff_of_dicts_with_diff_keys(self):
+        dict1 = {"unique1": 1, "shared1": 2, "shared2": 3}
```
JGSweets, line 129: probably should validate the other conditions that exist in the function being tested

dataprofiler/tests/profilers/test_unstructured_text_profile.py
```diff
@@ -139,6 +139,32 @@ def test_merge_profiles(self):
...
+        text_profile2.update(sample)
+        
+        expected_diff = {
+            'vocab': [['H', 'l', 'm', 'y', ':', '.', '!'],
+                      ['e', 'o', ' ', 'n', 'a', 'i', 's', 'G', 'r', 't'],
+                      ['B', 'b', 'd', '"', ',', "'", 'f']],
+            'vocab_count': [
+                {'!': 3, 'l': 2, 'm': 2, 'H': 1, 'y': 1, ':': 1, '.': 1},
+                {' ': -2, 'e': 'unchanged', 'n': -3, 'a': -3, 'o': 'unchanged',
+                 'i': 'unchanged', 's': 'unchanged', 'G': -2, 'r': -4, 't': -2},
```
AnhTruong, line 158: should we make `unchanged` as `0`, so the type is consistent with others in this dict?

dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -216,6 +216,43 @@ def __add__(self, other):
...
+                                              other_profile.__class__.__name__))
+
+        diff = {}
+
+        diff["vocab"] = utils.find_diff_of_lists_and_sets(
+            list(self.vocab_count.keys()),
+            list(other_profile.vocab_count.keys()))
+
+        diff["vocab_count"] = utils.find_diff_of_dicts_with_diff_keys(
+            dict(self.vocab_count.most_common(self._top_k_chars)),
```
AnhTruong, line 244: I'd prefer using the whole set instead of selecting the top ones. Otherwise, it'd be not consistent with the diff['vocab']

AnhTruong (commented):
some small comments and one question: what should we do when two profiles have different case sensitive options?

dataprofiler/tests/profilers/test_unstructured_text_profile.py
gme5078: This is consistent with the rest of differences

dataprofiler/profilers/unstructured_text_profile.py
gme5078: We dont keep track of the top k vocab so it cant be consistent

AnhTruong (approved):

JGSweets (approved):

AnhTruong (approved):

PR #342, opened 2021-07-16
Add correlation update with NaNs - Impute with mean

andrew-yin:
This PR enables correlation statistics to be updated with NaNs by replacing the NaN values with the mean of the batch of data the value belongs to.

This is an alternative/more computationally efficient method of handling NaNs in correlation update compared to #338.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1447,8 +1447,9 @@ class StructuredProfiler(BaseProfiler):
                 else:
                     clean_column_ids.append(idx)
 
-        data = pd.DataFrame(clean_samples)
-        data = data.apply(pd.to_numeric, errors='coerce')
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors='coerce')
+        means = {index:mean for index, mean in enumerate(batch_properties['mean'])}
+        data = data.fillna(value=means)
 
         # fill correlation matrix with nan initially
         n_cols = len(self._profile)
@@ -1464,12 +1465,15 @@ class StructuredProfiler(BaseProfiler):
                     corr_mat[id1][id2] = 1
                     continue
 
-                mean1 = batch_properties['mean'][id1]
-                mean2 = batch_properties['mean'][id2]
+                col1 = data.loc[:, id1]
+                col2 = data.loc[:, id2]
                 std1 = batch_properties['std'][id1]
                 std2 = batch_properties['std'][id2]
-                corr = utils.get_pairwise_corr(data.loc[:, id1], data.loc[:, id2],
-                                               mean1, mean2, std1, std2)
+
+                comoment = ((col1 * col2).sum() \
+                            - col1.sum() * col2.sum() / batch_properties['count'])
+                corr = (comoment / (batch_properties['count'] - 1)) / (std1 * std2)
+
                 corr_mat[id1][id2] = corr
                 corr_mat[id2][id1] = corr
 
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1437,6 +1437,9 @@ class StructuredProfiler(BaseProfiler):
 
         :param clean_samples: the input cleaned dataset
         :type clean_samples: dict()
+        :param batch_properties: mean/std/counts of each batch column necessary
+        for correlation computation
+        :type batch_properties: dict()
         """
         columns = self.options.correlation.columns
         clean_column_ids = []
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1468,14 +1468,14 @@ class StructuredProfiler(BaseProfiler):
         data = data.fillna(value=means)
 
         # Update the counts/std if needed (i.e. if null rows or exist)
-        if (len(data.index) != batch_properties['count']).any():
+        if (len(data) != batch_properties['count']).any():
             adjusted_stds = np.sqrt(
                 batch_properties['std']**2 * (batch_properties['count'] - 1) \
-                / (len(data.index) - 1)
+                / (len(data) - 1)
             )
             batch_properties['std'] = adjusted_stds
         # Set count key to a single number now that everything's been adjusted
-        batch_properties['count'] = len(data.index)
+        batch_properties['count'] = len(data)
 
         # fill correlation matrix with nan initially
         n_cols = len(self._profile)
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1442,7 +1442,9 @@ class StructuredProfiler(BaseProfiler):
         clean_column_ids = []
         if columns is None:
             for idx in range(len(self._profile)):
-                if self._profile[idx].profile['data_type'] not in ['int', 'float']:
+                data_type = self._profile[idx].\
+                    profiles["data_type_profile"].selected_data_type
+                if data_type not in ["int", "float"]:
                     clean_samples.pop(idx)
                 else:
                     clean_column_ids.append(idx)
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/utils.py
```diff
@@ -483,3 +483,33 @@ def find_diff_of_dicts(dict1, dict2):
...
+    :rtype: float
+    """
+    covariance = 0
+    for i1, i2 in zip(data1, data2):
+        if (not np.isnan(i1)) and (not np.isnan(i2)):
+            a = i1 - mean1
+            b = i2 - mean2
+            covariance += a * b
+
+    return (covariance / (len(data1.index) - 1)) / (std1 * std2)
```
JGSweets, line 515: is this not still ignoring the data points which have NaNs.

In otherwords if a value in a row is NaN for col1, but a real value in col2, shouldn't it be `mean1 * i2`?

Also, wouldn't std1 change as a result?

JGSweets: Additionally, I think instead of looping ourselves, we might be able to utilize vectorization to do this more quickly.
e..g
1. replace all Nans (in a copy of the dataset)
2. compute: `sum(fixed_column1 * fixed_column2)`
3. compute: updated stddev for columns which had updated nans (not actual std of new column, just updating relative to nan values (if possible) maybe this is expensive?

Thoughts?

andrew-yin: The function works as expected with handling means. The covariance can be defined as `sum((x_i - mean_x)(y_i - mean_y))/n`. So, if there's a NaN value and it's imputed to the mean of one of the columns, that term in the summation is zero and can be ignored when looping through no matter what the value in the other column is.

I agree that there is some confusion on std that could be addressed. I calculate the updated std prior to calling the function, so the function itself already receives an updated std. It would make more sense to update the std within this function for added clarity and improve the flexibility of this function, so I will make those changes.

Vectorization is also an idea that can definitely be implemented to improve the speed of the function. I'll play around with this.

JGSweets: If we know the location of the NaNs from the previous calculations, could we technically drop them without looping and then do the multiplication?

JGSweets: That way we don't have to call `np.isnan` again, this value may not be representative of all which the user want's considered a `NaN`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1431,7 +1431,7 @@ def _update_row_statistics(self, data, sample_ids=None):
             self.row_has_null_count = len(null_in_row_count)
             self.row_is_null_count = len(null_rows)
 
-    def _get_correlation(self, clean_samples):
+    def _get_correlation(self, clean_samples, batch_properties):
```
AnhTruong, line 1434: please add descriptions for batch_properties

dataprofiler/profilers/profile_builder.py
```diff
@@ -1442,14 +1442,16 @@ def _get_correlation(self, clean_samples):
...
+                if data_type not in ["int", "float"]:
                     clean_samples.pop(idx)
                 else:
                     clean_column_ids.append(idx)
 
-        data = pd.DataFrame(clean_samples)
-        data = data.apply(pd.to_numeric, errors='coerce')
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors='coerce')
+        means = {index:mean for index, mean in enumerate(batch_properties['mean'])}
+        data = data.fillna(value=means)
```
AnhTruong, line 1454: does this apply at the same nan indices detected by our DP?

AnhTruong (commented):
some comments

dataprofiler/profilers/profile_builder.py
andrew-yin: Yes, since the means array in batch properties is lined up with the column indices

dataprofiler/profilers/profile_builder.py
```diff
@@ -1442,25 +1442,40 @@ def _update_row_statistics(self, data, sample_ids=None):
...
+
+        # Update the counts/std if needed (i.e. if null rows or exist)
+        if (len(data.index) != batch_properties['count']).any():
+            adjusted_stds = np.sqrt(
+                batch_properties['std']**2 * (batch_properties['count'] - 1) \
+                / (len(data.index) - 1)
+            )
+            batch_properties['std'] = adjusted_stds
+        # Set count key to a single number now that everything's been adjusted
+        batch_properties['count'] = len(data.index)
```
JGSweets, line 1478: any reason specifically of the index?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -468,10 +519,48 @@ def test_correlation_update(self):
...
+        #                     [5, 3, 7, 6, 8, 1]
+        expected_corr_mat = np.array([
+            [1, -0.06987956, 0.32423975],
+            [-0.06987956, 1, -0.3613099],
+            [0.32423975, -0.3613099, 1]
+        ])
+        np.testing.assert_array_almost_equal(expected_corr_mat,
+                                             profile.correlation_matrix)
+
+        # Data with null rows and some imputed values
```
JGSweets, line 549: nice

dataprofiler/profilers/profile_builder.py
andrew-yin: I'm not sure what you quite mean, but `data.index` returns all the row indices, so the length of that is the number of total rows in the cleaned data excluding fully NaN rows but including partially NaN rows.

dataprofiler/profilers/profile_builder.py
AnhTruong: I meant the row indices, not the column indices. As we added a lot of nan indices detection, I'm just wondering if those detected indices are consistent with the indices used with `fillna` ?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -417,6 +449,24 @@ def test_merge_correlation(self, *mocks):
...
+        expected_corr_mat = np.array([
+            [1, -0.78571429],
+            [-0.78571429,  1]
+        ])
+        np.testing.assert_array_almost_equal(expected_corr_mat,
+                                             profile1.correlation_matrix)
+        profile2 = dp.StructuredProfiler(data2, options=profile_options)
+        merged_profile = profile1 + profile2
+        np.testing.assert_array_almost_equal(expected_corr_mat,
+                                             merged_profile.correlation_matrix)
```
AnhTruong, line 468: should it be None in this case?

dataprofiler/profilers/profile_builder.py
JGSweets: cleaned_samples has those removed, hence the `na` here is for covering when the columns are merged and na fills in the empty locations

dataprofiler/profilers/profile_builder.py
JGSweets: as oppose to `len(data)`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -468,10 +519,52 @@ def test_correlation_update(self):
...
+            [-0.06987956, 1, -0.3613099],
+            [0.32423975, -0.3613099, 1]
+        ])
+        np.testing.assert_array_almost_equal(expected_corr_mat,
+                                             profile.correlation_matrix)
+
+        # Data with null rows and some imputed values
+        data = pd.DataFrame({'a': [None, np.nan, 1, 7, 5, 9, 4, 10, np.nan, 2],
+                             'b': [10, 11, 1, 4, 2, 5, 'NaN', 3, np.nan, 8],
+                             'c': [1, 5, 3, 5, np.nan, 2, 6, 8, np.nan, 2]})
```
AnhTruong, line 552: can we also add the None value here as well?

dataprofiler/profilers/profile_builder.py
andrew-yin: this works better, forgot you could do this, i'll make those changes

dataprofiler/tests/profilers/test_profile_builder.py
andrew-yin: I'm not sure. The second profile has `total_samples > 0`, but no actual values. Should the correlation matrix exist for data in this case? What do you think? @AnhTruong

JGSweets: I think this is correct. It isn't that the values are preventing the calculation, it's that no data can be used to calculate.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -468,10 +519,52 @@ def test_correlation_update(self):
...
+
+        # Data with null rows and some imputed values
+        data = pd.DataFrame({'a': [None, np.nan, 1, 7, 5, 9, 4, 10, 'nan', 2],
+                             'b': [10, 11, 1, 4, 2, 5, 'NaN', 3, None, 8],
+                             'c': [1, 5, 3, 5, np.nan, 2, 6, 8, None, 2]})
+        data1 = data[:5]
+        data2 = data[5:]
+        profile = dp.StructuredProfiler(data1, options=profile_options)
+        profile.update_profile(data2)
+        # correlation between [*13/3*, *13/3*, 1, 7, 5]
```
JGSweets, line 557: good

dataprofiler/profilers/profile_builder.py
```diff
@@ -1442,25 +1442,40 @@ def _update_row_statistics(self, data, sample_ids=None):
...
                     clean_column_ids.append(idx)
 
-        data = pd.DataFrame(clean_samples)
-        data = data.apply(pd.to_numeric, errors='coerce')
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors='coerce')
+        means = {index:mean for index, mean in enumerate(batch_properties['mean'])}
+        data = data.fillna(value=means)
+
+        # Update the counts/std if needed (i.e. if null rows or exist)
+        if (len(data) != batch_properties['count']).any():
```
JGSweets, line 1471: small optimization, but we do `len(data)` a lot in this function, could we just set it to a var instead?

Won't block PR.

JGSweets (approved):

AnhTruong (approved):

PR #346, opened 2021-07-19
Add timing of profiles to StructuredProfiler and UnstructuredProfiler

JGSweets:
Adds timeit to `StructuredProfiler` and `UnstructuredProfiler` for the following calculations in each.
StructuredProfiler: correlations, row stats
UnstructuredProfiler: getting null values and base stats

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -487,6 +487,7 @@ class BaseProfiler(object):
         self._samples_per_update = samples_per_update
         self._min_true_samples = min_true_samples
         self.total_samples = 0
+        self.times = defaultdict(float)
 
         # TODO: allow set via options
         self._sampling_ratio = 0.2
@@ -546,6 +547,9 @@ class BaseProfiler(object):
             merged_profile.file_type = 'multiple files'
 
         merged_profile.total_samples = self.total_samples + other.total_samples
+
+        merged_profile.times = utils.add_nested_dictionaries(self.times,
+                                                             other.times)
 
         return merged_profile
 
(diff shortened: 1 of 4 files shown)


dataprofiler/tests/profilers/utils.py
```diff
@@ -132,3 +133,17 @@ def clean_report(report):
...
+    Creates a mock for the time.time function that increments the time for
+    every call.
+    """
+    def increment_counter():
+        counter = 0
+        while True:
+            counter += 1
+            yield counter
+    counter = increment_counter()
+    return mock.patch('time.time', side_effect=lambda: next(counter))
```
gme5078, line 149: nifty.

dataprofiler/profilers/profile_builder.py
```diff
@@ -970,15 +974,15 @@ def report(self, report_options=None):
...
             }),
             ("data_stats", OrderedDict()),
         ])
-
         report["data_stats"] = self._profile.profile
         return _prepare_report(report, output_format, omit_keys)
 
-    @staticmethod
-    def _clean_data_and_get_base_stats(data, sample_size,
+    @utils.method_timeit(name="clean_and_base_stats")
```
gme5078, line 984: I'm approving, but I dont think the average user will understand clean_and_base_stats

gme5078 (approved):

dataprofiler/profilers/profile_builder.py
JGSweets: i'm open to other names, i wasn't really sure what to put.

AnhTruong: may be `get_clean_and_base_stats `

dataprofiler/profilers/profile_builder.py
```diff
@@ -1490,6 +1498,7 @@ def _update_correlation(self, clean_samples, prev_dependent_properties):
             batch_corr, batch_properties["mean"],
             batch_properties["std"], batch_samples)
 
+    @utils.method_timeit(name='correlation')
```
AnhTruong, line 1501: should we make the name differently as we had correlation above already?
Also, how we name other stats timing to distinguish between update and merge?

JGSweets: we don't separate the cost between merge / update of correlations. All together, they are the cost of computing the statistic.

dataprofiler/profilers/profile_builder.py
JGSweets: I'm not sure that makes it more clear to the user as the intent is to label the cost of what is being calculated but not just the function

ChrisWallace2020 (approved):

gme5078 (approved):

ChrisWallace2020 (approved):

PR #352, opened 2021-07-21
Added JSON tests and fixed _load_data_from_file

gautomdas:
Consolidated JSON tests and made them work.

### DIFF
--- a/dataprofiler/data_readers/filepath_or_buffer.py
+++ b/dataprofiler/data_readers/filepath_or_buffer.py
@@ -43,6 +43,11 @@ class FileOrBufferHandler:
                 TextIOWrapper(self._filepath_or_buffer, encoding=self._encoding)
             self._is_wrapped = True
 
+        elif isinstance(self._filepath_or_buffer, StringIO) \
+                and self.open_method == 'rb':
+            self._filepath_or_buffer = \
+                BytesIO(self._filepath_or_buffer.read().encode())
+
         elif not data_utils.is_stream_buffer(self._filepath_or_buffer):
             # Raise AttributeError if attribute value not found.
             raise AttributeError(f'Type {type(self._filepath_or_buffer)} is '
--- a/dataprofiler/data_readers/filepath_or_buffer.py
+++ b/dataprofiler/data_readers/filepath_or_buffer.py
@@ -43,11 +43,6 @@ class FileOrBufferHandler:
                 TextIOWrapper(self._filepath_or_buffer, encoding=self._encoding)
             self._is_wrapped = True
 
-        elif isinstance(self._filepath_or_buffer, StringIO) \
-                and self.open_method == 'rb':
-            self._filepath_or_buffer = \
-                BytesIO(self._filepath_or_buffer.read().encode())
-
         elif not data_utils.is_stream_buffer(self._filepath_or_buffer):
             # Raise AttributeError if attribute value not found.
             raise AttributeError(f'Type {type(self._filepath_or_buffer)} is '
--- a/dataprofiler/data_readers/json_data.py
+++ b/dataprofiler/data_readers/json_data.py
@@ -244,8 +244,14 @@ class JSONData(SpreadSheetDataMixin, BaseData):
         :type input_file_path: str
         :return:
         """
-        self._file_encoding = data_utils.detect_file_encoding(input_file_path)
-        with FileOrBufferHandler(input_file_path, 'r', encoding=self.file_encoding) as input_file:
+        self._file_encoding = None
+        if not isinstance(input_file_path, StringIO):
+            self._file_encoding = \
+                data_utils.detect_file_encoding(input_file_path)
+                
+        with FileOrBufferHandler(input_file_path, 'r', encoding=self.file_encoding) \
+            as input_file:
+
             try:
                 data = json.load(input_file)
             except (json.JSONDecodeError, UnicodeDecodeError):
--- a/dataprofiler/data_readers/json_data.py
+++ b/dataprofiler/data_readers/json_data.py
@@ -245,14 +245,13 @@ class JSONData(SpreadSheetDataMixin, BaseData):
         :return:
         """
         self._file_encoding = data_utils.detect_file_encoding(input_file_path)
-        with open(input_file_path, encoding=self.file_encoding) as input_file:
+        with FileOrBufferHandler(input_file_path, 'r', encoding=self.file_encoding) as input_file:
             try:
                 data = json.load(input_file)
             except (json.JSONDecodeError, UnicodeDecodeError):
                 input_file.seek(0)
-                data = data_utils.generator_on_file(input_file)
                 data = data_utils.read_json(
-                    data_generator=data,
+                    data_generator=input_file,
                     selected_columns=self.selected_keys,
                     read_in_string=False
                 )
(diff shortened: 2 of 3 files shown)


dataprofiler/tests/data_readers/test_json_data.py
```diff
@@ -45,6 +45,21 @@ def setUpClass(cls):
...
+            # add BytesIO
+            buffer_info = input_file.copy()
+            with open(input_file['path'], 'rb') as fp:
+                buffer_info['path'] = BytesIO(fp.read())
+            cls.buffer_list.append(buffer_info)
+
+        cls.file_or_buf_list = cls.input_file_names + cls.buffer_list
+
+    @classmethod
+    def setUp(self):
```
JGSweets, line 60: since this is a class method, self should actually be `cls`
e.g. `def setUp(cls)` as it is not an object instance but reference to the class

dataprofiler/tests/data_readers/test_json_data.py
```diff
@@ -45,6 +45,21 @@ def setUpClass(cls):
...
+        for buffer in self.buffer_list:
```
JGSweets, line 61: this will have to update from `self `to `cls` as well

dataprofiler/tests/data_readers/test_json_data.py
```diff
@@ -45,6 +45,21 @@ def setUpClass(cls):
...
             dict(path=os.path.join(test_dir, "json/honeypot_intentially_mislabeled_file.parquet"), encoding='utf-8', count=14)
         ]
+
+        cls.buffer_list = []
+        for input_file in cls.input_file_names:
+            # add BytesIO
+            buffer_info = input_file.copy()
+            with open(input_file['path'], 'rb') as fp:
+                buffer_info['path'] = BytesIO(fp.read())
+            cls.buffer_list.append(buffer_info)
```
JGSweets, line 55: Since JSON can handle StringIO as well, we should add that too.

dataprofiler/data_readers/json_data.py
```diff
@@ -245,14 +245,13 @@ def _load_data_from_file(self, input_file_path):
         :return:
         """
         self._file_encoding = data_utils.detect_file_encoding(input_file_path)
-        with open(input_file_path, encoding=self.file_encoding) as input_file:
+        with FileOrBufferHandler(input_file_path, 'r', encoding=self.file_encoding) as input_file:
             try:
                 data = json.load(input_file)
             except (json.JSONDecodeError, UnicodeDecodeError):
                 input_file.seek(0)
-                data = data_utils.generator_on_file(input_file)
```
JGSweets, line 253: If this function is not used anywhere else, we can remove it from the utils.

dataprofiler/tests/data_readers/test_json_data.py
```diff
@@ -45,6 +45,21 @@ def setUpClass(cls):
...
+
+        cls.file_or_buf_list = cls.input_file_names + cls.buffer_list
+
+    @classmethod
+    def setUp(self):
+        for buffer in self.buffer_list:
+            buffer['path'].seek(0)
     
     def test_is_match_for_string_streams(self):
```
JGSweets, line 64: can we make a generic `is_match` that runs through `self.file_or_buf_list`? Then we can get rid of the duplicate functions as well.
can just call it `test_is_match`

dataprofiler/data_readers/filepath_or_buffer.py
```diff
@@ -43,6 +43,11 @@ def __enter__(self):
                 TextIOWrapper(self._filepath_or_buffer, encoding=self._encoding)
             self._is_wrapped = True
 
+        elif isinstance(self._filepath_or_buffer, StringIO) \
```
JGSweets, line 46: in what case does this occur? This seems a bit backwards to then convert it to bytes. 

Maybe the code elsewhere is wrong, but `rb` + `StringIO` shouldn't occur in our code.

dataprofiler/data_readers/json_data.py
```diff
@@ -244,15 +244,20 @@ def _load_data_from_file(self, input_file_path):
         :type input_file_path: str
         :return:
         """
-        self._file_encoding = data_utils.detect_file_encoding(input_file_path)
-        with open(input_file_path, encoding=self.file_encoding) as input_file:
+        self._file_encoding = None
+        if not isinstance(input_file_path, StringIO):
+            self._file_encoding = \
+                data_utils.detect_file_encoding(input_file_path)
```
JGSweets, line 250: good

dataprofiler/data_readers/json_data.py
```diff
@@ -244,15 +244,20 @@ def _load_data_from_file(self, input_file_path):
...
+                
+        with FileOrBufferHandler(input_file_path, 'r', encoding=self.file_encoding) \
+            as input_file:
```
JGSweets, line 253: I would suggest maybe breaking it up as:
```python
with FileOrBufferHandler(input_file_path, 'r', 
                                           encoding=self.file_encoding) as input_file:
```
If that fits.

JGSweets: Otherwise, for reference when indenting, we would want the indented line over two indents rather than one in this case so that it isn't inline with the code below which may cause confusion.

dataprofiler/tests/data_readers/test_json_data.py
```diff
@@ -45,50 +45,66 @@ def setUpClass(cls):
...
+            if isinstance(input_file, StringIO):
+                with open(input_file['path'], 'r',
+                        encoding=input_file['encoding']) as fp:
+                    byte_string = StringIO(fp.read())
+                    self.assertTrue(JSONData.is_match(byte_string))
+
+            elif isinstance(input_file, BytesIO):
+                with open(input_file['path'], 'rb') as fp:
+                    byte_string = BytesIO(fp.read())
+                    self.assertTrue(JSONData.is_match(byte_string))
```
JGSweets, line 85: I think we can reduce this down to just 3 lines where we loop through and assert that is match is true for the is_match call on the object

dataprofiler/tests/data_readers/test_json_data.py
```diff
@@ -45,50 +45,57 @@ def setUpClass(cls):
...
+    def test_is_match(self):
         """
         Determine if the json file can be automatically identified from
         byte stream
         """
-        for input_file in self.input_file_names:
-            with open(input_file['path'], 'rb') as fp:
-                byte_string = BytesIO(fp.read())
-                self.assertTrue(JSONData.is_match(byte_string))
+        for input_file in self.file_or_buf_list:
```
JGSweets, line 75: good

dataprofiler/data_readers/json_data.py
JGSweets: Won't block on this, as it can be addressed later.

JGSweets (dismissed):

JGSweets (approved):

dataprofiler/data_readers/json_data.py
```diff
@@ -244,15 +244,19 @@ def _load_data_from_file(self, input_file_path):
         :type input_file_path: str
         :return:
         """
-        self._file_encoding = data_utils.detect_file_encoding(input_file_path)
-        with open(input_file_path, encoding=self.file_encoding) as input_file:
+        self._file_encoding = None
+        if not isinstance(input_file_path, StringIO):
```
ChrisWallace2020, line 248: Why is this just StringIO and not also for ByteIO? Perhaps add a comment explaining too.

JGSweets: Strings don't need decoding bc they have already been decoded.

ChrisWallace2020 (approved):

PR #355, opened 2021-07-21
Add t-test to numerical stats differences

andrew-yin:
Performs a two-sample, two-tailed t-test for difference of means between numerical columns.

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -339,8 +339,8 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         if n1 <= 1 or n2 <= 1:
             warnings.warn("Insufficient sample size to "
                           "perform t-test.", RuntimeWarning)
-            return "null"
-
+            return None
+        
         results = {
             't-statistic': None,
             'conservative': {
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -348,9 +348,10 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
             }
         }
 
-        if n1 <= 1 or n2 <= 1:
-            warnings.warn("Insufficient sample size to "
-                          "perform t-test.", RuntimeWarning)
+        if n1 <= 1 or n2 <= 1 or np.isnan([mean1, mean2, var1, var2]).any() \
+                or None in [mean1, mean2, var1, var2]:
+            warnings.warn("T-test could not be performed due "
+                "to invalid profile statistics.", RuntimeWarning)
             return results
 
         s_delta = var1/n1 + var2/n2
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -294,6 +294,9 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
                                                    other_profile.variance),
             "stddev": utils.find_diff_of_numbers(self.stddev,
                                                  other_profile.stddev),
+            "t-test": self._perform_t_test(self.mean, self.variance, self.match_count,
+                                           other_profile.mean, other_profile.variance,
+                                           other_profile.match_count)
         }
         return differences
 
@@ -329,6 +332,53 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
             else self._correct_bias_kurtosis(
                     self.match_count,
                     self._biased_kurtosis)
+
+    @staticmethod
+    def _perform_t_test(mean1, var1, n1,
+                        mean2, var2, n2):
+        results = {}
+        s_delta = var1/n1 + var2/n2
+        t = (mean1 - mean2) / np.sqrt(s_delta)
+        conservative_df = min(n1, n2) - 1
+        welch_df = s_delta ** 2 / ((var1 / n1) ** 2 / (n1 - 1) + (var2 / n2) ** 2 / (n2 - 1))
+        results.update({
+            "t-statistic": t,
+            "conservative_df": conservative_df,
+            "welch_df": welch_df
+        })
+        try:
+            import scipy.stats
+        except ImportError:
+            # Failed, so we return the stats but don't perform the test
+            warnings.warn("Could not import necessary statistical packages. "
+                          "T-test will be incomplete.", RuntimeWarning)
+            return results
+        # If scipy import was successful, now perform the *two-sided* t-test
+        conservative_t = scipy.stats.t(conservative_df)
+        conservative_p_val = (1 - conservative_t.cdf(abs(t))) * 2
+        welch_t = scipy.stats.t(welch_df)
+        welch_p_val = (1 - welch_t.cdf(abs(t))) * 2
+        results.update({
+            'results': {
+                "conservative": {
+                    'p-value': conservative_p_val
+                },
+                "welch": {
+                    'p-value': welch_p_val
+                }
+            }
+        })
+
+        alphas = [0.01, 0.05, 0.1]
+        for alpha in alphas:
+            # If the p-val is below alpha, reject
+            conservative_result = "Reject" if conservative_p_val < alpha else "Accept"
+            welch_result = "Reject" if welch_p_val < alpha else "Accept"
+
+            results['results']["conservative"][alpha] = conservative_result
+            results['results']["welch"][alpha] = welch_result
+
+        return results
 
     def _update_variance(self, batch_mean, batch_var, batch_count):
         """
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -336,11 +336,6 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
     @staticmethod
     def _perform_t_test(mean1, var1, n1,
                         mean2, var2, n2):
-        if n1 <= 1 or n2 <= 1:
-            warnings.warn("Insufficient sample size to "
-                          "perform t-test.", RuntimeWarning)
-            return None
-        
         results = {
             't-statistic': None,
             'conservative': {
@@ -351,8 +346,13 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
                 'df': None,
                 'p-value': None
             }
-
         }
+
+        if n1 <= 1 or n2 <= 1:
+            warnings.warn("Insufficient sample size to "
+                          "perform t-test.", RuntimeWarning)
+            return results
+
         s_delta = var1/n1 + var2/n2
         t = (mean1 - mean2) / np.sqrt(s_delta)
         conservative_df = min(n1, n2) - 1
@@ -361,6 +361,7 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
         results['t-statistic'] = t
         results['conservative']['df'] = conservative_df
         results['welch']['df'] = welch_df
+        
         try:
             import scipy.stats
         except ImportError:
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -336,48 +336,48 @@ class NumericStatsMixin(with_metaclass(abc.ABCMeta, object)):
     @staticmethod
     def _perform_t_test(mean1, var1, n1,
                         mean2, var2, n2):
-        results = {}
+        if n1 <= 1 or n2 <= 1:
+            warnings.warn("Insufficient sample size to "
+                          "perform t-test.", RuntimeWarning)
+            return "null"
+
+        results = {
+            't-statistic': None,
+            'conservative': {
+                'df': None,
+                'p-value': None
+            },
+            'welch': {
+                'df': None,
+                'p-value': None
+            }
+
+        }
         s_delta = var1/n1 + var2/n2
         t = (mean1 - mean2) / np.sqrt(s_delta)
         conservative_df = min(n1, n2) - 1
-        welch_df = s_delta ** 2 / ((var1 / n1) ** 2 / (n1 - 1) + (var2 / n2) ** 2 / (n2 - 1))
-        results.update({
-            "t-statistic": t,
-            "conservative_df": conservative_df,
-            "welch_df": welch_df
-        })
+        welch_df = s_delta ** 2 / ((var1 / n1) ** 2 /
+                                   (n1 - 1) + (var2 / n2) ** 2 / (n2 - 1))
+        results['t-statistic'] = t
+        results['conservative']['df'] = conservative_df
+        results['welch']['df'] = welch_df
         try:
             import scipy.stats
         except ImportError:
             # Failed, so we return the stats but don't perform the test
             warnings.warn("Could not import necessary statistical packages. "
-                          "T-test will be incomplete.", RuntimeWarning)
+                          "To successfully perform the t-test, please run 'pip "
+                          "install scipy.' T-test results will be incomplete.",
+                          RuntimeWarning)
             return results
         # If scipy import was successful, now perform the *two-sided* t-test
         conservative_t = scipy.stats.t(conservative_df)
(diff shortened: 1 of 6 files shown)


dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -115,6 +115,15 @@ def test_diff_primitive_compilers(self):
 
         compiler1 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data1)
         compiler2 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data2)
+
+        def get_welch_df(var1, n1, var2, n2):
+            return (var1 / n1 + var2 / n2) ** 2 \
+                   / ((var1 / n1) ** 2 / (n1 - 1)
+                      + (var2 / n2) ** 2 / (n2 - 1))
```
ChrisWallace2020, line 122: I noticed you have this function multiple times, should probably move this out to a util so you can call it a bunch without having to define it

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,53 @@ def kurtosis(self):
...
+    def _perform_t_test(mean1, var1, n1,
+                        mean2, var2, n2):
+        results = {}
+        s_delta = var1/n1 + var2/n2
+        t = (mean1 - mean2) / np.sqrt(s_delta)
+        conservative_df = min(n1, n2) - 1
+        welch_df = s_delta ** 2 / ((var1 / n1) ** 2 / (n1 - 1) + (var2 / n2) ** 2 / (n2 - 1))
+        results.update({
+            "t-statistic": t,
+            "conservative_df": conservative_df,
```
JGSweets, line 346: is conservative the term? vs worst-case?

JGSweets: my thoughts are worst-case is clear in its definition whereas conservative has varying degrees

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,53 @@ def kurtosis(self):
...
                     self._biased_kurtosis)
 
+    @staticmethod
+    def _perform_t_test(mean1, var1, n1,
+                        mean2, var2, n2):
+        results = {}
+        s_delta = var1/n1 + var2/n2
+        t = (mean1 - mean2) / np.sqrt(s_delta)
+        conservative_df = min(n1, n2) - 1
+        welch_df = s_delta ** 2 / ((var1 / n1) ** 2 / (n1 - 1) + (var2 / n2) ** 2 / (n2 - 1))
```
ChrisWallace2020, line 343: Could use util function here if you pull it out of all the tests where you wrote a helper

dataprofiler/profilers/numerical_column_stats.py
JGSweets, line 346: if instead results were:
```python
{
    {
        "t-statistic": ,
        "conservative": {
            "df": <VALUE>,
            "p-value": <NONE_OR_VALUE>
        },
        "welch": {
            "df": <VALUE>,
            "p-value": <NONE_OR_VALUE>
        }
        
}
```
then we wouldn't have multiple sets of info and it would all be consistent whether it was computed or not.
`None` being the value for not computed.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,53 @@ def kurtosis(self):
...
+                "conservative": {
+                    'p-value': conservative_p_val
+                },
+                "welch": {
+                    'p-value': welch_p_val
+                }
+            }
+        })
+
+        alphas = [0.01, 0.05, 0.1]
```
JGSweets, line 372: probably include `1e-3`

However, do we think we should do this? or just report the p-value and let them do what they want with the value?
Thoughts?

dataprofiler/profilers/numerical_column_stats.py
JGSweets, line 343: I think this is more than 80 chars

dataprofiler/profilers/numerical_column_stats.py
JGSweets: ahh i see, you keep the p-value and the alpha comparison. Do you think it's clear to have p-value and the alpha keys at the same level vs:
```python
{
    "df": ...,
    "p-value": ...,
    "alpha":{
        0.1: ...,
        ...
    }
}
```

dataprofiler/profilers/numerical_column_stats.py
JGSweets: what happens if `n1 or n2 == 1`?

JGSweets: same for `n1 or n2 == 0`

dataprofiler/profilers/numerical_column_stats.py
andrew-yin: I haven't heard of this method called "worst-case" before. All source I've seen refer to this as the conservative estimate for the degrees of freedom.

andrew-yin: Agreed, this looks a lot cleaner

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,53 @@ def kurtosis(self):
                     self.match_count,
                     self._biased_kurtosis)
 
+    @staticmethod
+    def _perform_t_test(mean1, var1, n1,
+                        mean2, var2, n2):
+        results = {}
+        s_delta = var1/n1 + var2/n2
```
JGSweets, line 340: `n1 or n2 == 0` error or do we skip out prior and not allow this?

dataprofiler/profilers/numerical_column_stats.py
andrew-yin: Since the significance level largely depends on the context of the data and other factors, I think it slims down the differences if only the p-value was included and allowing the user to make their own interpretation. Most statistical software also only give you the p-value back.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,53 @@ def kurtosis(self):
...
+            "t-statistic": t,
+            "conservative_df": conservative_df,
+            "welch_df": welch_df
+        })
+        try:
+            import scipy.stats
+        except ImportError:
+            # Failed, so we return the stats but don't perform the test
+            warnings.warn("Could not import necessary statistical packages. "
+                          "T-test will be incomplete.", RuntimeWarning)
```
JGSweets, line 354: at the very least should say `pip install scipy` best case scenario we abstract `warn_on_profile` to be utilized for both profiling or any case where the user needs to install a set of requirements like `dataprofiler[ml]` or if we do `dataprofiler[stats]`

dataprofiler/tests/profilers/test_column_profile_compilers.py
JGSweets: since tests don't change, we should expect the results to be unchanged and instead of using a function should just be the values.

JGSweets: or are the values in these cases two extreme to reduce?

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -129,8 +138,30 @@ def test_diff_primitive_compilers(self):
...
                      'mean': 2.0,
                      'variance': 38.666666666666664,
-                     'stddev': 3.285085839971525
-                 }
+                     'stddev': 3.285085839971525,
+                     't-test': {
+                         't-statistic': (other1.mean - other2.mean) / (
+                             other1.variance / other1.match_count +
+                             other2.variance / other2.match_count) ** 0.5,
+                         'conservative_df': min(other1.match_count, other2.match_count) - 1,
```
JGSweets, line 146: i think we can specify this value rather than calc

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -173,6 +204,15 @@ def test_disabling_columns_during_primitive_diff(self):
         # Test disabled column in both compilers
         compiler2 = col_pro_compilers.ColumnPrimitiveTypeProfileCompiler(data2,
                                                                          options)
+
+        def get_welch_df(var1, n1, var2, n2):
```
JGSweets, line 208: as above

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -194,11 +234,35 @@ def test_disabling_columns_during_primitive_diff(self):
...
                     'std': -0.71, 
                     'sample_size': 2,
-                    'margin_of_error': -1.6}
+                    'margin_of_error': -1.6
+                },
+                't-test': {
+                    't-statistic': (other1.mean - other2.mean) / (
+                            other1.variance / other1.match_count +
+                            other2.variance / other2.match_count) ** 0.5,
+                    'conservative_df': min(other1.match_count, other2.match_count) - 1,
```
JGSweets, line 243: as above

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1331,6 +1336,28 @@ def test_diff(self):
                 'sample_size': -1,
                 'margin_of_error':
                     profile1['precision']['margin_of_error'] - profiler2['precision']['margin_of_error']
+            },
+            't-test': {
+                't-statistic': (profiler1.mean - profiler2.mean) / np.sqrt(
+                    profiler1.variance/profiler1.match_count +
+                    profiler2.variance/profiler2.match_count),
+                'conservative_df': min(profiler1.match_count, profiler2.match_count) - 1,
```
JGSweets, line 1344: as above

dataprofiler/profilers/numerical_column_stats.py
JGSweets: then let's do just p-value as you suggest

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,53 @@ def kurtosis(self):
                     self.match_count,
                     self._biased_kurtosis)
 
+    @staticmethod
+    def _perform_t_test(mean1, var1, n1,
+                        mean2, var2, n2):
+        if n1 <= 1 or n2 <= 1:
+            warnings.warn("Insufficient sample size to "
+                          "perform t-test.", RuntimeWarning)
+            return "null"
```
JGSweets, line 342: I think maybe just return `None` rather than text. JSON treats `None` as `null` during serialization.

JGSweets: or rather we can could also consider putting the if below the `results` and just return that such that static return in terms of keys

JGSweets (dismissed):

JGSweets (dismissed):

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -129,10 +130,21 @@ def test_diff_primitive_compilers(self):
...
-                 }
+                     'stddev': 3.285085839971525,
+                     't-test': {
+                         't-statistic': 0.4155260166386663,
+                         'conservative': {
+                             'df': 1,
+                             'p-value': 0.749287157907667
+                         },
+                         'welch': {
+                             'df': 3.6288111187629117,
```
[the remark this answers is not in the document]
JGSweets: degrees of freedom

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -459,8 +459,18 @@ def test_diff(self):
...
+                             },
+                             'welch': {
+                                 'df': 15.761400272034564,
+                                 'p-value': 0.07127621949432528
+                             }
+                         }
                          }
-        diff = profiler.diff(profiler2)
+        diff = profiler1.diff(profiler2)
         self.assertDictEqual(expected_diff, diff)
```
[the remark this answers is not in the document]
JGSweets: I'm not sure we need to ensure specific cases so long as we ensure the p-value and df are calculated correctly and the edge cases are handled correctly.

dataprofiler/tests/profilers/test_column_profile_compilers.py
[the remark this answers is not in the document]
AnhTruong: should we change to `dof`? I've seen that used in numpy or scipy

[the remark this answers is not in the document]
andrew-yin: Df stands for degrees of freedom, which determines the "curvature" of the t-distribution since the population variance is unknown: https://en.wikipedia.org/wiki/Degrees_of_freedom_(statistics)

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -577,17 +578,72 @@ def test_diff(self):
...
+                'welch': {
+                    'df': 25.945257024943864,
+                    'p-value': 0.6980401261750298
+                }
+            }
         }
         self.assertDictEqual(expected_diff, other1.diff(other2))
-        
+
+        # Invalid statistics (NaN variance, insufficient match count)
```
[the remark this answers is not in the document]
andrew-yin: `other2.matchcount = 1` means the sample size is too small, but I can move this to a different test.

dataprofiler/tests/profilers/test_text_column_profile.py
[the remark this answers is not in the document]
AnhTruong: The thing is, currently I don't know how we get the values of t-test in the unit tests. So, either we provide the calculations of those in the form of some comments, or we can test the correctness (ish) easily using the cases that I suggested.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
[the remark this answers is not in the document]
JGSweets: Yes, bc there are two conditions here which would cause it to skip the diff, but only 1 is being checked.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -584,7 +584,7 @@ def test_diff(self):
             'max': [4, None],
             'sum': 'unchanged',
             'mean': 0.3,
-            'variance': -8.362573099415204,
```
JGSweets, line 586: why the reduction?

JGSweets: also i believe the exact value is `(9*20/19) - (1*10/9)`

JGSweets: no need to do almost equal

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -584,7 +584,7 @@ def test_diff(self):
...
+            'variance': -8.3625731,
             'stddev': -2.0238425028660023,
```
JGSweets, line 588: similarly, this would be the sqrt of the value above.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
andrew-yin: I was getting issues with Torch on this number due to the decimals, but this looks like a good workaround.

JGSweets (approved):

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -330,6 +333,62 @@ def kurtosis(self):
...
+
+        invalid_stats = False
+        if n1 <= 1 or n2 <= 1:
+            warnings.warn("Insufficient sample size. "
+                          "T-test cannot be performed.", RuntimeWarning)
+            invalid_stats = True
+        if np.isnan([mean1, mean2, var1, var2]).any() or \
+                None in [mean1, mean2, var1, var2]:
+            warnings.warn("Null value(s) found in mean and/or variance values. "
+                          "T-test cannot be performed.", RuntimeWarning)
```
AnhTruong, line 359: 👍

AnhTruong (approved):

Issue #265, opened 2021-06-12
Disable output / silent option?


makkus:
Would it be possible to have an option to disable any output / info messages when running the Profiler? I've had a look at the sources, and it seems simple `print`s are used, e.g. for:

```
Finding the Null values in the columns... (with 9 processes)
...
```
Maybe there could be a 'silent=True' option in the Profiler base class, or something similar? Having this output in the (standard) log(s) would still be useful, but as it is it gets in the way of my 'normal' command-line output, unless I apply some dodgy workarounds...

JGSweets:
The readers currently use `logging`, possible we could update the rest of the library to do likewise.

JGSweets:
@makkus did you have further thoughts on this, e.g. a mechanism to achieve.
like similar to TF?
`dp.get_logger().set_level('INFO')`

makkus:
@JGSweets , sorry, what's TF? I don't really have a preference, and I guess it depends how you do it in the rest of the application. My personal use case would be to use DataProfiler as a library, so the main thing I'm concerned about is to not have any output to stdout/stderr. If I can can set the logging level and receive logging data for dataprofiler via the normal Python log mechanism, that'd be great, but it would be no dealbreaker for me to not have any logging output at all. This is just me though, not sure what your other users would like to see.

JGSweets:
TF as in TensorFlow.

Logging would allow you to set the log level for output to console.

makkus:
Right. I have never used TensorFlow, so can't comment there. But this sounds pretty much like the standard way to deal with logs in Python, right? Like letting me set something like:

```
import logging
logging.basicConfig(filename='example.log', encoding='utf-8', level=logging.DEBUG)
```
And then do logger-name-specific config if necessary (e.g. have different log levels for different libraries -- I personally seldomly need to do taht). If that's the case, in my opinion that would be the best option, yes.

ChrisWallace2020:
@makkus the 2 PRs above (#357, #368) should resolve the issue. 

The `prints` were replaced with `logging.info` so that the effective level may be set high enough to filter them out if you choose to do so. This can be set either through `logging` like normal or you can use the function `set_verbosity` we've written in `dp_logging.py` that takes in the `logging` level (`logging.DEBUG`, `logging.INFO`, `logging.WARNING` etc). DataProfiler-specific logging will output to `stdout` (if left verbose). 

I hope this addresses your issue, let me know if there is any further feedback.

makkus:
[...]

PR #357, opened 2021-07-23
Change print/stdout to logging.INFO so that user can set level to silence console output

ChrisWallace2020:
Refactor print/stdout calls in data readers and labelers (profiler coming in subsequent PR), add high level helper function to quickly toggle verbosity. Added test to ensure logs being written as expected, and modified other tests to accommodate new logging functionality.

Need to discuss how we want "global verbosity" to interplay with "verbose" keywords in specific functions, may need changes in subsequent PRs.

Original feature request: #265


dataprofiler/__init__.py
```diff
@@ -25,8 +27,22 @@
...
+
+
+def set_verbosity(verbose=True):
+    if verbose:
+        # INFO level will print/log update strings and other outputs as well
+        # as warnings/errors
+        logging.getLogger().setLevel(logging.INFO)
+    else:
+        # WARNING level will print/log only warnings/errors
+        logging.getLogger().setLevel(logging.WARNING)
```
JGSweets, line 48: I think we should take in / allow verbosity levels. 
e.g. in TensorFlow they allow you to get the actual logger and set the level yourself:
e.g. `dp.get_logger().set_level('INFO')`

This also allows someone to change its handlers or even where it's outputting.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/platform/tf_logging.py
Also, we may want to consider putting this in its own file and importing it in `__init__.py`

dataprofiler/data_readers/data_utils.py
```diff
@@ -4,13 +4,16 @@
...
 import dateutil
+import logging
 
 import pandas as pd
 import pyarrow.parquet as pq
 from chardet.universaldetector import UniversalDetector
 
 from .filepath_or_buffer import FileOrBufferHandler
 
+logger = logging.getLogger('DataProfiler.data_utils')
```
JGSweets, line 15: needs 2 spaces from import

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -875,11 +874,9 @@ def _validate_training(self, val_data, batch_size_test=32,
                 x_val, batch_size=batch_size_test, verbose=verbose_keras)[1])
             y_val_test.append(np.argmax(y_val, axis=-1))
             batch_id += 1
-            sys.stdout.flush()
             if verbose_log:
-                sys.stdout.write(
-                    "\rEPOCH %g, validation_batch_id %d" % (
-                        self._epoch_id, batch_id))
+                logger.info("\rEPOCH %g, validation_batch_id %d"
+                            % (self._epoch_id, batch_id))
```
JGSweets, line 879: previously, this would count as a single line, but this looks like it would create a new line per. Is there a way this would stream in?

dataprofiler/data_readers/data_utils.py
JGSweets: Also, if we always do `DataProfiler.{}` there might be collisions, no? Should we instead do something like 
`logger = logging.getLogger(__FILE__)` or in this case `logger = logging.getLogger('dataprofiler.data_readers.data_utils')`? where we simulate based on module?

ChrisWallace2020: Yeah thats a good point, I'll add.

dataprofiler/__init__.py
ChrisWallace2020: So do you want to maintain the simple `True`/`False` interface for quick enabling/disabling or just have them take the level, because then this essentially just becomes a wrapper for `logging.getLogger.setLevel` that doesn't add anything else.

dataprofiler/labelers/character_level_cnn_model.py
ChrisWallace2020: This is still just one line? I am confused I didn't change the content of the string message, I don't think I understand your question.

JGSweets: In console it wouldn't create a new line for each write, but keep the counter increasing in place

dataprofiler/__init__.py
JGSweets: getLogger abstracts the variable from the user and allows them to alter the logger for more functionality as illustrated in the issue.

Also, as seen in the tag case, allows them to make it thread safe I believe.

dataprofiler/__init__.py
```diff
@@ -1,3 +1,5 @@
+import logging
```
JGSweets, line 1: More that I think about it, should be in a new file bc we shouldn't have dp.logging

dataprofiler/__init__.py
JGSweets: Also, with new file, logger could be saved as a variable abstracted from the user.

JGSweets: Not sure this is best yet.

dataprofiler/data_readers/data.py
```diff
@@ -1,14 +1,14 @@
...
 
 from .csv_data import CSVData
 from .json_data import JSONData
 from .text_data import TextData
 from .parquet_data import ParquetData
 from .avro_data import AVROData
+from .. import dp_logging
 
-logger = logging.getLogger('DataProfiler.data')
+logger = dp_logging.get_logger().getChild(__name__.replace('dataprofiler.', ''))
```
JGSweets, line 11: Maybe not in this PR if we don't have other changes to make, but would it beneficial to abstract this get child functionality where you only have to call: `dp_logging.get_child_logger(__name__)`?

Basically, makes it simple for future logging repeatability.

dataprofiler/dp_logging.py
```diff
@@ -0,0 +1,31 @@
+import logging
+import threading
+
+_logger = None
+_logger_lock = threading.Lock()
+
+
+def get_logger():
```
JGSweets, line 8: add description

dataprofiler/dp_logging.py
```diff
@@ -0,0 +1,31 @@
...
+            logging.Formatter("%(levelname)s:%(name)s: %(message)s"))
+        logger.addHandler(stream_handle)
+
+        _logger = logger
+        return _logger
+    finally:
+        _logger_lock.release()
+
+
+def set_verbosity(v):
```
JGSweets, line 35: add description

dataprofiler/data_readers/data.py
ChrisWallace2020: Yeah thats a good idea, I'll add if other features get requested

dataprofiler/__init__.py
```diff
@@ -6,6 +6,7 @@
                                     StructuredDataLabeler, \
                                     UnstructuredDataLabeler
 from .validators.base_validators import Validator
+from .dp_logging import set_verbosity
```
JGSweets, line 9: i would also allow the `get_logger`, e.g. `dp.get_logger` in this functionality.

dataprofiler/data_readers/data.py
ChrisWallace2020: Added

dataprofiler/dp_logging.py
ChrisWallace2020: Added

dataprofiler/__init__.py
ChrisWallace2020: Added

dataprofiler/tests/test_dp_logging.py
```diff
@@ -0,0 +1,71 @@
+import unittest
+import logging
+
+from dataprofiler import DataLabeler, dp_logging
+from dataprofiler.labelers.regex_model import RegexModel
+
+
+class TestDPLogging(unittest.TestCase):
+
```
JGSweets, line 9: we should add a test that asserts the default is set to `INFO`.

dataprofiler/tests/labelers/test_regex_model.py
```diff
@@ -183,14 +182,17 @@ def test_predict(self, mock_stdout):
                                [0, 1, 0],
                                [0, 1, 0]])
          ]}
-        model_output = model.predict(['   ', 'hello'])
+        with self.assertLogs('DataProfiler.labelers.regex_model',
+                             level='INFO') as cm:
+            model_output = model.predict(['   ', 'hello'])
         self.assertIn('pred', model_output)
+        log_output = ''.join(cm.output)
```
JGSweets, line 189: if we aren't using this here, i would suggest moving it to just above the assert it is being used with for clarity.

dataprofiler/tests/labelers/test_regex_model.py
```diff
@@ -231,12 +233,13 @@ def test_predict(self, mock_stdout):
...
-        mock_stdout.truncate(0)
-        model_output = model.predict(['hello world.'], verbose=False)
-        self.assertNotIn('Data Samples', mock_stdout.getvalue())
+        # test verbose = False doesn't make any INFO logs
+        with self.assertRaisesRegex(AssertionError,
+                                    'no logs of level INFO or higher triggered '
+                                    'on DataProfiler.labelers.regex_model'):
+            with self.assertLogs('DataProfiler.labelers.regex_model',
+                                 level='INFO'):
+                model.predict(['hello world.'], verbose=False)
```
JGSweets, line 242: Interesting solve. the other option would be to assertLog, put a dummy log in and then assert it is the only log. I'd have to think on what's more appropriate. It just feels weird to make an assert knowing it should fail.

dataprofiler/tests/test_dp_logging.py
```diff
@@ -0,0 +1,71 @@
...
+        # Check that 3 'EPOCH i' updates written for each epoch
+        for i in range(7):
+            # 3 entries indices corresponding to given epoch
+            idxs = [i * 3, i * 3 + 1, i * 3 + 2]
+            for idx in idxs:
+                log = cm.output[idx]
+                epoch_msg = (f"INFO:DataProfiler.labelers."
+                             f"character_level_cnn_model:\rEPOCH {i}")
+                self.assertEqual(epoch_msg, log[:len(epoch_msg)])
```
JGSweets, line 31: this seems highly dependent on the code below rather than just validating that text is output at different levels as expected. Also, these seem to check to make sure the logs were made rather than the message is actually sent to the stdout which is what verbosity may or may not be preventing.

dataprofiler/tests/test_dp_logging.py
ChrisWallace2020: Added

dataprofiler/tests/labelers/test_regex_model.py
ChrisWallace2020: Moved

dataprofiler/tests/test_dp_logging.py
ChrisWallace2020: Made it not code-dependant, need to figure out how to capture stdout in test case, currently having issues mocking.

dataprofiler/dp_logging.py
```diff
@@ -0,0 +1,56 @@
...
+        _logger = logger
+        return _logger
+    finally:
+        # Unlock before returning
+        _logger_lock.release()
+
+
+def set_verbosity(level):
+    """
+    Set verbosity level for DataProfiler logger
```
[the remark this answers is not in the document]
ChrisWallace2020: The "ints" are `logging.INFO`, `logging.DEBUG` etc that they just store as ints (10, 20, 30), I will list the options in the docstring though

dataprofiler/__init__.py
```diff
@@ -6,6 +6,7 @@
                                     StructuredDataLabeler, \
                                     UnstructuredDataLabeler
 from .validators.base_validators import Validator
+from .dp_logging import set_verbosity, get_logger
```
lettergram, line 9: I'd just import dp_logging

dataprofiler/tests/test_dp_logging.py
```diff
@@ -0,0 +1,68 @@
...
+        self.assertIn("this is warning 2", mock_stdout.getvalue())
+
+        # Going back to INFO still works as expected
+        dp_logging.set_verbosity(logging.INFO)
+
+        dummy_logger.info("this is info 1")
+        self.assertIn("this is info 1", mock_stdout.getvalue())
+
+        dummy_logger.warning("this is warning 1")
+        self.assertIn("this is warning 1", mock_stdout.getvalue())
```
JGSweets, line 68: are you sure this is valid, as in `mock_stdout` won't have been cleared meaning that `"this is info 1"` and `"this is warning 1"` will already be in the output

ChrisWallace2020: Whoops, copy and paste error, fixing now

JGSweets (approved):

gme5078 (approved):

PR #360, opened 2021-07-26
Allow internal data function calls directly from our data class

JGSweets:
This allows the user to directly access functions of the wrapped data inside of our data class. e.g. if the data internally is a pandas dataframe:
```python
data = CSVData(...)  # has pandas dataframe
df_without_na= data.dropna() # will not be in place unless user specifies it

# or iterating through rows
for row in data.iterrows():
    print(row)
```

### DIFF
--- a/dataprofiler/data_readers/base_data.py
+++ b/dataprofiler/data_readers/base_data.py
@@ -172,3 +172,32 @@ class BaseData(object):
         :return: length of the dataset
         """
         return len(self)
+
+    def __getattribute__(self, name):
+        """
+        Overrides getattr for the class such that functions can be applied
+        directly to the data class if the function is not part of the data
+        class.
+        e.g. if data is BaseData where self.data = [1, 2, 3, 1]
+        ```
+        data.count(1)  # returns 2, bc data.data has the function 'count'
+        ```
+        """
+        try:
+            returned = object.__getattribute__(self, name)
+        except AttributeError as attr_error:
+            class_name = self.__class__.__name__
+            data_class_name = self.data.__class__.__name__
+            if (not f"'{class_name}' object has no attribute '{name}'"
+                    == str(attr_error)):
+                raise
+            try:
+                returned = object.__getattribute__(self.data, name)
+            except AttributeError as attr_error:
+                if (not f"'{data_class_name}' object has no attribute '{name}'"
+                        == str(attr_error)):
+                    raise
+                raise AttributeError(f"'{class_name}' nor '{data_class_name}'"
+                                     f"objects have no attribute '{name}'")
+        return returned
+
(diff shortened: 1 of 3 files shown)


dataprofiler/tests/data_readers/test_base_data.py
```diff
@@ -0,0 +1,40 @@
...
+        # initialize the data class
+        data = BaseData(input_file_path="", data=FakeDataClass(), options={})
+
+        # if the function exists in BaseData fail the test because the results
+        # may become inaccurate.
+        self.assertFalse(hasattr(BaseData, 'func1'))
+
+        with self.assertRaisesRegex(AttributeError,
+                                    "'BaseData' nor 'FakeDataClass'"
+                                    "objects have no attribute 'test'"):
```
AnhTruong, line 32: should be `"'Neither BaseData' nor 'FakeDataClass'"
                                    "objects has attribute 'test'"`

JGSweets: fixed

dataprofiler/data_readers/base_data.py
```diff
@@ -172,3 +172,33 @@ def length(self):
...
+        data.count(1)  # returns 2, bc data.data has the function 'count'
+        ```
+        """
+        try:
+            returned = object.__getattribute__(self, name)
+        except AttributeError as attr_error:
+            class_name = self.__class__.__name__
+            data_class_name = self.data.__class__.__name__
+            if (not f"'{class_name}' object has no attribute '{name}'"
+                    == str(attr_error)):
```
gme5078, line 192: Its seems strange to use error string for the conditional, but I'm not sure any other way to do this check

dataprofiler/tests/data_readers/test_base_data.py
```diff
@@ -0,0 +1,40 @@
...
+                                    "Neither 'BaseData' nor 'FakeDataClass' "
+                                    "objects have attribute 'test'"):
+            data.test
+
+        # validate it will take BaseData attribute over the data attribute
+        self.assertIsNone(data.data_type)
+
+        # validate will auto call the data function if it doesn't exist in
+        # BaseData
+        self.assertEqual("success", data.func1())
```
gme5078, line 40: personally, did you try this with csv/pandas functions?

JGSweets: yes.

dataprofiler/data_readers/base_data.py
JGSweets: i validate it is an attribute error with the except and then validate the string, there might be a better way.

gme5078 (approved):

AnhTruong (approved):

PR #361, opened 2021-07-27
GH-Pages Workflow Improvement

gme5078:
This is just to automate some of the processes. 
These copies/removals are to ensure there are no security risks.


gme5078:
This has been fixed already with an updated Furo Theme

PR #363, opened 2021-07-27
List now maintains same visual data representation

JGSweets:
Previously if inputing list into the profiler, the data representation would change ints to floats when illustrating samples:
```python
import dataprofiler as dp

data = [1, 2, None]
profiler = dp.Profiler(data)

# output would have: ['1.0', '2.0'] as opposed to ['1', '2']
print(profiler.profile[0].sample)
```

This fixes that to show the latter repsentation.


ChrisWallace2020 (dismissed):

JGSweets:
Bug for 2D data as input. need to continue refactoring

gme5078 (approved):

ChrisWallace2020 (approved):

PR #365, opened 2021-07-28
Min True Samples Error Checking

gme5078:
Quick fix to prevent weird errors

### DIFF
--- a/dataprofiler/tests/profilers/test_profile_builder.py
+++ b/dataprofiler/tests/profilers/test_profile_builder.py
@@ -939,6 +939,17 @@ class TestStructuredProfiler(unittest.TestCase):
                                                samples_per_update=1)
         self.assertEqual(2, sparse_profile._min_col_samples_used)
 
+        # Test invalid input
+        msg = "`min_true_samples` must be an integer or `None`."
+        with self.assertRaisesRegex(ValueError, msg):
+            profile = dp.StructuredProfiler(sparse_df, 
+                                            min_true_samples="Bloop")
+
+        # Test None input (defaults to zero)
+        profile = dp.StructuredProfiler(sparse_df, min_true_samples=None)
+        self.assertEqual(None, profile._min_true_samples)
+
+
     def test_save_and_load(self):
         datapth = "dataprofiler/tests/data/"
         test_files = ["csv/guns.csv", "csv/iris.csv"]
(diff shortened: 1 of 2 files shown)


dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -939,6 +939,24 @@ def test_min_col_samples_used(self, *mocks):
                                                samples_per_update=1)
         self.assertEqual(2, sparse_profile._min_col_samples_used)
 
+    @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler')
```
JGSweets, line 942: can also mock out update, no need to profile anything in this.

JGSweets: Probably should test for Unstructured and dp.Profiler too, no?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -939,6 +939,24 @@ def test_min_col_samples_used(self, *mocks):
...
+    def test_min_true_samples(self, *mocks):
+        empty_df = pd.DataFrame([])
+
+        # Test invalid input
+        msg = "`min_true_samples` must be an integer or `None`."
+        with self.assertRaisesRegex(ValueError, msg):
+            profile = dp.StructuredProfiler(empty_df, 
+                                            min_true_samples="Bloop")
+
+        # Test None input (defaults to zero)
```
ChrisWallace2020, line 952: You say "defaults to zero" but then assert that its None. I would clarify in the comment that it doesn't get set to 0 in the init, otherwise the comment seems to clash with the assertion being made

dataprofiler/tests/profilers/test_profile_builder.py
gme5078: update not mocked since I wanted to make sure None still worked without error (it gets manipulated in update). Also the test takes 30ms so its not really going to get much faster. 

I added the tests for unstructured and dp.Profiler too

dataprofiler/tests/profilers/test_profile_builder.py
gme5078: Fair, changed the wording to make it somewhat more clear

ChrisWallace2020 (approved):

JGSweets (approved):

PR #367, opened 2021-07-28
Graphs: Create function which accepts a profiler and creates histogram bar charts

az85252:
I have created some functions which will allow multiple IntColumns and FloatColumns in a StructuredProfiler class object to be able to have their histograms plotted.

### DIFF
--- a/dataprofiler/reports/graphs.py
+++ b/dataprofiler/reports/graphs.py
@@ -9,16 +9,16 @@ import seaborn as sns
 
 def plot_histograms(profiler, columns=None):
     """
-        Take a input of StructuredProfiler class and a list of specified column
-        names and then plots the histograms of those that are int or float
-        columns.
+    Take a input of StructuredProfiler class and a list of specified column
+    names and then plots the histograms of those that are int or float
+    columns.
 
-        :param profiler: StructuredProfiler variable
-        :param columns: list of column names to be plotted
-        :type profiler: StructuredProfiler
-        :type columns: list
-        :return: returns fig
-        :rtype: fig
+    :param profiler: StructuredProfiler variable
+    :param columns: list of column names to be plotted
+    :type profiler: StructuredProfiler
+    :type columns: list
+    :return: returns fig
+    :rtype: fig
     """
 
     # get all inds to graph, raise error if user specified doesn't exist (part 1)
@@ -70,17 +70,17 @@ def plot_histograms(profiler, columns=None):
     return fig
 
 
-def plot_col_histogram(data_type_profiler, ax=None, title=""):
+def plot_col_histogram(data_type_profiler, ax=None, title=None):
     """
-        Take a input of a Int or Float Column and plots the histogram
+    Take a input of a Int or Float Column and plots the histogram
 
-        :param data_type_profiler: the Int or Float column we pass in
-        :param ax: ax as in seaborn ax
-        :param title: name of a individual histogram
-        :type data_type_profiler: Union[IntColumn, FloatColumn]
-        :type ax: list
-        :type title: str
-        :return:
+    :param data_type_profiler: the Int or Float column we pass in
+    :param ax: ax as in seaborn ax
+    :param title: name of a individual histogram
+    :type data_type_profiler: Union[IntColumn, FloatColumn]
+    :type ax: list
+    :type title: str
+    :return: ax
     """
 
     histogram = data_type_profiler._get_best_histogram_for_profile()
@@ -88,7 +88,10 @@ def plot_col_histogram(data_type_profiler, ax=None, title=""):
         raise ValueError("The column profiler, " + str(
             data_type_profiler.name) + ", provided had no data and "
                                        "therefore could not be plotted.")
-    plot = sns.histplot(x=histogram['bin_edges'][:-1], bins=histogram['bin_edges'],
+    ax = sns.histplot(x=histogram['bin_edges'][:-1], bins=histogram['bin_edges'],
                  weights=histogram['bin_counts'], ax=ax)
-    plot.set_title(title)
+    if title is None:
+        title = str(data_type_profiler.name)
+    ax.set_title(title)
+    return ax
 
--- a/dataprofiler/reports/graphs.py
+++ b/dataprofiler/reports/graphs.py
@@ -72,6 +72,10 @@ def plot_histograms(profiler, columns=None):
 
 def plot_col_histogram(data_type_profiler, ax=None, title=""):
     histogram = data_type_profiler._get_best_histogram_for_profile()
+    if not histogram['bin_counts'] or not histogram['bin_edges']:
+        raise ValueError("The column profiler, " + str(
+            data_type_profiler.name) + ", provided had no data and "
+                                       "therefore could not be plotted.")
     plot = sns.histplot(x=histogram['bin_edges'][:-1], bins=histogram['bin_edges'],
                  weights=histogram['bin_counts'], ax=ax)
     plot.set_title(title)
--- a/dataprofiler/reports/graphs.py
+++ b/dataprofiler/reports/graphs.py
@@ -14,8 +14,8 @@ def plot_histograms(profiler, columns=None):
     columns.
 
     :param profiler: StructuredProfiler variable
-    :param columns: list of column names to be plotted
     :type profiler: StructuredProfiler
+    :param columns: list of column names to be plotted
     :type columns: list
     :return: returns fig
     :rtype: fig
@@ -36,13 +36,17 @@ def plot_histograms(profiler, columns=None):
             inds_to_graph.extend(profiler._col_name_to_idx[col])
         sorted(inds_to_graph)
     # get all columns which are either int or float (part2)
-    for col_ind in reversed(inds_to_graph):
-        # get data_type
-        data_compiler = profiler.profile[col_ind].profiles['data_type_profile']
-        data_type = data_compiler.selected_data_type
-        data_type_profiler = data_compiler._profiles[data_type]
-        if data_type not in ['int', 'float']:
-            inds_to_graph.pop()
+    def is_index_graphable_column(ind_to_graph):
+        """
+        This function filters ind_to_graph by returning false if there is a
+        data type that is not a int or float, otherwise true
+        """
+        col_profiler = profiler.profile[ind_to_graph]
+        data_compiler = col_profiler.profiles['data_type_profile']
+        if data_compiler.selected_data_type not in ['int', 'float']:
+            return False
+        return True
+    inds_to_graph = list(filter(is_index_graphable_column, inds_to_graph))
 
     if not inds_to_graph:
         warnings.warn("No plots were constructed"
@@ -75,10 +79,10 @@ def plot_col_histogram(data_type_profiler, ax=None, title=None):
     Take a input of a Int or Float Column and plots the histogram
 
     :param data_type_profiler: the Int or Float column we pass in
-    :param ax: ax as in seaborn ax
-    :param title: name of a individual histogram
     :type data_type_profiler: Union[IntColumn, FloatColumn]
+    :param ax: ax as in seaborn ax
     :type ax: list
+    :param title: name of a individual histogram
     :type title: str
     :return: ax
     """
--- a/dataprofiler/reports/graphs.py
+++ b/dataprofiler/reports/graphs.py
@@ -17,7 +17,8 @@ def plot_histograms(profiler, columns=None):
         :param columns: list of column names to be plotted
         :type profiler: StructuredProfiler
         :type columns: list
-        :return:
+        :return: returns fig
+        :rtype: fig
     """
 
     # get all inds to graph, raise error if user specified doesn't exist (part 1)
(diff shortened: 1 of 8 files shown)


dataprofiler/tests/reports/test_graphs.py
```diff
@@ -28,20 +28,23 @@ def setUpClass(cls):
...
     @mock.patch("dataprofiler.reports.graphs.plot_col_histogram")
     def test_normal(self, plot_col_mock, seaborn_mock, plt_mock):
-        x = graphs.plot_histograms(self.profiler, [1])
+        graphsplot = graphs.plot_histograms(self.profiler, [1])
         print(graphs)
         self.assertEqual(1, plot_col_mock.call_count)
-        self.assertEqual(1, plot_col_mock.call_args_list[0].args[0].name)
-        self.assertIsNotNone(x)
+        self.assertEqual(1, plot_col_mock.call_args_list[0][0][0].name)
+        self.assertIsNotNone(graphsplot)
```
JGSweets, line 47: can we assert the return is a plt.Figure:
`self.assertIsInstance(x, plt.Figure)`

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,84 @@
+import unittest
+import warnings
+
+from dataprofiler.profilers import IntColumn
+from dataprofiler.reports import graphs
+from unittest import mock
+import dataprofiler as dp
+import pandas as pd
```
JGSweets, line 8: cleanup imports for proper order

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(0, plot_col_mock.call_args_list[0][0][0].name)
+        # grabs the second argument passed into the plot col call and validates
+        # it is the column profiler and its name matches what we expect it to
+        self.assertEqual(1, plot_col_mock.call_args_list[1][0][0].name)
+        self.assertIsNotNone(graphsplot)
+
+    @mock.patch("dataprofiler.reports.graphs.plot_col_histogram")
+    def test_normal(self, plot_col_mock, seaborn_mock, plt_mock):
+        graphsplot = graphs.plot_histograms(self.profiler, [1])
+        print(graphs)
```
JGSweets, line 44: remove

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,84 @@
...
+class TestPlotColHistogram(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.data = pd.Series([1, 2, 4, 2, 5, 35, 32], dtype=str)
+        cls.profiler = IntColumn(cls.data.name)
+        cls.profiler.update(cls.data)
+
+    def test_normal(self, plt_mock):
+        graphs.plot_col_histogram(self.profiler)
```
JGSweets, line 77: assert return is the instance type we expect, e.g. `plt.Axes`

JGSweets: Also, need to validate the data in the figure is what we expect it to.

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,84 @@
...
+        self.assertEqual(2, plot_col_mock.call_count)
+        # grabs the first argument passed into the plot col call and validates
+        # it is the column profiler and its name matches what we expect it to
+        self.assertEqual(0, plot_col_mock.call_args_list[0][0][0].name)
+        # grabs the second argument passed into the plot col call and validates
+        # it is the column profiler and its name matches what we expect it to
+        self.assertEqual(1, plot_col_mock.call_args_list[1][0][0].name)
+        self.assertIsNotNone(graphsplot)
+
+    @mock.patch("dataprofiler.reports.graphs.plot_col_histogram")
```
JGSweets, line 41: we can prob move this to above the class since we can apply to all the functions here

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,84 @@
...
+
+from dataprofiler.profilers import IntColumn
+from dataprofiler.reports import graphs
+from unittest import mock
+import dataprofiler as dp
+import pandas as pd
+
+
+@mock.patch("dataprofiler.reports.graphs.plt.show")
+@mock.patch("dataprofiler.reports.graphs.sns")
```
JGSweets, line 12: we can prob remove the sns mock since we are mocking the function which calls it: ` @mock.patch("dataprofiler.reports.graphs.plot_col_histogram")`

JGSweets: this will change the function inputs for all tests below

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,84 @@
...
+                dp.StructuredProfiler(data=None, options=self.options))
+
+
+@mock.patch("dataprofiler.reports.graphs.plt.show")
+class TestPlotColHistogram(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        cls.data = pd.Series([1, 2, 4, 2, 5, 35, 32], dtype=str)
+        cls.profiler = IntColumn(cls.data.name)
```
JGSweets, line 73: For this, let's do:
`cls.profiler = IntColumn('test')`
Otherwise, the title is an empty str

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,82 @@
+import unittest
+from unittest import mock
+
+import pandas as pd
+from matplotlib import pyplot as plt
+import dataprofiler as dp
```
JGSweets, line 6: need a break between this and the line above to separate the imports

setup.py
```diff
@@ -27,6 +27,10 @@
 # Get the install_requirements from requirements.txt
 with open(path.join(here, 'requirements-ml.txt'), encoding='utf-8') as f:
     ml_packages = f.read().splitlines()
+
+# Get the install_requirements from requirements.txt
```
ChrisWallace2020, line 31: Should change the comment to specify that its `requirements-reports.txt`

dataprofiler/reports/graphs.py
```diff
@@ -0,0 +1,97 @@
...
+
+    :param profiler: StructuredProfiler variable
+    :param columns: list of column names to be plotted
+    :type profiler: StructuredProfiler
+    :type columns: list
+    :return: returns fig
+    :rtype: fig
+    """
+
+    # get all inds to graph, raise error if user specified doesn't exist (part 1)
```
ChrisWallace2020, line 24: I would remove the `(part x)` comments, personally, I think just listing out the steps is sufficient indication of a "step" of computation. Not a huge issue though, won't block because of it.

dataprofiler/reports/graphs.py
```diff
@@ -0,0 +1,97 @@
...
+def plot_histograms(profiler, columns=None):
+    """
+    Take a input of StructuredProfiler class and a list of specified column
+    names and then plots the histograms of those that are int or float
+    columns.
+
+    :param profiler: StructuredProfiler variable
+    :param columns: list of column names to be plotted
+    :type profiler: StructuredProfiler
+    :type columns: list
```
ChrisWallace2020, line 19: You should organize them by parameter, then type of that parameter, then parameter, type of that parameter, and so on... So instead of this the list (top to bottom) should be `:param profiler:, :type profiler:, :param columns:, :type columns:`)

dataprofiler/reports/graphs.py
```diff
@@ -0,0 +1,97 @@
...
+
+def plot_col_histogram(data_type_profiler, ax=None, title=None):
+    """
+    Take a input of a Int or Float Column and plots the histogram
+
+    :param data_type_profiler: the Int or Float column we pass in
+    :param ax: ax as in seaborn ax
+    :param title: name of a individual histogram
+    :type data_type_profiler: Union[IntColumn, FloatColumn]
+    :type ax: list
```
ChrisWallace2020, line 81: See comment about docstring parameter order above

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,82 @@
+import unittest
+from unittest import mock
+
+import pandas as pd
+from matplotlib import pyplot as plt
+
+import dataprofiler as dp
+from dataprofiler.profilers import IntColumn
+from dataprofiler.reports import graphs
```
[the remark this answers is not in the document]
JGSweets: yes, `dp.graphs` but if nothing generally wrong with this PR, we might get this in and do a cleanup

dataprofiler/reports/graphs.py
```diff
@@ -0,0 +1,97 @@
...
+            inds_to_graph.extend(profiler._col_name_to_idx[col])
+        sorted(inds_to_graph)
+    # get all columns which are either int or float (part2)
+    for col_ind in reversed(inds_to_graph):
+        # get data_type
+        data_compiler = profiler.profile[col_ind].profiles['data_type_profile']
+        data_type = data_compiler.selected_data_type
+        data_type_profiler = data_compiler._profiles[data_type]
+        if data_type not in ['int', 'float']:
+            inds_to_graph.pop()
```
ChrisWallace2020, line 45: Have you considered using `filter` on a list, to just hand it a boolean function and it filters out only values that return True. I think that could clean up this section a good bit.

ChrisWallace2020: Also it seems `data_type_profiler` never gets used, you could probably remove that

ChrisWallace2020: And I'm not sure this is working as intended, because you iterate backwards, and pop if something isn't int/float, but that wont work because if you get to an item that is int/float, then the next isn't, you'll actually pop the one that is.

ChrisWallace2020: Regardless of how you solve it though, should add a test case where some of the users specified columns aren't int/float, and verify that they get properly ignored, and nothing that is actually int/float gets omitted on accident.

JGSweets:
```python
    def is_index_graphable_column(ind_to_graph):
        col_profiler = profiler.profile[ind_to_graph]
        data_compiler = col_profiler.profiles['data_type_profile']
        if data_compiler.selected_data_type not in ['int', 'float']:
            return False
        return True
    inds_to_graph = list(filter(is_index_graphable_column, inds_to_graph))
```
so like this?

dataprofiler/reports/graphs.py
```diff
@@ -0,0 +1,97 @@
...
+                      " because no int or float columns were found in columns")
+        return
+    # get proper tile format for graph
+    n = len(inds_to_graph)
+    cols = math.ceil(math.sqrt(n))
+    rows = math.ceil(n / cols)
+    fig, axs = plt.subplots(rows, cols)  # this will need to be flattened into a list
+    # flatten axes for inputing graphs into the plot
+    if not isinstance(axs, (np.ndarray)):
+        axs = np.array([axs])
```
ChrisWallace2020, line 58: Firstly the `np.ndarray` doesn't need those parentheses, but also why is this in an `if` statement? Is the return of `plt.subplots` not always of the same type?

ChrisWallace2020: I would also maybe add comments about what `fig` and `axs` are since the names are a bit vague

JGSweets: Yes, plt.subplots doesn't always return the same object or object dimensions

az85252: There are instances where axs  has rows=1 or cols=1 and axes does not become a np array it becomes a number.

dataprofiler/reports/graphs.py
```diff
@@ -0,0 +1,101 @@
...
+    def is_index_graphable_column(ind_to_graph):
+        """
+        This function filters ind_to_graph by returning false if there is a
+        data type that is not a int or float, otherwise true
+        """
+        col_profiler = profiler.profile[ind_to_graph]
+        data_compiler = col_profiler.profiles['data_type_profile']
+        if data_compiler.selected_data_type not in ['int', 'float']:
+            return False
+        return True
```
ChrisWallace2020, line 48: You could simplify this to just `return data_compiler.selected_dataype in ['int', 'float']`, not crucial, but would be cleaner

dataprofiler/reports/graphs.py
ChrisWallace2020: Mmmm, makes sense

dataprofiler/reports/graphs.py
az85252: hmm with that I am getting:

AttributeError: 'ColumnPrimitiveTypeProfileCompiler' object has no attribute 'selected_dataype'

JGSweets: Stay with the conditions because others maybe added

dataprofiler/tests/reports/test_graphs.py
```diff
@@ -0,0 +1,82 @@
...
+        self.assertIsInstance(graphsplot, plt.Figure)
+
+    def test_bad_column_name(self, plot_col_mock, plt_mock):
+        with self.assertRaisesRegex(ValueError,
+                                    "Column \"a\" is not found as a profiler column"):
+            graphs.plot_histograms(self.profiler, [0, "a"])
+
+    def test_no_column_plottable(self, plot_col_mock, plt_mock):
+        with self.assertWarnsRegex(Warning, "No plots were constructed"
+                                            " because no int or float columns were found in columns"):
```
ChrisWallace2020, line 54: A few lines in this test class seem to be longer than 80 characters and need to be slimmed down.

JGSweets (approved):

ChrisWallace2020 (approved):

PR #375, opened 2021-07-30
Change DataProfiler to Accept URLs

gautomdas:
Added methods to read URL

### DIFF
--- a/dataprofiler/data_readers/data.py
+++ b/dataprofiler/data_readers/data.py
@@ -54,7 +54,7 @@ class Data(object):
         if not options:
             options = dict()
 
-        if is_valid_url(input_file_path):
+        if isinstance(input_file_path, str) and is_valid_url(input_file_path):
             input_file_path = url_to_bytes(input_file_path, options)
 
         for data_class_info in cls.data_classes:
--- a/dataprofiler/data_readers/data.py
+++ b/dataprofiler/data_readers/data.py
@@ -6,6 +6,7 @@ from .json_data import JSONData
 from .text_data import TextData
 from .parquet_data import ParquetData
 from .avro_data import AVROData
+from .data_utils import is_valid_url, url_to_bytes
 from .. import dp_logging
 
 logger = dp_logging.get_child_logger(__name__)
@@ -52,6 +53,9 @@ class Data(object):
 
         if not options:
             options = dict()
+
+        if is_valid_url(input_file_path):
+            input_file_path = url_to_bytes(input_file_path, options)
 
         for data_class_info in cls.data_classes:
             data_class = data_class_info["data_class"]
--- a/dataprofiler/data_readers/data.py
+++ b/dataprofiler/data_readers/data.py
@@ -54,7 +54,7 @@ class Data(object):
         if not options:
             options = dict()
 
-        if isinstance(input_file_path, str) and is_valid_url(input_file_path):
+        if is_valid_url(input_file_path):
             input_file_path = url_to_bytes(input_file_path, options)
 
         for data_class_info in cls.data_classes:
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -4,6 +4,8 @@ import json
 from io import open, StringIO, BytesIO, TextIOWrapper
 from collections import OrderedDict
 import dateutil
+import requests
+from urllib.parse import urlparse
 
 import pandas as pd
 import pyarrow.parquet as pq
@@ -610,3 +612,45 @@ def is_stream_buffer(filepath_or_buffer):
     if isinstance(filepath_or_buffer, (StringIO, BytesIO)):
         return True
     return False
+
+def is_valid_url(url_as_string):
+    """
+    Determines whether a given string is a valid URL
+
+    :param url_as_string: string to be tested if URL 
+    :type url_as_string: str
+    :return: true if string is a valid URL
+    :rtype: boolean
+    """
+    result = urlparse(url_as_string)
+    # this is the minimum characteristics needed for a valid URL
+    return all([result.scheme, result.netloc])
+
+
+def url_to_bytes(url_as_string, options):
+    """
+    Reads in URL and converts it to a byte stream
+
+    :param url_as_string: string to read as URL 
+    :type url_as_string: str
+    :return: true if string is a valid URL
+    :rtype: boolean
+    """
+    all_chunks = b''
+
+    verify_url = False
+    if 'verify' in options:
+        verify_url = options['verify']
+
+    with requests.get(url_as_string, stream=True, verify=verify_url) as url:
+        url.raise_for_status()
+        total_bytes = 0
+        c_size = 8192
+        for chunk in url.iter_content(chunk_size=c_size):
+            all_chunks += chunk
+            total_bytes += c_size
+            # check if total bytes read has been greater than 1GB
+            if total_bytes >= 1024 ** 3:
+                raise ValueError('The downloaded file from the url may not be larger than 1GB')
+
+    return BytesIO(all_chunks)
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -622,6 +622,9 @@ def is_valid_url(url_as_string):
     :return: true if string is a valid URL
     :rtype: boolean
     """
+    if not isinstance(url_as_string, str):
+        return False
+
     result = urllib.parse.urlparse(url_as_string)
     # this is the minimum characteristics needed for a valid URL
     return all([result.scheme, result.netloc])
@@ -633,8 +636,10 @@ def url_to_bytes(url_as_string, options):
 
     :param url_as_string: string to read as URL 
     :type url_as_string: str
-    :return: true if string is a valid URL
-    :rtype: boolean
+    :param options: options for the url 
+    :type options: dict
+    :return: BytesIO stream of data downloaded from URL
+    :rtype: BytesIO stream
     """
     stream = BytesIO()
 
@@ -644,22 +649,20 @@ def url_to_bytes(url_as_string, options):
 
     with requests.get(url_as_string, stream=True, verify=verify_url) as url:
         url.raise_for_status()
-        if 'Content-length' in url.headers and \
-            int(url.headers['Content-length']) >= 1024 ** 3:
+        if 'Content-length' in url.headers \
+                and int(url.headers['Content-length']) >= 1024 ** 3:
 
-            raise ValueError('The downloaded file from the url may not be \
-                larger than 1GB')
+            raise ValueError('The downloaded file from the url may not be larger than 1GB')
 
         total_bytes = 0
         c_size = 8192
 
-        for chunk in url.iter_content(chunk_size=8192):
+        for chunk in url.iter_content(chunk_size=c_size):
             stream.write(chunk)
             total_bytes += c_size
 
-            if total_bytes >= 1024 ** 3:
+            if total_bytes > 1024 ** 3:
                 raise ValueError('The downloaded file from the url may not be larger than 1GB')
-
 
     stream.seek(0)
     return stream
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -5,7 +5,7 @@ from io import open, StringIO, BytesIO, TextIOWrapper
 from collections import OrderedDict
 import dateutil
 import requests
-from urllib.parse import urlparse
+import urllib
 
 import pandas as pd
 import pyarrow.parquet as pq
@@ -622,7 +622,7 @@ def is_valid_url(url_as_string):
     :return: true if string is a valid URL
     :rtype: boolean
     """
-    result = urlparse(url_as_string)
+    result = urllib.parse.urlparse(url_as_string)
     # this is the minimum characteristics needed for a valid URL
     return all([result.scheme, result.netloc])
 
@@ -636,7 +636,7 @@ def url_to_bytes(url_as_string, options):
     :return: true if string is a valid URL
     :rtype: boolean
     """
-    all_chunks = b''
+    stream = BytesIO()
 
     verify_url = False
     if 'verify' in options:
@@ -644,13 +644,22 @@ def url_to_bytes(url_as_string, options):
 
     with requests.get(url_as_string, stream=True, verify=verify_url) as url:
         url.raise_for_status()
+        if 'Content-length' in url.headers and \
+            int(url.headers['Content-length']) >= 1024 ** 3:
+
+            raise ValueError('The downloaded file from the url may not be \
+                larger than 1GB')
+
         total_bytes = 0
         c_size = 8192
-        for chunk in url.iter_content(chunk_size=c_size):
-            all_chunks += chunk
+
+        for chunk in url.iter_content(chunk_size=8192):
+            stream.write(chunk)
             total_bytes += c_size
-            # check if total bytes read has been greater than 1GB
+
             if total_bytes >= 1024 ** 3:
                 raise ValueError('The downloaded file from the url may not be larger than 1GB')
 
-    return BytesIO(all_chunks)
+
+    stream.seek(0)
+    return stream
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -643,7 +643,7 @@ def url_to_bytes(url_as_string, options):
     """
     stream = BytesIO()
 
-    verify_url = False
+    verify_url = True
     if 'verify' in options:
         verify_url = options['verify']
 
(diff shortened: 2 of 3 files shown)


dataprofiler/data_readers/data_utils.py
```diff
@@ -4,6 +4,8 @@
 from io import open, StringIO, BytesIO, TextIOWrapper
 from collections import OrderedDict
 import dateutil
+import requests
+from urllib.parse import urlparse
```
JGSweets, line 8: instead of importing specifically urlparse, we can just import urllib so we know where it's coming form below.

```python
import urllib

.
.
urllib.parse.urlparse(...)
```

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,45 @@ def is_stream_buffer(filepath_or_buffer):
...
+def is_valid_url(url_as_string):
+    """
+    Determines whether a given string is a valid URL
+
+    :param url_as_string: string to be tested if URL 
+    :type url_as_string: str
+    :return: true if string is a valid URL
+    :rtype: boolean
+    """
+    result = urlparse(url_as_string)
```
JGSweets, line 625: this may not be a string, we don't truly know what is being passed in. Hence we should have a try except to catch when this fails.

JGSweets: e.g. a user could pass in BytesIO into Data and this may error, no?

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,45 @@ def is_stream_buffer(filepath_or_buffer):
...
+    verify_url = False
+    if 'verify' in options:
+        verify_url = options['verify']
+
+    with requests.get(url_as_string, stream=True, verify=verify_url) as url:
+        url.raise_for_status()
+        total_bytes = 0
+        c_size = 8192
+        for chunk in url.iter_content(chunk_size=c_size):
+            all_chunks += chunk
```
JGSweets, line 650: instead of storing into all_chunks, I think we can load directly into a BytesIO

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,45 @@ def is_stream_buffer(filepath_or_buffer):
...
+def url_to_bytes(url_as_string, options):
+    """
+    Reads in URL and converts it to a byte stream
+
+    :param url_as_string: string to read as URL 
+    :type url_as_string: str
+    :return: true if string is a valid URL
+    :rtype: boolean
+    """
+    all_chunks = b''
```
JGSweets, line 639: this could be the init of the BytesIO,
stream = BytesIO()

dataprofiler/data_readers/data_utils.py
JGSweets: this could then be `stream.write(chunk)`

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,45 @@ def is_stream_buffer(filepath_or_buffer):
...
+        total_bytes = 0
+        c_size = 8192
+        for chunk in url.iter_content(chunk_size=c_size):
+            all_chunks += chunk
+            total_bytes += c_size
+            # check if total bytes read has been greater than 1GB
+            if total_bytes >= 1024 ** 3:
+                raise ValueError('The downloaded file from the url may not be larger than 1GB')
+
+    return BytesIO(all_chunks)
```
JGSweets, line 656: if we do the above where we write to BytesIO, we can then seek to 0 prior to returning the stream:

```python
stream.seek(0)
return stream
```

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +15,52 @@ def test_read_data(self):
...
+        def infinite_loop(*args, **kwargs):
+            max_file_list = ([b'test']
+                             * (int(max_allows_file_size) // c_size + 1))
+            for value in max_file_list:
+                yield value
+        # mock the iter_content to return up to 1GB + so error raises
+        mock_request_get.return_value.__enter__.return_value\
+            .iter_content.side_effect = infinite_loop
+
+        with self.assertRaisesRegex(ValueError, 'The downloaded file from the url may not be larger than 1GB'):
```
JGSweets, line 60: make sure within 80 chars.

JGSweets: I think we should maybe do a RuntimeError? @lettergram

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +15,52 @@ def test_read_data(self):
...
+        """Ensure error logs for trying to save empty data."""
+
+        for url in self.input_url_names:
+            data_obj = Data(url['path'])
+            self.assertEqual(data_obj.data_type, url['data_type'])
+
+    @mock.patch('requests.get')
+    def test_read_url_overflow(self, mock_request_get):
+        # assumed chunk size
+        c_size = 8192
```
JGSweets, line 49: Thinking about this further, I think we should do two things in this test:
1. validate that if <= 1GB we don't error.
2. validate if > 1GB we error.

Right now we test > 1GB, but it is possible the code is in an always error mode. So we should also set the mock to something that is < 1GB loops,
e.g. (pseudo code below, not sure if works)
```python
mock_request_get.return_value.__enter__.return_value\
            .iter_content.return_value = [b'test']
```
also thinking about this further, we prob don't need this fake function for the test of > 1GB either:
```python
mock_request_get.return_value.__enter__.return_value\
            .iter_content.return_value = [b'test'] * (int(max_allows_file_size) // c_size + 1))
```
should work as well.

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +15,52 @@ def test_read_data(self):
...
+        max_allows_file_size = 1024 ** 3 # 1GB
+        def infinite_loop(*args, **kwargs):
+            max_file_list = ([b'test']
+                             * (int(max_allows_file_size) // c_size + 1))
+            for value in max_file_list:
+                yield value
```
JGSweets, line 55: prob can remove this function `infinite_loop` based on above suggestion

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +15,52 @@ def test_read_data(self):
...
+        # mock the iter_content to return up to 1GB + so error raises
+        mock_request_get.return_value.__enter__.return_value\
+            .iter_content.side_effect = infinite_loop
+
+        with self.assertRaisesRegex(ValueError, 'The downloaded file from the url may not be larger than 1GB'):
+            data_obj = Data('https://test.com')
+
+
```
JGSweets, line 63: extra space here

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +15,52 @@ def test_read_data(self):
...
+            dict(path=http_url+'parquet/iris.parq', count=150, data_type='parquet'),
+            dict(path=http_url+'parquet/titanic.parq', count=1316, data_type='parquet')
+        ]
+
+    def test_read_from_url(self):
+        """Ensure error logs for trying to save empty data."""
+
+        for url in self.input_url_names:
+            data_obj = Data(url['path'])
+            self.assertEqual(data_obj.data_type, url['data_type'])
```
JGSweets, line 44: let's mock out actual urls and utilize the same testing structure as below.

if we want an actual file like we do in the other readers, reference the file, and load it into a chunked list:
```python
c_size = 8192
def chunked_file(filepath):
    with open(file, 'rb') as fp:
        bytes = fp.read(c_size)
        while byte != b"":
            yield bytes
            bytes = f.read(c_size)
mock_request_get.return_value.__enter__.return_value\
            .iter_content.side_effect  = chunk_file(filepath)
```
not 100% sure ^^ correct

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +15,52 @@ def test_read_data(self):
...
+                 num_columns=10, encoding='utf-8', data_type='csv'),
+            dict(path=http_url+'csv/iris-utf-16.csv',
+                 count=150, delimiter=',', has_header=[0],
+                 num_columns=6, encoding='utf-16', data_type='csv'),
+            dict(path=http_url+'avro/users.avro', count=4,  data_type='avro'),
+            dict(path=http_url+'avro/userdata1.avro', count=1000,  data_type='avro'),
+            dict(path=http_url+'json/iris-utf-8.json', encoding='utf-8', count=150, data_type='json'),
+            dict(path=http_url+'json/iris-utf-16.json', encoding='utf-16', count=150, data_type='json'),
+            dict(path=http_url+'parquet/iris.parq', count=150, data_type='parquet'),
+            dict(path=http_url+'parquet/titanic.parq', count=1316, data_type='parquet')
```
JGSweets, line 36: prolly just limit to:
```
dict(path=test_dir+'/csv/diamonds.csv',
     count=53940, delimiter=',', has_header=[0],
     num_columns=10, encoding='utf-8', data_type='csv'),
dict(path=test_dir+'/avro/users.avro', count=4,  data_type='avro'),
dict(path=test_dir+'/json/iris-utf-16.json', encoding='utf-16', count=150, data_type='json'),
dict(path=test_dir+'/parquet/iris.parq', count=150, data_type='parquet'),
```
that way we test each data_type and a few different encodings.

JGSweets: ^^ prob can also remove superfluous info as well e.g. num_cols, has_header. however we might be able to check things like count to validate read properly.

we also need to add a `text` data type

dataprofiler/data_readers/data.py
```diff
@@ -53,6 +54,9 @@ def __new__(cls, input_file_path=None, data=None, data_type=None, options=None):
         if not options:
             options = dict()
 
+        if isinstance(input_file_path, str) and is_valid_url(input_file_path):
```
JGSweets, line 57: i think the string check should be in the `is_valid_url` as oppose to here as the `is_valid_url` will fail if anything but a string is passed to it.

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,54 @@ def is_stream_buffer(filepath_or_buffer):
...
+        c_size = 8192
+
+        for chunk in url.iter_content(chunk_size=8192):
+            stream.write(chunk)
+            total_bytes += c_size
+
+            if total_bytes >= 1024 ** 3:
+                raise ValueError('The downloaded file from the url may not be larger than 1GB')
+
+
```
JGSweets, line 663: extra space

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -1,10 +1,12 @@
 import unittest
+from unittest import mock
+import os
 
 import pandas as pd
 
 from dataprofiler.data_readers.data import Data
 
-
+test_root_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
```
JGSweets, line 9: need space between this and next line

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +16,60 @@ def test_read_data(self):
...
+
+        try:
+            # mock the iter_content to return just under 1GB so no error raises
+            mock_request_get.return_value.__enter__.return_value.iter_content.\
+                return_value = [b'test'] * (int(max_allows_file_size) // c_size - 1)
+            data_obj = Data('https://test.com')
+
+        except ValueError:
+            self.fail("URL string unexpected overflow error.")
+
```
JGSweets, line 64: extra space here

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +16,60 @@ def test_read_data(self):
...
+    @mock.patch('requests.get')
+    def test_read_url_content_overflow(self, mock_request_get):
+        # assumed chunk size
+        c_size = 8192
+        max_allows_file_size = 1024 ** 3 # 1GB
+
+        try:
+            # mock the iter_content to return just under 1GB so no error raises
+            mock_request_get.return_value.__enter__.return_value.iter_content.\
+                return_value = [b'test'] * (int(max_allows_file_size) // c_size - 1)
```
JGSweets, line 59: this has `c_size - 1` and the other is `c_size + 1`, that's a separation of 2. Not a big deal, but I feel like it should be only separated by 1.

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,54 @@ def is_stream_buffer(filepath_or_buffer):
...
+def url_to_bytes(url_as_string, options):
+    """
+    Reads in URL and converts it to a byte stream
+
+    :param url_as_string: string to read as URL 
+    :type url_as_string: str
+    :return: true if string is a valid URL
+    :rtype: boolean
```
ChrisWallace2020, line 637: Copied docstring from above, need to make this one include proper return and also description/type of options here

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,54 @@ def is_stream_buffer(filepath_or_buffer):
...
+    stream = BytesIO()
+
+    verify_url = False
+    if 'verify' in options:
+        verify_url = options['verify']
+
+    with requests.get(url_as_string, stream=True, verify=verify_url) as url:
+        url.raise_for_status()
+        if 'Content-length' in url.headers and \
+            int(url.headers['Content-length']) >= 1024 ** 3:
```
ChrisWallace2020, line 648: Second line under `/` break should be tabbed forward 1 more, also its good practice to include the `and` with the 2nd line so that the second line reads `... and <bool>`

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,54 @@ def is_stream_buffer(filepath_or_buffer):
...
+
+            raise ValueError('The downloaded file from the url may not be \
+                larger than 1GB')
+
+        total_bytes = 0
+        c_size = 8192
+
+        for chunk in url.iter_content(chunk_size=8192):
```
ChrisWallace2020, line 656: Should probably pass `c_size` directly here in case the value ever gets changed.

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +16,60 @@ def test_read_data(self):
...
+    def setUpClass(cls):
+        test_dir = os.path.join(test_root_path, 'data')
+        cls.input_file_names = [
+            dict(path=os.path.join(test_dir, 'csv/diamonds.csv'),
+                 count=53940, delimiter=',', has_header=[0],
+                 num_columns=10, encoding='utf-8', data_type='csv'),
+            dict(path=os.path.join(test_dir, 'avro/users.avro'), count=4,  data_type='avro'),
+            dict(path=os.path.join(test_dir, 'json/iris-utf-16.json'), encoding='utf-16', count=150, data_type='json'),
+            dict(path=os.path.join(test_dir, 'parquet/iris.parq'), count=150, data_type='parquet'),
+            dict(path=os.path.join(test_dir, 'txt/code.txt'), count=150, data_type='text'),
```
ChrisWallace2020, line 31: Format these entries similar to the first one so that they all are under 80 chars

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -14,6 +16,60 @@ def test_read_data(self):
...
+                bytes = fp.read(c_size)
+                while bytes != b"":
+                    yield bytes
+                    bytes = fp.read(c_size)
+
+        for input_file in self.input_file_names:
+            mock_request_get.return_value.__enter__.return_value.iter_content.side_effect = \
+                lambda chunk_size: chunk_file(input_file['path'], chunk_size)
+            data_obj = Data('https://test.com')
+            self.assertEqual(data_obj.data_type, input_file['data_type'])
```
ChrisWallace2020, line 48: What exactly is this testing? Because this url isn't real so why would we expect the data_types to match all of the files in `self.input_file_names`? Some comments here would be really helpful.

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,57 @@ def is_stream_buffer(filepath_or_buffer):
...
+    :param url_as_string: string to read as URL 
+    :type url_as_string: str
+    :param options: options for the url 
+    :type options: dict
+    :return: BytesIO stream of data downloaded from URL
+    :rtype: BytesIO stream
+    """
+    stream = BytesIO()
+
+    verify_url = False
```
JGSweets, line 646: by default this should be true

dataprofiler/data_readers/data_utils.py
```diff
@@ -610,3 +612,57 @@ def is_stream_buffer(filepath_or_buffer):
...
+    :param options: options for the url 
+    :type options: dict
+    :return: BytesIO stream of data downloaded from URL
+    :rtype: BytesIO stream
+    """
+    stream = BytesIO()
+
+    verify_url = True
+    if 'verify' in options:
+        verify_url = options['verify']
```
JGSweets, line 648: while `verify` is clear here, it won't be at the top level for options, change to `verify_ssl`

JGSweets (approved):

ChrisWallace2020 (approved):

PR #376, opened 2021-08-02
Add median absolute deviation

AnhTruong:
- update from histogram
-- Subtract bin edges from median
-- Take absolute of bin edges and corresponding bin counts (flip negative part of the histogram around 0)
-- Iterate through two histograms until getting accumulated count at 0.5

- add option for median absolute deviation


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,162 @@ def _get_percentile(self, percentiles):
             quantiles[499] = median_value
         return quantiles
 
+    @BaseColumnProfiler._timeit(name="median_absolute_deviation")
```
JGSweets, line 958: remove timing

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -681,10 +683,13 @@ def _select_method_for_histogram(self, current_exact_var, current_est_var,
 
         return selected_method
 
-    def _histogram_to_array(self):
+    def _histogram_to_array(self, histogram=None):
```
JGSweets, line 686: why is this changing?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,162 @@ def _get_percentile(self, percentiles):
...
+        # find the break point to fold the deviation list
+        id_zero = None
+        for i in range(len(bin_edges)):
+            if bin_edges[i] > 0:
+                id_zero = i
+                break
+
+        # if all bin edges are positive (no break point),
+        # interpolate the histogram to find the median value
+        if id_zero is None:
```
JGSweets, line 988: in what case does this occur?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,162 @@ def _get_percentile(self, percentiles):
...
+            bin_counts = bin_counts.astype(float)
```
JGSweets, line 989: I believe this was already done above

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,162 @@ def _get_percentile(self, percentiles):
...
+            normalized_bin_counts = bin_counts / np.sum(bin_counts)
```
JGSweets, line 990: same above

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,162 @@ def _get_percentile(self, percentiles):
...
+            cumsum_bin_counts = np.cumsum(normalized_bin_counts)
```
JGSweets, line 991: this is going to  be expensive when we only want a single point out.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,162 @@ def _get_percentile(self, percentiles):
...
+
+            median_bin_inds = cumsum_bin_counts == 0.5
```
JGSweets, line 993: why do we need to check this if we already interp below?

dataprofiler/profilers/numerical_column_stats.py
JGSweets: can the same be the case for the opposite?

dataprofiler/profilers/numerical_column_stats.py
JGSweets: not sure if another reasonable way though rn

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, options=None):
         self.min = None
         self.max = None
         self.sum = 0
+        self.median_abs_dev = np.nan
```
JGSweets, line 56: remove

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +955,161 @@ def _get_percentile(self, percentiles):
             quantiles[499] = median_value
         return quantiles
 
+    def _get_median_abs_dev(self):
```
JGSweets, line 958: convert into a property

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: removed. That's my trial fix.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: good point. Yes, it's the same. Added that case.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: removed

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: don't know if we have another way. Should be within 1000 bins anyway.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: should we initialize it as nan?

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: this is to interpolate in case we have only positive or negative histogram. So, we don't have to go through the long loop below it. Btw, the median should be 0 in this case, which has been updated.

dataprofiler/profilers/profiler_options.py
```diff
@@ -391,6 +398,10 @@ def _validate_helper(self, variable_path='NumericalOptions'):
             errors.append("{}: The numeric stats must toggle on sum,"
                           " variance, and skewness if kurtosis is "
                           "toggled on.".format(variable_path))
+        if histogram_disabled and not mad_disabled:
+            errors.append("{}: The numeric stats must toggle on histogram "
+                          "and quantiles if median absolute deviation is "
+                          "toggled on.".format(variable_path))
```
ChrisWallace2020, line 404: Idk if I am just missing it but I think you should add this as a test case somewhere in the options test to ensure this gets caught.

dataprofiler/profilers/numerical_column_stats.py
ChrisWallace2020: No I think he's saying we should just remove it since `self.median_abs_deviation` is a property that calculates the value, and it gets plugged directly into the profile, so this attribute is never used, so it can be removed.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -950,6 +952,166 @@ def _get_percentile(self, percentiles):
...
+
+        bin_counts = self._stored_histogram['histogram']['bin_counts']
+        bin_edges = self._stored_histogram['histogram']['bin_edges']
+
+        # normalize bin counts
+        bin_counts = bin_counts.astype(float)
+        normalized_bin_counts = bin_counts / np.sum(bin_counts)
+
+        # calculate deviations from median
+        median = self.quantiles[499]
```
JGSweets, line 973: this is not a guarantee, should use _get_percentile

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -264,6 +265,7 @@ def profile(self):
             kurtosis=self.np_type_to_type(self.kurtosis),
             histogram=self._get_best_histogram_for_profile(),
             quantiles=self.quantiles,
+            median_absolute_deviation=self.median_abs_deviation,
```
JGSweets, line 268: update to match

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1112,6 +1112,150 @@ def median_abs_deviation(self):
...
+        # find the break point to fold the deviation list
+        id_zero = None
+        for i in range(len(bin_edges)):
+            if bin_edges[i] > 0:
+                id_zero = i
+                break
+
+        # if all bin edges are positive or negative (no break point),
+        # the median value is actually 0
+        if id_zero is None or id_zero == 1:
```
JGSweets, line 1145: Should the negative break point be `1` or `len(bin_edges) - 1`?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1112,6 +1112,150 @@ def median_abs_deviation(self):
...
+        for i in range(min(len(bin_counts_long), 1000)):
+            print(i, (bin_edges_long[i], bin_edges_long[i+1]), bin_counts_long[i])
+        for i in range(min(len(bin_counts_short), 1000)):
+            print(i, (bin_edges_short[i], bin_edges_short[i + 1]), bin_counts_short[i])
+        # iterate through two histograms until getting the points
+        # with cumsum counts as 0.5
+        cumsum_count, cumsum_count_prev = 0, 0
+        id = 0
+        bin_width = bin_edges_long[-1] - bin_edges_long[-2]
+        while id <= len(bin_counts_short) - 1 and cumsum_count < 0.5:
```
JGSweets, line 1176: could simplify to first part to: `id < len(bin_counts_short)` NBD though.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1112,6 +1112,150 @@ def median_abs_deviation(self):
...
+            [normalized_bin_counts[id_zero - 1] * (bin_edges[id_zero] /
+                (bin_edges[id_zero] - bin_edges[id_zero - 1]))],
+            normalized_bin_counts[id_zero:])
+        bin_edges_pos = np.append([0], bin_edges[id_zero:])
+
+        bin_counts_neg = np.append([normalized_bin_counts[id_zero - 1] -
+             normalized_bin_counts[id_zero - 1] * (bin_edges[id_zero] /
+                (bin_edges[id_zero] - bin_edges[id_zero - 1]))],
+            normalized_bin_counts[:id_zero - 1][::-1])
+        bin_edges_neg = np.append([0], -bin_edges[:id_zero][::-1])
```
JGSweets, line 1164: this is where you split to pos and negative, but also more importantly, split the bin if it is within a bin. I'm curious of the case when the bin occurs directly at a bin_edge.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1112,6 +1112,150 @@ def median_abs_deviation(self):
...
+                    counts_ahead = bin_counts_short[id]
+                    edges_behind = [bin_edges_long[id], bin_edges_long[id + 1]]
+                    counts_behind = bin_counts_long[id]
+
+                # check if the left part includes the median
+                cumsum_count_left = cumsum_count_prev
+                if id > 1:
+                    cumsum_count_left = cumsum_count_prev + \
+                        (edges_ahead[0] - edges_behind[0]) / bin_width * \
+                        counts_behind
```
JGSweets, line 1204: I'm concerned this isn't correct, for example:
if `id==2` bc when short/long 2 were added, cumsum was > 0.5, the left part would actually be an overlap of short 2 and long 1. This would need a subtraction of part of long 1 and included in the interpolation. the interpolation would have to start from the trailing edge of short 2 (which it does) and go to the leading edge of long 1 (i'm not sure it is doing this). Additionally, cumsum_prev would have to have the subtraction as mentioned before and hte left would need the addition  relative to both.

Currently, it is including all of the long part but if it falls within it, then the long part needs to be interpolated too.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: fixed with the new update

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: changed

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: I don't get it. Basically, id_zero is None with all edges are negative and 1 when all edges are positive

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: somewhat like you said, but to be precise:
- I specify behind vs ahead bins and counts: behind hist is the one with lower edges for any index
- the left part is the part starting from cumsum_prev up to the edges_ahead[0]
- the counts of the left part is a proportion of counts_behind with the ratio given by that formular
I'm confident that is the correct one

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1001,6 +1003,145 @@ def _get_percentile(self, percentiles):
...
+
+        bin_counts = self._stored_histogram['histogram']['bin_counts']
+        bin_edges = self._stored_histogram['histogram']['bin_edges']
+
+        # normalize bin counts
+        bin_counts = bin_counts.astype(float)
+        normalized_bin_counts = bin_counts / np.sum(bin_counts)
+
+        # calculate deviations from median
+        median = self._get_percentile([50])[0]
```
andrew-yin, line 1024: I think this can be modified after #389 is merged (using self.median). Also, maybe there should be an validate check in options to not compute this if median is disabled. Thoughts?

JGSweets: good point. would still store the value so not repeatedly called.

AnhTruong: I've added an if statement to process this

dataprofiler/profilers/numerical_column_stats.py
JGSweets: i above is equal to  a value within: `range(len(bin_edges))` which are integers between `0` and the `len(bin_edges) - 1`

JGSweets: I see so you are saying if all are positive, it is already > 0 at the first bin. but `None` bc the iteration no positive # existed.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1021,6 +1029,148 @@ def _get_percentile(self, percentiles):
             quantiles[percentiles == 50] = median_value
         return quantiles.tolist()
 
+    @property
+    def median_absolute_deviation(self):
```
JGSweets, line 1033: fix name to matchs scipy

AnhTruong: which existing? I made it consistent across the file and the option.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -73,6 +74,8 @@ def __init__(self, options=None):
             self.bias_correction = options.bias_correction.is_enabled
             self._top_k_modes = options.mode.top_k_modes
             self._median_is_enabled = options.median.is_enabled
+            self._median_abs_dev_is_enabled = \
+                options.median_absolute_deviation.is_enabled
```
JGSweets, line 78: this should also match and be `median_abs_deviation`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1021,6 +1029,179 @@ def _get_percentile(self, percentiles):
             quantiles[percentiles == 50] = median_value
         return quantiles.tolist()
 
+    @staticmethod
+    def fold_histogram(bin_counts, bin_edges, value):
```
JGSweets, line 1033: make private, no need for external use.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1021,6 +1029,179 @@ def _get_percentile(self, percentiles):
...
+
+        if len(bin_edges_neg) > 1 and bin_edges_neg[1] == 0:
+            bin_edges_neg = bin_edges_neg[1:]
+            bin_counts_neg = bin_counts_neg[1:]
+        return [bin_counts_pos, bin_edges_pos], [bin_counts_neg, bin_edges_neg]
+
+    @property
+    def median_absolute_deviation(self):
+        """
+        Get median absolute deviation.
```
JGSweets, line 1086: should mention how this is deriving it and that it is actually an estimate from the histogram and not an exact value of the dataset

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1021,6 +1029,102 @@ def _get_percentile(self, percentiles):
...
+        return [bin_counts_pos, bin_edges_pos], [bin_counts_neg, bin_edges_neg]
+
+    @property
+    def median_abs_deviation(self):
+        """
+        Get median absolute deviation.
+
+        :return: median absolute deviation
+        """
+        if not self._has_histogram or not self._median_abs_dev_is_enabled:
```
JGSweets, line 1090: will mostly work but will fail on data like: `[-9,-8,4,5,6,7,19,20]` where the median falls between multiple buckets because of averaging between them.

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: changed to median_abs_deviation

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: done

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: added

dataprofiler/profilers/numerical_column_stats.py
AnhTruong: agree, added the fix

AnhTruong: and added the test

JGSweets:
@AnhTruong this isn't WIP anymore right?

AnhTruong:
> @AnhTruong this isn't WIP anymore right?

No, I forgot to remove that label.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -12,7 +12,7 @@
 import copy
 import abc
 import warnings
-import sys
+from itertools import chain, zip_longest
```
lettergram, line 15: I think we may want to just import itertools and use the functions accordingly, that way we know what namespace they're from. I can imagine a function `chain`

AnhTruong: fixed

JGSweets (approved):
I do have concern we are choosing 0.5, but I think this approximation can be addressed in a future update.

lettergram (approved):

PR #378, opened 2021-08-02
Allow text to determine encoding automatically

JGSweets:
TextFiles previously only read ascii/utf-8 unless the user specified otherwise. Now the system will auto determine for Text the same way it does for CSVs and JSON. AVRO and Parquet do not determine file encoding.


dataprofiler/data_readers/base_data.py
```diff
@@ -93,7 +99,12 @@ def data_format(self, value):
     @property
     def file_encoding(self):
         if not self._file_encoding:
-            return sys.getdefaultencoding()
+            # set to default, detect if not StringIO
+            self._file_encoding = sys.getdefaultencoding()
```
lettergram, line 103: Shouldn't we detect the file encoding first?

Also, can this error? I'm not sure `sys.getdefaultcoding()` is correct here

https://stackoverflow.com/questions/15530635/why-is-sys-getdefaultencoding-different-from-sys-stdout-encoding-and-how-does

It appears to be discouraged 

https://stackoverflow.com/questions/3828723/why-should-we-not-use-sys-setdefaultencodingutf-8-in-a-py-script

dataprofiler/data_readers/data_utils.py
```diff
@@ -399,9 +401,9 @@ def _decode_is_valid(encoding):
                                         chunk_size=512, threshold=0.2,
                                         cp_isolation=None, cp_exclusion=None,
                                         preemptive_behaviour=True, explain=False)
-                result = result.best().first()
+                result = result.best()
             if result:
-                encoding = result.encoding
+                encoding = result.first().encoding
```
lettergram, line 406: What if there's not first?

dataprofiler/tests/data/txt/utf8.txt
```diff
@@ -1,4 +1,4 @@
-première is first
+première is first. ....
 première is slightly different
-Кириллица is Cyrillic
-𐐀 am Deseret
+Кириллица is Cyrillic. . . test
+𐐀 am Deseret. . test
```
lettergram, line 4: Do we test a blank file or file with just some enters or something?

dataprofiler/data_readers/data_utils.py
JGSweets: Currently, this is error'ing bc result itself is returning `null`, hence result.first() doesn't exist. maybe ill check for both bc i might be on a. newer version

dataprofiler/tests/data/txt/utf8.txt
JGSweets: adding `empty.txt` to the file check list for our tests

dataprofiler/data_readers/base_data.py
JGSweets: updated to use locale and fallback on sys.getfilesystemencoding

dataprofiler/data_readers/csv_data.py
```diff
@@ -581,15 +576,16 @@ def is_match(cls, file_path, options=None):
         if options is None:
             options = dict()
 
-        file_encoding = None
-        if not isinstance(file_path, StringIO):
-            file_encoding = data_utils.detect_file_encoding(file_path=file_path)
+        if 'encoding' not in options and not isinstance(file_path, StringIO):
+            options['encoding'] = data_utils.detect_file_encoding(
```
ChrisWallace2020, line 580: Is it appropriate to be modifying options that the user passes us rather than simply executing logic in accordance with them? I don't know if this occurs elsewhere but it feels strange to change the options as they are an indication of user intention. I am open to discussion though.

JGSweets: it is a technique we use so that when we pass information to the Profiler, the options don't have to recompute the encoding. Also, note that it doesn't overwrite the options.

JGSweets: In future, we could do something like:
```python
is_match = BlankData.is_match(..., options)
# or
is_match, options = BlankData.is_match(..., options, update_options=True)
```

dataprofiler/data_readers/base_data.py
```diff
@@ -93,7 +100,22 @@ def data_format(self, value):
...
+            # get system default, but if set to ascii, just update to utf-8
+            file_encoding = "utf-8"
+            try:
+                file_encoding = locale.getpreferredencoding(False)
+            except:
+                file_encoding = sys.getfilesystemencoding()
+            finally:
+                if file_encoding.lower() in ['ANSI_X3.4-1968', 'ascii']:
+                    file_encoding = 'utf-8'
+                self._file_encoding = file_encoding
```
lettergram, line 112: Shouldn't this be tabbed out?

lettergram (approved):

dataprofiler/data_readers/base_data.py
```diff
@@ -93,7 +100,22 @@ def data_format(self, value):
...
         if not self._file_encoding:
-            return sys.getdefaultencoding()
+            # get system default, but if set to ascii, just update to utf-8
+            file_encoding = "utf-8"
+            try:
+                file_encoding = locale.getpreferredencoding(False)
+            except:
+                file_encoding = sys.getfilesystemencoding()
+            finally:
+                if file_encoding.lower() in ['ANSI_X3.4-1968', 'ascii']:
```
AnhTruong, line 110: can lowercased string be ever in uppercased `'ANSI_X3.4-1968'`

dataprofiler/tests/data_readers/test_base_data.py
```diff
@@ -38,3 +78,27 @@ def func1(self):
...
+
+            data = BaseData(input_file_path=input_file['path'], data=None,
+                            options={})
+            self.assertEqual(input_file['encoding'].lower(),
+                             data.file_encoding.lower(),
+                             input_file['path'])
+
+        # test when data is specified without input_file_object
+        file_encoding = locale.getpreferredencoding(False)
+        if file_encoding.lower() in ['ascii', 'ANSI_X3.4-1968']:
```
AnhTruong, line 100: same question as above

AnhTruong (commented):
just two small comments. Otherwise lgtm

dataprofiler/tests/data_readers/test_base_data.py
JGSweets: good find

AnhTruong (approved):

dataprofiler/data_readers/base_data.py
JGSweets: fixed

ChrisWallace2020 (approved):

PR #380, opened 2021-08-03
Added URL Documentation

gautomdas:
Added documentation on how to read a URL in.


docs/source/index.rst
```diff
@@ -171,6 +171,7 @@ Supported Data Formats
 * Parquet file
 * Text file
 * Pandas DataFrame
+* URL string
```
JGSweets, line 174: * A URL that points to one of the supported file types above

JGSweets: match the above you have

JGSweets (approved):

PR #382, opened 2021-08-03
Add mode to numeric stats

andrew-yin:
Uses the histogram to estimate the mode of the data.


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -52,6 +52,8 @@ def __init__(self, options=None):
                              "of type NumericalOptions.")
         self.min = None
         self.max = None
+        self.mode = np.nan
+        self.mode_count = None # By default, return all modes
```
JGSweets, line 56: i would make this private and set it to `_max_k_modes'

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -110,6 +114,7 @@ def __init__(self, options=None):
             "histogram_and_quantiles":
                 NumericStatsMixin._get_histogram_and_quantiles,
             "num_zeros": NumericStatsMixin._get_num_zeros,
+            "mode": NumericStatsMixin._get_mode,
```
JGSweets, line 117: instead of calculating at profile time, I think we can do this at report time (since so simple). we won't need to time it then.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -574,6 +588,33 @@ def _correct_bias_kurtosis(match_count, biased_kurtosis):
...
+
+        # Get the bin(s) with the highest frequency
+        highest_idxs = np.argwhere(bin_counts == bin_counts.max()).flatten()
+        if self.mode_count is not None and len(highest_idxs) > self.mode_count:
+            highest_idxs = highest_idxs[:self.mode_count]
+
+        mode = (bin_edges[highest_idxs] + bin_edges[highest_idxs + 1]) / 2
+        # return single number instead of list if only one mode
+        if len(mode) == 1:
+            mode = mode[0]
```
JGSweets, line 615: hmm, problem is this changes the data type depending on result which might be confusing. maybe just leave it as list.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -264,7 +264,7 @@ def test_profiled_mode(self):
         df = pd.Series([1, 1, 1, 1, 1, 1, 1]).apply(str)
         profiler = FloatColumn(df.name)
         profiler.update(df)
-        self.assertAlmostEqual(1, profiler.mode, places=2)
+        np.testing.assert_array_almost_equal([1], profiler.mode, decimal=2)
```
JGSweets, line 267: what is the value such that this is not just `1`

andrew-yin: Could you clarify your question?

JGSweets: why are we having to assert almost equal when the mode should be `[1]`

`self.assertListEqual([1], profiler.mode)`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -307,6 +319,10 @@ def mean(self):
             return 0
         return float(self.sum) / self.match_count
 
+    @property
+    def mode(self):
+        return self._estimate_mode_from_histogram()
```
JGSweets, line 324: this is abstracting it, but do you think it is necessary?

Also, add a function description.

andrew-yin: I think there could be potential future features that may need access to the mode, in which case it could be easier to call and use with this.

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: The mode is estimating it off the histogram, so I believe it's not exactly equal to one as the bin it falls in is 1.00, 1.002 or something similar.

JGSweets: in this case, i believe it should be exact since just 1 value in the column. Otherwise might be different.

I just find the `assert_array_almost_equal` strange as it is setting a tolerance around it which is variable to the size of the input values and not normalized.

JGSweets: thoughts?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -803,8 +832,8 @@ def test_profile(self):
             },
             times=defaultdict(float, {'histogram_and_quantiles': 1.0,
                                       'precision': 1.0, 'max': 1.0, 'min': 1.0,
-                                      'skewness': 1.0, 'kurtosis': 1.0,
-                                      'sum': 1.0, 'variance': 1.0,
+                                      'skewness': 1.0,
```
JGSweets, line 835: any reason for the move here?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -68,6 +69,8 @@ def __init__(self, options=None):
         self.num_negatives = 0
         if options:
             self.bias_correction = options.bias_correction.is_enabled
+            if options.mode.count is not None:
```
JGSweets, line 72: `_max_k_modes`?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -68,6 +69,8 @@ def __init__(self, options=None):
...
+                self._max_k_modes = options.mode.count
```
JGSweets, line 73: `_max_k_modes`?

dataprofiler/profilers/numerical_column_stats.py
JGSweets: I would suggest the option matching the internal variable.

dataprofiler/tests/profilers/test_float_column_profile.py
andrew-yin: Must've been an accidental refactor, will revert

JGSweets (commented):
Also, don't forget to add to readme for report.

dataprofiler/tests/profilers/test_float_column_profile.py
AnhTruong: for this case specifically with unique values, it is exactly 1, but for other cases, using almost equal is fine to me.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -258,6 +259,33 @@ def test_profiled_max(self):
...
+        profiler = FloatColumn(df.name)
+        profiler.update(df)
+        np.testing.assert_array_almost_equal([1.25], profiler.mode, decimal=2)
+
+        # with negative values
+        df = pd.Series([-1.1, 1.9, 1.9, 1.9, 2.1, 2.01, 2.01, 2.01]).apply(str)
+        profiler = FloatColumn(df.name)
+        profiler.update(df)
+        np.testing.assert_array_almost_equal([1.9, 2.01], profiler.mode,
+                                             decimal=2)
```
AnhTruong, line 287: should we add the case where all values are unique?

AnhTruong: also, should we add test for different values of the option top_k_modes?

AnhTruong (commented):
some small comments

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -574,6 +596,30 @@ def _correct_bias_kurtosis(match_count, biased_kurtosis):
                    (biased_kurtosis + 3) - 3 * (match_count - 1))
         return kurtosis
 
+    def _estimate_mode_from_histogram(self):
+        """
+        Estimates the mode of the current data using the
+        histogram. If there are multiple modes, returns
+        K of them (where K is defined in options given, but
+        1 by default)
```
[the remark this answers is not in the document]
andrew-yin: This documentation needs to be updated

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, options=None):
                              "of type NumericalOptions.")
         self.min = None
         self.max = None
+        self._max_k_modes = None # By default, return all modes
```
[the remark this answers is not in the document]
andrew-yin: I can set the default to 5 for now.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, options=None):
                              "of type NumericalOptions.")
         self.min = None
         self.max = None
+        self._top_k_modes = 5 # By default, return all modes
```
JGSweets, line 55: are we assuming 5 is all modes by default?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -574,6 +590,41 @@ def _correct_bias_kurtosis(match_count, biased_kurtosis):
...
+            return np.nan
+
+        bin_counts = self._stored_histogram['histogram']['bin_counts']
+        bin_edges = self._stored_histogram['histogram']['bin_edges']
+
+        # Get the K bin(s) with the highest frequency (one-pass):
+        cur_max = -1
+        highest_idxs = []
+        count = 0
+        for i in range(0, len(bin_counts)):
```
JGSweets, line 613: not a big deal, bu tin future, no need for the `0` in the `range`

dataprofiler/profilers/numerical_column_stats.py
andrew-yin: oops, need to update documentation

JGSweets (approved):

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -574,6 +590,41 @@ def _correct_bias_kurtosis(match_count, biased_kurtosis):
...
+            if bin_counts[i] > cur_max:
+                highest_idxs = [i]
+                cur_max = bin_counts[i]
+                count = 1
+            elif bin_counts[i] == cur_max and count < self._top_k_modes:
+                highest_idxs.append(i)
+                count += 1
+            if count == self._top_k_modes:
+                    break
```
ChrisWallace2020, line 622: This break is tabbed too far forward

JGSweets: good catch

JGSweets: this needs to be a continue because if we had top_k_5 and we had the bin setup it would give the wrong result:
`bin_counts = [5] * 5 + [6] * 6` it would breakout after the 5th instance of 5 when the mode has a count of 6.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -574,6 +590,41 @@ def _correct_bias_kurtosis(match_count, biased_kurtosis):
...
+
+        # Get the K bin(s) with the highest frequency (one-pass):
+        cur_max = -1
+        highest_idxs = []
+        count = 0
+        for i in range(0, len(bin_counts)):
+            if bin_counts[i] > cur_max:
+                highest_idxs = [i]
+                cur_max = bin_counts[i]
+                count = 1
```
ChrisWallace2020, line 617: Not critical, but I would leave comments explaining why the `count` and `highest_idxs` are being reset here, since I thought `_top_k_modes` meant the top `k` most frequent data points, not the top `k` data points of the (singular) highest frequency.

JGSweets (approved):

ChrisWallace2020 (approved):

PR #384, opened 2021-08-05
Catch Divide by 0 bug for unique row ratio

ChrisWallace2020:
Calculating unique row ratio with 0 samples raises divide by 0 error, catch it in this case so no error is raised

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1555,7 +1555,8 @@ class StructuredProfiler(BaseProfiler):
         return _prepare_report(report, output_format, omit_keys)
 
     def _get_unique_row_ratio(self):
-        return len(self.hashed_row_dict) / self.total_samples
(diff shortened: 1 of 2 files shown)


JGSweets (commented):
Add test

dataprofiler/profilers/profile_builder.py
```diff
@@ -1555,15 +1555,19 @@ def report(self, report_options=None):
         return _prepare_report(report, output_format, omit_keys)
 
     def _get_unique_row_ratio(self):
-        return len(self.hashed_row_dict) / self.total_samples
+        if self.total_samples:
+            return len(self.hashed_row_dict) / self.total_samples
+        return 0
```
ChrisWallace2020, line 1560: This is the only new logic, the changes below are simply refactors

JGSweets (approved):

AnhTruong (approved):

PR #392, opened 2021-08-16
Add chi-squared tests to profile

andrew-yin:
Computes a matrix of p-values of chi-squared tests for homogeneity between all categorical columns.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1878,8 +1878,7 @@ class StructuredProfiler(BaseProfiler):
             :param col_idx: Column index
             :type col_idx: int
             """
-            compiler = self._profile[col_idx]
-            data_stats_compiler = compiler.profiles["data_stats_profile"]
+            data_stats_compiler = self._profile[col_idx].profiles["data_stats_profile"]
             profiler = data_stats_compiler._profiles["category"]
             return profiler if profiler.is_match else None
 
@@ -1888,22 +1887,24 @@ class StructuredProfiler(BaseProfiler):
         chi2_mat = np.full((n_cols, n_cols), np.nan)
         # Compute chi_sq for each
         for i in range(n_cols):
-            profile1 = _get_categorical_profile(i)
-            if profile1 is None:
+            data_stats_compiler1 = self._profile[i].profiles["data_stats_profile"]
+            profiler1 = data_stats_compiler1._profiles["category"]
+            if not profiler1.is_match:
                 continue
             for j in range(i, n_cols):
                 if i == j:
                     chi2_mat[i][j] = 1
                     continue
-                profile2 = _get_categorical_profile(j)
-                if profile2 is None:
+                data_stats_compiler2 = self._profile[j].profiles["data_stats_profile"]
+                profiler2 = data_stats_compiler2._profiles["category"]
+                if profiler2 is None:
                     continue
 
                 results = utils.perform_chi_squared_test(
-                    profile1.categorical_counts,
-                    profile1.sample_size,
-                    profile2.categorical_counts,
-                    profile2.sample_size
+                    profiler1.categorical_counts,
+                    profiler1.sample_size,
+                    profiler2.categorical_counts,
+                    profiler2.sample_size
                 )
                 chi2_mat[i][j] = results["p-value"]
                 chi2_mat[j][i] = results["p-value"]
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1888,7 +1888,7 @@ class StructuredProfiler(BaseProfiler):
                 if not profiler2.is_match:
                     continue
 
-                results = utils.perform_chi_squared_test(
+                results = utils.perform_chi_squared_test_for_homogeneity(
                     profiler1.categorical_counts,
                     profiler1.sample_size,
                     profiler2.categorical_counts,
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1869,7 +1869,7 @@ class StructuredProfiler(BaseProfiler):
         Calculates the p-value from a chi-squared test
         for homogeneity between all categorical columns.
         """
-        def _get_categorical_profile(self, col_idx):
+        def _get_categorical_profile(col_idx):
             """
             Returns the CategoricalColumn associated with the given
             column index. If the column is not determined to be categorical,
@@ -1888,14 +1888,14 @@ class StructuredProfiler(BaseProfiler):
         chi2_mat = np.full((n_cols, n_cols), np.nan)
         # Compute chi_sq for each
         for i in range(n_cols):
-            profile1 = self._get_categorical_profile(i)
+            profile1 = _get_categorical_profile(i)
             if profile1 is None:
                 continue
             for j in range(i, n_cols):
                 if i == j:
                     chi2_mat[i][j] = 1
                     continue
-                profile2 = self._get_categorical_profile(j)
+                profile2 = _get_categorical_profile(j)
                 if profile2 is None:
                     continue
 
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1553,7 +1553,7 @@ class StructuredProfiler(BaseProfiler):
                 "file_type": self.file_type,
                 "encoding": self.encoding,
                 "correlation_matrix": self.correlation_matrix,
-                "chi2_matrix": self.chi2_matrix,
+                "chi2_homogeneity_matrix": self.chi2_matrix,
                 "profile_schema": defaultdict(list),
                 "times": self.times,
             }),
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1865,45 +1865,51 @@ class StructuredProfiler(BaseProfiler):
         return corr_mat
 
     def _update_chi2(self):
+        """
+        Calculates the p-value from a chi-squared test
+        for homogeneity between all categorical columns.
+        """
         n_cols = len(self._profile)
-        cat_cols = {}
-        cat_col_idxs = []
-
-        # Get categorical columns
-        for i in range(n_cols):
-            compiler = self._profile[i]
-            data_stats_compiler = compiler.profiles["data_stats_profile"]
-            profiler = data_stats_compiler._profiles["category"]
-            if profiler.is_match:
-                categories = profiler._categories
-                sample_size = profiler.sample_size
-                cat_cols[i] = (categories, sample_size)
-                cat_col_idxs.append(i)
 
         # Fill matrix with nan initially
         chi2_mat = np.full((n_cols, n_cols), np.nan)
         # Compute chi_sq for each
-        for i in range(len(cat_col_idxs)):
-            for j in range(i, len(cat_col_idxs)):
+        for i in range(n_cols):
+            profile1 = self._get_categorical_profile(i)
+            if profile1 is None:
+                continue
+            for j in range(i, n_cols):
                 if i == j:
                     chi2_mat[i][j] = 1
                     continue
-                idx1 = cat_col_idxs[i]
-                idx2 = cat_col_idxs[j]
-
-                categories1 = cat_cols[idx1][0]
-                categories2 = cat_cols[idx2][0]
-                sample_size1 = cat_cols[idx1][1]
-                sample_size2 = cat_cols[idx2][1]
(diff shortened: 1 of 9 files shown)


dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -120,14 +129,15 @@ def test_validate_helper(self):
...
+        option.chi2 = StructuredOptions()
 
         expected_error = set()
         for key in self.boolean_keys:
             ckey = key.capitalize()
             if key == "data_labeler": ckey = "DataLabeler"
             elif key == "category": ckey = "Categorical"
             elif key == "datetime": ckey = "DateTime"
-            if key == "multiprocess":
+            if key == "multiprocess" or key == "chi2":
```
JGSweets, line 140: interesting, how does `correlation` not cause an error here.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1857,6 +1864,47 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
 
         return corr_mat
 
+    def _update_chi2(self):
```
JGSweets, line 1867: Please add a description.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1857,6 +1864,47 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
...
+        cat_cols = {}
+        cat_col_idxs = []
+
+        # Get categorical columns
+        for i in range(n_cols):
+            compiler = self._profile[i]
+            data_stats_compiler = compiler.profiles["data_stats_profile"]
+            profiler = data_stats_compiler._profiles["category"]
+            if profiler.is_match:
+                categories = profiler._categories
```
JGSweets, line 1878: let's not use the private variable here and just use `cateogries`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1857,6 +1864,47 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
...
+                if i == j:
+                    chi2_mat[i][j] = 1
+                    continue
+                idx1 = cat_col_idxs[i]
+                idx2 = cat_col_idxs[j]
+
+                categories1 = cat_cols[idx1][0]
+                categories2 = cat_cols[idx2][0]
+                sample_size1 = cat_cols[idx1][1]
+                sample_size2 = cat_cols[idx2][1]
```
JGSweets, line 1897: these 2 can be done in a single line via:
`categories1, sample_size1 = cat_cols[idx1]`
likewise for the set 2

dataprofiler/profilers/profile_builder.py
```diff
@@ -1857,6 +1864,47 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
 
         return corr_mat
 
+    def _update_chi2(self):
+        n_cols = len(self._profile)
+        cat_cols = {}
+        cat_col_idxs = []
+
+        # Get categorical columns
+        for i in range(n_cols):
```
JGSweets, line 1873: I wonder if it would be better to just loop through the columns instead of collecting them each time and then looping through the collected categories.

something like, define an internal function which gets a categorical column or returns None
```python
def get_cateogrical_profile_if_match(col):
    data_stats_compiler = compiler.profiles["data_stats_profile"]
     profiler = data_stats_compiler._profiles["category"]
     return profiler if profiler.is_match else None
```
then you could loop through the columns
```
for i in range(n_cols):
    col1_cat = get_cateogrical_profile_if_match(self._profile[i])
    if col1_cat is None:
        continue
    for j in range(i, n_cols):
        col2_cat = get_cateogrical_profile_if_match(self._profile[j])
        .
        .
        .
```
thoughts? 
it minimizes the code at least to a specific area, but not sure that's necessary.

dataprofiler/profilers/profile_builder.py
andrew-yin: I use _categories to get the counts. Any other ways to do this?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1857,6 +1864,53 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
...
+        Calculates the p-value from a chi-squared test
+        for homogeneity between all categorical columns.
+        """
+        n_cols = len(self._profile)
+
+        # Fill matrix with nan initially
+        chi2_mat = np.full((n_cols, n_cols), np.nan)
+        # Compute chi_sq for each
+        for i in range(n_cols):
+            profile1 = self._get_categorical_profile(i)
```
JGSweets, line 1878: i would just define it within this function as i don't think _get_categorical_profile would be used anywhere else.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1857,6 +1864,52 @@ def _merge_correlation_helper(corr_mat1, mean1, std1, n1,
...
+        def _get_categorical_profile(col_idx):
+            """
+            Returns the CategoricalColumn associated with the given
+            column index. If the column is not determined to be categorical,
+            returns None.
+
+            :param col_idx: Column index
+            :type col_idx: int
+            """
+            compiler = self._profile[col_idx]
```
JGSweets, line 1881: I think this is actually the column profiler not the compiler which is below as `data_stats_compiler`. 

Do you think it would it make more sense to pass in the the column profiler into this function vs the index? Would the code below become more intuitive?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -2777,7 +2939,9 @@ def test_profiler_factory_class_creates_correct_profiler(self, *mocks):
         data_str = 'test'
         self.assertIsInstance(Profiler(data_str), UnstructuredProfiler)
 
-    def test_save_and_load_structured(self):
+    @mock.patch('dataprofiler.profilers.profile_builder.'
+                'StructuredProfiler._update_chi2')
+    def test_save_and_load_structured(self, *mocks):
```
JGSweets, line 2944: What's happening here? to mock out the tests seems dangerous as it if saving is failing in the tests, it likely will on the real data too. 

Is there a reason we need to mock it out?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -963,8 +1122,10 @@ def test_min_true_samples(self, *mocks):
         # Test valid input
         profile = dp.StructuredProfiler(empty_df, min_true_samples=10)
         self.assertEqual(10, profile._min_true_samples)
-        
-    def test_save_and_load(self):
+
+    @mock.patch('dataprofiler.profilers.profile_builder.'
```
JGSweets, line 1126: What's happening here? to mock out the tests seems dangerous as it if saving is failing in the tests, it likely will on the real data too. 

Is there a reason we need to mock it out?

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: in clean reports hist/correlations is updated so that it can be tested, does that need to happen for chi_2 too? `clean_report`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -2778,7 +2965,7 @@ def test_profiler_factory_class_creates_correct_profiler(self, *mocks):
         self.assertIsInstance(Profiler(data_str), UnstructuredProfiler)
 
     def test_save_and_load_structured(self):
-        datapth = "dataprofiler/tests/data/"
```
JGSweets, line 2781: fix

JGSweets (approved):

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -603,6 +605,189 @@ def test_correlation_update(self):
...
+
+        # Single column
+        data = pd.DataFrame({'a': ["y", "y", "n", "n", "y"]})
+        profiler = dp.StructuredProfiler(data)
+        expected_mat = np.array([1])
+        self.assertEqual(expected_mat, profiler.chi2_matrix)
+
+        data = pd.DataFrame({'a': ["y", "y", "y", "y", "n", "n", "n"],
+                             'b': ["y", "maybe", "y", "y", "n", "n", "maybe"],
+                             'c': ["n", "maybe", "n", "n", "n", "y", "y"]})
```
AnhTruong, line 626: do we deal with non values in this PR?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -603,6 +605,189 @@ def test_correlation_update(self):
...
+                'ColumnDataLabelerCompiler')
+    @mock.patch('dataprofiler.profilers.profile_builder.DataLabeler',
+                spec=StructuredDataLabeler)
+    def test_merge_chi2(self, *mocks):
+        data = pd.DataFrame({'a': ["y", "y", "y", "y", "n", "n", "n"],
+                             'b': ["y", "maybe", "y", "y", "n", "n", "maybe"],
+                             'c': ["n", "maybe", "n", "n", "n", "y", "y"]})
+
+        data1 = data[:4]
+        data2 = data[4:]
```
AnhTruong, line 675: can we add merge from 
- 2 empty data
- 1 empty and 1 non-empty

AnhTruong (commented):
some comments

dataprofiler/profilers/profiler_options.py
```diff
@@ -1019,6 +1019,8 @@ def __init__(self, null_values=None):
         :vartype data_labeler: DataLabelerOptions
         :ivar correlation: option set for correlation profiling.
         :vartype correlation: CorrelationOptions
+        :ivar chi2: option set for chi2 matrix
```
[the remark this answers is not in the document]
JGSweets: chi squared is a known test, i'd rather not change it personally.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: non values? the categorical profiler should not receive nan values if that's what you mean.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: our correlation tests do this too, right?

dataprofiler/tests/profilers/test_profile_builder.py
andrew-yin: Agreed with Jeremy, NaN values shouldn't affect the chi2 calculation.

dataprofiler/tests/profilers/test_profile_builder.py
andrew-yin: Merging doesn't allow for empty data:
```python
# In the case of __add__ with one of the schemas not initialized
if strict and (len_schema1 == 0 or len_schema2 == 0):
    raise ValueError("Cannot merge empty profiles.")
```

dataprofiler/profilers/profiler_options.py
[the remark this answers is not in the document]
AnhTruong: bc chi2 is specific method, but correlation is more for the goal. Ideally, I'd prefer numeric_correlation and categorical_correlation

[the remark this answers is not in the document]
AnhTruong: probably for the next PR

dataprofiler/profilers/utils.py
```diff
@@ -614,3 +615,67 @@ def wrapper(self, *args, **kw):
...
+        results["chi2-statistic"] = np.inf
+        results["p-value"] = 0
+        return results
+
+    # Calculate chi-sq statistic
+    chi2_statistic = 0
+    for cat, count in cat_counts.items():
+        expected1 = sample_size1 * count / total
+        expected2 = sample_size2 * count / total
+        chi2_statistic += (categories1.get(cat, 0) - expected1) \
```
AnhTruong, line 671: do we have to take into account the co-occurrence of two categories from two columns, e.g., observed values from the contingency table in https://en.wikipedia.org/wiki/Chi-squared_test

dataprofiler/tests/profilers/test_profile_builder.py
AnhTruong: I think we can use the same logic as with correlation
```
corr_mat1 = self.correlation_matrix
        corr_mat2 = other.correlation_matrix
        n1 = self.total_samples - self.row_is_null_count
        n2 = other.total_samples - other.row_is_null_count
        if n1 == 0:
            return corr_mat2
        if n2 == 0:
            return corr_mat1

            return None
```

AnhTruong (approved):

JGSweets (approved):

PR #395, opened 2021-08-19
Raise known exception when ssl_verify fails so user knows they can correct it.

andrew-yin:
Catches the SSLError that occurs if verify = True

### DIFF
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -668,9 +668,10 @@ def url_to_bytes(url_as_string, options):
 
                 if total_bytes > 1024 ** 3:
                     raise ValueError('The downloaded file from the url may not be larger than 1GB')
-    except requests.exceptions.SSLError:
-        raise ValueError("The URL given has an untrusted SSL certificate. Although highly discouraged, "
-                         "you can proceed with profiling by setting 'verify_url' to False in options.")
+    except requests.exceptions.SSLError as e:
+        raise RuntimeError("The URL given has an untrusted SSL certificate. Although highly discouraged, "
+                           "you can proceed with reading the data by setting 'verify_url' to False in "
+                           "options (i.e. options=dict(verify_url=False)).") from e
 
     stream.seek(0)
     return stream
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -651,22 +651,28 @@ def url_to_bytes(url_as_string, options):
     if 'verify_ssl' in options:
         verify_url = options['verify_ssl']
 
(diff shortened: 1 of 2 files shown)


dataprofiler/data_readers/data_utils.py
```diff
@@ -651,22 +651,26 @@ def url_to_bytes(url_as_string, options):
...
+            for chunk in url.iter_content(chunk_size=c_size):
+                stream.write(chunk)
+                total_bytes += c_size
 
-            if total_bytes > 1024 ** 3:
-                raise ValueError('The downloaded file from the url may not be larger than 1GB')
+                if total_bytes > 1024 ** 3:
+                    raise ValueError('The downloaded file from the url may not be larger than 1GB')
+    except requests.exceptions.SSLError:
+        raise ValueError("The URL given has an untrusted SSL certificate. Although highly discouraged, "
```
JGSweets, line 672: I think i like the exception thrown by: https://stackoverflow.com/questions/696047/re-raise-exception-with-a-different-type-and-message-preserving-existing-inform
where instead it basically still throws the previous error, so the user can debug it, but also throws its own error:
```python
try:
    ...
except requests.exceptions.SSLError as e:
        raise RuntimeError(.....) from e
```
I chose `RuntimeError` because i think it clarifies it wasn't necessary something wrong with user values, but something happened during runtime and will also still trigger the `SSLError`.
Additionally, i might mention in the description e.g. `options=dict(verify_url=False)`. Not sure this last part is necessary, but may make it more clear.

dataprofiler/data_readers/data_utils.py
```diff
@@ -651,22 +651,26 @@ def url_to_bytes(url_as_string, options):
...
+                         "you can proceed with profiling by setting 'verify_url' to False in options.")
```
JGSweets, line 673: this is specifically for reading data so instead of `profiling by ...` maybe `reading data by setting the dp.Data's option dict to contain {'verify_url': False}`.
Or something similar.

dataprofiler/tests/data_readers/test_data.py
```diff
@@ -113,5 +114,14 @@ def test_read_url_header_overflow(self, mock_request_get):
             # stub URL, mock_request_get replaces the content requests.get will see
             data_obj = Data('https://test.com')
 
+    @mock.patch('requests.get')
+    def test_read_url_verify_ssl(self, mock_request_get):
+        mock_request_get.side_effect = requests.exceptions.SSLError()
+
+        with self.assertRaisesRegex(ValueError, \
+            "The URL given has an untrusted SSL certificate."):
```
JGSweets, line 122: Tab should be inline with the error. Also, should we test the whole string?

JGSweets (approved):

lettergram (approved):

PR #404, opened 2021-08-24
refactor: update histograms to allow column ind specification

JGSweets:
Allows user to specify column inds and fixes a bug

### DIFF
--- a/dataprofiler/reports/graphs.py
+++ b/dataprofiler/reports/graphs.py
@@ -7,7 +7,7 @@ import numpy as np
 import seaborn as sns
 
 
-def plot_histograms(profiler, columns=None):
+def plot_histograms(profiler, columns=None, column_inds=None):
     """
     Take a input of StructuredProfiler class and a list of specified column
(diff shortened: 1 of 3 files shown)


dataprofiler/reports/graphs.py
```diff
@@ -7,25 +7,42 @@
...
     :type profiler: StructuredProfiler
-    :param columns: list of column names to be plotted
-    :type columns: list
-    :return: returns fig
-    :rtype: fig
+    :param columns: List of column names to be plotted. Can only specify columns
+        or column_inds, but not both
+    :type columns: list[Union[int,str]]
+    :param column_inds: List of column indexes to be plotted
+    :type column_inds: list[int]
```
lettergram, line 22: Why is this necessary, given you can control the `columns` that are being passed in?

JGSweets: columns may have duplicates, but column indexes are unique, updated naming to fix

PR #405, opened 2021-08-30
fixes for csv reading of html and verify_ssl naming

JGSweets:
Error and naming of verify_ssl was different, this is fixed.

Reading data as a string would not skip lines which were purely blank and hence limited data resulted in HTML being read as CSV.


dataprofiler/data_readers/csv_data.py
```diff
@@ -625,7 +625,7 @@ def is_match(cls, file_path, options=None):
                 count = len([i.start() for i in re.finditer(delimiter_regex, line)])
             else:
                 # If no delimiter, see if spaces are regular intervals
-                count = len([i.start() for i in re.finditer(space_regex, line)])
+                count = len([i.start() for i in re.finditer(space_regex, line.rstrip())])
```
JGSweets, line 628: fixes issue with delimiters being counted when empty to its right.

dataprofiler/data_readers/data_utils.py
```diff
@@ -570,12 +570,12 @@ def load_as_str_from_file(file_path, file_encoding=None, max_lines=10,
 
         sample_size_bytes = min(max_bytes, chunk_size_bytes)
         
-        # Read the file until the appropriate number of occurances                    
+        # Read the file until the appropriate number of occurrences
```
JGSweets, line 573: spelling

dataprofiler/data_readers/data_utils.py
```diff
@@ -667,11 +672,14 @@ def url_to_bytes(url_as_string, options):
                 total_bytes += c_size
 
                 if total_bytes > 1024 ** 3:
-                    raise ValueError('The downloaded file from the url may not be larger than 1GB')
```
JGSweets, line 670: pep8

dataprofiler/data_readers/csv_data.py
lettergram: Could it be empty to its left?

dataprofiler/data_readers/data_utils.py
```diff
@@ -560,44 +560,47 @@ def load_as_str_from_file(file_path, file_encoding=None, max_lines=10,
...
                 data_as_str += sample_lines[:loc].decode(file_encoding)
             else:
                 data_as_str += sample_lines[:loc]
 
-            total_occurances += occurance
-            if total_occurances >= max_lines:
+            # count minus blank lines
+            total_occurrences += (
+                    occurrence
+                    - sample_lines[:loc].count(search_query_value * 2))
```
AnhTruong, line 602: why *2 here?

dataprofiler/data_readers/csv_data.py
AnhTruong: did we have the test for this one?

AnhTruong (commented):
some comments

dataprofiler/data_readers/data_utils.py
JGSweets: two `\n\n` would indicate an empty line

dataprofiler/data_readers/csv_data.py
JGSweets: added

lettergram: still no general stripping? `line.strip()`

JGSweets: I believe we are only addressing the case where it is a bunch of whitespace tacked on to the end, if we did both sides it might mess up the count.

lettergram: I see, I don't know if pulling data from the right is appropriate either? 

Was there a particular error that occurred?

JGSweets: Yes, specifically the test I added. where it was a column of numbers but an extra space existed after the numbers

AnhTruong (approved):

lettergram (approved):

Issue #411, opened 2021-09-13
Adding custom regex for custom labels


ian-contiamo:
**Describe the outcome you'd like:**

I would like to add custom regex patterns corresponding to custom labels. For instance, I would add a regex to recognize German passport numbers, and add the corresponding `GERMAN_PASSPORT` label.

Is this possible at the moment? Or is it a feature you have on your roadmap?

**Additional context:**

Regex models are implemented in the code, but I see no obvious way of adding new regex patterns. In your documentation, the emphasis is on training and adding new neural networks, but adding custom regex detection would be a much simpler way to customize and extend labeling.

JGSweets:
We allow for any model to be added and used for label detection. Our main focus is utilizing deep learning to enhance the detection beyond regex capabilities for more complex tasks.

We do include a regex model in the repo which can could be imported and updated if desired. One would have to manually add their own regex model parameters for new labels to be detected. It is on the roadmap to add an example of the regex model. I'll talk with @lettergram about pushing that out sooner rather than later.

ian-contiamo:
Thanks, an example would be perfect.

JGSweets:
@ian-contiamo Can you check out the new example to see if this meets your needs?

ian-contiamo:
Thanks @JGSweets, it's obviously of great help!

I'm still wrapping my head around how you do things, and I will try to figure out how to add new labels rather than replace the existing ones altogether.

PR #413, opened 2021-09-13
Adding an example for a regex labeler

JGSweets:
Adds feature documentation for an example labeler and adds a post processor specifically for using the regex labeler with structured data.

Intended to address: https://github.com/capitalone/DataProfiler/issues/411

### DIFF
--- a/dataprofiler/labelers/data_processing.py
+++ b/dataprofiler/labelers/data_processing.py
@@ -1792,17 +1792,25 @@ class StructRegexPostProcessor(BaseDataPostprocessor,
 
     def process(self, data, labels=None, label_mapping=None, batch_size=None):
         """Data preprocessing function."""
+        
+        # predictions come from regex_processor in the split format which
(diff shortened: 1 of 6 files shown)


dataprofiler/labelers/data_processing.py
```diff
@@ -1705,3 +1705,104 @@ def _save_processor(self, dirpath):
...
+        # prediction is the same as the pred for structured
+        for i in range(len(results['pred'])):
+            results['pred'][i] = np.mean(results['pred'][i], axis=0)
+
+        # set the confidences to be the aggregate of the predictions
+        if 'conf' in results:
+            results['conf'] = np.vstack(results['pred'])
+
+        # get argmax from predictions
+        results['pred'] = np.argmax(results['pred'], axis=1)
```
lettergram, line 1807: doesn't this argmax the entire dataset?

lettergram (approved):

PR #426, opened 2021-10-28
Bug fix to handle final word in word level argmax

lettergram:


dataprofiler/labelers/data_processing.py
```diff
@@ -825,8 +825,8 @@ def _word_level_argmax(self, data, predictions, label_mapping,
             label_count = {label_mapping[default_label]: 0}
             for idx in range(len(sample)):
 
-                # Split on separator
-                if sample[idx] in separator_dict:
+                # Split on separator or last sample
+                if sample[idx] in separator_dict or idx == len(sample)-1:
```
AnhTruong, line 829: should we add a test for this?

lettergram: Very good point, I'll add a test to evaluate this case

dataprofiler/labelers/data_processing.py
```diff
@@ -825,8 +825,8 @@ def _word_level_argmax(self, data, predictions, label_mapping,
             label_count = {label_mapping[default_label]: 0}
             for idx in range(len(sample)):
```
cbbruss, line 826: Would an `enumerate` not be cleaner?

JGSweets (approved):

ksneab7 (approved):

dataprofiler/labelers/data_processing.py
JGSweets: agreed, but not going to block the PR if @lettergram doesn't get to it prior to merge

PR #428, opened 2021-11-19
fix: numpy removing trailing nulls from strings

JGSweets:
If a string ends in a null character, it is trimmed by numpy during training. This stops that for occurring by setting the dtrype to object


lettergram (approved):

ksneab7 (commented):
Thanks for investigating and fixing this!!

PR #435, opened 2022-01-28
Add is_pred_labels to struct labeler and increment version

JGSweets:
Allows user to specify is_pred_labels (False, True) in post processor for structured.
Now labeler can return indexes instead of labels

### DIFF
--- a/dataprofiler/labelers/data_processing.py
+++ b/dataprofiler/labelers/data_processing.py
@@ -830,10 +830,10 @@ class CharPostprocessor(BaseDataPostprocessor,
                 is_end = (idx == len(sample)-1 and start_idx > 0)
 
                 if not is_separator:
-                    label = entities_in_sample[idx]                    
+                    label = entities_in_sample[idx]
                     if label not in label_count:
                         label_count[label] = 0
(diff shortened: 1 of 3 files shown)


dataprofiler/labelers/data_processing.py
```diff
@@ -1231,6 +1232,9 @@ def __init__(self, default_label='UNKNOWN', pad_label='PAD',
         :param flatten_separator: separator used to put between flattened
             samples.
         :type flatten_separator: str
+        :param is_pred_labels: (default: true) if true, will convert the model
+            indexes to the label strings given the label_mapping
```
ksneab7, line 1236: nit-pick, but indices not indexes.

ksneab7 (approved):
Very small comment about a function headers english but looks good to me

dataprofiler/labelers/data_processing.py
JGSweets: https://stackoverflow.com/questions/1378781/proper-terminology-should-i-say-indexes-or-indices

taylorfturner (approved):
LGTM

PR #446, opened 2022-03-29
(Data Profiler) Differences: Address int64 issue in differences

micdavis:
Fix: updated test_(int/float)_column_profile.py test_diff to cover json.dumps() functionality. Updated numerical_column_stats.py _perform_t_test to cast conservative and welch values to floats before assignment.

Jira Task: https://jira.kdc.capitalone.com/browse/C4APPRES-482


dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -6,7 +6,7 @@
 
 import pandas as pd
 import numpy as np
-
+import json
```
JGSweets, line 9: json is an first-party package, it should be grouped with the imports at the top of the file
```python
# first party imports

# third party

# our lib
```

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -6,7 +6,7 @@
 import pandas as pd
 import numpy as np
 import warnings
-
+import json
```
JGSweets, line 9: same for imports here

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1569,6 +1569,10 @@ def test_diff(self):
             }
         }
         profile_diff = profiler1.diff(profiler2)
+        try:
+            json.dumps(profile_diff)
+        except TypeError:
+            self.fail('Object of type int64 is not JSON serializable in the diff')
```
JGSweets, line 1575: Instead of specifying the error that occurred before, it could be different. We could instead explain why it is an issue, the `Diff Profile is not JSON serializable`, we could also add in the exception itself:
```python
...
except TypeError as e:
            self.fail('<EXPLAINATION>. Exception raised: {}'.format(str(e)))
```
Then we can say the problem and what happened.

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -1006,6 +1006,10 @@ def test_diff(self):
             }
         }
         profile_diff = profiler1.diff(profiler2)
+        try:
```
JGSweets, line 1009: as above for fail

ksneab7 (approved):

ksneab7 (approved):

Issue #441, opened 2022-02-09
Datetime Profiler cannot current detect dates with days which have suffixes


JGSweets:
**General Information:**
 - Library version: v0.7.6

**Describe the bug:**
While this date can be detected: `Nov 15, 2013` adding a suffix to the day does not allow it to be predicted, e.g. `Nov 15th, 2013`.

**To Reproduce:**
```
DateTimeColumn._get_datetime_profile(pd.Series(['Nov 15th, 2013']))

# output:
# {
#     'date_formats': [],
#     'min': None,
#     'max': None,
#     'min_obj': datetime.datetime(9999, 12, 31, 23, 59, 59, 999999),
#     'max_obj': datetime.datetime(1, 1, 1, 0, 0),
#     'match_count': 0
#  }
```

**Expected behavior:**
```
DateTimeColumn._get_datetime_profile(pd.Series(['Nov 15th, 2013']))

# output:
# {
#     'date_formats': ['%b %d, %Y'],  # something in this should indicate a suffix, currently doesn't.
#     'min': 'Nov 15th, 2013',
#     'max': 'Nov 15th, 2013',
#     'min_obj': Timestamp('2013-11-15 00:00:00'),
#     'max_obj': Timestamp('2013-11-15 00:00:00'),
#     'match_count': 1
# }
```

JGSweets:
probably a short term fix for removing `th`, `nd`, and `st` so they can at least be recognized.

PR #458, opened 2022-04-11
DateTimeColumn: Handle Datetime day suffixes

micdavis:
Fix: Updated datetime_column_profile.py to remove day suffixes from the date when we check its format. Updated the test_datetime_column_profile.py test to test for this change. Addresses this issue #441

### DIFF
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 import warnings
 
 import numpy as np
@@ -35,13 +36,6 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
         "%b-%d-%y",  # Mar-13-13
         "%m%d%Y",  # 03142013
         "%H:%M:%S.%f"  # 05:46:30.258509
-    ]
-
-    _day_suffixes = [
-        "st",
-        "nd",
-        "rd",
-        "th"
     ]
 
     def __init__(self, name, options=None):
@@ -171,11 +165,12 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
         return converted_date
 
     @staticmethod
-    def _replace_day_suffix(date, suffix_list):
+    def _replace_day_suffix(date):
         try:
-            for suffix in suffix_list:
-                date = date.replace(suffix, "")
-        except (AttributeError):
+            suffix = '|'.join(["st", "nd", "rd", "th"])
+            pattern = re.compile("(\d{1,2})(" + suffix + ")")
+            date = pattern.sub(r'\1', date)
+        except (TypeError):
             date = np.nan
         return date
 
@@ -206,7 +201,7 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
                 break
             valid_dates = df_series.apply(
                 lambda x: cls._validate_datetime(
-                    cls._replace_day_suffix(x, cls._day_suffixes), date_format
+                    cls._replace_day_suffix(x), date_format
                 )
             )
 
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -175,6 +175,15 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
 
     @staticmethod
     def _replace_day_suffix(date, pattern):
+        """
+        Checks the date for a suffix after the day. Removed suffix if present
+
+        :param date: a string that is possibly a date
+        :type date: str
+        :param pattern: a precompiled regex pattern that is used to check for day suffixes
+        :type pattern: Pattern
+        :return: either the date string passed in, or Nan
+        """
         try:
             date = pattern.sub(r'\1', date)
         except (TypeError):
--- a/dataprofiler/profilers/datetime_column_profile.py
+++ b/dataprofiler/profilers/datetime_column_profile.py
@@ -37,6 +37,13 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
         "%H:%M:%S.%f"  # 05:46:30.258509
     ]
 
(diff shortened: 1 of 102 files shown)


dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -163,6 +170,15 @@ def _validate_datetime(date, date_format):
 
         return converted_date
 
+    @staticmethod
```
micdavis, line 173: Do static methods such as this need to be covered in the [test_datetime_column_profile.py](https://github.com/capitalone/DataProfiler/pull/458/files#diff-9bba6fcc721a948265b1158d5a65dc4cd34a55984ff110d83782d9da29abc6b1).

micdavis (commented):
One potential issue is that if a date is formatted as such:
`15thMarth13` then it will be picked up as `2013-03-15`.

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -163,6 +170,15 @@ def _validate_datetime(date, date_format):
...
+    def _replace_day_suffix(date, suffix_list):
+        try:
+            for suffix in suffix_list:
+                date = date.replace(suffix, "")
+        except (AttributeError):
+            date = np.nan
```
JGSweets, line 179: what error is occurring?

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -163,6 +170,15 @@ def _validate_datetime(date, date_format):
 
         return converted_date
 
+    @staticmethod
+    def _replace_day_suffix(date, suffix_list):
+        try:
+            for suffix in suffix_list:
+                date = date.replace(suffix, "")
```
JGSweets, line 177: is this faster than using regex with
re.compile?

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -189,7 +205,9 @@ def _get_datetime_profile(cls, df_series):
             if is_row_datetime.all():
                 break
             valid_dates = df_series.apply(
```
JGSweets, line 207: if we assume these are strings, could we use pandas regex application?

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -37,6 +37,13 @@ class DateTimeColumn(BaseColumnPrimitiveTypeProfiler):
         "%H:%M:%S.%f"  # 05:46:30.258509
     ]
 
+    _day_suffixes = [
+        "st",
+        "nd",
+        "rd",
+        "th"
```
taylorfturner, line 44: would want a final `,` here too... per PEP8 I think

taylorfturner (commented):
one syntax comment

dataprofiler/profilers/datetime_column_profile.py
JGSweets: I think it's covered by the other functions output, with the `th`, but we should have an individual test with each day_suffix in a dataset being profiled to validate each rather than updating existing tests.

dataprofiler/profilers/datetime_column_profile.py
micdavis: When a non-string type is passed into this function and replace is called on the variable, it throws an AttributeError.

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -163,6 +173,14 @@ def _validate_datetime(date, date_format):
 
         return converted_date
 
+    @staticmethod
+    def _replace_day_suffix(date, pattern):
+        try:
+            date = pattern.sub(r'\1', date)
```
JGSweets, line 179: `_compiled_day_suffix_regex`

dataprofiler/profilers/datetime_column_profile.py
```diff
@@ -163,6 +173,14 @@ def _validate_datetime(date, date_format):
 
         return converted_date
 
+    @staticmethod
+    def _replace_day_suffix(date, pattern):
```
JGSweets, line 177: just needs a description

dataprofiler/tests/profilers/test_datetime_column_profile.py
```diff
@@ -371,6 +373,25 @@ def test_datetime_column_with_wrong_options(self):
...
+        """
+        Tests datetime examples with daytime suffixes.
+        :return:
+        """
+        data = [
+            'Mar 1st, 2020',
+            'Feb 22nd, 2019',
+            'October 23rd, 2018',
+            '12thMar13'
+        ]
```
JGSweets, line 386: ideally we should check to make sure each is properly updated. This could technically be done by checking `samples` maybe?

JGSweets: `profiler.sample`

JGSweets (commented):
LGTM, small fixes

dataprofiler/tests/profilers/test_datetime_column_profile.py
JGSweets: rethinking that it is recognizing all 4 as dates since `match_count`

taylorfturner:
@micdavis , run `isort . ` in the root of the repo just to make sure those imports are sorted correctly

JGSweets (approved):

dataprofiler/tests/profilers/test_datetime_column_profile.py
```diff
@@ -268,6 +268,8 @@ def test_profile(self):
             expected = defaultdict(float, {'datetime': 2.0})
             self.assertEqual(expected, profiler.profile['times'])
```
taylorfturner, line 270: nit pick but you can probably get rid of this extra space between the two `self.assert` in the same test function

taylorfturner (commented):
- one specific comment 
- one general for `isort .`
other than those, I'll apporove

JGSweets (approved):

PR #481, opened 2022-06-07
Fixes overflow bug if moments are large

JGSweets:
Overflow error would occur if M2 was large and powered.
```
OverflowError: (34, 'Numerical result out of range')
```

This PR fixes the above error. However, a long term fix would be potentially iterating through large data to ensure smaller moments.

### DIFF
--- a/dataprofiler/profilers/utils.py
+++ b/dataprofiler/profilers/utils.py
@@ -18,24 +18,24 @@ from dataprofiler import settings
 def dict_merge(dct, merge_dct):
     # Recursive dictionary merge
     # Copyright (C) 2016 Paul Durivage <pauldurivage+github@gmail.com>
-    # 
+    #
     # This program is free software: you can redistribute it and/or modify
     # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


dataprofiler/profilers/utils.py
```diff
@@ -318,7 +318,8 @@ def biased_skew(df_series):
     if (M2 == 0):
         return 0.0
 
-    skew = np.sqrt(n) * M3 / M2 ** 1.5
+    with np.errstate(all='ignore'):
```
JGSweets, line 321: fix (1) is here

dataprofiler/profilers/utils.py
```diff
@@ -354,7 +355,8 @@ def biased_kurt(df_series):
     if (M2 == 0):
         return -3.0
 
-    kurt = n * M4 / M2 ** 2 - 3
```
JGSweets, line 357: fix (2/2) is here

ksneab7 (approved):

taylorfturner (approved):

PR #506, opened 2022-07-05
Add New preprocessor for using an encoding map

JGSweets:
Allows a user to set the encoding map to be used in the preprocesor

### DIFF
--- a/dataprofiler/labelers/data_processing.py
+++ b/dataprofiler/labelers/data_processing.py
@@ -222,7 +222,7 @@ class CharPreprocessor(BaseDataPreprocessor, metaclass=AutoSubRegistrationMeta):
 
     def __init__(self, max_length=3400, default_label='UNKNOWN',
                  pad_label='PAD', flatten_split=0, flatten_separator=" ",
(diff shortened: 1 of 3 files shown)


dataprofiler/labelers/data_processing.py
```diff
@@ -647,14 +648,131 @@ def process(self, data, labels=None, label_mapping=None, batch_size=32):
...
+            data, labels, label_mapping, batch_size)
+        encoding_map = self._parameters['encoding_map']
+        max_length = self._parameters['max_length']
+        for char_processed_data in char_processor_gen:
+            x = char_processed_data
+            if labels:
+                x = char_processed_data[0]
+                y = char_processed_data[1]
+
+            processed_x = np.zeros((len(x), max_length), dtype='int64')
```
JGSweets, line 762: uses the previous char generator and then just encodes it after.

taylorfturner (approved):

taylorfturner (approved):

micdavis (approved):

PR #527, opened 2022-07-11
Graph Reader/Loader added to GraphData

MisterPNP:
The previous GraphData only contained is_match, a function allowing to differentiate between tabular and graph CSV datasets.

This PR adds the functionality of reading/loading the graph data from the input CSV into a NetworkX graph. This graph contains nodes, edges, and possible attributes for both. The use of NetworkX is very useful for downstream profiling of the data due to NetworkX's integrated functions to analyze and pull information from graphs.


taylorfturner:
@MisterPNP `git fetch upstream` so your local version of `capitalone:main` is up to date prior to your `git checkout -b ...`

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+        column_names = []
+
+        with FileOrBufferHandler(file_path) as csv_file:
+            csv_reader = csv.reader(csv_file, delimiter = options.get("delimiter", ","))
+            
+            # fetch only column names
+            for row in csv_reader:
+                column_names.append(row)
+                break
+        column_names = column_names[0]
```
JGSweets, line 61: right now this presumes the first row is the header. Some CSVs have comments at the top. We cause use the info from `CSVData.is_match` and options to determine this. (will also need a test, e.g. single graph data file which has comments before the column names start).
to make this more generic we could do:
```python
@classmethod
def csv_column_names(cls, file_path, delimiter, quotechar, header): 
    column_names = []
    if header is None:
        return column_names

    for _ in range(header + 1):
        row = next(csv_reader)
    .
    .
    .
```
csv.reader can be updated to handle quotechar too:
```python
csv_reader = csv.reader(csv_file, delimiter=delimiter, quotechar=quotechar)
```

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+        Current formats checked:
+            - attributed edge list
+
+        This works by finding whether the file contains a target and a source node
+        '''
+        if options is None:
+            options = dict()
+        if not CSVData.is_match(file_path, options):
+            return False
+        column_names = cls.csv_column_names(file_path, options)
```
JGSweets, line 83: this would need to update with the changes above

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def _format_data_networkx(self):
+        '''
+        Formats the input file into a networkX graph
+        '''
+        source_keywords = ['source', 'src', 'origin']
+        target_keywords = ['target', 'destination', 'dst']
+
+        if self.input_file_path:
+            column_names = self.csv_column_names(self.input_file_path, self.options)
```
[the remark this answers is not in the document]
JGSweets: We should also only determine these if the class property is `None`

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+
+            if self.options.get("source_node") is None or self.options.get("destination_node") is None:
```
JGSweets, line 111: as above, only determine 1 / the other if the class prop is still `None`

dataprofiler/data_readers/graph_data.py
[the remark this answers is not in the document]
JGSweets:
```python
if not self._column_names:
    self._column_names = self.csv_column_names(self.input_file_path, ...)
```

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+                source_node = self.options.get("source_node")
+                destination_node = self.options.get("destination_node")
+
+            # read lines from csv
+            csv_as_list = []
+            with FileOrBufferHandler(self.input_file_path) as csv_file:
+                csv_reader = csv.reader(csv_file, delimiter = self.options.get("delimiter", self.options.get("delimiter", ",")))
+                for row in csv_reader:
+                    row = [item.replace(" ", "") for item in row]
+                    csv_as_list.append(row)
```
JGSweets, line 124: maybe we should just do the same as CSVData here.
```python
data_utils.read_csv_df(
    input_file_path,
    self._delimiter, self._header, self._column_names,
    read_in_string=True,
    encoding=self.file_encoding
)
```

JGSweets: then we can load from pandas_edgelist? or loop through as you desire

JGSweets: https://networkx.org/documentation/stable/reference/generated/networkx.convert_matrix.from_pandas_edgelist.html#networkx.convert_matrix.from_pandas_edgelist
Not sure what versions it is viable within

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+from .filepath_or_buffer import FileOrBufferHandler
+
+class GraphData(BaseData):
+        
+    def __init__(self, input_file_path=None, data=None, options=None):
+        BaseData.__init__(self, input_file_path, data, options)
+        if options is None:
+            options = dict()
+
+        return self._load_data()
```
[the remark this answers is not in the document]
JGSweets: I see in `_load_data` you already have a value error, hence could just do:
```python
if data is not None:
    self._load_data(data)
```
Though we should prob raise an error that we cannot load this type yet or something, thoughts?

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+
+        This works by finding whether the file contains a target and a source node
+        '''
+        if options is None:
+            options = dict()
+        if not CSVData.is_match(file_path, options):
+            return False
+        column_names = cls.csv_column_names(file_path, options)
+        source_keywords = ['source', 'src', 'origin']
+        target_keywords = ['target', 'destination', 'dst']
```
JGSweets, line 85: should these be class props, or settable?

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+                attributes = dict()
+                for column in range(0, len(csv_as_list[0])):
+
+                    if csv_as_list[line][column] is None:
+                        continue
+                    if column is not source_node or column is not destination_node:
+                        attributes[column_names[column]] = (csv_as_list[line][column])
+                        list_edges.append([csv_as_list[line][destination_node], csv_as_list[line][source_node], attributes])
+                    if column is source_node or column is destination_node:
+                        list_nodes.add(csv_as_list[line][column])
```
JGSweets, line 141: can we add nodes while in here rather than collect all then add? would potentially save memory?

dataprofiler/tests/data/csv/graph-data-input-json.json
```diff
@@ -0,0 +1 @@
+{"name":"John", "age":30, "car":null}
```
taylorfturner, line 1: left over from last PR ... probably need `"null"`

dataprofiler/data_readers/graph_data.py
```diff
@@ -0,0 +1,159 @@
...
+
+import networkx as nx
+from numpy import source
+
+from .base_data import BaseData
+from .csv_data import CSVData
+from .filepath_or_buffer import FileOrBufferHandler
+
+class GraphData(BaseData):
+
```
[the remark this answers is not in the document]
taylorfturner: +1 ... for clarity, just like in `avro_data.py`, just put `data_type = "graph"` (with one line space between the `def __init__`

PR #531, opened 2022-07-12
Reformatted code with trailing-whitespace hook.

jakleh:


taylorfturner (approved):

taylorfturner (commented):
failing tests.... need to confirm tests are testing the data with the white space / we can update tests to reflect the no white space. Or we need to ignore certain files for whitespace if the tests rely on that whitespace

JGSweets (changes requested):
I don't think we should affect test data, that may have intent on the whitespace for testing purposes

PR #536, opened 2022-07-13
HOT FIX: update README and remove isort.cfg

taylorfturner:
- Removing `isort.cfg`. Setting now moved to `setup.cfg`
- Also adding note in README about installing `requirements-dev.txt`


.isort.cfg
```diff
@@ -1,4 +0,0 @@
-[settings]
```
taylorfturner, line 1: now in `setup.cfg`

micdavis (approved):

PR #543, opened 2022-07-18
Binary Tree merging of multiple profiles 

taylorfturner:
Adding function to `utils.py` to take in a list of profile objects (assuming all same profile type) and merging them all together using their built in `__add__` method.

### DIFF
--- a/dataprofiler/profilers/utils.py
+++ b/dataprofiler/profilers/utils.py
@@ -7,6 +7,7 @@ import multiprocessing as mp
 import os
 import time
 import warnings
+from lib2to3.pytree import type_repr
 
 import numpy as np
 import psutil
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/utils.py
```diff
@@ -686,3 +686,21 @@ def perform_chi_squared_test_for_homogeneity(
...
+
+
+def merge_list_of_profiles(list_of_profiles):
+    """Using the profile object's `__add__` method, to merge all the
+        provided profiles in the list into a single profile, which
+        is returned by the function.
+
+    :param list_of_profiles: List of Data Profiler profile objects.
+        Acceptable profile types are `StructuredProfiler` and
+        `UnstructuredProfiler`.
```
JGSweets, line 698: technically, this could be anything with an `add`? or might we have to curate specific code for this top level?

taylorfturner: True it could be anything with an `__add__`. That's completely fair. Making it more reusable at any level is 🥇

PR #544, opened 2022-07-19
Calculate correlation only between selected columns

Ta7ar:
In the current implementation, correlation calculation between specific columns can be set with,
`profile_options = dp.ProfilerOptions()`
`profile_options.set({"correlation.columns": ["column name 1", "column name 2"]})`
But in the calculation ([Line 1834 in `profile_builder.py`](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profile_builder.py#L1834)), correlation is correctly measured only if `correlation.columns` is not set (i.e. `correlation.columns` is `None`) and raises an error if `correlation.columns` is set. 
This PR fixes the issue and allows the user to calculate correlation between only the selected columns.

_Todo:_ Raise error if column name passed does not exist in the dataset 

_Test Coverage:_
`dataprofiler/tests/profilers/test_profile_builder.py                                              1572      2    99%`
Where the coverage is 99%

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1844,7 +1844,7 @@ class StructuredProfiler(BaseProfiler):
         columns = self.options.correlation.columns
         column_ids = list(range(len(self._profile)))
         if columns is not None:
-            column_ids = [self._col_name_to_idx[col_name][0] for col_name in columns]
+            column_ids = [idx for col_name in columns for idx in self._col_name_to_idx[col_name]]
         clean_column_ids = []
         for idx in column_ids:
             data_type = (
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1842,21 +1842,23 @@ class StructuredProfiler(BaseProfiler):
         :type batch_properties: dict()
         """
         columns = self.options.correlation.columns
+        column_ids = list(range(len(self._profile)))
+        if columns is not None:
+            column_ids = [self._col_name_to_idx[col_name][0] for col_name in columns]
         clean_column_ids = []
-        if columns is None:
-            for idx in range(len(self._profile)):
-                data_type = (
-                    self._profile[idx].profiles["data_type_profile"].selected_data_type
-                )
-                if data_type not in ["int", "float"]:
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1842,21 +1842,23 @@ def _get_correlation(self, clean_samples, batch_properties):
         :type batch_properties: dict()
         """
         columns = self.options.correlation.columns
+        column_ids = list(range(len(self._profile)))
+        if columns is not None:
+            column_ids = [self._col_name_to_idx[col_name][0] for col_name in columns]
```
JGSweets, line 1847: what happens if a pd.DF has two columns with the same name? Is there a way to specify one of the two columns? Thoughts?

Ta7ar: Good callout, I don't think this current implementation handles that. Should be an easy feature to add I think.

taylorfturner: we either 
- handle processing dataframes with two or more column names that are the same 
OR 
- we raise an error that all column names in the dataframe must be unique

Ta7ar: **what happens if a pd.DF has two columns with the same name?**
If there are columns with the same name, the current implementation only takes the first of the repeating columns. Could implement the functionality to use all of the repeating columns.

**Is there a way to specify one of the two columns?**
In the current implementation no. Would need to work on CorrelationOptions to add that functionality.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1842,21 +1842,23 @@ def _get_correlation(self, clean_samples, batch_properties):
         :type batch_properties: dict()
         """
         columns = self.options.correlation.columns
+        column_ids = list(range(len(self._profile)))
+        if columns is not None:
+            column_ids = [idx for col_name in columns for idx in self._col_name_to_idx[col_name]]
```
taylorfturner, line 1847: how does this not return an `'int' object is not iterable`?

taylorfturner: resolved on test locally

JGSweets (approved):

PR #545, opened 2022-07-19
Reformatted dataprofiler/tests/labelers/test_base_model.py using flake8; added flake8 configuration to setup.cfg.

jakleh:


taylorfturner (changes requested):
your branch is behind `capitalone:main`. You'll need to rebase I think

taylorfturner (changes requested):
Must better on the `diff` report for sure! Will hold off approving until the docstrings are written out.

PR #550, opened 2022-07-21
Reformatted dataprofiler/dp_logging.py using flake8; changed first docstring in float, text, and order column profiler modules.

jakleh:


dataprofiler/dp_logging.py
```diff
@@ -1,3 +1,5 @@
+"""coding=utf-8."""
```
JGSweets, line 1: should we not be more explicit here?

dataprofiler/dp_logging.py
```diff
@@ -38,7 +40,9 @@ def get_logger():
 
 def set_verbosity(level):
     """
-    Set verbosity level for DataProfiler logger. Must set it to one of
+    Set verbosity level for DataProfiler logger.
+
+    Must set it to one of
```
taylorfturner, line 45:
```suggestion
    Must set it to one of the following values:
```

dataprofiler/dp_logging.py
```diff
@@ -50,7 +54,7 @@ def set_verbosity(level):
 
 def get_child_logger(name):
     """
-    Returns logger for given filepath
+    Return logger for given filepath.
```
taylorfturner, line 57:
```suggestion
    Return logger for the given filepath.
```

dataprofiler/profilers/base_column_profilers.py
```diff
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
 """
-coding=utf-8
-
 Profiles the data.
```
JGSweets, line 3: I would suggest these are abstract classes for profiles.

taylorfturner, line 3:
```suggestion
Lowest level column profiler class to be used further up in the inheritance tree of the package.
```

taylorfturner: see @JGSweets comment on this line too

JGSweets (approved):

PR #558, opened 2022-07-22
Reformatted dataprofiler/profilers/unstructured_labeler_profile.py us…

jakleh:
…ing flake8.

### DIFF
--- a/dataprofiler/profilers/unstructured_labeler_profile.py
+++ b/dataprofiler/profilers/unstructured_labeler_profile.py
@@ -1,3 +1,4 @@
+"""Profile analysis for applying labels within unstructured profiling."""
 from collections import defaultdict
 
 from ..labelers.data_labelers import DataLabeler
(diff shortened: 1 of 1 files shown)


dataprofiler/profilers/unstructured_labeler_profile.py
```diff
@@ -94,16 +95,17 @@ def __add__(self, other):
 
     def report(self, remove_disabled_flag=False):
         """
-        Private abstract method for returning report.
+        Return report.
```
taylorfturner, line 98: this technically returns the `self.profile` object? Even thought we have this called `report()` we are returning a `profile` object?

jakleh: The previous docstring also said that it returns a report, but I'll go ahead and change it if what you said is the case.

taylorfturner: 👍

JGSweets (approved):

taylorfturner (approved):

PR #568, opened 2022-07-26
Added SyntheticDataOptions and NaN replication required metrics

Ta7ar:
Added `SyntheticDataOptions` which can be set using `profile_options.set({"synthetic_data.is_enabled": True})`. This enables calculation of metrics specifically needed for `capitalone/synthetic-data`.

Added calculation of metrics needed by `capitalone/synthetic-data` for replicating NaN values in the generated dataset.

_Todo_: Write tests once PR passes initial review.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1438,6 +1438,9 @@ class StructuredProfiler(BaseProfiler):
         self.correlation_matrix = None
         self.chi2_matrix = None
 
+        # capitalone/synthetic-data specific metrics
+        self._nan_replication_dependent_properties = []
+
         if data is not None:
             self.update_profile(data)
 
@@ -1739,6 +1742,10 @@ class StructuredProfiler(BaseProfiler):
                 quantiles = calculate_quantiles(num_quantile_groups, quantiles)
                 report["data_stats"][i]["statistics"]["quantiles"] = quantiles
 
+        if self.options.synthetic_data.is_enabled:
+            report["synthetic_data"] = {
+                "nan_replication": self._nan_replication_dependent_properties
+            }
         return _prepare_report(report, output_format, omit_keys)
 
     def _get_unique_row_ratio(self):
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2119,65 @@ def _update_chi2(self):
...
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
+
+        for col_id in range(len(self._profile)):
+            null_count = getattr(self._profile[col_id], "null_count")
+            if null_count == 0:
+                # No missing values to replicate
+                continue
+
+            # Partition data based on whether target column value is NaN or not
+            col_is_nan = data.iloc[:, col_id].isnull()
```
JGSweets, line 2146: could we get this info from the profile of the column already?

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2119,65 @@ def _update_chi2(self):
...
+            # Drop the target column from partitions
+            partition_nan = partition_nan.drop(columns=col_id)
+            partition_not_nan = partition_not_nan.drop(columns=col_id)
+
+            # Calculate mean vectors of each partition
+            # i.e mean vectors of rows where target col is or is not NaN
+            mean_nan, mean_not_nan = (
+                partition_nan.mean().to_numpy(),
+                partition_not_nan.mean().to_numpy(),
+            )
```
JGSweets, line 2159: do we not have this info already from the col profiles? Is this only relative to the column we are iterating through?

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2119,65 @@ def _update_chi2(self):
...
+            self._nan_replication_dependent_properties.append(
+                {
+                    "column_id": col_id,
+                    "column_name": col_name,
+                    # Array index serves as class label
+                    # 0 indicates not NaN, 1 indicates NaN
+                    "class_priors": [prior_not_nan, prior_nan],
+                    "class_mean": [mean_not_nan, mean_nan],
+                }
+            )
```
JGSweets, line 2179: how do we handle this if we update the profile? wouldn't this be  incorrect/

dataprofiler/profilers/profile_builder.py
Ta7ar: This creates a boolean array needed for indexing. We do calculate the list of indices where there is a null value so I guess I could use that to create this array.

Ta7ar: But that would still be an O(n) operation I think (initialize an array of size `sample_size` with all values being `False` and then flip values to `True` according to list of indices where there is null)

dataprofiler/profilers/profile_builder.py
Ta7ar: No we do not have this info pre calculated and this is pretty much the only extra calculation needed for replicating missing values in synthetic data. Basically we split the dataset into two partitions, one where the column we are processing (call it the `target column`) is NaN and where it is not NaN. Then take the means of each of the partitions (while not keeping the `target column` in the mean vectors calculated)

dataprofiler/profilers/profile_builder.py
Ta7ar: Yep working on making it handle updates.

Ta7ar:
Closing this PR because it diverged from main branch.

PR #571, opened 2022-07-29
SyntheticDataOptions and metrics calculations for NaN replication

Ta7ar:
Added `SyntheticDataOptions` which can be set using `profile_options.set({"synthetic_data.is_enabled": True})`. This enables calculation of metrics specifically needed for `capitalone/synthetic-data`.

Added calculation of metrics needed by `capitalone/synthetic-data` for replicating NaN values in the generated dataset.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1438,6 +1438,9 @@ class StructuredProfiler(BaseProfiler):
         self.correlation_matrix = None
         self.chi2_matrix = None
 
+        # capitalone/synthetic-data specific metrics
+        self._nan_replication_metrics = {}
+
         if data is not None:
             self.update_profile(data)
 
@@ -1522,6 +1525,14 @@ class StructuredProfiler(BaseProfiler):
                 merged_profile.chi2_matrix = None
             else:
                 merged_profile.chi2_matrix = merged_profile._update_chi2()
+
+        if (
+            self.options.synthetic_data.is_enabled
+            and other.options.synthetic_data.is_enabled
+        ):
+            merged_profile._nan_replication_metrics = (
+                self._merge_nan_replication_metrics(other)
+            )
 
         return merged_profile
 
@@ -1739,6 +1750,10 @@ class StructuredProfiler(BaseProfiler):
                 quantiles = calculate_quantiles(num_quantile_groups, quantiles)
                 report["data_stats"][i]["statistics"]["quantiles"] = quantiles
 
+        if self.options.synthetic_data.is_enabled:
+            report["synthetic_data"] = {
+                "nan_replication": self._nan_replication_metrics
+            }
         return _prepare_report(report, output_format, omit_keys)
 
     def _get_unique_row_ratio(self):
@@ -2112,6 +2127,166 @@ class StructuredProfiler(BaseProfiler):
 
         return chi2_mat
 
+    def _get_null_and_sample_count(self):
+        """
+        Gets the null_count and sample_size of each column.
+
+        :return: Dictionary where column id maps to a list of [null_count, sample_size]
+        :rtype: dict
+        """
+        counts = {}
+        for col_id in range(len(self._profile)):
+            null_count = getattr(self._profile[col_id], "null_count", 0)
+            sample_size = getattr(self._profile[col_id], "sample_size", 0)
+            counts[col_id] = [null_count, sample_size]
+        return counts
+
+    def _calc_nan_replication_metrics(self, clean_samples, prev_counts):
+        """
+        Calculate metrics needed for replicating NaN values in capitalone/synthetic-data.
+
+        Metrics include class priors and class specific means of each column.
+
+        Required for running LDA based binary classifier where predicted class label indicates whether
+        a value of a column should be NaN (1) or not (0).
+
+        :param clean_samples: input cleaned dataset
+        :type clean_samples: dict
+        :param prev_counts: null_count and sample_size of each column before update
+        :type prev_counts: dict
+        """
+        if not self.options.synthetic_data.replicate_nan.is_enabled:
+            return
+
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
+
+        for col_id in range(len(self._profile)):
+            null_count = getattr(self._profile[col_id], "null_count")
+            if null_count == 0:
+                # No missing values to replicate
+                continue
+
+            # Calculate class priors
+            # i.e probability that a value in target col is or is not NaN
+            sample_size = getattr(self._profile[col_id], "sample_size")
+            true_count = sample_size - null_count
+
+            # null_count and sample_size get updated on profile_update
+            # Therefore the priors are calculated from the updated values
+            prior_nan = null_count / sample_size
+            prior_not_nan = true_count / sample_size
+
+            # Partition data based on whether target column value is NaN or not
+            col_is_nan = data.iloc[:, col_id].isnull()
+            partition_nan = data[col_is_nan]
+            partition_not_nan = data[~col_is_nan]
+
+            # Drop the target column from partitions
+            partition_nan = partition_nan.drop(columns=col_id)
+            partition_not_nan = partition_not_nan.drop(columns=col_id)
+
+            # Calculate mean vectors of each partition
+            # i.e mean vectors of rows where target col is or is not NaN
+            mean_not_nan, mean_nan = (
+                partition_not_nan.mean().to_numpy(),
+                partition_nan.mean().to_numpy(),
+            )
+
+            # Combine with old means
+            if col_id in self._nan_replication_metrics:
+                prev_null_count, prev_sample_size = prev_counts[col_id]
+                prev_true_count = prev_sample_size - prev_null_count
+
+                # null_count, sample_size and true_count of new data chunk
+                added_null_count, added_sample_size = (
+                    null_count - prev_null_count,
+                    sample_size - prev_sample_size,
+                )
+                added_true_count = added_sample_size - added_null_count
+
+                prev_class_means = self._nan_replication_metrics[col_id]["class_mean"]
+                prev_mean_not_nan, prev_mean_nan = prev_class_means
+
+                mean_not_nan = utils.combine_means(
+                    prev_mean_not_nan, mean_not_nan, prev_true_count, added_true_count
+                )
+                mean_nan = utils.combine_means(
+                    prev_mean_nan, mean_nan, prev_null_count, added_null_count
+                )
+
+            col_name = getattr(self._profile[col_id], "name")
+            self._nan_replication_metrics[col_id] = {
+                "column_name": col_name,
+                # Array index serves as class label
+                # 0 indicates not NaN, 1 indicates NaN
+                "class_prior": [prior_not_nan, prior_nan],
+                "class_mean": [mean_not_nan, mean_nan],
+            }
+
+    def _merge_nan_replication_metrics(self, other):
+        """
+        Merge NaN replication metrics between two data profiles
+
+        :param other: profile being added to this one.
+        :type other: StructuredProfiler
+        :return: merged nan replication metrics
+        :rtype: dict
+        """
+
+        if (
+            not self.options.synthetic_data.replicate_nan.is_enabled
+            or not other.options.synthetic_data.replicate_nan.is_enabled
+        ):
+            return {}
+
+        merged_properties = {}
+        for col_id in range(len(self._profile)):
+            self_null_count = getattr(self._profile[col_id], "null_count")
+            other_null_count = getattr(other._profile[col_id], "null_count")
+            null_count = self_null_count + other_null_count
+            if null_count == 0:
+                continue
+
+            merged_properties[col_id] = {}
+            self_sample_size = getattr(self._profile[col_id], "sample_size")
+            other_sample_size = getattr(other._profile[col_id], "sample_size")
+            sample_size = self_sample_size + other_sample_size
+            true_count = sample_size - null_count
+
+            prior_nan = null_count / sample_size
+            prior_not_nan = true_count / sample_size
+
+            merged_properties[col_id]["class_prior"] = [prior_not_nan, prior_nan]
+
+            col_name = getattr(self._profile[col_id], "name")
+            merged_properties[col_id]["column_name"] = col_name
+
+            # 'if null_count == 0' check guarantees that either self or other has _nan_replication_metrics for col_id
+            if col_id not in self._nan_replication_metrics:
+                other_mean = other._nan_replication_metrics[col_id]["class_mean"]
+                merged_properties[col_id]["class_mean"] = other_mean
+
+            elif col_id not in other._nan_replication_metrics:
+                self_mean = self._nan_replication_metrics[col_id]["class_mean"]
+                merged_properties[col_id]["class_mean"] = self_mean
+
+            else:
+                self_true_count = self_sample_size - self_null_count
+                other_true_count = other_sample_size - other_null_count
+
+                self_mean = self._nan_replication_metrics[col_id]["class_mean"]
+                other_mean = other._nan_replication_metrics[col_id]["class_mean"]
+
+                mean_not_nan = utils.combine_means(
+                    self_mean[0], other_mean[0], self_true_count, other_true_count
+                )
+                mean_nan = utils.combine_means(
+                    self_mean[1], other_mean[1], self_null_count, other_null_count
+                )
+                merged_properties[col_id]["class_mean"] = [mean_not_nan, mean_nan]
+
+        return merged_properties
+
     def _update_profile_from_chunk(self, data, sample_size, min_true_samples=None):
         """
         Iterate over the columns of a dataset and identify its parameters.
@@ -2178,6 +2353,10 @@ class StructuredProfiler(BaseProfiler):
         # Record the previous mean/std values of columns that need to
         # have correlation updated.
         corr_prev_dependent_properties = self._get_correlation_dependent_properties()
+
+        # Record previous null_count, sample_size of each column
+        # For updating self._nan_replication_metrics
+        prev_null_and_sample_count = self._get_null_and_sample_count()
 
         # Create StructuredColProfilers upon initialization
         # Record correlation between columns in data and index in _profile
@@ -2330,6 +2509,12 @@ class StructuredProfiler(BaseProfiler):
         if self.options.chi2_homogeneity.is_enabled:
             self.chi2_matrix = self._update_chi2()
         self._update_row_statistics(data, samples_for_row_stats)
+
+        # Calculate metrics specific to capitalone/synthetic-data
+        if self.options.synthetic_data.is_enabled:
+            self._calc_nan_replication_metrics(
+                clean_sampled_dict, prev_null_and_sample_count
+            )
 
     def save(self, filepath=None):
         """
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -932,6 +932,34 @@ class CorrelationOptions(BaseInspectorOptions):
         return errors
 
 
+class SyntheticDataOptions(BaseInspectorOptions):
+    """For configuring options for synthetic data generation"""
+
+    def __init__(self):
+        """
+        Initialize options for synthetic data generation
+        :ivar is_enabled: boolean option to enable/disable synthetic data generation options
+        :vartype is_enabled: BooleanOption
+        :ivar replicate_nan: boolean option to enable/disable calculating metrics needed for
+        replicating nan values in synthetic data generator
+        :vartype replicate_nan: BooleanOption
+        """
+
+        self.replicate_nan = BooleanOption(is_enabled=True)
+        super().__init__(is_enabled=False)
+
+    def _validate_helper(self, variable_path="SyntheticDataOptions"):
+        """
+        Validate the options do not conflict and cause errors.
+
+        :param variable_path: current path to variable set.
+        :type variable_path: str
+        :return: list of errors (if raise_error is false)
+        :rtype: list(str)
+        """
+        return super()._validate_helper(variable_path)
+
+
 class DataLabelerOptions(BaseInspectorOptions):
     """For configuring options for Data Labeler Column."""
 
@@ -1156,6 +1184,8 @@ class StructuredOptions(BaseOption):
         :vartype correlation: CorrelationOptions
         :ivar chi2_homogeneity: option set for chi2_homogeneity matrix
         :vartype chi2_homogeneity: BooleanOption()
+        :ivar synthetic_data: option set for metrics calculation for synthetic data gen
+        :vartype synthetic_data: SyntheticDataOptions
         :ivar null_values: option set for defined null values
         :vartype null_values: Union[None, dict]
         """
@@ -1170,6 +1200,7 @@ class StructuredOptions(BaseOption):
         self.data_labeler = DataLabelerOptions()
         self.correlation = CorrelationOptions()
         self.chi2_homogeneity = BooleanOption(is_enabled=True)
+        self.synthetic_data = SyntheticDataOptions()
         # Non-Option variables
         self.null_values = null_values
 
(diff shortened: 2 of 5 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1739,6 +1750,10 @@ def report(self, report_options=None):
                 quantiles = calculate_quantiles(num_quantile_groups, quantiles)
                 report["data_stats"][i]["statistics"]["quantiles"] = quantiles
 
+        if self.options.synthetic_data.is_enabled:
+            report["synthetic_data"] = {
+                "nan_replication": self._nan_replication_metrics
+            }
```
taylorfturner, line 1756: would put this in `report["data_stats"][<column_index>]["nan_replication_metrics"]`

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -27,7 +27,7 @@ def test_default_profiler_options(self, *mocks):
         self.assertTrue(profile.options.data_labeler.is_enabled)
         for column in profile.options.properties:
             # TODO: remove the check for correlation option once it's updated to True
-            if column == "correlation":
+            if column == "correlation" or column == "synthetic_data":
```
taylorfturner, line 30: change option name ... `nan_replication_metrics`

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+        :param prev_counts: null_count and sample_size of each column before update
+        :type prev_counts: dict
+        """
+        if not self.options.synthetic_data.replicate_nan.is_enabled:
+            return
+
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
+
+        for col_id in range(len(self._profile)):
+            null_count = getattr(self._profile[col_id], "null_count")
```
taylorfturner, line 2164: add something like `compiler = self._profile[id]` so you don't have to keep repeating the `self._profile[col_id]` call

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+            if null_count == 0:
+                # No missing values to replicate
+                continue
+
+            # Calculate class priors
+            # i.e probability that a value in target col is or is not NaN
+            sample_size = getattr(self._profile[col_id], "sample_size")
```
taylorfturner, line 2171: see 2164 comment

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+                prev_mean_not_nan, prev_mean_nan = prev_class_means
+
+                mean_not_nan = utils.combine_means(
+                    prev_mean_not_nan, mean_not_nan, prev_true_count, added_true_count
+                )
+                mean_nan = utils.combine_means(
+                    prev_mean_nan, mean_nan, prev_null_count, added_null_count
+                )
+
+            col_name = getattr(self._profile[col_id], "name")
```
taylorfturner, line 2217: see 2164 comment

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+
+        if (
+            not self.options.synthetic_data.replicate_nan.is_enabled
+            or not other.options.synthetic_data.replicate_nan.is_enabled
+        ):
+            return {}
+
+        merged_properties = {}
+        for col_id in range(len(self._profile)):
+            self_null_count = getattr(self._profile[col_id], "null_count")
```
taylorfturner, line 2244: see 2164 comment

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+            other_null_count = getattr(other._profile[col_id], "null_count")
+            null_count = self_null_count + other_null_count
+            if null_count == 0:
+                continue
+
+            merged_properties[col_id] = {}
+            self_sample_size = getattr(self._profile[col_id], "sample_size")
```
taylorfturner, line 2251: see 2164 comment

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+            other_sample_size = getattr(other._profile[col_id], "sample_size")
+            sample_size = self_sample_size + other_sample_size
+            true_count = sample_size - null_count
+
+            prior_nan = null_count / sample_size
+            prior_not_nan = true_count / sample_size
+
+            merged_properties[col_id]["class_prior"] = [prior_not_nan, prior_nan]
+
+            col_name = getattr(self._profile[col_id], "name")
```
taylorfturner, line 2261: see 2164 comment

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+        """
+        Gets the null_count and sample_size of each column.
+
+        :return: Dictionary where column id maps to a list of [null_count, sample_size]
+        :rtype: dict
+        """
+        counts = {}
+        for col_id in range(len(self._profile)):
+            null_count = getattr(self._profile[col_id], "null_count", 0)
+            sample_size = getattr(self._profile[col_id], "sample_size", 0)
```
taylorfturner, line 2140: see 2164 comment

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+            col_name = getattr(self._profile[col_id], "name")
+            merged_properties[col_id]["column_name"] = col_name
+
+            # 'if null_count == 0' check guarantees that either self or other has _nan_replication_metrics for col_id
+            if col_id not in self._nan_replication_metrics:
+                other_mean = other._nan_replication_metrics[col_id]["class_mean"]
+                merged_properties[col_id]["class_mean"] = other_mean
+
+            elif col_id not in other._nan_replication_metrics:
+                self_mean = self._nan_replication_metrics[col_id]["class_mean"]
```
taylorfturner, line 2270: `self._nan_replication_metrics[col_id]["class_mean"]` repated...

`self._nan_replication_metrics[col_id]["class_mean"]` --> `class_mean = self._nan_replication_metrics[col_id]["class_mean"]` and then just use `class_mean` in the rest of the function

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2127,166 @@ def _update_chi2(self):
...
+                merged_properties[col_id]["class_mean"] = self_mean
+
+            else:
+                self_true_count = self_sample_size - self_null_count
+                other_true_count = other_sample_size - other_null_count
+
+                self_mean = self._nan_replication_metrics[col_id]["class_mean"]
+                other_mean = other._nan_replication_metrics[col_id]["class_mean"]
```
taylorfturner, line 2278: see comment on 2270 ... same comment for `other_mean`

dataprofiler/profilers/profiler_options.py
```diff
@@ -932,6 +932,34 @@ def _validate_helper(self, variable_path="CorrelationOptions"):
...
+        Validate the options do not conflict and cause errors.
+
+        :param variable_path: current path to variable set.
+        :type variable_path: str
+        :return: list of errors (if raise_error is false)
+        :rtype: list(str)
+        """
+        return super()._validate_helper(variable_path)
+
+
```
taylorfturner, line 962: as discussed, remove and put `nan_replication_metrics` as `self.nan_replication_metrics` in the  `class StructuredOptions(BaseOption)` in `profiler_options.py`

dataprofiler/profilers/profiler_options.py
```diff
@@ -1170,6 +1200,7 @@ def __init__(self, null_values=None):
         self.data_labeler = DataLabelerOptions()
         self.correlation = CorrelationOptions()
         self.chi2_homogeneity = BooleanOption(is_enabled=True)
+        self.synthetic_data = SyntheticDataOptions()
```
taylorfturner, line 1203: update to `StructuredOptions` I think

dataprofiler/profilers/utils.py
```diff
@@ -759,3 +759,20 @@ def merge_profile_list(list_of_profiles, pool_count=5):
...
+    :param mean2: Second mean vector
+    :type mean2: np.ndarray
+    :param n1: Number of samples mean1 was calculated from
+    :type n1: int
+    :param n2: Number of samples mean2 was calculated from
+    :type n2: int
+    :return: Combined mean vector
+    :rtype: np.ndarray
+    """
+    return (mean1 * n1 + mean2 * n2) / (n1 + n2)
```
taylorfturner, line 778: why not `return (mean1 + mean2) / 2`?

Ta7ar: That equation would only work if `n1 == n2`. If sample sizes the means were calculated from are different then we need to use `(mean1 * n1 + mean2 * n2) / (n1 + n2)`

dataprofiler/profilers/profiler_options.py
Ta7ar: This options is an attribute of the `StructuredOptions` class (the code highlighted is in the `def __init__` of the `StructuredOptions` class) so it wouldn’t really make sense to use `StructuredOptions` here. Was thinking of a simple `BooleanOptions`.

Ta7ar:
Updates in the recent commit include:
* Reduced code redundancy
* Removed `SyntheticDataOptions` class and use `BooleanOptions` class instead
* Renamed options from `synthetic_data` to `null_replication_metrics` (`null` instead of the previous `nan` because `DataProfiler` reports `null_count`, `null_types`, `null_types_index` so `null` sounds more cohesive)
* Change `class_mean` type from `np.ndarray` to `list` (serializable)
* `null_replication_metrics` reported in `data_stats` array
* Updated test cases

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,161 @@ def _update_chi2(self):
...
+
+        :param other: profile being added to this one.
+        :type other: StructuredProfiler
+        :return: merged nan replication metrics
+        :rtype: dict
+        """
+        merged_properties = {}
+        for col_id in range(len(self._profile)):
+            self_compiler = self._profile[col_id]
+            other_compiler = other._profile[col_id]
```
taylorfturner, line 2238: I may have done `self_profile` and `other_profile` but won't hold the PR up over this

dataprofiler/profilers/utils.py
taylorfturner: Ref: https://www.statisticshowto.com/combined-mean/

taylorfturner: not weighted mean

taylorfturner (approved):

Ta7ar:
Test coverage:
`dataprofiler/tests/profilers/test_profile_builder.py                                              1612      2    99%`

ksneab7 (approved):

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,161 @@ def _update_chi2(self):
...
+            prior_not_null = true_count / sample_size
+
+            # Partition data based on whether target column value is null or not
+            col_is_null = data.iloc[:, col_id].isnull()
+            partition_null = data[col_is_null]
+            partition_not_null = data[~col_is_null]
+
+            # Drop the target column from partitions
+            partition_null = partition_null.drop(columns=col_id)
+            partition_not_null = partition_not_null.drop(columns=col_id)
```
JGSweets, line 2188: Can we index the values with the boolean without dropping? Calculate the means without dropping?

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,161 @@ def _update_chi2(self):
...
+            sample_size = getattr(profile, "sample_size")
+            true_count = sample_size - null_count
+
+            # null_count and sample_size get updated on profile_update
+            # Therefore the priors are calculated from the updated values
+            prior_null = null_count / sample_size
+            prior_not_null = true_count / sample_size
+
+            # Partition data based on whether target column value is null or not
+            col_is_null = data.iloc[:, col_id].isnull()
```
JGSweets, line 2182: My concern with this call is `isnull` only generates a subset of the nulls determined by the clean operation in struct col. This could cause confusion.

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,161 @@ def _update_chi2(self):
...
+
+            # Drop the target column from partitions
+            partition_null = partition_null.drop(columns=col_id)
+            partition_not_null = partition_not_null.drop(columns=col_id)
+
+            # Calculate mean vectors of each partition
+            # i.e mean vectors of rows where target col is or is not null
+            mean_not_null, mean_null = (
+                partition_not_null.mean().tolist(),
+                partition_null.mean().tolist(),
```
JGSweets, line 2194: what if we were to calc this relative to the mean we know for each column against union of null to non-null in the other columns?

JGSweets: Also, do we need to calc the mean of both? We should know the mean of the entire column, and the mean of one partial, hence we could calculate the other without O(n)

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,161 @@ def _update_chi2(self):
...
+                added_null_count, added_sample_size = (
+                    null_count - prev_null_count,
+                    sample_size - prev_sample_size,
+                )
+                added_true_count = added_sample_size - added_null_count
+
+                prev_class_means = self._null_replication_metrics[col_id]["class_mean"]
+                prev_mean_not_null, prev_mean_null = prev_class_means
+
+                mean_not_null = utils.combine_means(
```
JGSweets, line 2212: What if we store sums and counts as oppose to re'calcing? Would that be better? Similar to how we store in numerical stats?

JGSweets (changes requested):
Would like to validate some ideas prior to approval.

dataprofiler/profilers/profile_builder.py
Ta7ar: The only reason we drop the `col_id` (which is the column where we will be replicating null values in the synthetic data) is because we do not want to have the mean of that column in the mean vector. So we could calculate the mean first and then drop the column but that would be more expensive (calculating the mean of a column we don't need)

dataprofiler/profilers/profile_builder.py
Ta7ar: Could you elaborate on what you mean by "`isnull` only generates a subset of the nulls"? From what I understand the function takes in `clean_sampled_dict` (line 2390) as an argument where each column in the said dict had the `clean_data_and_get_base_stats` (line 470) function called which removes "any nulls in column". So I don't think `isnull` would give a subset, it should give all null values as detected by `clean_data_and_get_base_stats`.

dataprofiler/profilers/profiler_options.py
```diff
@@ -932,6 +932,34 @@ def _validate_helper(self, variable_path="CorrelationOptions"):
         return errors
 
 
+class SyntheticDataOptions(BaseInspectorOptions):
```
JGSweets: The paradigm we've tried to follow previously is for the Options to match the naming of the property or grouping of properties being calculate. In StructuredProfiler it is called `nan_replication_metrics`. I think it would make sense for the options to reflect as: `NanReplicationMetricsOptions`. I'm open to suggestion on this.

I do see a pathway for choosing all synth data options such that it turns off anything that doesn't need to be calculated for synth data.

dataprofiler/profilers/profile_builder.py
JGSweets: I forget though if slicing also copies though.

dataprofiler/profilers/profiler_options.py
taylorfturner: I'm open to changing as well. The idea putting it in `StructuredProfiler` is that it is specific to structured data for right now.  I'd rather just create a `SyntheticDataOptions` and not an options class for just this one `nan_replication_metrics`. 

I do see a pathway for `SyntheticDataOptions` as well.

dataprofiler/profilers/profiler_options.py
```diff
@@ -1170,6 +1172,7 @@ def __init__(self, null_values=None):
         self.data_labeler = DataLabelerOptions()
         self.correlation = CorrelationOptions()
         self.chi2_homogeneity = BooleanOption(is_enabled=True)
+        self.null_replication_metrics = BooleanOption(is_enabled=False)
```
JGSweets, line 1175: nice

Ta7ar:
Changes made in latest commit:
* General improvements in runtime and space complexity.
  * `mean_not_null` no longer explicitly calculated by iterating over dataset. Instead `sum_null` is calculated by iterating over rows where target column value is `null`. Then `sum_null` is subtracted from `total_row_sum` (pre calculated) to obtain `sum_not_null` from which `mean_not_null` gets calculated by element wise division.
  * Copies of dataset are no longer created from partitioning data and dropping columns. Instead required calculations take place through explicit dataframe indexing.
* Removed `_get_null_and_sample_counts` and `utils.combine_means` as they are no longer needed.
* Updated test cases to test newly added field `class_sum` in `null_replication_metrics`.
* Indices of null values obtained from `null_types_index` instead of `df[col_id].isnull()`

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,171 @@ def _update_chi2(self):
...
+
+            # Gets list of null indices of the entire dataset
+            null_type_dict = getattr(profile, "null_types_index")
+            null_indices = set.union(*null_type_dict.values())
+            null_indices = np.asarray(list(null_indices))
+
+            # Keep only the null indices inside the chunk (reverse index shift)
+            if profile._index_shift is not None:
+                null_indices -= profile._index_shift
+                null_indices = [index for index in null_indices if index >= 0]
```
taylorfturner, line 2178:
```suggestion
                null_indices = [index - profile._index_shift for index in null_indices if index >= 0]
```

dataprofiler/profilers/profile_builder.py
```diff
@@ -1438,6 +1438,9 @@ def __init__(self, data, samples_per_update=None, min_true_samples=0, options=No
         self.correlation_matrix = None
         self.chi2_matrix = None
 
+        # capitalone/synthetic-data specific metrics
+        self._null_replication_metrics = dict()
```
taylorfturner, line 1442: instantiating class variables with `None` is best practice. Interesting we do it on line 1435. 

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

taylorfturner: we do it on line 1438 and 1439

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,171 @@ def _update_chi2(self):
...
+            # Convert numpy arrays to lists (serializable)
+            sum_null = sum_null.tolist()
+            sum_not_null = sum_not_null.tolist()
+
+            mean_null = mean_null.tolist()
+            mean_not_null = mean_not_null.tolist()
+
+            merged_properties[col_id] = {
+                # Array index serves as class label
+                # 0 indicates not null, 1 indicates null
```
taylorfturner, line 2290: nit picky and not going to hold the PR up over it... just move the comments to above the dict and not inside the dict.

dataprofiler/profilers/profile_builder.py
taylorfturner: or something like 

```
shift = profile._index_shift
[index - shift for index in null_indices if index >= 0]
```

Ta7ar: to condense those two lines into one, the list comprehension would be,
`null_indices = [index - profile._index_shift for index in null_indices if index - profile._index_shift >= 0]`
Could implement it but feels a bit too long for a list comprehension

taylorfturner: oh fair point.. yeah keep as is

taylorfturner (approved):

taylorfturner (approved):

taylorfturner (approved):

dataprofiler/profilers/profile_builder.py
```diff
@@ -2112,6 +2130,174 @@ def _update_chi2(self):
...
+        """
+        Calculate metrics needed for replicating null values in capitalone/synthetic-data.
+
+        Required for running LDA based binary classifier where predicted class label indicates whether
+        a value of a column should be NaN (1) or not (0).
+
+        :param clean_samples: input cleaned dataset
+        :type clean_samples: dict
+        """
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
```
JGSweets, line 2143: is this call still needed? I assume yes for the conversion to numeric? Do we need to apply the entire array? is this expensive?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1927,6 +1927,97 @@ def test_unique_row_ratio_empty_profiler(self):
         profiler = StructuredProfiler(pd.DataFrame([]))
         self.assertEqual(0, profiler._get_unique_row_ratio())
 
+    def test_null_replication_metrics_calculation(self):
+        data = pd.DataFrame(
+            {
+                "a": [3, 2, np.nan, 7, np.nan],
```
JGSweets, line 1933: should we validate that the clean, etc. works for nulls that aren't just caught by `np.nan`. e.g. blank spaces, or None?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1927,6 +1927,97 @@ def test_unique_row_ratio_empty_profiler(self):
...
+        # Test Profile updates
+        data_2 = pd.DataFrame(
+            {
+                "a": [3, 2, np.nan, 7, 5],
+                "b": [10, 10, 1, 4, 2],
+                "c": [2, 5, 3, np.nan, np.nan],
+            }
+        )
+
+        profiler.update_profile(data_2)
```
JGSweets, line 1972: shouldn't the values after this and the values after `merge` be the same?

dataprofiler/profilers/profile_builder.py
Ta7ar: Honestly, not too sure about that. Doing it because that is what correlation calculation does and I need the same type of data for null replication as correlation calculation uses.

dataprofiler/tests/profilers/test_profile_builder.py
Ta7ar: As long as the DataProfiler successfully detects null values the calculation shouldn't be affected. Could definitely add that to the tests though.

dataprofiler/tests/profilers/test_profile_builder.py
Ta7ar: Initially `profiler` has `data`, then update `profiler` with `data_2` (so `profiler` has `data` and `data_2`). Then initialize `profiler2` with `data` then merge it with `profiler`, therefore `merged_profile` contains `2*data+data_2` whereas `profile` after update contains `data+data_2` so it shouldn't be the same.

JGSweets: That makes sense. I thought it was going to be `data + data2`

I might suggest doing that as we should validate that merge / update should provide the same results.

one way would be to do merge prior to update to validate.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1927,6 +1927,97 @@ def test_unique_row_ratio_empty_profiler(self):
...
+                "c": [2, 5, 3, 5, 7],
+            }
+        )
+        profile_options = dp.ProfilerOptions()
+        profile_options.set(
+            {
+                "data_labeler.is_enabled": False,
+                "structured_options.multiprocess.is_enabled": False,
+                "null_replication_metrics.is_enabled": True,
+            }
```
JGSweets, line 1944: nice

taylorfturner (approved):

PR #579, opened 2022-08-04
Add property array to continuous distribution profile in GraphProfiler/ Integer Node ID identified as Integer in GraphData

MisterPNP:
Add property array to continuous distribution profile in GraphProfiler, necessary for tasks requiring replication of distribution.

### DIFF
--- a/dataprofiler/profilers/graph_profiler.py
+++ b/dataprofiler/profilers/graph_profiler.py
@@ -273,7 +273,13 @@ class GraphProfile(object):
 
     @BaseColumnProfiler._timeit(name="continuous_distribution")
     def _get_continuous_distribution(self, graph, continuous_attributes):
-        """Compute the continuous distribution of graph edge continuous attributes."""
+        """
+        Compute the continuous distribution of graph edge continuous attributes.
+
+        Returns properties array in the profile:
+        - 6-property length: norm, uniform, expon, logistic
+        - 7-property length: gamma, lognorm
+        """
         attributes = self._find_all_attributes(graph)
         continuous_distributions = dict()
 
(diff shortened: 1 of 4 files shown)


dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -63,20 +64,66 @@ def setUpClass(cls):
...
+                np_index = 0
+                for np_item in item:
+                    self.assertAlmostEqual(
+                        np_item, self.expected_props[index][np_index]
+                    )
+                    np_index += 1
+            else:
+                self.assertAlmostEqual(self.expected_props[index], item)
+            index += 1
+
```
taylorfturner, line 100: instead of doing the index that way, thoughts on using `enumerate()` instead... make the code a bit cleaner / less variable definitions to keep track of too

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -63,20 +64,66 @@ def setUpClass(cls):
...
         self.assertDictEqual(self.expected_profile, graph_profile.report())
```
taylorfturner, line 127: same above... `enumerate()`

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -85,7 +132,25 @@ def test_graph_data_object(self):
...
+                np_index = 0
+                for np_item in item:
+                    self.assertAlmostEqual(
+                        np_item, self.expected_props[index][np_index]
+                    )
+                    np_index += 1
+            else:
+                self.assertAlmostEqual(self.expected_props[index], item)
+            index += 1
+
```
taylorfturner, line 153: same... `enumerate()`

dataprofiler/tests/profilers/test_graph_profiler.py
taylorfturner: `for iter_value, item in enumerate(properties)`

dataprofiler/profilers/graph_profiler.py
```diff
@@ -299,10 +309,27 @@ def _get_continuous_distribution(self, graph, continuous_attributes):
...
+                if best_fit == "lognorm":
+                    shape, loc, scale = fit
+                    properties = [shape, loc, scale, mean, variance, skew, kurtosis]
+                elif best_fit == "gamma":
+                    a, loc, scale = fit
+                    properties = [a, loc, scale, mean, variance, skew, kurtosis]
+                else:
+                    loc, scale = fit
+                    properties = [loc, scale, mean, variance, skew, kurtosis]
+
```
taylorfturner, line 326: we can clean this up a bit ... we know that `[loc, scale, mean, variance, skew, kurtosis]` will always be in the list... so the if / else should really just add `a` or `shape` to the list .... else `if not lognorm or gamma` just keep moving and have propertyes == `[loc, scale, mean, variance, skew, kurtosis]`

taylorfturner (changes requested):
couple changes for more DRY code

dataprofiler/tests/profilers/test_graph_profiler.py
taylorfturner: also this for loop is the same as the one below... so we can probably clean the for loop up too

taylorfturner (approved):

JGSweets (approved):

taylorfturner (approved):

PR #599, opened 2022-08-10
Add Graph Github Pages

MisterPNP:
Add Graph Github Pages


taylorfturner (approved):

taylorfturner (changes requested):
Graph Reader for DataReader section in docs

docs/source/data_readers.rst
```diff
@@ -125,6 +125,28 @@ Possible `options`:
 * data_format: user selected format in which to return data. Currently only supports "text".
 * samples_per_line - chunks by which to read in the specified dataset
 
+GraphData
```
JGSweets, line 128: I think this needs go before text as text is the catchall

JGSweets: also need to add up above in the list of data readers

taylorfturner (approved):

PR #626, opened 2022-09-09
New Data Labeler: ColumnNameModel Build

taylorfturner:


dataprofiler/labelers/column_name_model.py
```diff
@@ -31,8 +32,8 @@ def __init__(self, label_mapping=None, parameters=None):
         # parameter initialization
         if not parameters:
             parameters = {}
-        parameters.setdefault('negative_dataframe', )
-        parameters.setdefault('positive_dataframe', )
+        parameters.setdefault('false_positive_df', None)
```
JGSweets, line 35: nice naming

taylorfturner: lol thx

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,232 @@
...
+                errors.append("`{}` is not an accepted parameter.".format(param))
+        if errors:
+            raise ValueError("\n".join(errors))
+
+    def _compare_negative(self, list_of_column_names, check_values_dict, negative_threshold):
+        """Filter out column name examples that are false positives"""
+        scores = self._model(
+                list_of_column_names,
+                check_values_dict,
+                self._make_lower_case(),
```
JGSweets, line 91: is this something we are going to parameterize? why not just do `str.lower`?

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _need_to_reconstruct_model(self):
+        pass
+
+    def reset_weights(self):
+        """Reset weights."""
+        pass
+
+    def _make_lower_case(str, **kwargs):
+        return str.lower()
```
JGSweets, line 137: should it be `str.lower()` or `str.lower`?

requirements-ml.txt
```diff
@@ -1,5 +1,6 @@
 scikit-learn>=0.23.2
 keras>=2.4.3
+rapidfuzz>=2.6.1
```
JGSweets, line 3: we might need to put some error catching if this is the case, e.g. user doesn't install [ml]

dataprofiler/labelers/column_name_model.py
taylorfturner: Good question -- thought the same myself. The parameter for the `process.cdist` function must be `callable` so that is a big requirement. `str.lower()` isn't callable since that returns a `str` object type. 

Documentation to the `process.cdist` function [here](https://maxbachmann.github.io/RapidFuzz/Usage/process.html#cdist)

dataprofiler/labelers/column_name_model.py
taylorfturner:
```In [1]: str.lower
Out[1]: <function str.lower()>
```
should return a `str` object type... so think we actually do want to call the `lower()` as a function call

requirements-ml.txt
taylorfturner: true true!

dataprofiler/labelers/base_model.py
```diff
@@ -134,6 +134,7 @@ def get_class(cls, class_name):
         # Import possible internal models
         from .character_level_cnn_model import CharacterLevelCnnModel  # NOQA
         from .regex_model import RegexModel  # NOQA
+        from .column_name_model import ColumnNameModel  # NOQA
```
taylorfturner, line 137: adding the model to be loadable as a class

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,225 @@
+"""Contains class for regex data labeling model."""
```
taylorfturner, line 1: entirely new model based regex model type

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -0,0 +1,292 @@
...
+                { 
+                    'attribute': 'role',
+                    'label': 'name',
+                },
+                {
+                    'attribute': 'send_address',
+                    'label': 'address',
+                },
+            ]
+        }
```
JGSweets, line 46: format fixing

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -0,0 +1,292 @@
...
+
```
JGSweets, line 47: does this need 2 spaces?

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -0,0 +1,292 @@
+import json
```
taylorfturner, line 1: test suite for each method in the new `ColumnNameModel` class

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # test show confidences
+        expected_output = {
+            "ssn": 
+                {
+                    "pred": "ssn",
+                    "conf": 100.0
+                }
+            }
+        model_output = model.predict(data=["ssn", "role_name", "wallet_address"], show_confidences=True)
```
JGSweets, line 195: do we have a test for false as well?

dataprofiler/tests/labelers/test_column_name_model.py
taylorfturner: fixing in next commit

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -0,0 +1,292 @@
...
+        self.assertTrue(len(logs.output))
+
+        # test show confidences
+        expected_output = {
+            "ssn": 
+                {
+                    "pred": "ssn",
+                    "conf": 100.0
+                }
+            }
```
JGSweets, line 194: should we just check the whole thing rather than an individual?

dataprofiler/tests/labelers/test_column_name_model.py
taylorfturner: which part?

dataprofiler/tests/labelers/test_column_name_model.py
JGSweets: can use a loop on the keys to test each. But i'm not sure if we will need to do the `np.array_equal` with the `np.assertTrue` that was only bc of extreme float condition testing but these don't have the same float issues?

dataprofiler/tests/labelers/test_column_name_model.py
JGSweets: end of the params and before the func

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -0,0 +1,290 @@
...
+                { 
+                    'attribute': 'role',
+                    'label': 'name',
+                },
+                {
+                    'attribute': 'send_address',
+                    'label': 'address',
+                },
+            ]
+        }
```
JGSweets, line 46: untab, yes?

dataprofiler/tests/labelers/test_column_name_model.py
taylorfturner: pre-commit forces two spaces ....

requirements-ml.txt
taylorfturner:
```
try:
    from rapidfuzz import fuzz, process
except ImportError:
    raise ImportError("Please ensure you run `pip install requirements-ml.txt`")
```

in `column_name_model.py`, this is an option but I'd rather just document setup and installation well. In the docs we should just be explicit that user must install `requirements-ml.txt` for the `column_name_model.py`

dataprofiler/tests/labelers/test_column_name_model.py
taylorfturner: yes, the below is a `FALSE` scenario
```
with self.assertLogs(
            "DataProfiler.labelers.column_name_model", level="INFO"
        ) as logs:
            model_output = model.predict(data=["ssn", "role_name", "wallet_address"])
```

dataprofiler/tests/labelers/test_column_name_model.py
taylorfturner: good call

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,248 @@
+"""Contains class for column name data labeling model."""
+import json
+import os
+
+import numpy as np
+
+try:
+    from rapidfuzz import fuzz, process
+except ImportError:
+    raise ImportError("Please ensure you run `pip install requirements-ml.txt`")
```
JGSweets, line 10: this is pretty much guaranteed to run which means this would happen all the time for any model, even the regex one.
However, doing something like:
https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/reports/graphs.py#L12
with:
https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/reports/graphs.py#L21
would alllow modules to import, but only error on function / class calls

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,248 @@
+"""Contains class for column name data labeling model."""
+import json
+import os
+
+import numpy as np
+
+try:
+    from rapidfuzz import fuzz, process
```
JGSweets, line 8: I'm wondering if we should just import `rapidfuzz` bc otherwise down below we don't really know where `fuzz` and especially `process` come from. the latter being extremely generic.

taylorfturner: totally fair -- great point

requirements-ml.txt
JGSweets: yes, def be explicit about it in the docs

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,253 @@
...
+                    """`{}` must be a list of dictionaries with at
+                    least 'attribute' as the key""".format(
+                        param
+                    )
+                )
+            elif param == "true_positive_dict" and (
+                not isinstance(value, list) or not isinstance(value[0], dict)
+            ):
+                errors.append(
+                    """`{}` must be a list of dictionaries each with the following
```
micdavis, line 80: doesn't check for `attribute` and `label` key in list of dicts. Does it need to?

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,253 @@
...
+        :type show_confidences:
+        :param verbose: Flag to determine whether to print status or not
+        :type verbose: bool
+        :return: char level predictions and confidences
+        :rtype: dict
+        """
+        false_positive_dict = self._parameters["false_positive_dict"]
+        if false_positive_dict:
+            data = self._compare_negative(
+                data, false_positive_dict, negative_threshold=50
```
micdavis, line 206: is this `negative_threshold` value arbitrary? Should it be a constant at the top of the file or in a config somewhere?

dataprofiler/tests/labelers/test_column_name_model.py
JGSweets: ahh i missed it, sorry since it wasn't declared i missed it.

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,253 @@
...
+            )
+            if verbose:
+                logger.info("compare_negative process complete")
+
+        output = self._compare_positive(
+            data,
+            self._parameters["true_positive_dict"],
+            positive_threshold=85,
```
micdavis, line 214: same question as above regarding `positive_threshold`

dataprofiler/labelers/column_name_model.py
taylorfturner: good catch -- nah this should be a param and not hard coded

dataprofiler/labelers/column_name_model.py
taylorfturner: fixing -- good catch

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,258 @@
+"""Contains class for column name data labeling model."""
```
taylorfturner, line 1: whole new model -- based on regex_model.py to some extent

requirements-ml.txt
taylorfturner: added the `require_model` too on this

dataprofiler/labelers/column_name_model.py
```diff
@@ -0,0 +1,258 @@
...
+        pass
+
+    def _need_to_reconstruct_model(self):
+        pass
+
+    def reset_weights(self):
+        """Reset weights function."""
+        pass
+
+    @require_module(["rapidfuzz"])
```
JGSweets, line 157: do we have a test for this?

JGSweets: Also, I think this is going to spout a graph profiling error as oppose to a labeler error

tyfarnan (approved):

Issue #610, opened 2022-08-12
Static Type Hinting: `dataprofiler/data_readers/`


taylorfturner:
**Is your feature request related to a problem? Please describe.**
Improve the project in the area of 
- Make collaboration and code contribution easier
- Getting started with the code when a contributor is new to the code base

**Describe the outcome you'd like:**
Utilizing types from [mypy](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html) documentation for python3, implement type hinting in the `dataprofiler/data_readers/` folder. In this case, because there are multiple files in the folder, at the very minimum the changes should be partitioned out into multiple Pull Requests (PRs). Potentially this could go so far as 1 PR --> 1 file update with type hints.

**Additional context:**
- [mypy](https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html) documentation
- [Why use python types?](https://medium.com/vacatronics/why-you-should-consider-python-type-hints-770e5cb1570f)

taylorfturner:
Because this folder is so large, I recommend making more PRs that are smaller (3 - 4 files at a time / keep the net diff number lower when possible).

taylorfturner:
@Sanketh7 just these last three PRs and this will be complete? Is that right?

Sanketh7:
> @Sanketh7 just these last three PRs and this will be complete? Is that right?

There are 2 more files that need static typing but these are very short and I'll have a PR for them up soon.

taylorfturner:
All merged. Closing. Thanks @Sanketh7

PR #629, opened 2022-09-14
Post Processor for ColumnNameModel

taylorfturner:


taylorfturner:
#610

dataprofiler/tests/labelers/test_data_processing.py
```diff
@@ -2928,3 +2930,69 @@ def test_random_state_constructor(self):
...
+                    "attirbute": "test",
+                    "label": "test",
+                }
+            ],
+            "positive_threshold_config": 85,
+        }
+        post_processor = ColumnNameModelPostprocessor(parameters=params)
+
+        # call save processor func
+        post_processor._save_processor(".")
```
taylorfturner, line 2998: I despise this test... really need it mocked

dataprofiler/labelers/column_name_model.py
```diff
@@ -36,6 +36,8 @@ def __init__(self, parameters=None):
             parameters = {}
         parameters.setdefault("false_positive_dict", None)
         parameters.setdefault("true_positive_dict", None)
+        parameters.setdefault("include_label", True)
+        parameters.setdefault("negative_threshold_config", None)
```
taylorfturner, line 40: params for the `column_name_model` and not `predict()` parameters

dataprofiler/tests/labelers/test_column_name_model.py
```diff
@@ -113,51 +94,22 @@ def test_help(self, mock_stdout):
...
-        model_output = model.predict(
-            data=["ssn", "role_name", "wallet_address"], show_confidences=True
-        )
-        self.assertTrue(np.array_equal(expected_output, model_output))
+        # `show_confidences` is disabled currently
+        # should raise error if set to `True`
+        with self.assertRaises(NotImplementedError):
+            model.predict(
+                data=["ssn", "role_name", "wallet_address"], show_confidences=True
+            )
```
taylorfturner, line 112: disabled `show_confidences` right now. Need to clean this up and enable it as a toggle

dataprofiler/labelers/column_name_model.py
```diff
@@ -112,35 +136,6 @@ def _compare_negative(
...
-            if scores[i][0] > positive_threshold:
-                output_dictionary[list_of_column_names[i]] = {}
-                output_dictionary[list_of_column_names[i]]["pred"] = check_values_dict[
-                    scores[i][1]
-                ]["label"]
-                if show_confidences:
-                    output_dictionary[list_of_column_names[i]]["conf"] = scores[i][0]
-
-        return output_dictionary
-
```
taylorfturner, line 143: completely got rid of the `_compare_positive` becuase we don't need it anymore after moving some things to the post processor

dataprofiler/labelers/column_name_model.py
```diff
@@ -186,9 +181,6 @@ def predict(
         batch_size=None,
         show_confidences=False,
         verbose=True,
-        include_label=True,
-        negative_threshold_config=50,
-        positive_threshold_config=85,
```
taylorfturner, line 191: parameters to the class __init__ not the predict method. Keeping the API consistent across all the models

dataprofiler/labelers/column_name_model.py
```diff
@@ -208,18 +201,29 @@ def predict(
...
+            scorer=rapidfuzz.fuzz.token_sort_ratio,
+            include_label=self._parameters["include_label"],
         )
+
+        if show_confidences:
+            raise NotImplementedError(
+                """`show_confidences` parameter is disabled
+                for Proof of Concept implementation. Confidence
+                values are enabled by default."""
+            )
```
taylorfturner, line 225: Need to think through how to enable this

dataprofiler/labelers/data_processing.py
```diff
@@ -2079,3 +2079,129 @@ def process(self, data, labels=None, label_mapping=None, batch_size=None):
...
+                the order in which entities are given priority must be set
+            random_state: Random state setting to be used for randomly
+                selecting a prediction when two labels have equal opportunity
+                for a given sample.
+        :type parameters: dict
+        :return: None
+        """
+        # allowed_parameters = self.__class__.__init__.__code__.co_varnames[
+        #     1 : self.__class__.__init__.__code__.co_argcount
+        # ]
```
taylorfturner, line 2116: don't think we need this anymore for this model @JGSweets

dataprofiler/labelers/data_processing.py
```diff
@@ -2079,3 +2079,129 @@ def process(self, data, labels=None, label_mapping=None, batch_size=None):
...
+            cls.__name__
+            + "\n\n"
+            + "Parameters:\n"
+            + param_docs[param_start_ind:param_end_ind]
+            + "\nProcess Output Format:\n"
+            "    Each sample receives a label.\n"
+            "    Original data - ['My', 'String', ...]\n"
+            "    Output labels - ['<LABEL_1>', '<LABEL_2>', "
+            "..(num samples)]"
+        )
```
taylorfturner, line 2176: needs an update

dataprofiler/tests/labelers/test_data_processing.py
```diff
@@ -2928,3 +2930,69 @@ def test_random_state_constructor(self):
...
+            ],
+            "positive_threshold_config": 85,
+        }
+
+        processor = ColumnNameModelPostprocessor(parameters=params)
+        process_output = processor.process(data)
+
+        self.assertEqual(process_output, expected_output)
+
+    def test_save_processor(self):
```
JGSweets, line 2985: so this would need to be mocked bc otherwise it saves the file to disk

JGSweets: `@mock.patch("builtins.open")`

JGSweets: if we use the other examples, can then use:
`mock_file = setup_save_mock_open(mock_open)`

JGSweets: Check out https://github.com/capitalone/DataProfiler/pull/629/files#diff-fd672afb661ad4d53cbd966058bbb9f993407275e507567950ae75290e45cc8eR2896
for the rest, just above in regex's post_save

dataprofiler/labelers/column_name_model.py
taylorfturner: keeping the `predict()` API consistent

dataprofiler/labelers/column_name_model.py
```diff
@@ -77,14 +87,28 @@ def _validate_parameters(self, parameters):
                 not isinstance(value, list)
                 or not isinstance(value[0], dict)
                 and "attribute" not in value[0].keys()
-                and "list" not in value[0].keys()
+                and "label" not in value[0].keys()
```
taylorfturner, line 90: typo

dataprofiler/tests/labelers/test_column_name_model.py
taylorfturner: an option is to put this in the post processor part of the labeler, but that breaks the API paradigm for the `model's` `predict()` method ...

dataprofiler/labelers/data_processing.py
```diff
@@ -2079,3 +2079,113 @@ def process(self, data, labels=None, label_mapping=None, batch_size=None):
...
+):
+    """Subclass of BaseDataPostprocessor for postprocessing regex data."""
+
+    def __init__(self, parameters):
+        """Initialize the ColumnNameModelPostProcessor class."""
+        # set parameters
+        if not parameters:
+            parameters = {}
+        parameters.setdefault("true_positive_dict", None)
+        parameters.setdefault("positive_threshold_config", None)
```
JGSweets, line 2095: this ultimately should mimic the other procesors where we take in paramters as individual args: e.g.

```python
def __init__(
        self,
        default_label="UNKNOWN",
        pad_label="PAD",
        flatten_separator=" ",
        use_word_level_argmax=False,
        output_format="character_argmax",
        separators=(" ", ",", ";", "'", '"', ":", "\n", "\t", "."),
        word_level_min_percent=0.75,
    ):
        """
        Initialize the CharPostprocessor class.

        :param default_label: Key for label_mapping that is the default label
        :type default_label: string (could be int, char, etc.)
        :param pad_label: Key for label_mapping that is the pad label
        :type pad_label: string (could be int, char, etc.)
        :param flatten_separator: separator used to put between flattened
            samples.
        :type flatten_separator: str
        :param use_word_level_argmax: whether to require the argmax value of
            each character in a word to determine the word's entity
        :type use_word_level_argmax: bool
        :param output_format: (character_argmax vs NER) where character_argmax
            is a list of encodings for each character in the input text and NER
            is in the dict format which specifies start,end,label for each
            entity in a sentence
        :type output_format: str
        :param separators: list of characters to use for separating words within
            the character predictions
        :type separators: tuple(str)
        :param word_level_min_percent: threshold on generating dominant
            word_level labeling
        :type word_level_min_percent: float
        """
        super().__init__(
            default_label=default_label,
            pad_label=pad_label,
            flatten_separator=flatten_separator,
            use_word_level_argmax=use_word_level_argmax,
            output_format=output_format,
            separators=separators,
            word_level_min_percent=word_level_min_percent,
        )
```

JGSweets: the super should auto handle parameters where need be,
if we want `true_positive_dict` to be a dict by default, we could always do a 
```python
if true_positive_dict is None:
    true_positive_dict = {}
```

dataprofiler/labelers/data_processing.py
```diff
@@ -2079,3 +2079,113 @@ def process(self, data, labels=None, label_mapping=None, batch_size=None):
...
+    BaseDataPostprocessor, metaclass=AutoSubRegistrationMeta
+):
+    """Subclass of BaseDataPostprocessor for postprocessing regex data."""
+
+    def __init__(self, parameters):
+        """Initialize the ColumnNameModelPostProcessor class."""
+        # set parameters
+        if not parameters:
+            parameters = {}
+        parameters.setdefault("true_positive_dict", None)
```
JGSweets, line 2094: do we want these defaults to both be `None`, will that error by default or do we want to use any defaults?

dataprofiler/labelers/data_processing.py
```diff
@@ -2079,3 +2079,113 @@ def process(self, data, labels=None, label_mapping=None, batch_size=None):
...
+        parameters.setdefault("positive_threshold_config", None)
+
+        super().__init__(**parameters)
+
+    def _validate_parameters(self, parameters):
```
JGSweets, line 2099: I think we are missing tests for this

JGSweets (changes requested):

micdavis (approved):

dataprofiler/labelers/data_processing.py
taylorfturner: added

dataprofiler/labelers/data_processing.py
taylorfturner: - `parameters.setdefault("true_positive_dict", None)` default for this is a no go in my opinion unless we want "out of the box" examples of attributes and labels, which I'd say is out of scope for this PR
- `parameters.setdefault("positive_threshold_config", None)`: however a default for this is a fair point, but really only matters if the `true_positive_dict` is set as well...

dataprofiler/labelers/data_processing.py
taylorfturner: ah good point... I was following the `model` paradigm still. I'll update this

micdavis (dismissed):

micdavis (dismissed):

JGSweets (approved):

micdavis (approved):

PR #630, opened 2022-09-15
Static Typing profilers/utils.py

tonywu315:


tonywu315:
#608

dataprofiler/profilers/utils.py
```diff
@@ -455,26 +488,18 @@ def find_diff_of_dates(stat1, stat2):
...
-    if diff.days >= 0:
-        return "+" + str(diff)
-
-    return "-" + str(abs(diff))
+    if stat1 is None or stat2 is None:
+        return [i if i is None else i.strftime("%x %X") for i in [stat1, stat2]]
+    elif stat1 != stat2:
+        diff = stat1 - stat2
+        return ("+" if diff.days >= 0 else "-") + str(abs(diff))
+    return "unchanged"
```
JGSweets, line 499: any reason for removing the 
`diff = find_diff_of_numbers(stat1, stat2)`
and handling this here?

tonywu315: `find_diff_of_numbers()` only takes in ints and floats. If I make it also take in `datetime`, then mypy complains because ints/floats can't be subtracted with `datetime`.

dataprofiler/profilers/utils.py
```diff
@@ -462,26 +491,18 @@ def find_diff_of_dates(stat1, stat2):
...
-    if isinstance(diff, list):
-        # Deal with the case where one stat is None
-        return [None if i is None else i.strftime("%x %X") for i in diff]
-
-    # Must be timedelta object
-    if diff.days >= 0:
-        return "+" + str(diff)
-
-    return "-" + str(abs(diff))
+    if stat1 is None or stat2 is None:
```
JGSweets, line 497: I believe this will actually error bc we have an error in our test.
Line 107 in : https://github.com/capitalone/DataProfiler/blob/69d71736ca032b20a290c5573f9b3c7a9d2b1577/dataprofiler/tests/test_utils.py
```python
self.assertEqual("unchanged", utils.find_diff_of_numbers(None, None))
```
should be:
```python
self.assertEqual("unchanged", utils.find_diff_of_dates(None, None))
```

As a result, this passes currently, but would with the proper test.

Can we add this fix to the test in the PR

JGSweets (changes requested):
Fix to `diff_of_dates` / test.

dataprofiler/profilers/utils.py
JGSweets: Fair, should we fix numbers to take in any type that can have a subtraction operation? Not sure if that's actually feasible. Otherwise, can do as is but fix the above.

JGSweets: Found this:
https://stackoverflow.com/questions/37669222/how-can-i-hint-that-a-type-is-comparable-with-typing
```python
from __future__ import annotations

from abc import abstractmethod
from typing import MutableSequence, Protocol, TypeVar


class Comparable(Protocol):
    """Protocol for annotating comparable types."""

    @abstractmethod
    def __lt__(self: CT, other: CT) -> bool:
        pass


CT = TypeVar("CT", bound=Comparable)

def comparison_sort(s: MutableSequence[CT]) -> None:
    pass


comparison_sort([1, 2, 3])  # OK
comparison_sort([1.0, 2.0, 3.0])  # OK
comparison_sort(["42", "420", "2137"])  # OK
comparison_sort([1, 2, "3"])  # mypy error 
```

Could we alter `Comparable` name and abstractions to use `__eq__ `and `__sub__`. which would allow a datetime to go into diff_of_numbers?

Maybe that's not appropriate either bc it would take in anything with that.

tonywu315: I added this and it seems to work fine. The only thing is that I need to make `__sub__` return `Any` because subtracting `datetime` results in a `timedelta`.

dataprofiler/profilers/utils.py
```diff
@@ -364,11 +391,28 @@ def biased_kurt(df_series):
...
+    @abstractmethod
+    def __sub__(self: T, other: T) -> Any:
+        """Compare two subtractables."""
+        pass
+
+
+T = TypeVar("T", bound=Subtractable)
+
+
+def find_diff_of_numbers(stat1: Optional[T], stat2: Optional[T]) -> Any:
```
JGSweets, line 415: I think that's fine for now, but in the log run we should try to avoid using `Any` where possible.

Ideally we could create a typing file we use that makes these descriptive declarations. As things like `any` or `Optional[T]` don't make sense to someone reading it without diving deep into the code.

Will approve though, a future state for which to strive.

JGSweets (approved):

micdavis (approved):

PR #640, opened 2022-09-16
refactor: validate parameters and the returns of functions

stefanycoimbra:
Imports additions to validate the types of functions' parameters and its returns using mypy:
Line 4: from typing import List, Callable, TypeVar, cast, Any
Line 5: from typing import NoReturn

Initialize the generic type to use with decorators:
Line 8: F = TypeVar('F', bound=Callable[..., Any])

Other changes:
Line 11: add the "NoReturn" for the function warn_missing_module()
Lines 30 and 41: add the return type as F (generic type)
Lines 55 and 57: add the cast to using decorators with wrappers (see in https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators)

### DIFF
--- a/dataprofiler/reports/utils.py
+++ b/dataprofiler/reports/utils.py
@@ -27,7 +27,7 @@ def warn_missing_module(graph_func: str, module_name: str) -> NoReturn:
 
 
 
-def require_module(names: list[str]) -> F:
+def require_module(names: List[str]) -> F:
     """
     Check if a set of modules exists in sys.modules prior to running function.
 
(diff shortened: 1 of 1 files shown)


JGSweets (dismissed):
Very cool! LGTM

dataprofiler/reports/utils.py
```diff
@@ -21,7 +26,8 @@ def warn_missing_module(graph_func, module_name):
     warnings.warn(warning_msg, RuntimeWarning, stacklevel=3)
 
 
-def require_module(names):
+
+def require_module(names: list[str]) -> F:
```
JGSweets, line 30: I think it's because this should be List?

JGSweets (approved):

taylorfturner (approved):
Epic 🚀

PR #643, opened 2022-09-19
Static Typing profilers/profile_builder.py

tonywu315:


tonywu315:
#608

JGSweets:
@tonywu315 FYI, I think your code is out of date. You will likely need to rebase this branch onto `C1/main`.

dataprofiler/profilers/profile_builder.py
```diff
@@ -955,10 +986,6 @@ def _restore_data_labelers(self, data_labeler=None):
                 self.options.set({"data_labeler.is_enabled": False})
                 self.options.set({"data_labeler.data_labeler_object": data_labeler})
 
-            except Exception as e:
```
JGSweets, line 958: Why was this removed?

JGSweets (changes requested):

dataprofiler/profilers/profile_builder.py
JGSweets: nvm

dataprofiler/profilers/profile_builder.py
```diff
@@ -955,10 +986,6 @@ def _restore_data_labelers(self, data_labeler=None):
...
-                utils.warn_on_profile("data_labeler", e)
-                self.options.set({"data_labeler.is_enabled": False})
-
```
taylorfturner, line 961: is this change needed as a result of static typing update?

taylorfturner (commented):
one main comment around removing an exception catch

dataprofiler/profilers/profile_builder.py
JGSweets: NVM. I think this is actually broken code.  I believe there's a path which makes it not a duplicate.

Could definitely be refactored.

Essentially, if DP is not None and an error was raised, it would try to set it again in the first case and re raise I beleive.

JGSweets: Maybe it is ideal to re-raise bc we shouldn't be setting something that fails.

dataprofiler/profilers/profile_builder.py
taylorfturner: I'd rather keep these PRs only static typing and not changing code ... unless its absolutely necessary

dataprofiler/profilers/profile_builder.py
```diff
@@ -2353,19 +2431,19 @@ def tqdm(level):
...
+            sample_id_list[0] = sample_id_list[0][:sample_size]
+            # Sort the sample_id_list and replace prior
+            sample_id_list[0] = sorted(sample_id_list[0])
 
         # Numpy arrays allocate to heap and can be shared between processes
         # Non-locking multiprocessing fails on machines without POSIX (windows)
         # The function handles that situation, but will be single process
         # Newly introduced features (python3.8) improves the situation
-        sample_ids = np.array(sample_ids)
+        sample_ids = np.array(sample_id_list)
```
JGSweets, line 2446: I believe the intent is to overwrite the existing to drop it from memory.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1269,12 +1310,12 @@ def _clean_data_and_get_base_stats(self, data, sample_size, min_true_samples=Non
         data = data.apply(str)
 
         # get memory size
-        base_stats = {"memory_size": utils.get_memory_size(data, unit="M")}
+        base_stats: Dict = {"memory_size": utils.get_memory_size(data, unit="M")}
 
         # Setup sample generator
         sample_ind_generator = utils.shuffle_in_chunks(len_data, chunk_size=sample_size)
 
-        true_sample_list = set()
```
JGSweets, line 1277: I believe the intent is to overwrite to allow the garbage collection for memory

dataprofiler/profilers/profile_builder.py
JGSweets: that second exception i don't think gets hit ever so I think it's fine to remove.

tonywu315: I'll make a separate PR for any code changes in the future.

taylorfturner: perfect, thanks, tony!

dataprofiler/profilers/profile_builder.py
```diff
@@ -2358,14 +2436,14 @@ def tqdm(level):
         # If there are no minimum true samples, you can sort to save time
         if min_true_samples in [None, 0]:
             sample_ids[0] = sample_ids[0][:sample_size]
-            # Sort the sample_ids and replace prior
+            # Sort the sample_id_list and replace prior
```
taylorfturner, line 2439: revert this back @tonywu315

taylorfturner (changes requested):

dataprofiler/profilers/profile_builder.py
taylorfturner, line 2439:
```suggestion
            # Sort the sample_id and replace prior
```

taylorfturner (approved):

micdavis (approved):

JGSweets (approved):

PR #644, opened 2022-09-19
Static Typing profilers/profiler_options.py

tonywu315:


dataprofiler/profilers/profiler_options.py
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env python
 """Specify the options when running the data profiler."""
+from __future__ import annotations
```
JGSweets, line 3: is this needed?

JGSweets:
@tonywu315 FYI, I think your code is out of date. You will likely need to rebase this branch onto `C1/main`.

JGSweets:
@tonywu315 FYI, I think your code is out of date. You will likely need to rebase this branch onto `C1/main`.

dataprofiler/profilers/profiler_options.py
tonywu315: On line 36, I added
```
if TYPE_CHECKING:
    from .profile_builder import BaseProfiler
```
because BaseProfiler has a circular dependency with utils. Adding this means that I need to postpone the evaluation of type annotations, since BaseProfiler is not defined until later. In Python 3.7, this can be done by importing annotations from `__future__` or by using `"BaseProfiler"` instead of `BaseProfiler` for types.

tonywu315: Oh I thought this was for the utils PR. For this one, it's because a lot of the class types aren't defined until later.

JGSweets (approved):

ksneab7 (approved):

micdavis (approved):

PR #648, opened 2022-09-20
Static Typing profilers/numerical_column_stats.py

tonywu315:
#608


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1583,7 +1651,7 @@ def is_int(x):
             return a == b
 
     @staticmethod
-    def np_type_to_type(val):
+    def np_type_to_type(val: Any) -> Union[int, float]:
```
JGSweets, line 1654: Just want to validate this is correct as the value may be an `np.float`. Additionally, the case for `np.int`
https://numpy.org/devdocs/reference/typing.html

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1594,6 +1662,6 @@ def np_type_to_type(val):
         """
         if isinstance(val, np.integer):
             return int(val)
-        if isinstance(val, np.float):
```
JGSweets, line 1597: Is this intentional to remove?
what if instead we did `np.floating`? would handling float32 mess up the input as well?

tonywu315: `np.float` is a deprecated alias for `float` since [NumPy 1.20.0](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations). I think using `np.floating` is a good idea.

dataprofiler/profilers/numerical_column_stats.py
tonywu315: The function now checks if `val` is `np.floating` or `np.integer` which should cover everything.

micdavis (dismissed):

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1302,16 +1340,16 @@ def median_abs_deviation(self):
...
 
-    def _get_quantiles(self):
+    def _get_quantiles(self) -> None:
         """
         Retrieve quantile set based on specified number of quantiles in self.quantiles.
 
         :return: list of quantiles
         """
-        percentiles = np.linspace(0, 100, len(self.quantiles) + 2)[1:-1]
+        percentiles: List[float] = np.linspace(0, 100, len(self.quantiles) + 2)[1:-1]
```
tyfarnan, line 1349: does np.linspace return a list? or an ndarray?

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: wonder if this should be `Union[int, float]`

dataprofiler/profilers/numerical_column_stats.py
taylorfturner:
```In [4]: np.linspace(0,1)
Out[4]: 
array([0.        , 0.02040816, 0.04081633, 0.06122449, 0.08163265,
       0.10204082, 0.12244898, 0.14285714, 0.16326531, 0.18367347,
       0.20408163, 0.2244898 , 0.24489796, 0.26530612, 0.28571429,
       0.30612245, 0.32653061, 0.34693878, 0.36734694, 0.3877551 ,
       0.40816327, 0.42857143, 0.44897959, 0.46938776, 0.48979592,
       0.51020408, 0.53061224, 0.55102041, 0.57142857, 0.59183673,
       0.6122449 , 0.63265306, 0.65306122, 0.67346939, 0.69387755,
       0.71428571, 0.73469388, 0.75510204, 0.7755102 , 0.79591837,
       0.81632653, 0.83673469, 0.85714286, 0.87755102, 0.89795918,
       0.91836735, 0.93877551, 0.95918367, 0.97959184, 1.        ])

In [5]: type(np.linspace(0,100))
Out[5]: numpy.ndarray```

Comment (taylorfturner) [reply dataprofiler/profilers/numerical_column_stats.py:1349]: looks like `ndarray`

Comment (taylorfturner) [inline dataprofiler/profilers/numerical_column_stats.py:1349]:
```suggestion
        percentiles: np.ndarray = np.linspace(0, 100, len(self.quantiles) + 2)[1:-1]
```

dataprofiler/profilers/numerical_column_stats.py
tonywu315: The function is used in places where the type is neither `np.floating` or `np.integer`, and `val` is returned in this case. For example, line 305 has `np_type_to_type(self.mode)` and `self.mode` is `List[float]`.

taylorfturner: cool cool ... let's hold off on this then for now `Any` will work

JGSweets: oof, that's weird.

JGSweets (approved):

ksneab7 (approved):

PR #657, opened 2022-09-20
Added static typing to data_readers/avro_data.py

Sanketh7:
#610


dataprofiler/data_readers/avro_data.py
```diff
@@ -11,9 +13,9 @@
 class AVROData(JSONData, BaseData):
     """AVROData class to save and load spreadsheet data."""
 
-    data_type = "avro"
+    data_type: Optional[str] = "avro"
```
taylorfturner, line 16: wouldn't this just be `str` not `Optional[str]` ... not sure we would want to allow for `None` https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

dataprofiler/data_readers/avro_data.py
```diff
@@ -74,18 +81,20 @@ def is_match(cls, file_path, options=None):
 
         # get current position of stream
         if data_utils.is_stream_buffer(file_path):
+            assert not isinstance(file_path, str)
```
taylorfturner, line 84: is this required by `mypy`?

dataprofiler/data_readers/avro_data.py
```diff
@@ -74,18 +81,20 @@ def is_match(cls, file_path, options=None):
...
             starting_location = file_path.tell()
 
         is_valid_avro = fastavro.is_avro(file_path)
 
         # return to original position in stream
         if data_utils.is_stream_buffer(file_path):
+            assert not isinstance(file_path, str)
```
taylorfturner, line 91: is this required by `mypy`?

dataprofiler/data_readers/avro_data.py
JGSweets: @taylorfturner I wonder if it has to do with the Abstract being `None` until defined. But agreed, if it can be str to be str

Sanketh7: When I annotate with `str`, mypy complains because it's set to `None` in `BaseData`.

dataprofiler/data_readers/avro_data.py
Sanketh7: It's needed for mypy but I now changed it into an extra condition in the if statement instead of an assertion.

dataprofiler/data_readers/avro_data.py
```diff
@@ -125,13 +132,13 @@ def _get_nested_keys_from_dicts(cls, dicts):
         :type dicts: list(dict)
         :return: a dictionary containing nested keys
         """
-        nested_keys = {}
+        nested_keys: Dict
```
JGSweets, line 135: I think this still needs to be defined as `= {}`

JGSweets (approved):

micdavis (approved):

PR #660, opened 2022-09-20
Static Typing profiler

tonywu315:
#608

### DIFF
--- a/dataprofiler/profilers/unstructured_text_profile.py
+++ b/dataprofiler/profilers/unstructured_text_profile.py
@@ -633,7 +633,7 @@ class TextProfiler(object):
 
     def report(self, remove_disabled_flag: bool = False) -> Dict:
         """Report profile attribute of class; potentially pop val from self.profile."""
-        calcs_dict_keys = self.__calculations.keys()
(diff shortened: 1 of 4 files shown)


dataprofiler/profilers/unstructured_text_profile.py
```diff
@@ -623,9 +631,9 @@ def diff(self, other_profile, options=None):
 
         return diff
 
-    def report(self, remove_disabled_flag=False):
+    def report(self, remove_disabled_flag: bool = False) -> Dict:
         """Report profile attribute of class; potentially pop val from self.profile."""
-        calcs_dict_keys = self._TextProfiler__calculations.keys()
```
tonywu315, line 628: I'm not sure if this was a typo because I couldn't find `self._TextProfiler__calculations` anywhere else.

JGSweets: Not a typo, if you go into a base class you can still call the previous classes specific variable that way in python when it has the dunder.

micdavis (approved):

ksneab7 (approved):

PR #661, opened 2022-09-21
Static Typing profilers/column profile

tonywu315:
#608

### DIFF
--- a/dataprofiler/profilers/order_column_profile.py
+++ b/dataprofiler/profilers/order_column_profile.py
@@ -1,4 +1,9 @@
 """Index profile analysis for individual col within structured profiling."""
+from __future__ import annotations
+
+from typing import Dict, Optional, Tuple, cast
+
+from pandas import DataFrame, Series
 
 from . import BaseColumnProfiler, utils
 from .profiler_options import OrderOptions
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/order_column_profile.py
```diff
@@ -1,4 +1,9 @@
 """Index profile analysis for individual col within structured profiling."""
+from __future__ import annotations
+
+from typing import Dict, Optional, Tuple, cast
+
+from pandas import DataFrame, Series
```
taylorfturner, line 7: this formatting looks a bit off, but maybe it is right. Did the `pre-commit` run when you did `git commit -m`

taylorfturner (commented):
just one comment around imports

dataprofiler/profilers/order_column_profile.py
tonywu315: `isort` did run

taylorfturner: perfect

taylorfturner (approved):

micdavis (approved):

dataprofiler/profilers/order_column_profile.py
```diff
@@ -209,9 +218,9 @@ def _merge_order(
         ) or order == "random":
             piecewise = False
 
-        return order, first_value, last_value, piecewise
+        return order, cast(int, first_value), cast(int, last_value), piecewise
```
JGSweets, line 221: Does this need to be casted or is optional an acceptable output arg?

tonywu315: I think `first_value` and `last_value` are always not None because one of the conditionals on lines 205 to 213 will run.

PR #662, opened 2022-09-23
added static typing to data_utils.py

Sanketh7:
#610


dataprofiler/data_readers/data_utils.py
```diff
@@ -187,10 +212,14 @@ def read_json_df(data_generator, selected_columns=None, read_in_string=False):
         k += 1
     if not lines and k:
         raise ValueError("No JSON data could be read from these data.")
-    return json_to_dataframe(lines, selected_columns, read_in_string)
+    return json_to_dataframe(cast(List[Dict], lines), selected_columns, read_in_string)
```
JGSweets, line 215: instead of casting, ca we not define it on 191?

dataprofiler/data_readers/data_utils.py
```diff
@@ -239,17 +268,17 @@ def read_json(data_generator, selected_columns=None, read_in_string=False):
         k += 1
     if not lines and k:
         raise ValueError("No JSON data could be read from these data.")
-    return lines
+    return cast(List[Dict], lines)
```
JGSweets, line 271: as above, can we define it ahead of time instead of casting?

JGSweets:
@Sanketh7 looks like the branch needs updating as well.

dataprofiler/data_readers/data_utils.py
```diff
@@ -31,7 +43,9 @@ def data_generator(data_list):
         yield item
 
 
-def generator_on_file(file_object):
+def generator_on_file(
+    file_object: Union[StringIO, BytesIO]
+) -> Generator[Union[str, bytes], None, None]:
```
taylorfturner, line 48: is this after `black` formatting pre-commit?

dataprofiler/data_readers/data_utils.py
```diff
@@ -99,7 +115,11 @@ def unicode_to_str(data, ignore_dicts=False):
     return data
 
 
-def json_to_dataframe(json_lines, selected_columns=None, read_in_string=False):
+def json_to_dataframe(
+    json_lines: List[Dict],
+    selected_columns: Optional[List[str]] = None,
+    read_in_string: bool = False,
+) -> Tuple[pd.DataFrame, pd.Series]:
```
taylorfturner, line 122: is this after black formatting pre-commit?

dataprofiler/data_readers/data_utils.py
```diff
@@ -137,7 +157,11 @@ def json_to_dataframe(json_lines, selected_columns=None, read_in_string=False):
     return df, original_df_dtypes
 
 
-def read_json_df(data_generator, selected_columns=None, read_in_string=False):
+def read_json_df(
+    data_generator: Generator,
+    selected_columns: Optional[List[str]] = None,
+    read_in_string: bool = False,
+) -> Tuple[Iterator[pd.DataFrame], pd.Series]:
```
taylorfturner, line 164: is this after black formatting pre-commit?

dataprofiler/data_readers/data_utils.py
```diff
@@ -299,13 +327,18 @@ def read_csv_df(
...
 
     return data
 
 
-def read_parquet_df(file_path, selected_columns=None, read_in_string=False):
+def read_parquet_df(
+    file_path: str,
+    selected_columns: Optional[List[str]] = None,
+    read_in_string: bool = False,
+) -> Tuple[pd.DataFrame, pd.Series]:
```
taylorfturner, line 341: is this after black formatting pre-commit?

taylorfturner (commented):
comments mainly around formatting and +1 to @JGSweets comment around not casting

JGSweets (dismissed):

dataprofiler/data_readers/data_utils.py
Sanketh7: The issue is that `unicode_to_str` returns `Union[str, List, Dict]`. So `lines` becomes `List[Union[str, List, Dict]]`. However, it seems like `unicode_to_str` in this case should only return `Dict` so I've added a conditional check that should always pass before appending to `list` to tell mypy it's always a `Dict`.

dataprofiler/data_readers/data_utils.py
Sanketh7: Same as above.

dataprofiler/data_readers/data_utils.py
Sanketh7: yes

dataprofiler/data_readers/data_utils.py
Sanketh7: yes

dataprofiler/data_readers/data_utils.py
Sanketh7: yes

dataprofiler/data_readers/data_utils.py
Sanketh7: yes

dataprofiler/data_readers/data_utils.py
JGSweets: i see, in the future I wonder if we can address the fact that we use the object_hook and then wrap the output as well. seems like redundancy that might fix the issue to begin with:
```
obj = unicode_to_str(
    json.loads(
        raw_line, object_hook=unicode_to_str, object_pairs_hook=OrderedDict
    ),
    ignore_dicts=True,
)
```

JGSweets (approved):

dataprofiler/data_readers/data_utils.py
```diff
@@ -180,7 +204,8 @@ def read_json_df(data_generator, selected_columns=None, read_in_string=False):
                 ),
                 ignore_dicts=True,
             )
-            lines.append(obj)
+            if isinstance(obj, dict):  # shold always pass but needed for mypy
+                lines.append(obj)
```
taylorfturner, line 208: based on `unicode_to_str` could be `str`, `list`, or `dict`?

dataprofiler/data_readers/data_utils.py
```diff
@@ -232,7 +261,8 @@ def read_json(data_generator, selected_columns=None, read_in_string=False):
                 ),
                 ignore_dicts=True,
             )
-            lines.append(obj)
+            if isinstance(obj, dict):  # should always pass but needed for mypy
+                lines.append(obj)
```
taylorfturner, line 265: based on unicode_to_str could be str, list, or dict?

JGSweets: that's what it looks like in the func. The input can also be a list, dict, or str. 

basically converts ensures any str inside of the objs are str and not unicode

taylorfturner: @Sanketh7 why just a `dict` here?

JGSweets: Good catch and Q @taylorfturner

dataprofiler/data_readers/data_utils.py
Sanketh7: json.loads with our object_hook set to dict makes it so unicode_to_str gets a dict (and returns a dict). I don't think mypy can detect that unless we explicitly tell it that it will be always be a dict.

JGSweets: Hmm, my concern is a silent error could happen if somehow data returned that wasn't a dict. We'd lose our on that information unless our intent was to purposefully skip anything that wasn't a dict. This might be a case where casting is better and we can handle a real fix later.

taylorfturner (commented):
one comment around the dict isinstance

taylorfturner (dismissed):

dataprofiler/data_readers/data_utils.py
```diff
@@ -299,13 +327,18 @@ def read_csv_df(
 
     # if the buffer was wrapped, detach it before returning
     if is_buf_wrapped:
+        assert isinstance(file_path, TextIOWrapper)
```
micdavis, line 330: Should the boolean `is_buf_wrapped` be removed and we can just check `if isinstance(file_path, TextIOWrapper)`?

dataprofiler/data_readers/data_utils.py
```diff
@@ -602,7 +648,7 @@ def load_as_str_from_file(
 
             # Return either the last index of sample_lines OR
             # the index of the newline char that matches remaining_lines
-            search_query_value = "\n"
+            search_query_value: Union[str, bytes] = "\n"
             if isinstance(sample_lines, bytes):
                 search_query_value = b"\n"
```
micdavis, line 653: does it make more sense to cast to string immediately? On line 653 could we do `search_query_value = case(str, b"\n")` then we don't need line 660 and search_query_value can just be statically typed as `str`

dataprofiler/data_readers/data_utils.py
Sanketh7: That should work and would simplify the static typing.

dataprofiler/data_readers/data_utils.py
```diff
@@ -180,7 +204,7 @@ def read_json_df(data_generator, selected_columns=None, read_in_string=False):
                 ),
                 ignore_dicts=True,
             )
-            lines.append(obj)
+            lines.append(cast(Dict, obj))
```
JGSweets, line 207: did we try overloading `unicode_to_str`? would that fix this so we don't have to cast?

dataprofiler/data_readers/data_utils.py
```diff
@@ -286,26 +314,27 @@ def read_csv_df(
...
         # in 3.6 this avoids read_csv wrapping the stream and closing too early
         file_path = TextIOWrapper(file_path, encoding=encoding)
-        is_buf_wrapped = True
 
     fo = pd.read_csv(file_path, **args)
     data = fo.read()
 
     # if the buffer was wrapped, detach it before returning
-    if is_buf_wrapped:
+    if isinstance(file_path, TextIOWrapper):
```
JGSweets, line 326: does this mean we can remove entire if statement though?
I'm concerned the statement means not just the is_wrapped statement.

Can we leave the code as is till we understand and add an issue to GH to address / remove?

JGSweets (changes requested):
concerned about code alteration

dataprofiler/data_readers/data_utils.py
JGSweets: Or rather it looks like the code still achieves the same thing, but if we do this, we should at the very least not remove the comment from line 289

dataprofiler/data_readers/data_utils.py
JGSweets: Instead of casting, could nth loc just accept bytes?

dataprofiler/data_readers/data_utils.py
Sanketh7: I did some experimentation and added this:
```py
@overload
def unicode_to_str(data: Dict, ignore_dicts: bool) -> Dict: ...
@overload
def unicode_to_str(data: str, ignore_dicts: bool) -> str: ...
@overload
def unicode_to_str(data: List, ignore_dicts: bool) -> List: ...
def unicode_to_str(
    data: Union[str, List, Dict], ignore_dicts: bool = False
) -> Union[str, List, Dict]:
# function body here
```
The issue I'm getting now is when we call this
```py
json.loads(raw_line, object_hook=unicode_to_str, object_pairs_hook=OrderedDict)
```
I get 
```
Argument "object_hook" to "loads" has incompatible type overloaded function; expected "Optional[Callable[[Dict[Any, Any]], Any]]"
```
from mypy. My guess is that we can solve this by casting before we pass in the function to `object_hook` but that would somewhat defeat the purpose of overloading in the first place.

dataprofiler/data_readers/data_utils.py
Sanketh7: I can add that comment right before the if statement on line 326.

Sanketh7: Actually for now, I can leave the original code and just cast to TextIOWrapper to satisfy mypy.

dataprofiler/data_readers/data_utils.py
JGSweets: @Sanketh7 I think you don't need to specify:
```
def unicode_to_str(
    data: Union[str, List, Dict], ignore_dicts: bool = False
) -> Union[str, List, Dict]:
```
b/c of the three other overloads. I assume the defaults would still need to be assigned though

dataprofiler/data_readers/data_utils.py
```diff
@@ -243,13 +271,13 @@ def read_json(data_generator, selected_columns=None, read_in_string=False):
...
 
 def read_csv_df(
-    file_path,
-    delimiter,
-    header,
-    selected_columns=[],
-    read_in_string=False,
-    encoding="utf-8",
-):
+    file_path: Union[str, BytesIO, TextIOWrapper],
```
JGSweets, line 274: similar to below, if we override allowing each diff input, does that remove the need to cast?

Sanketh7: I don't think so because as far I can tell, overloading doesn't let us have multiple implementations. It just lets us specify relationships between input types and return types.

dataprofiler/data_readers/data_utils.py
Sanketh7: Even with that removal, I'm still getting the same issue. I think it has to do with the function taking on 3 possible types while `object_hook` is only compatible with one of those types.

dataprofiler/data_readers/data_utils.py
taylorfturner: in certain conditions, would still need to cast I think because the code doesn't know that under this condition the type should be this. So that's where `cast()` comes in to play under specific `if` conditions... at least that's how I'm reading it

taylorfturner (commented):
just one comment... can't approve at this point since I clicked `update branch` but I think we are close on this once we nail down @JGSweets and @micdavis comment

dataprofiler/data_readers/data_utils.py
```diff
@@ -232,7 +260,7 @@ def read_json(data_generator, selected_columns=None, read_in_string=False):
                 ),
                 ignore_dicts=True,
             )
-            lines.append(obj)
+            lines.append(cast(Dict, obj))  # should always be Dict
```
ksneab7, line 263: Is this comment needed? isnt that what the type cast is doing?

taylorfturner: that's fair ... not really needed per se. I'm fine leaving just for clarity in the code

Sanketh7: I added it because `cast` is inherently unsafe but it turns out `obj` will always be a `Dict` here.

ksneab7 (dismissed):

dataprofiler/data_readers/data_utils.py
```diff
@@ -99,7 +115,11 @@ def unicode_to_str(data, ignore_dicts=False):
     return data
 
 
-def json_to_dataframe(json_lines, selected_columns=None, read_in_string=False):
+def json_to_dataframe(
+    json_lines: List[Dict],
```
JGSweets, line 119: looking deeper into this, I believe this might be incorrect.

` json_lines: List[Dict]` -> json lines can be a `List[Union[Dict, List, str]` Need to validate if it can be a number, but it might be able to also be a `float` `int` as well.

JGSweets: This might be causing the issues below.

dataprofiler/data_readers/data_utils.py
JGSweets: @Sanketh7

dataprofiler/data_readers/data_utils.py
```diff
@@ -163,7 +187,7 @@ def read_json_df(data_generator, selected_columns=None, read_in_string=False):
         each call as well as original dtypes of the dataframe columns.
     :rtype: typle(Iterator(pd.DataFrame), pd.Series(dtypes)
     """
-    lines = list()
+    lines: List[Dict] = list()
```
taylorfturner, line 190: not necessarily the case... could be a list of `str` / `list` / `'dict`, @Sanketh7?

dataprofiler/data_readers/data_utils.py
```diff
@@ -69,7 +83,9 @@ def convert_int_to_string(x):
         return str(x)
 
 
-def unicode_to_str(data, ignore_dicts=False):
+def unicode_to_str(
+    data: Union[str, List, Dict], ignore_dicts: bool = False
+) -> Union[str, List, Dict]:
```
JGSweets, line 88: need to fix docstrings below.

dataprofiler/data_readers/data_utils.py
Sanketh7: https://docs.python.org/3/library/json.html#json-to-py-table
Could it be any of these? Or are there constraints on what kind of data we support?

JGSweets: That's a fair question, I actually believe after looking through the code that it could be a list of any of those. We can use an alias to describe this maybe?
`JSON = t.Union[str, int, float, bool, None]` I'm not including the recursive option for now.
https://github.com/python/typing/issues/182

dataprofiler/data_readers/data_utils.py
JGSweets: This relates to the comment above about it being JSON

dataprofiler/data_readers/data_utils.py
JGSweets: Is this bc of the `OrderedDict`?

JGSweets: What's our guarantee that it is a dict?

JGSweets: I've determined this could be a list. however, not a string.

dataprofiler/data_readers/data_utils.py
JGSweets: Okay, after further review, while this can accept files which are lists or dicts

JGSweets: that's bc:
```python
    elif first_item_type == dict:
        df = pd.json_normalize(json_lines)
    else:
        df = pd.DataFrame(json_lines)
```

dataprofiler/data_readers/data_utils.py
JGSweets: The code that could result in this would be:
```python
data = dp.Data(StringIO("[1]\n[2]"))
data.head()
```
Currently, there's an error,  in the reader not supporting this. However, it should.

taylorfturner: should be resolved in PR #691

dataprofiler/data_readers/data_utils.py
```diff
@@ -611,7 +657,8 @@ def load_as_str_from_file(
             while start_loc < len_sample_lines - 1 and total_occurrences < max_lines:
                 loc, occurrence = find_nth_loc(
                     sample_lines[start_loc:],
-                    search_query=search_query_value,
+                    search_query=cast(str, search_query_value),
```
JGSweets, line 660: is this fixed and not needing casting if we fix to accept bytes?

dataprofiler/data_readers/data_utils.py
Sanketh7: We could make `find_nth_loc` accept bytes but then we'd have to deal with the cases where the search string is of type `str` and the search query is of type `bytes` (and vice versa). One way I found to deal with this is to use the type `AnyStr` in `find_nth_loc` which basically creates a generic type that can only be `str` or `bytes`. However, the issue I run into is that the current way the code is set up, we would have to pass in a `Union[str, bytes]` to `find_nth_loc` but you can't pass that in for `AnyStr` (see https://github.com/python/mypy/issues/1533).

dataprofiler/data_readers/data_utils.py
Sanketh7: I was able to resolve it in https://github.com/capitalone/DataProfiler/pull/662/commits/7d5b8fb4f120504a8d84ea0da8a907d66798711b using 
```py
JSONType = Union[str, int, float, bool, None, List['JSONType'], Dict[str, 'JSONType']]
```
Note that this uses recursive types which are new to mypy. Currently you have to use the flag `--enable-recursive-aliases` but according to https://github.com/python/mypy/issues/731#issuecomment-1213482527 it will soon be a standard feature.

dataprofiler/data_readers/data_utils.py
Sanketh7: This issue no longer applies due to the new way of handling json types.

dataprofiler/data_readers/data_utils.py
Sanketh7: No longer a need for casting from my latest commit.

dataprofiler/data_readers/data_utils.py
Sanketh7: Fixed to accept all json types.

dataprofiler/data_readers/data_utils.py
Sanketh7: Done

dataprofiler/data_readers/data_utils.py
Sanketh7: See https://github.com/capitalone/DataProfiler/pull/662#discussion_r998298127

dataprofiler/data_readers/data_utils.py
```diff
@@ -17,8 +29,10 @@
 
 logger = dp_logging.get_child_logger(__name__)
 
+JSONType = Union[str, int, float, bool, None, List["JSONType"], Dict[str, "JSONType"]]
```
taylorfturner, line 32: just to make it a bit cleaner... let's move this to `_typing.py` https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/_typing.py

taylorfturner (dismissed):

dataprofiler/data_readers/data_utils.py
taylorfturner: awesome! Thx

micdavis (dismissed):

dataprofiler/data_readers/data_utils.py
```diff
@@ -90,7 +105,7 @@ def unicode_to_str(data, ignore_dicts=False):
     # if data is a dictionary
     if isinstance(data, dict) and not ignore_dicts:
         return {
-            unicode_to_str(key, ignore_dicts=True): unicode_to_str(
+            cast(str, unicode_to_str(key, ignore_dicts=True)): unicode_to_str(
```
JGSweets, line 108: there's no guarantee this is a string, this could be an int, right?

dataprofiler/data_readers/data_utils.py
JGSweets: Can we choose not to include recursive aliases? I think I mentioned above leaving out the `List['JSONType'], Dict[str, 'JSONType']` for now. Can we just include `Dict` in there instead?

JGSweets: I'm concerned about conditionals being used @taylorfturner

dataprofiler/data_readers/data_utils.py
```diff
@@ -99,7 +114,11 @@ def unicode_to_str(data, ignore_dicts=False):
     return data
 
 
-def json_to_dataframe(json_lines, selected_columns=None, read_in_string=False):
+def json_to_dataframe(
+    json_lines: List[JSONType],
```
JGSweets, line 118: We should actually validate this, @taylorfturner can `df = pd.DataFrame(json_lines)` take in any list of JSON?

dataprofiler/data_readers/data_utils.py
Sanketh7: Changed to be nonrecursive.

dataprofiler/data_readers/data_utils.py
Sanketh7: Removed the cast.

dataprofiler/data_readers/data_utils.py
taylorfturner: It works assuming `List[JSONType]` but should it be `Dict[str, List[JSONType]]`? 

```
from typing import Dict, List, Union
import pandas as pd

JSONType = Union[str, int, float, bool, None, List, Dict]

test_iter_list = []
test_iter_list.append(['test', 'test'])
test_iter_list.append([1,2,3])
test_iter_list.append([1.02, 2.02, 3.03])
test_iter_list.append([True, False])
test_iter_list.append([None, None])
test_iter_list.append([['test', 'test'], ['test', 'test']])
test_iter_list.append([{'test': 'test', 'test': 'test'}])


for test_iter in test_iter_list: 
    pd.DataFrame(test_iter)
```

taylorfturner: CC @JGSweets and @Sanketh7

taylorfturner: @Sanketh7 need to update the docstring for this on L126 `:type json_lines: list(dict)`

taylorfturner: > 
This does work. So I think we are GTG 
`for test_iter in test_iter_list: json_to_dataframe(test_iter)`

dataprofiler/data_readers/data_utils.py
taylorfturner: awesome -- thx, @Sanketh7

taylorfturner: is it removed @Sanketh7 ?  looks like it is still there ...

Sanketh7: Should be removed by https://github.com/capitalone/DataProfiler/pull/662/commits/8de0ceff3ff0503ae4422de6c2b804700e34f484

dataprofiler/data_readers/data_utils.py
taylorfturner: List[list] works
```
In [307]: data = data_generator("""[["test", "test"], ["test", "test"]]""".splitlines())
In [308]: read_json(data)
Out[308]: [[['test', 'test'], ['test', 'test']]]
```

JGSweets (approved):

micdavis (approved):

PR #664, opened 2022-09-26
Add static typing to data labeler and abstract classes

tonywu315:
#609

### DIFF
--- a/dataprofiler/labelers/base_data_labeler.py
+++ b/dataprofiler/labelers/base_data_labeler.py
@@ -1,9 +1,11 @@
 """Contains abstract classes from which labeler classes will inherit."""
+from __future__ import annotations
 
 import json
 import os
 import sys
 import warnings
+from typing import Dict, List, Optional, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -19,9 +21,9 @@ default_labeler_dir = pkg_resources.resource_filename("resources", "labelers")
 class BaseDataLabeler(object):
     """Parent class for data labeler objects."""
 
-    _default_model_loc = None
+    _default_model_loc: str = None  # type: ignore
 
-    def __init__(self, dirpath=None, load_options=None):
+    def __init__(self, dirpath: str = None, load_options: Dict = None) -> None:
         """
         Initialize DataLabeler class.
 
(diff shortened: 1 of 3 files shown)


dataprofiler/labelers/base_data_labeler.py
```diff
@@ -19,9 +21,9 @@
 class BaseDataLabeler(object):
     """Parent class for data labeler objects."""
 
-    _default_model_loc = None
+    _default_model_loc: str = None  # type: ignore
```
JGSweets, line 24: can this not be `Optional` then?

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -552,9 +586,13 @@ def _load_postprocessor(self, processor_class, dirpath):
                     str(processor_class)
                 )
             )
-        self.set_postprocessor(processor_class.load_from_disk(dirpath))
+        self.set_postprocessor(
+            cast(data_processing.BaseDataProcessor, processor_class).load_from_disk(
```
JGSweets, line 590: Is the problem `processor_class` could still be a string at this point according to mypy despite the `if`

tonywu315: yes, I'm not sure why mypy can't detect it

dataprofiler/labelers/base_data_labeler.py
tonywu315: it shouldn't be for any child class

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -267,23 +273,29 @@ def set_labels(self, labels):
...
-        batch_size=32,
-        predict_options=None,
-        error_on_mismatch=False,
-        verbose=1,
-    ):
+        data: Union[pd.DataFrame, pd.Series, np.ndarray],
+        batch_size: int = 32,
+        predict_options: Dict[str, bool] = None,
+        error_on_mismatch: bool = False,
+        verbose: bool = True,
```
tonywu315, line 280: I changed the default of `verbose` from `1` to `True` because it is only used in `self._model.predict()`, which would takes in a bool.

taylorfturner: nice

taylorfturner (approved):

dataprofiler/labelers/base_data_labeler.py
JGSweets: good point

JGSweets (approved):

PR #666, opened 2022-09-27
Added static typing to data_readers/base_data.py and data_readers/json_data.py

Sanketh7:
#610


taylorfturner:
@Sanketh7 conflict on `avro_data.py` FYI

dataprofiler/data_readers/avro_data.py
```diff
@@ -73,19 +91,19 @@ def is_match(cls, file_path, options=None):
             options = dict()
 
         # get current position of stream
-        if data_utils.is_stream_buffer(file_path):
+        if data_utils.is_stream_buffer(file_path) and not isinstance(file_path, str):
```
taylorfturner, line 94: why this add?

dataprofiler/data_readers/avro_data.py
```diff
@@ -73,19 +91,19 @@ def is_match(cls, file_path, options=None):
...
             starting_location = file_path.tell()
 
         is_valid_avro = fastavro.is_avro(file_path)
 
         # return to original position in stream
-        if data_utils.is_stream_buffer(file_path):
+        if data_utils.is_stream_buffer(file_path) and not isinstance(file_path, str):
```
taylorfturner, line 100: why this add?

dataprofiler/data_readers/json_data.py
```diff
@@ -268,36 +276,36 @@ def _load_data_from_file(self, input_file_path):
...
         :return: dataframe in json format
         """
-        data = self._get_data_as_df(data)
-        data = data.to_json(orient="records")
-        char_per_line = min(len(data), self.SAMPLES_PER_LINE_DEFAULT)
-        return list(map("".join, zip(*[iter(data)] * char_per_line)))
+        data_df: pd.DataFrame = self._get_data_as_df(data)
+        data_json = data_df.to_json(orient="records")
+        char_per_line = min(len(data_json), self.SAMPLES_PER_LINE_DEFAULT)
+        return list(map("".join, zip(*[iter(data_json)] * char_per_line)))
```
taylorfturner, line 306: we are no longer overwriting here... was mypy complaining about something with how it was prior?

dataprofiler/data_readers/json_data.py
```diff
@@ -419,4 +434,4 @@ def reload(self, input_file_path=None, data=None, options=None):
         """
         self._selected_keys = None
         super(JSONData, self).reload(input_file_path, data, options)
-        self.__init__(self.input_file_path, data, options)
+        JSONData.__init__(self, self.input_file_path, data, options)
```
taylorfturner, line 437: was mypy complaining here?

dataprofiler/data_readers/json_data.py
```diff
@@ -268,36 +276,36 @@ def _load_data_from_file(self, input_file_path):
...
         :return: dataframe in record format
         """
-        data = self._get_data_as_df(data)
-        data = data.to_dict(orient="records", into=OrderedDict)
-        for i, sample in enumerate(data):
-            data[i] = json.dumps(
+        data_df: pd.DataFrame = self._get_data_as_df(data)
+        data_records: List = data_df.to_dict(orient="records", into=OrderedDict)
+        for i, sample in enumerate(data_records):
+            data_records[i] = json.dumps(
```
taylorfturner, line 290: no longer overwriting variable names... was mypy complaining about this and that is the reason for the change?

taylorfturner (commented):
comments around code changes and reasons for those changes

dataprofiler/data_readers/avro_data.py
Sanketh7: is_stream_buffer checks if file_path is a StringIO or BytesIO but mypy still complained that file_path could still be a string and thus .seek() doesn't exist for strings.

dataprofiler/data_readers/json_data.py
Sanketh7: Yes, mypy complained about calling self.__init__ but I think this change should do the same thing.

dataprofiler/data_readers/json_data.py
Sanketh7: The issue is that if we define `data` of type `A`, we can't reassign it to type `B`. If we try to change the type annotation later, mypy complains that `data` already exists. I guess a workaround would be to make `data` a union of the types it can be but I was thinking that could be messy.

dataprofiler/data_readers/json_data.py
JGSweets: Could that be problematic for anything that inherits it? For instance if avro didn't have a reload, it would only call JSONData reload which I think could cause a problem.

dataprofiler/data_readers/avro_data.py
JGSweets: Does that mean we have an issue somewhere to fix then?

dataprofiler/data_readers/json_data.py
Sanketh7: The original code kept reassigning the types of data stored in `data` but mypy doesn't let you do that (each variable can only have 1 type in its lifetime). We could make `data` be typed as a union of all the types it takes on but I was thinking that would be confusing.

JGSweets: The concern would be now we have the data 3x in memory at one time. We could use a defined type which is the union. This might be a decent case for casting or allowing data to be Any

dataprofiler/data_readers/avro_data.py
taylorfturner: yeah I would look further up the code to see where `file_path` could be a string and try to solve it that way rather than adding the `and not isinstance(file_path, str)`

dataprofiler/data_readers/json_data.py
taylorfturner: +1

dataprofiler/data_readers/json_data.py
taylorfturner: >  data 3x in memory at one time
Yeah, this is what the code was trying to avoid as it was written last time. 

Couple options in order of how robust they are:
- `Union[]` is an option
- casting
- `Any`

dataprofiler/data_readers/avro_data.py
Sanketh7: For the operation within the if statement, `file_path` can't be a string. `is_stream_buffer` checks if it's a file buffer but mypy doesn't seem to be able to detect it because it would complain on the next line that `.tell()` doesn't exist for strings. So far the only solution I could find was explicitly telling mypy it wouldn't be a string.

dataprofiler/data_readers/json_data.py
taylorfturner: +1

dataprofiler/data_readers/avro_data.py
Sanketh7: `file_path` is one of the function's parameters and some doctstrings (like the constructor's) says it should be a string. This doesn't seem to be a problem because `fastavro.is_avro(file_path)` accepts `Union[str, IO]`. So the code doesn't seem to be broken but rather mypy isn't playing well with the current way of detecting if `file_path` is a file buffer.

JGSweets: We really shouldn't be altering runtime code to make static typing work.

dataprofiler/data_readers/avro_data.py
taylorfturner: outdated... no longer moot to this PR. already in `main`

dataprofiler/data_readers/avro_data.py
taylorfturner: outdated diff... this is already in `main`

taylorfturner: yep somehow got merged in through this PR https://github.com/capitalone/DataProfiler/pull/657/files#diff-da94771d0aad55dc1a9833ead30c7108a37e9eaef5344bae68657c858eab1062R94

JGSweets: Can we remove in this PR?

dataprofiler/data_readers/avro_data.py
JGSweets: Can re remove in this PR?

JGSweets: There should be a way to resolve this correctly. just need to find the right answer.

dataprofiler/data_readers/avro_data.py
taylorfturner: I think we should fix / remove in this PR -- whatever the solution may be

dataprofiler/data_readers/avro_data.py
taylorfturner: agree

dataprofiler/data_readers/avro_data.py
```diff
@@ -91,13 +91,19 @@ def is_match(
             options = dict()
 
         # get current position of stream
-        if data_utils.is_stream_buffer(file_path) and not isinstance(file_path, str):
+        if data_utils.is_stream_buffer(file_path):
```
Sanketh7, line 94: Because mypy doesn't detect type narrowing inside `data_utils.is_stream_buffer`, I figured casting like this would be a good compromise without having to add extra runtime checks.

Sanketh7: Oops I didn't leave the comment on the right line -- the relevant line is line 95.

dataprofiler/data_readers/structured_mixins.py
```diff
@@ -25,12 +25,14 @@ class SpreadSheetDataMixin(object):
...
-    def __init__(self, input_file_path: str, data: Any, options: Dict) -> None:
+    def __init__(
+        self, input_file_path: Optional[str], data: Any, options: Dict
+    ) -> None:
         """Initialize spreadsheet mixin object."""
         self._data_formats: Dict = dict()
         self._data_formats["dataframe"] = self._get_data_as_df
         self._original_df_dtypes: Optional[pd.Series]
-        self.input_file_path: str = input_file_path
+        self.input_file_path: Optional[str] = input_file_path
```
taylorfturner, line 35: why optional str on `input_file_path`?

JGSweets: as above

dataprofiler/data_readers/structured_mixins.py
```diff
@@ -25,12 +25,14 @@ class SpreadSheetDataMixin(object):
     :return: None
     """
 
-    def __init__(self, input_file_path: str, data: Any, options: Dict) -> None:
+    def __init__(
+        self, input_file_path: Optional[str], data: Any, options: Dict
```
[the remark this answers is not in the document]
taylorfturner: got it -- makes sense

dataprofiler/data_readers/structured_mixins.py
taylorfturner: got it -- makes sense

JGSweets (approved):

PR #668, opened 2022-09-29
Add static typing to labeler utils

tonywu315:
#609

### DIFF
--- a/dataprofiler/labelers/classification_report_utils.py
+++ b/dataprofiler/labelers/classification_report_utils.py
@@ -1,11 +1,12 @@
 """Contains functions for classification."""
 import warnings
+from typing import Dict, List, Optional, Set, Tuple, Union, cast
 
 import numpy as np
 import sklearn.metrics._classification
 
 
-def convert_confusion_matrix_to_MCM(conf_matrix):
+def convert_confusion_matrix_to_MCM(conf_matrix: Union[List, np.ndarray]):
     """
     Convert a confusion matrix into the MCM format.
 
@@ -52,14 +53,14 @@ def convert_confusion_matrix_to_MCM(conf_matrix):
 
 
 def precision_recall_fscore_support(
-    MCM,
-    beta=1.0,
-    labels=None,
-    pos_label=1,
-    average=None,
-    warn_for=("precision", "recall", "f-score"),
-    sample_weight=None,
-):
+    MCM: np.ndarray,
+    beta: float = 1.0,
+    labels: np.ndarray = None,
+    pos_label: Union[str, int] = 1,
+    average: str = None,
+    warn_for: Union[Tuple[str, ...], Set[str]] = ("precision", "recall", "f-score"),
+    sample_weight: np.ndarray = None,
+) -> Tuple[np.ndarray, np.ndarray, np.ndarray, Optional[np.ndarray]]:
     """
     Perform same functionality as recision_recall_fscore_support function.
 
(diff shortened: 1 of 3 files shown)


dataprofiler/labelers/labeler_utils.py
```diff
@@ -184,24 +188,26 @@ def evaluate_accuracy(
                 f1_report["macro avg"][metric] = np.nan
             else:
                 f1_report["macro avg"][metric] *= (
-                    float(len(label_names)) / num_labels_with_positive_support
+                    float(len(label_names or [])) / num_labels_with_positive_support
```
tonywu315, line 191: `label_names` may be `None` which would cause an error

dataprofiler/labelers/labeler_utils.py
```diff
@@ -184,24 +188,26 @@ def evaluate_accuracy(
...
     if "macro avg" in f1_report:
-        f1 = f1_report["macro avg"]["f1-score"]  # this is micro for the report
+        f1: float = f1_report["macro avg"]["f1-score"]  # this is micro for the report
     else:
         # this is the only remaining option for the report
         f1 = f1_report["accuracy"]
 
     if verbose:
-        f1_report_str = f1_report_dict_to_str(f1_report, label_names)
+        f1_report_str = f1_report_dict_to_str(f1_report, label_names or [""])
```
tonywu315, line 201: same here

dataprofiler/labelers/classification_report_utils.py
```diff
@@ -342,15 +343,15 @@ def classification_report(
     p, r, f1, s = precision_recall_fscore_support(
         MCM, labels=labels, average=None, sample_weight=sample_weight
     )
-    rows = zip(target_names, p, r, f1, s)
+    rows = zip(target_names, p, r, f1, cast(np.ndarray, s))
 
     if y_type.startswith("multilabel"):
-        average_options = ("micro", "macro", "weighted", "samples")
+        average_options: Tuple[str, ...] = ("micro", "macro", "weighted", "samples")
```
taylorfturner, line 349: why the `...` do we not want to enforce a len of four? so something like `Tuple[str, str, str, str]`?

dataprofiler/labelers/labeler_utils.py
taylorfturner: thoughts on putting a condition here instead of doing `or []`... add an `if not label_names: label_names = []` then do `float(len([label_names]))`

dataprofiler/labelers/labeler_utils.py
taylorfturner: what about same idea from comment earlier... add a condition to check `label_names` value...

dataprofiler/labelers/classification_report_utils.py
tonywu315: The next line is a tuple with length 3

taylorfturner (approved):

dataprofiler/labelers/classification_report_utils.py
taylorfturner: https://docs.python.org/3/library/typing.html#typing.Tuple

taylorfturner: ```To specify a variable-length tuple of homogeneous type, use literal ellipsis, e.g. Tuple[int, ...]. A plain [Tuple](https://docs.python.org/3/library/typing.html#typing.Tuple) is equivalent to Tuple[Any, ...], and in turn to [tuple](https://docs.python.org/3/library/stdtypes.html#tuple).```

micdavis (approved):

PR #669, opened 2022-09-29
Allow diff to set format options for prepare report

JGSweets:
Now can use:
```python
profile1.diff(profile2, options=dict(output_format='pretty', omit_keys=[....])
```

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1177,7 +1177,7 @@ class UnstructuredProfiler(BaseProfiler):
         return merged_profile
 
     def diff(  # type: ignore[override]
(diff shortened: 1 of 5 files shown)


taylorfturner (approved):

dataprofiler/profilers/profile_builder.py
```diff
@@ -1208,7 +1211,11 @@ def diff(  # type: ignore[override]
         report["data_stats"] = self._profile.diff(
             other_profile._profile, options=options
         )
-        return _prepare_report(report)
+        return _prepare_report(
```
micdavis, line 1214: just curious, why do we return options here if they are passed in? I wouldn't have thought to do that.

micdavis (commented):
just that one comment

micdavis (approved):

PR #672, opened 2022-09-30
Add static typing to labeler models

tonywu315:
#609

### DIFF
--- a/dataprofiler/labelers/base_model.py
+++ b/dataprofiler/labelers/base_model.py
@@ -5,10 +5,11 @@ import abc
 import copy
 import inspect
 import warnings
-from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
 import pandas as pd
+import tensorflow as tf
 
 
 class AutoSubRegistrationMeta(abc.ABCMeta):
@@ -48,7 +49,7 @@ class BaseModel(object, metaclass=abc.ABCMeta):
         :return: None
         """
         # initialize class
-        self._model = None
+        self._model: tf.keras.Model = None
         self._validate_parameters(parameters)
         self._parameters: Dict = parameters
         self._label_mapping: Dict[str, int] = None  # type: ignore
@@ -314,7 +315,11 @@ class BaseModel(object, metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def predict(
-        self, data: Iterator, batch_size: int, show_confidences: bool, verbose: bool
+        self,
+        data: Union[pd.DataFrame, pd.Series, np.ndarray],
+        batch_size: int,
+        show_confidences: bool,
+        verbose: bool,
     ) -> Dict:
         """
         Predict the data with the current model.
@@ -366,10 +371,10 @@ class BaseTrainableModel(BaseModel, metaclass=abc.ABCMeta):
         train_data: Union[pd.DataFrame, pd.Series, np.ndarray],
         val_data: Union[pd.DataFrame, pd.Series, np.ndarray],
         batch_size: int = 32,
-        epochs: int = 1,
         label_mapping: Dict[str, int] = None,
         reset_weights: bool = False,
-    ) -> None:
+        verbose: bool = True,
+    ) -> Tuple[Dict, Optional[float], Dict]:
         """
         Train the current model with the training data and validation data.
 
--- a/dataprofiler/labelers/char_load_tf_model.py
+++ b/dataprofiler/labelers/char_load_tf_model.py
@@ -30,7 +30,7 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
     requires_zero_mapping = False
 
     def __init__(
-        self, model_path: str, label_mapping: Dict, parameters: Dict = None
+        self, model_path: str, label_mapping: Dict[str, int], parameters: Dict = None
     ) -> None:
         """
         Initialize Loadable TF Model.
--- a/dataprofiler/labelers/char_load_tf_model.py
+++ b/dataprofiler/labelers/char_load_tf_model.py
@@ -1,12 +1,16 @@
 """Contains class for training data labeler model."""
+from __future__ import annotations
+
 import copy
 import json
 import os
 import sys
 import time
 from collections import defaultdict
+from typing import Dict, Iterable, List, Optional, Tuple, Union, cast
 
 import numpy as np
+import pandas as pd
 import tensorflow as tf
 
 from .. import dp_logging
@@ -25,7 +29,9 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
     # boolean if the label mapping requires the mapping for index 0 reserved
     requires_zero_mapping = False
 
-    def __init__(self, model_path, label_mapping=None, parameters=None):
+    def __init__(
+        self, model_path: str, label_mapping: Dict, parameters: Dict = None
+    ) -> None:
         """
         Initialize Loadable TF Model.
 
@@ -54,7 +60,7 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
 
         BaseModel.__init__(self, label_mapping, parameters)
 
-    def __eq__(self, other):
+    def __eq__(self, other: BaseModel) -> bool:  # type: ignore
         """
         Check if two models are equal with one another.
 
@@ -74,7 +80,7 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
             return False
         return True
 
-    def _validate_parameters(self, parameters):
+    def _validate_parameters(self, parameters: Dict) -> None:
         """
         Validate the parameters sent in.
 
@@ -110,7 +116,9 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
         if errors:
             raise ValueError("\n".join(errors))
 
-    def set_label_mapping(self, label_mapping):
+    def set_label_mapping(
+        self, label_mapping: Union[List[str], Dict[str, int]]
+    ) -> None:
         """
         Set the labels for the model.
 
@@ -139,7 +147,7 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
             )
         super().set_label_mapping(label_mapping)
 
-    def _need_to_reconstruct_model(self):
+    def _need_to_reconstruct_model(self) -> bool:
         """
         Determine whether or not the model needs to be reconstructed.
 
@@ -153,7 +161,7 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
             or default_ind != self._model_default_ind
         )
 
-    def save_to_disk(self, dirpath):
+    def save_to_disk(self, dirpath: str) -> None:
         """
         Save whole model to disk with weights.
 
@@ -177,13 +185,14 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
         self._model.save(dirpath)
 
     @classmethod
-    def load_from_disk(cls, dirpath):
+    def load_from_disk(cls, dirpath: str) -> CharLoadTFModel:
         """
         Load whole model from disk with weights.
 
         :param dirpath: directory path where you want to load the model from
         :type dirpath: str
-        :return: None
+        :return: loaded CharLoadTFModel
+        :rtype: CharLoadTFModel
         """
         # load parameters
         model_param_dirpath = os.path.join(dirpath, "model_parameters.json")
@@ -215,7 +224,7 @@ class CharLoadTFModel(BaseTrainableModel, metaclass=AutoSubRegistrationMeta):
         ]
         return loaded_model
 
-    def _construct_model(self):
+    def _construct_model(self) -> None:
         """
         Model constructor for the data labeler.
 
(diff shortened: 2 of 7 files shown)


dataprofiler/labelers/base_model.py
```diff
@@ -48,7 +49,7 @@ def __init__(self, label_mapping: Union[List, Dict], parameters: Dict) -> None:
         :return: None
         """
         # initialize class
-        self._model = None
+        self._model: tf.keras.Model = None
```
taylorfturner, line 52: can you do `annotations` for this?

dataprofiler/labelers/base_model.py
```diff
@@ -366,10 +371,10 @@ def fit(
         train_data: Union[pd.DataFrame, pd.Series, np.ndarray],
         val_data: Union[pd.DataFrame, pd.Series, np.ndarray],
         batch_size: int = 32,
-        epochs: int = 1,
         label_mapping: Dict[str, int] = None,
         reset_weights: bool = False,
-    ) -> None:
+        verbose: bool = True,
+    ) -> Tuple[Dict, Optional[float], Dict]:
```
taylorfturner, line 377: why is `verbose` being added and `epochs` being removed?

dataprofiler/labelers/base_model.py
```diff
@@ -5,10 +5,11 @@
 import copy
 import inspect
 import warnings
-from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Tuple, Type, Union
 
 import numpy as np
 import pandas as pd
+import tensorflow as tf
```
taylorfturner, line 12: can you do annotations for this so we don't have to import?

dataprofiler/labelers/regex_model.py
```diff
@@ -1,11 +1,15 @@
...
+
 import copy
 import json
 import os
 import re
 import sys
+from typing import Dict, Union
 
 import numpy as np
+import pandas as pd
```
taylorfturner, line 12: can we not do an `annotations` here too?

taylorfturner (commented):
generally good after inital read... just some comments

dataprofiler/labelers/base_model.py
taylorfturner: does mypy complain about that?

dataprofiler/labelers/base_model.py
taylorfturner: https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

dataprofiler/labelers/base_model.py
tonywu315: mypy complains that `Signature of "fit" incompatible with supertype "BaseTrainableModel"` for all of the subclasses

dataprofiler/tests/labelers/test_integration_struct_data_labeler.py
```diff
@@ -57,7 +57,7 @@ def test_fit_with_default_model(self):
         self.assertEqual(3, len(model_predictions[0]))  # history, f1, f1_report
         self.assertIsInstance(model_predictions[0][0], dict)  # history
         self.assertIsNone(model_predictions[0][1])  # no f1 since no validation
-        self.assertListEqual(model_predictions[0][2], [])  # empty f1_report
+        self.assertDictEqual(model_predictions[0][2], {})  # empty f1_report
```
taylorfturner, line 60: good catch

dataprofiler/labelers/base_model.py
```diff
@@ -366,10 +371,10 @@ def fit(
         train_data: Union[pd.DataFrame, pd.Series, np.ndarray],
         val_data: Union[pd.DataFrame, pd.Series, np.ndarray],
         batch_size: int = 32,
-        epochs: int = 1,
```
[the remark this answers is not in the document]
tonywu315: added `epochs` back

dataprofiler/labelers/char_load_tf_model.py
```diff
@@ -376,7 +386,7 @@ def fit(
             history[metric_label] = model_results[i]
 
         if val_data:
-            f1, f1_report = self._validate_training(val_data)
+            f1, f1_report = self._validate_training(val_data)  # type: ignore
```
JGSweets, line 391: why the `ignore` here?

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -58,23 +62,23 @@ def create_glove_char(n_dims, source_file=None):
...
     pca = decomposition.PCA(n_components=n_dims)
-    reduced_embds = pca.fit_transform(embd_matrix)
+    reduced_embds: np.ndarray = pca.fit_transform(embd_matrix)
 
     # write to file
     dir_name = os.path.dirname(source_file)
     embd_file_name = os.path.join(dir_name, "glove-reduced-{}D.txt".format(n_dims))
     with open(embd_file_name, "w") as file:
-        for word, embd in zip(embd_words, reduced_embds):
+        for word, embd in zip(embd_words, reduced_embds):  # type: ignore
```
JGSweets, line 71: why the ignore?

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -653,7 +666,7 @@ def fit(
             history[metric_label] = model_results[i]
 
         if val_data:
-            f1, f1_report = self._validate_training(val_data)
+            f1, f1_report = self._validate_training(val_data)  # type: ignore
```
JGSweets, line 671: as above

dataprofiler/labelers/character_level_cnn_model.py
tonywu315: I just added typing to `embd_words`, so its unnecessary now

dataprofiler/labelers/character_level_cnn_model.py
tonywu315: same reason as above

dataprofiler/labelers/char_load_tf_model.py
JGSweets: Is there a way we can suggest to mypy that it can't be? is there where overload signatures come into play?
https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -592,13 +605,14 @@ def _reconstruct_model(self):
...
         self,
-        train_data,
-        val_data=None,
-        batch_size=32,
-        label_mapping=None,
-        reset_weights=False,
-        verbose=True,
-    ):
+        train_data: Union[pd.DataFrame, pd.Series, np.ndarray],
+        val_data: Union[pd.DataFrame, pd.Series, np.ndarray] = None,
```
taylorfturner, line 609:
```suggestion
        val_data: Optional[Union[pd.DataFrame, pd.Series, np.ndarray]] = None,
```

taylorfturner: ?

taylorfturner (changes requested):
comments (recommend aliasing where possible) and request around setting something to `None` by default without `Optional[]` typing

dataprofiler/labelers/char_load_tf_model.py
tonywu315: I tried adding overloading, but it doesn't work because `pandas` does not have any type stubs right now, so `pd.DataFrame` evalutes to `Any`. I found this potential [solution](https://stackoverflow.com/questions/63214123/mypy-overload-function-with-numpy-ndarray-and-pandas-dataframe-signature-parame).

dataprofiler/labelers/character_level_cnn_model.py
tonywu315: added

dataprofiler/labelers/char_load_tf_model.py
```diff
@@ -18,14 +22,18 @@
 logger = dp_logging.get_child_logger(__name__)
 labeler_utils.hide_tf_logger_warnings()
 
+Data = Union[pd.DataFrame, pd.Series, np.ndarray]
```
taylorfturner, line 25: @tonywu315 love it --can we make this used everywhere? I see other places in other files where `Union[pd.DataFrame, pd.Series, np.ndarray]` is used but not with an alias. I'd recommend doing this so we don't have to repeat in other files too

tonywu315: added to `utils.py`

taylorfturner (approved):

dataprofiler/labelers/char_load_tf_model.py
```diff
@@ -420,11 +436,8 @@ def _validate_training(
         :type verbose_keras: bool
         return (f1-score, f1 report).
         """
-        f1 = None
-        f1_report = None
-
         if val_data is None:
-            return f1, f1_report
+            return None, None
```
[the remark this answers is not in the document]
taylorfturner: ah I'm tracking... lemme  read this is a bit more

taylorfturner (dismissed):

dataprofiler/labelers/base_model.py
```diff
@@ -5,10 +5,12 @@
 import copy
 import inspect
 import warnings
-from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Type, Union
 
-import numpy as np
-import pandas as pd
+if TYPE_CHECKING:
+    import tensorflow as tf
```
JGSweets, line 11: This i'm concerned about, what happens if someone doesn't have TF installed?

This is why we put it in the class itself to avoid the req unless executed. I understand this won't hit at runtime, but I guess someone could operate on the code w/o TF?

JGSweets: Actually, `line 53` shouldn't not be a `tf.keras.model` truly it is `Any`. So we shouldn't need this line anyways?

tonywu315: I changed it

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -653,7 +670,7 @@ def fit(
             history[metric_label] = model_results[i]
 
         if val_data:
-            f1, f1_report = self._validate_training(val_data)
+            f1, f1_report = self._validate_training(val_data)  # type: ignore
```
taylorfturner, line 673: let's see how we might be able to remove the `#type: ignore` and maybe remove `if val_data` on line 672

tonywu315: removed redundant `if val_data`

dataprofiler/labelers/char_load_tf_model.py
```diff
@@ -227,7 +237,7 @@ def _construct_model(self):
         default_ind = self.label_mapping[self._parameters["default_label"]]
         model_loc = self._parameters["model_path"]
 
-        self._model = tf.keras.models.load_model(model_loc)
+        self._model: tf.keras.Model = tf.keras.models.load_model(model_loc)
         softmax_output_layer_name = self._model.outputs[0].name.split("/")[0]
         softmax_layer_ind = labeler_utils.get_tf_layer_index_from_name(
```
micdavis, line 242: would it make more sense to cast `softmax_layer_ind` to int at assignment rather than down below

dataprofiler/labelers/utils.py
```diff
@@ -1,7 +1,12 @@
 """Contains functions for checking for installations/dependencies."""
 import sys
 import warnings
-from typing import Any, Callable, List
+from typing import Any, Callable, List, Union
+
+import numpy as np
+import pandas as pd
+
+DataArray = Union[pd.DataFrame, pd.Series, np.ndarray]
```
JGSweets, line 9: If we are going to create our own. I might suggest we start following a similar pardigm to `numpy` and utilize our own `_typing` folder or `_typing.py` file at the root for universal typing that can be applied throughout the repo.

https://github.com/numpy/numpy/tree/main/numpy/_typing

dataprofiler/labelers/char_load_tf_model.py
```diff
@@ -351,9 +363,9 @@ def fit(
             if reset_weights:
                 self.reset_weights()
 
-        history = defaultdict()
-        f1 = None
-        f1_report = []
+        history: Dict = defaultdict()
+        f1: Optional[float] = None
+        f1_report: Dict = {}
```
taylorfturner, line 368:
```suggestion
        f1_report: Optional[Dict] = None
```

Maybe this so that if it does end up returning `{}` we can just check `if not f1_report and not f1` instead of having to check for a blank dict()?

dataprofiler/labelers/char_load_tf_model.py
taylorfturner: good call out

dataprofiler/labelers/char_load_tf_model.py
taylorfturner: also I think recommendation is to use non mutable default values... `dict` is a mutable type. So I think most folks just set defaults to `None`

dataprofiler/labelers/char_load_tf_model.py
```diff
@@ -351,9 +361,9 @@ def fit(
...
```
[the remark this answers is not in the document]
tonywu315: reverted

dataprofiler/labelers/char_load_tf_model.py
tonywu315: good idea

dataprofiler/labelers/utils.py
tonywu315: added `_typing.py`

taylorfturner (approved):

dataprofiler/labelers/char_load_tf_model.py
JGSweets: That's the case if it is in the function definition. Not when being defined within the func itself.

PR #673, opened 2022-10-03
Add static typing to labelers/data_processing.py

tonywu315:
#609

### DIFF
--- a/dataprofiler/labelers/base_model.py
+++ b/dataprofiler/labelers/base_model.py
@@ -5,7 +5,7 @@ import abc
 import copy
 import inspect
 import warnings
-from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, Iterator, List, Optional, Tuple, Type, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -60,7 +60,7 @@ class BaseModel(object, metaclass=abc.ABCMeta):
         super().__init_subclass__(**kwargs)
         cls._register_subclass()
 
-    def __eq__(self, other: BaseModel) -> bool:  # type: ignore
+    def __eq__(self, other: object) -> bool:
         """
         Check if two models are equal with one another.
 
@@ -75,6 +75,7 @@ class BaseModel(object, metaclass=abc.ABCMeta):
         """
         if (
             type(self) != type(other)
+            or not isinstance(other, BaseModel)
             or self._parameters != other._parameters
             or self._label_mapping != other._label_mapping
         ):
@@ -271,7 +272,7 @@ class BaseModel(object, metaclass=abc.ABCMeta):
 
         :return: None
         """
-        param_docs: str = inspect.getdoc(cls._validate_parameters)  # type: ignore
+        param_docs = cast(str, inspect.getdoc(cls._validate_parameters))
         param_start_ind = param_docs.find("parameters:\n") + 12
         param_end_ind = param_docs.find(":type parameters:") - 1
 
--- a/dataprofiler/labelers/data_processing.py
+++ b/dataprofiler/labelers/data_processing.py
@@ -11,7 +11,7 @@ import random
 import types
 import warnings
 from collections import Counter
-from typing import Any, Dict, Generator, Iterable, List, Tuple, Type, Union
+from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple, Type, Union
 
 import numpy as np
 import pkg_resources
@@ -453,12 +453,15 @@ class CharPreprocessor(BaseDataPreprocessor, metaclass=AutoSubRegistrationMeta):
         # create None generator if labels is not set
         if labels is None:
 
-            def gen_none():
+            def gen_none() -> Generator[None, None, None]:
                 """Generate infinite None(s). Must be closed manually."""
                 while True:
                     yield None
 
             labels = gen_none()
+
+        if label_mapping is None:
+            label_mapping = {}
 
         # loop through each sample
         for sample_buffer, label_set in zip(data, labels):
@@ -657,7 +660,13 @@ class CharPreprocessor(BaseDataPreprocessor, metaclass=AutoSubRegistrationMeta):
         if batch_data["samples"]:
             yield batch_data
 
-    def process(self, data, labels=None, label_mapping=None, batch_size=32):
+    def process(
+        self,
+        data: np.ndarray,
+        labels: np.ndarray = None,
+        label_mapping: Dict[str, int] = None,
+        batch_size: int = 32,
+    ) -> Generator[Union[Tuple[np.ndarray, np.ndarray], np.ndarray], None, None]:
         """
         Flatten batches of data.
 
@@ -719,7 +728,7 @@ class CharPreprocessor(BaseDataPreprocessor, metaclass=AutoSubRegistrationMeta):
                 [[sentence] for sentence in batch_data["samples"]], dtype=object
             )
             if labels is not None:
-                num_classes = max(label_mapping.values()) + 1
+                num_classes = max(label_mapping.values()) + 1  # type: ignore
 
                 Y_train = tf.keras.utils.to_categorical(
                     batch_data["labels"], num_classes
@@ -734,14 +743,14 @@ class CharEncodedPreprocessor(CharPreprocessor, metaclass=AutoSubRegistrationMet
 
     def __init__(
         self,
-        encoding_map=None,
-        max_length=5000,
-        default_label="UNKNOWN",
-        pad_label="PAD",
-        flatten_split=0,
-        flatten_separator=" ",
-        is_separate_at_max_len=False,
-    ):
+        encoding_map: Dict[str, int] = None,
+        max_length: int = 5000,
+        default_label: str = "UNKNOWN",
+        pad_label: str = "PAD",
+        flatten_split: float = 0,
+        flatten_separator: str = " ",
+        is_separate_at_max_len: bool = False,
+    ) -> None:
         """
         Initialize the CharEncodedPreprocessor class.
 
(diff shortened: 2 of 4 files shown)


dataprofiler/labelers/data_processing.py
```diff
@@ -107,40 +116,35 @@ def get_parameters(self, param_list=None):
...
+    def set_params(self, **kwargs: Any) -> None:
         """Set the parameters if they exist given kwargs."""
         # first check if any parameters are invalid
         self._validate_parameters(kwargs)
 
         for param in kwargs:
             self._parameters[param] = kwargs[param]
 
-    @abc.abstractmethod
-    def process(self, *args):
```
tonywu315, line 119: `mypy` creates a errors for `process` in each subclass because the type signature is different, so I removed it here

dataprofiler/labelers/data_processing.py
```diff
@@ -1500,7 +1575,7 @@ def __eq__(self, other):
         :rtype: bool
         """
         if (
-            type(self) != type(other)
+            not isinstance(other, StructCharPostprocessor)
```
tonywu315, line 1578: check to see if other is also `StructCharPostprocessor`

dataprofiler/labelers/data_processing.py
```diff
@@ -2055,14 +2155,19 @@ def _save_processor(self, dirpath):
         ) as fp:
             json.dump(params, fp)
 
-    def process(self, data, labels=None, label_mapping=None, batch_size=None):
+    def process(
+        self,
+        data: np.ndarray,
+        results: Dict,
+        label_mapping: Dict[str, int],
```
tonywu315, line 2162: Renamed `labels` to `results` to be consistent with other postprocessors `process` and the superclass

dataprofiler/labelers/data_processing.py
```diff
@@ -1911,25 +2003,30 @@ def priority_prediction(results, entity_priority_order):
...
             results["pred"][i] = pred / np.linalg.norm(
                 pred, axis=1, ord=1, keepdims=True
             )
 
-    def process(self, data, labels=None, label_mapping=None, batch_size=None):
+    def process(
+        self,
+        data: np.ndarray,
+        results: Dict,
+        label_mapping: Dict[str, int],
```
tonywu315, line 2023: Renamed `labels` to `results` to be consistent with other postprocessors `process` and the superclass. Also, removed default `None` from `label_mapping` because that causes an error later in the function.

dataprofiler/labelers/data_processing.py
taylorfturner: hmm what is you just remove `@abc.abstractmethod` as the decorator. We would still want the `def process()` in the base class definition so it enforces that function in all child classes.

dataprofiler/labelers/data_processing.py
```diff
@@ -1307,6 +1369,7 @@ def help(cls):
         :return: None
         """
         param_docs = inspect.getdoc(cls._validate_parameters)
+        assert param_docs is not None
```
taylorfturner, line 1372: all this will raise is an `AssertionError`, which isn't terribly explicit for the user of the tool. Couple thoughts: 
- `inspect.getdoc(cls._validate_parameters)` should always return `not None` based on that fact there is a docstring
- at the same time, I don't mind checking that is not `None`, but if we do keep this, I'd prefer to have a couple things. 1) an explanation of why it failed and 2) recommend path for resolving the issue in a `warning` that the code raises

dataprofiler/labelers/data_processing.py
```diff
@@ -1500,7 +1575,7 @@ def __eq__(self, other):
         :rtype: bool
         """
         if (
-            type(self) != type(other)
```
taylorfturner, line 1503: why get rid of this check?

dataprofiler/labelers/data_processing.py
```diff
@@ -1557,6 +1632,7 @@ def help(cls):
         :return: None
         """
         param_docs = inspect.getdoc(cls._validate_parameters)
+        assert param_docs is not None
```
taylorfturner, line 1635: Same comment from above on this ... 


> all this will raise is an AssertionError, which isn't terribly explicit for the user of the tool. Couple thoughts:

> inspect.getdoc(cls._validate_parameters) should always return not None based on that fact there is a docstring
at the same time, I don't mind checking that is not None, but if we do keep this, I'd prefer to have a couple things. 1) an explanation of why it failed and 2) recommend path for resolving the issue in a warning that the code raises

dataprofiler/tests/labelers/test_data_processing.py
```diff
@@ -2958,7 +2958,7 @@ def test_process(self):
         }
 
         processor = ColumnNameModelPostprocessor()
-        process_output = processor.process(data, labels=model_output)
+        process_output = processor.process(data, model_output)
```
taylorfturner, line 2961:
```suggestion
        process_output = processor.process(data, results=model_output)
```

dataprofiler/labelers/data_processing.py
taylorfturner: good call

dataprofiler/labelers/data_processing.py
taylorfturner: good call

dataprofiler/labelers/data_processing.py
```diff
@@ -1875,6 +1966,7 @@ def help(cls):
...
```
taylorfturner, line 1969: Same comment from above on this ...

> all this will raise is an AssertionError, which isn't terribly explicit for the user of the tool. Couple thoughts:
> 
> inspect.getdoc(cls._validate_parameters) should always return not None based on that fact there is a docstring
> at the same time, I don't mind checking that is not None, but if we do keep this, I'd prefer to have a couple things. 1) an explanation of why it failed and 2) recommend path for resolving the issue in a warning that the code raises

dataprofiler/labelers/data_processing.py
```diff
@@ -2129,6 +2234,7 @@ def help(cls):
...
```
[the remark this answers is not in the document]
taylorfturner: > 
> all this will raise is an AssertionError, which isn't terribly explicit for the user of the tool. Couple thoughts:
> 
> inspect.getdoc(cls._validate_parameters) should always return not None based on that fact there is a docstring
> at the same time, I don't mind checking that is not None, but if we do keep this, I'd prefer to have a couple things. 1) an explanation of why it failed and 2) recommend path for resolving the issue in a warning that the code raises

taylorfturner (dismissed):
good changes and a couple requests / questions

dataprofiler/labelers/data_processing.py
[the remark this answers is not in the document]
tonywu315: i cast it to str now

dataprofiler/labelers/data_processing.py
tonywu315: I changed it to `not isinstance(other, StructCharPostprocessor)`, which allows `mypy` to infer the type of `other`. It should be equivalent.

dataprofiler/labelers/data_processing.py
taylorfturner: @tonywu315, I'd prefer to keep this, but if its causing an issue with `mypy` glad to talk about that. CC @JGSweets

dataprofiler/tests/labelers/test_data_processing.py
```diff
@@ -2958,7 +2958,7 @@ def test_process(self):
         }
 
         processor = ColumnNameModelPostprocessor()
-        process_output = processor.process(data, labels=model_output)
+        process_output = processor.process(data, results=model_output)
```
taylorfturner, line 2961: awesome!

dataprofiler/labelers/data_processing.py
taylorfturner: per @tonywu315, `Signature of "process" incompatible with supertype "BaseDataProcessor"` is how `mypy` complains about the `def process()`

tonywu315: I would be possible to keep it if we added default values to `results` and `label_mapping` for all of the postprocessors, which would make them not violate Liskov substitution principle.

taylorfturner:
@tonywu315 tests failing on 3.7 and 3.8

micdavis (approved):

dataprofiler/labelers/base_model.py
```diff
@@ -74,6 +74,7 @@ def __eq__(self, other: BaseModel) -> bool:  # type: ignore
         """
         if (
             type(self) != type(other)
+            or not isinstance(other, BaseModel)
```
JGSweets, line 77: This theoretically should be covered by above, but does mypy get confused?

dataprofiler/labelers/data_processing.py
```diff
@@ -62,18 +82,22 @@ def __eq__(self, other):
         :return: Whether or not self and other are equal
         :rtype: bool
         """
-        if type(self) != type(other) or self._parameters != other._parameters:
+        if (
+            type(self) != type(other)
+            or not isinstance(other, BaseDataProcessor)
```
JGSweets, line 87: same, another mypy confused?

dataprofiler/labelers/base_model.py
tonywu315: yeah mypy thinks `other` is still an object

dataprofiler/labelers/data_processing.py
taylorfturner: `type(self) != type(other)` should validate that `not isinstance(other, BaseDataProcessor)` would return `True`. Just a redundant check

dataprofiler/labelers/base_model.py
taylorfturner: `type(self) != type(other)` should validate that `not isinstance(other, BaseDataProcessor)` would return `True`. Just a redundant check. I'm fine with it --> better than `#type: ignore` and casting IMO

JGSweets (approved):

PR #677, opened 2022-10-05
Added static typing to *_data classes in data_readers

Sanketh7:
#610


taylorfturner:
looks like tests are failing because `options` is `None` 

```
>       self.SAMPLES_PER_LINE_DEFAULT: int = options.get("record_samples_per_line", 1)
E       AttributeError: 'NoneType' object has no attribute 'get'
```

dataprofiler/data_readers/csv_data.py
```diff
@@ -71,15 +75,15 @@ def __init__(self, input_file_path=None, data=None, options=None):
...
-        self._delimiter = options.get("delimiter", None)
-        self._quotechar = options.get("quotechar", None)
-        self._selected_columns = options.get("selected_columns", list())
-        self._header = options.get("header", "auto")
-        self._checked_header = "header" in options and self._header != "auto"
-        self._default_delimiter = ","
-        self._default_quotechar = '"'
+        self.SAMPLES_PER_LINE_DEFAULT: int = options.get("record_samples_per_line", 1)
+        self._selected_data_format: str = options.get("data_format", "dataframe")
+        self._delimiter: Optional[str] = options.get("delimiter", None)
```
JGSweets, line 80: we should remove the extra spaces at the end of these too, is pre-commit setup or did it not run?

dataprofiler/data_readers/csv_data.py
```diff
@@ -57,9 +59,11 @@ def __init__(self, input_file_path=None, data=None, options=None):
         :type options: dict
         :return: None
         """
-        options = self._check_and_return_options(options)
+        if options is not None:
```
JGSweets, line 62: I think We can remove this if statement, no?

JGSweets: that should fix the bug too

dataprofiler/data_readers/graph_data.py
```diff
@@ -169,12 +169,12 @@ def is_match(cls, file_path, options=None):
...
         return False
 
-    def _format_data_networkx(self):
+    def _format_data_networkx(self) -> nx.Graph:
         """Format the input file into a networkX graph."""
         networkx_graph = nx.Graph()
 
         # read lines from csv
-        if not self._checked_header or not self._delimiter:
+        if (not self._checked_header or not self._delimiter) and self.input_file_path is not None:
```
JGSweets, line 177: why does this need to be not `None`? was it not working before?

Does `load_as_str_from_file` need to update to accept `Optional` as it handles it internally?

JGSweets: `input_file_path` cannot be None, fair. Is there a better way to define at the beginning of the func as opposed to repeating throughout the func?

Could we at the beginning of the func do:
```python
assert self.input_file_path is not None
```

dataprofiler/data_readers/csv_data.py
Sanketh7: Done

dataprofiler/data_readers/csv_data.py
Sanketh7: Looks like pre-commit didn't run for some reason. I manually ran it and pushed formatting changes.

dataprofiler/data_readers/graph_data.py
```diff
@@ -12,9 +13,14 @@
 class GraphData(BaseData):
     """GraphData class to identify, read, and load graph data."""
 
-    data_type = "graph"
+    data_type: Optional[str] = "graph"
```
taylorfturner, line 16: in this file... shouldn't this always be `data_type: str`?

dataprofiler/data_readers/graph_data.py
```diff
@@ -237,41 +263,41 @@ def _format_data_networkx(self):
...
             attributes = dict()
             for column in range(0, len(csv_as_list[0])):
-                if csv_as_list[line][column] is None:
+                if csv_as_list[line_index][column] is None:
                     continue
                 if (
                     column is not self._source_node
                     or column is not self._destination_node
-                ):
+                ) and self._column_names is not None:
```
taylorfturner, line 275: **Note**: code change here

dataprofiler/data_readers/graph_data.py
```diff
@@ -212,23 +234,27 @@ def _format_data_networkx(self):
...
+        if (
+            self._column_names is None
+            and isinstance(self._header, int)
+            and self.file_encoding is not None
+        ):
             self._column_names = self.csv_column_names(
                 self.input_file_path, self._header, self._delimiter, self.file_encoding
             )
-        if self._source_node is None:
+        if self._source_node is None and self._column_names is not None:
```
taylorfturner, line 245: Note: code change here

dataprofiler/data_readers/graph_data.py
```diff
@@ -212,23 +234,27 @@ def _format_data_networkx(self):
...
             self._source_node = self._find_target_string_in_column(
                 self._column_names, self._source_keywords
             )
-        if self._destination_node is None:
+        if self._destination_node is None and self._column_names is not None:
```
taylorfturner, line 249: Note: code change here

dataprofiler/data_readers/text_data.py
```diff
@@ -13,9 +14,14 @@
 class TextData(BaseData):
     """TextData class to save and load text files."""
 
-    data_type = "text"
+    data_type: Optional[str] = "text"
```
taylorfturner, line 17: same as above... in this file it would only be a `str` ?

taylorfturner (commented):
couple material comments and some notes around code additions

ksneab7 (dismissed):

dataprofiler/data_readers/csv_data.py
```diff
@@ -18,9 +20,14 @@
 class CSVData(SpreadSheetDataMixin, BaseData):
     """SpreadsheetData class to save and load spreadsheet data."""
 
-    data_type = "csv"
+    data_type: Optional[str] = "csv"
```
JGSweets, line 23: I think this might be an issue with the original code. The value shouldn't be optional, it is only `None for the base class b/c it needs to be set by the derived. Hence, setting it to optional seems like it isn't true.

Maybe in the base we could make it an abstract property?

```python
import abc

class Base(abc.ABC):
    @property
    @abc.abstractmethod
    def data_type(self) -> str: ...

    @data_type.setter
    @abc.abstractmethod
    def data_type(self, value: str) -> None: ...

class CSVData(Base):
    data_type: str = "csv
```

Would something like this work?

taylorfturner: +1 yeah I'd prefer too that the child classes aren't `Optional[str]` type for `data_type` attribute CC @Sanketh7

dataprofiler/data_readers/csv_data.py
```diff
@@ -58,8 +65,9 @@ def __init__(self, input_file_path=None, data=None, options=None):
         :return: None
         """
         options = self._check_and_return_options(options)
+        options = cast(Dict, options)
```
JGSweets, line 68: shouldn't the line above already indicate that it is a dict?

dataprofiler/data_readers/csv_data.py
```diff
@@ -89,39 +97,40 @@ def __init__(self, input_file_path=None, data=None, options=None):
...
+    def _check_and_return_options(options: Optional[Dict]) -> Dict:
         """
         Ensure options are valid inputs to the data reader.
 
         :param options: dictionary of options for the csv reader to validate
         :type options: dict
         :return: None
         """
-        options = super()._check_and_return_options(options)
+        options = super(CSVData, CSVData)._check_and_return_options(options)
```
JGSweets, line 133: What's going on here? this seems abnormal.

dataprofiler/data_readers/csv_data.py
```diff
@@ -186,7 +198,7 @@ def _guess_delimiter_and_quotechar(
         vocab = Counter(data_as_str)
         if "\n" in vocab:
             vocab.pop("\n")
-        for char in omitted + [quotechar]:
+        for char in omitted + ([quotechar] if quotechar is not None else []):
```
JGSweets, line 201: I think this line gets a little confusing if it's all in one. If this if is necessary, can we do it before the `for`?
```python
omitted_list: list[str] = ommitted
if quotechar is not None:
    omitted_list: list[str] = omitted + [quotechar]
```

JGSweets: then we can use that in the for

dataprofiler/data_readers/csv_data.py
```diff
@@ -534,13 +546,13 @@ def _load_data_from_str(self, data_as_str):
             )
         return data_utils.read_csv_df(
             data_buffered,
-            self.delimiter,
-            self.header,
+            cast(str, self.delimiter),
```
JGSweets, line 549: do these need to be casted?

dataprofiler/data_readers/graph_data.py
Sanketh7: The issue is that `BaseData` sets the field to `None`.

dataprofiler/data_readers/text_data.py
Sanketh7: `BaseData` sets that field to `None`

dataprofiler/data_readers/csv_data.py
Sanketh7: This last solution works well and I've committed that change.

dataprofiler/data_readers/csv_data.py
Sanketh7: I had to change `_check_and_return_options` to be a static method because `BaseData` implements it as a static method. However, `super()` with no arguments only really works with instance and class methods. I ended up following https://stackoverflow.com/questions/26788214/super-and-staticmethod-interaction#:~:text=When%20you%20call%20a%20class,or%20class%20it%20was%20called. to figure out how to make it work with static methods.

dataprofiler/data_readers/csv_data.py
Sanketh7: You're right. I probably forgot to delete this from a previous implementation.

dataprofiler/data_readers/csv_data.py
```diff
@@ -58,8 +65,9 @@ def __init__(self, input_file_path=None, data=None, options=None):
         :return: None
         """
         options = self._check_and_return_options(options)
+        options = cast(Dict, options)
         BaseData.__init__(self, input_file_path, data, options)
-        SpreadSheetDataMixin.__init__(self, input_file_path, data, options)
+        SpreadSheetDataMixin.__init__(self, input_file_path, data, cast(Dict, options))
```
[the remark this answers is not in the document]
Sanketh7: Same as above.

dataprofiler/data_readers/csv_data.py
Sanketh7: Oops. Looks like it makes a test fail.
```
=================================== FAILURES ===================================
_______________ TestBaseDataClass.test_can_apply_data_functions ________________
self = <dataprofiler.tests.data_readers.test_base_data.TestBaseDataClass testMethod=test_can_apply_data_functions>

    def test_can_apply_data_functions(self):
        class FakeDataClass:
            # matches the `data_type` value in BaseData for validating priority
            data_type = "FakeData"
    
            def func1(self):
                return "success"
    
        # initialize the data class
        data = BaseData(input_file_path="", data=FakeDataClass(), options={})
    
        # if the function exists in BaseData fail the test because the results
        # may become inaccurate.
        self.assertFalse(hasattr(BaseData, "func1"))
    
        with self.assertRaisesRegex(
            AttributeError,
            "Neither 'BaseData' nor 'FakeDataClass' " "objects have attribute 'test'",
        ):
            data.test
    
        # validate it will take BaseData attribute over the data attribute
>       self.assertIsNone(data.data_type)
E       AssertionError: 'FakeData' is not None

dataprofiler/tests/data_readers/test_base_data.py:96: AssertionError
```

dataprofiler/data_readers/csv_data.py
Sanketh7: Looks like with one of my recent commits to data_utils, `self.delimiter` doesn't need to be casted. However, `self.header` does because `read_csv_df` only supports `Optional[int]` but `self.header` could also be a string the `"auto"` case. As far as I can tell, this case doesn't exist at this point at runtime but it doesn't seem like mypy can detect that.

dataprofiler/data_readers/csv_data.py
Sanketh7: I assume we could change the test to detect an attribute error instead of checking if the field is None?

JGSweets: Interesting, basically bc data_type doesn't exist in BaseData, this test is no longer validating priority.
Attribute error still wouldn't be correct bc FakeData as it.
We would need to test a property they both have. I think instead we can update the test to use `input_file_path`
```python
        class FakeDataClass:
            # matches the `data_type` value in BaseData for validating priority
            options = {"not_empty": "data"}
    
            def func1(self):
                return "success"
```
In the test we can assert the `options` is empty.

dataprofiler/data_readers/csv_data.py
JGSweets: Instead could we use `BaseData. _check_and_return_options(options)`?

dataprofiler/data_readers/csv_data.py
JGSweets: @Sanketh7 in case you missed

Sanketh7: Done

dataprofiler/data_readers/csv_data.py
Sanketh7: Made the change here: https://github.com/capitalone/DataProfiler/pull/677/commits/2e850014d17df3baa1c6415ae0d2c6d76ecb893f

I didn't remove the data_type field from FakeDataClass because I wasn't sure if that would cause unintended side effects.

dataprofiler/data_readers/csv_data.py
taylorfturner: @JGSweets thoughts on this? Any additional thoughts around this implementation? 

Found this example which helps provide an example for how @Sanketh7 implement this: https://stackoverflow.com/questions/26788214/super-and-staticmethod-interaction#:~:text=class%20Second(First)%3A%0A%20%20%40staticmethod%0A%20%20def%20getlist()%3A%0A%20%20%20%20l%20%3D%20super(Second%2C%20Second).getlist()%20%20%23%20note%20the%202nd%20argument%0A%20%20%20%20l.append(%27second%27)

dataprofiler/data_readers/csv_data.py
```diff
@@ -716,7 +733,12 @@ def is_match(cls, file_path, options=None):
         # Assume not a CSV
         return False
 
-    def reload(self, input_file_path=None, data=None, options=None):
+    def reload(
+        self,
+        input_file_path: Optional[str] = None,
+        data: Optional[pd.DataFrame] = None,
```
taylorfturner, line 739: `:type data: multiple types` is the docstring for `data`... but static typing is only saying `None` or `pd.DataFrame`. I think at least one or the other should be update to make sure the code and docstring match

dataprofiler/data_readers/csv_data.py
```diff
@@ -737,4 +759,4 @@ def reload(self, input_file_path=None, data=None, options=None):
             header=self.header, delimiter=self.delimiter, quotechar=self.quotechar
         )
         super(CSVData, self).reload(input_file_path, data, options)
-        self.__init__(self.input_file_path, data, options)
+        self.__init__(self.input_file_path, data, options)  # type: ignore
```
taylorfturner, line 762: Can we get rid of this `#type: ignore` by resolving an issue upstream?

dataprofiler/data_readers/graph_data.py
```diff
@@ -12,9 +13,14 @@
...
     """GraphData class to identify, read, and load graph data."""
 
-    data_type = "graph"
+    data_type: str = "graph"
 
-    def __init__(self, input_file_path=None, data=None, options=None):
+    def __init__(
+        self,
+        input_file_path: Optional[str] = None,
+        data: Optional[nx.Graph] = None,
```
taylorfturner, line 21: same comment from above in `csv_data` --> `:type data: multiple types` is type on `data` but we only allow `nx.graph` or `None` in this `__init__`. Should update so they match

dataprofiler/data_readers/parquet_data.py
```diff
@@ -118,6 +137,9 @@ def is_match(cls, file_path, options=None):
 
         # get current position of stream
         if data_utils.is_stream_buffer(file_path):
+            file_path = cast(
+                Union[StringIO, BytesIO], file_path
+            )  # guaranteed by is_stream_buffer
```
taylorfturner, line 142: let's see if we can get rid of these `cast` statements

dataprofiler/data_readers/parquet_data.py
```diff
@@ -128,11 +150,19 @@ def is_match(cls, file_path, options=None):
 
         # return to original position in stream
         if data_utils.is_stream_buffer(file_path):
+            file_path = cast(
+                Union[StringIO, BytesIO], file_path
+            )  # guaranteed by is_stream_buffer
```
taylorfturner, line 155: let's see if we can get rid of these `cast` statements

dataprofiler/data_readers/parquet_data.py
```diff
@@ -148,4 +178,4 @@ def reload(self, input_file_path=None, data=None, options=None):
         :return: None
         """
         super(ParquetData, self).reload(input_file_path, data, options)
-        self.__init__(self.input_file_path, data, options)
+        self.__init__(self.input_file_path, data, options)  # type: ignore
```
taylorfturner, line 181: any change we can get rid of this `#type: ignore`

dataprofiler/data_readers/text_data.py
```diff
@@ -45,6 +51,7 @@ def __init__(self, input_file_path=None, data=None, options=None):
             raise ValueError("Input data type is not string.")
 
         options = self._check_and_return_options(options)
+        options = cast(Dict, options)
```
taylorfturner, line 54: any chance we can get rid of this `cast`?

dataprofiler/tests/data_readers/test_base_data.py
```diff
@@ -93,7 +94,7 @@ def func1(self):
             data.test
 
         # validate it will take BaseData attribute over the data attribute
-        self.assertIsNone(data.data_type)
```
taylorfturner, line 96: why is this getting removed?

dataprofiler/data_readers/csv_data.py
Sanketh7: The issue is that mypy doesn't like using `self.__init__` because it doesn't know which constructor will end up being called (and therefore doesn't know what types are needed). However, the current code assumes that behavior so I just told mypy to ignore that line.

dataprofiler/data_readers/parquet_data.py
Sanketh7: Done.

dataprofiler/data_readers/text_data.py
Sanketh7: Looks like that cast is no longer needed. Removed it.

dataprofiler/tests/data_readers/test_base_data.py
Sanketh7: In order to make `data_type` have type `str` instead of `Optional[str]`, I had to make it so it doesn't get set to anything in `BaseData` (and just has a type definition line for mypy). This makes it so you get an exception if you try getting the `data_type` in `BaseData` so @JGSweets recommended looking at another field to test the same behavior. In this case, I looked to make sure `options` is taken from `BaseData` instead of `FakeData` which would mean that `options` is an empty dict.

dataprofiler/data_readers/text_data.py
taylorfturner: awesome

dataprofiler/data_readers/csv_data.py
JGSweets: I'm fine with it. We can refactor in the future if necessary.

dataprofiler/tests/data_readers/test_base_data.py
JGSweets: correct

JGSweets (approved):

micdavis (approved):

Issue #633, opened 2022-09-16
PSI added to diff of numerical stat columns


JGSweets:
**Is your feature request related to a problem? Please describe.**
Within https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/numerical_column_stats.py#L350
Need to add PSI - https://medium.com/model-monitoring-psi/population-stability-index-psi-ab133b0a5d42
Should be a helper function to calculate which is called withtin the `def diff` function.

**Describe the outcome you'd like:**
receiving PSI in the `diff` command of NumericalStatsMixin
Tests around the addition.

**Additional context:**

taylorfturner:
I'm working this issue

taylorfturner:
PR merged

PR #688, opened 2022-10-13
Adding `PSI` to `diff` report 

taylorfturner:
Adding PSI to `diff` report as part of issue #633 

- adding PSI `def _calculate_psi()` and implement in `NumericStatsMixin.diff()`
- updated unit testing to reflect `psi` output as part of `diff()`


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -349,6 +349,8 @@ def report(self, remove_disabled_flag: bool = False) -> Dict:
                     continue
                 elif profile_key == "times":
                     continue
+                elif profile_key == "psi":
```
taylorfturner, line 352: add option for disabling flag `psi`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -391,6 +393,7 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
                 other_profile.variance,
                 other_profile.match_count,
             ),
+            "psi": self._calculate_psi(),
```
taylorfturner, line 396: call static method in `def diff()`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -516,6 +519,10 @@ def _perform_t_test(
         results["welch"]["p-value"] = float(welch_p_val)
         return results
 
+    @staticmethod
+    def _calculate_psi():
+        raise NotImplementedError()
+
     def _update_variance(
```
taylorfturner, line 526: this is where PSI actually gets calculated

dataprofiler/profilers/numerical_column_stats.py
JGSweets: can we combine this with `times` and create a `in [...]`?

taylorfturner: we could... I kinda see `times` and `psi` as two distinct keys ... whereas `stddev` and `variance` together make sense becuase they are derivatives of each other

JGSweets: they are different, but the logic is the same. if the key is this, continue. whereas stddev / variance are dependent hence, if one is not there, we have to pop the other.

taylorfturner: update

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -391,7 +439,16 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
                 other_profile.match_count,
             ),
+            "psi": self._calculate_psi(
+                self.mean,
+                self.sum,
+                self._stored_histogram["histogram"],
+                other_profile.mean,
+                other_profile.sum,
+                other_profile._stored_histogram["histogram"],
+            ),
```
taylorfturner, line 449: simply adding `psi` as a new `key` in the `differences `dict` that is returned from the `def diff()` method

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -516,6 +573,48 @@ def _perform_t_test(
...
+                continue
+
+            iter_psi = (other_percent - self_percent) * np.log(
+                other_percent / self_percent
+            )
+            if iter_psi and iter_psi != float("inf"):
+                psi_value += iter_psi
+
+        return psi_value
+
```
taylorfturner, line 617: `staticmethod` for calculating the `psi` value between two column profile objects

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1059,42 +1158,9 @@ def _update_histogram(self, df_series: pd.Series) -> None:
         self._stored_histogram["current_loss"] = histogram_loss
         self._stored_histogram["total_loss"] += histogram_loss
 
-    def _histogram_for_profile(
-        self, histogram_method: str
+    def _regenerate_histogram(
```
taylorfturner, line 1161: broke out `_histogram_for_profile()` function into two functions in the code so that we could call `_regenerate_histogram` on either `self` or `other_profile`.

The idea here is that to calculate PSI the edges in the historgram need to be the same between `self` and `other_profile`. In order to accomplish, prior to calling `_calculate_psi` potentially one or both `obj._stored_histograms` need to be updated. 

This is the purpose of the refactor and breaking out into `_regenerate_histogram`.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1159,6 +1225,49 @@ def _histogram_for_profile(
...
+            return (
+                self._stored_histogram["histogram"],
+                self._stored_histogram["total_loss"],
+            )
+
+        return self._regenerate_histogram(
+            bin_counts=bin_counts,
+            bin_edges=bin_edges,
+            suggested_bin_count=suggested_bin_count,
+        )
```
taylorfturner, line 1269: Calling `_regenerate_histogram` here now. The old code in the function is now in `def _regenerate_histogram()`.

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -164,16 +164,17 @@ def test_diff_primitive_compilers(self):
...
+                "mode": [[-1.15, 0.55, 2.25, 14.15], [], [5, -1]],
+                "median_absolute_deviation": -1.30,
                 "variance": 38.666666666666664,
                 "stddev": 3.285085839971525,
                 "t-test": {
                     "t-statistic": 0.4155260166386663,
                     "conservative": {"df": 1, "p-value": 0.749287157907667},
                     "welch": {"df": 3.6288111187629117, "p-value": 0.7011367179395704},
                 },
+                "psi": 0.17328679513998632,
```
taylorfturner, line 177: `test_diff_primitive_compilers` `expected_diff` results

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -577,9 +577,9 @@ def test_diff(self):
             "max": -1.0,
             "sum": -9.0,
             "mean": profile1["mean"] - profile2["mean"],
-            "median": -2.5,
-            "mode": [[1], [], [2, 4]],
-            "median_absolute_deviation": -0.5,
+            "median": -2.40,
+            "mode": [[1.15], [], [2, 4]],
```
taylorfturner, line 581: changes as `suggested_bin_counts` changed for PSI calculation

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -1053,14 +1053,15 @@ def test_diff(self):
             "stddev": -7.899494936611665,
             "sum": -4.0,
             "variance": -94.0,
-            "median": -4,
-            "mode": [[2, 6, 4], [], [1, 15]],
-            "median_absolute_deviation": -5,
+            "median": -3.80,
+            "mode": [[2.2, 4.2, 5.8], [], [1, 15]],
+            "median_absolute_deviation": -5.39,
```
taylorfturner, line 1058: changes as `suggested_bin_counts` changed for PSI calculation

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1684,8 +1684,8 @@ def test_diff(self):
             "sum": 3.5,
             "variance": profile1["variance"] - profile2["variance"],
             "median": 4.25,
-            "mode": [[2.5, 12.5, 5], [], [1, 15, 0.5, 0]],
-            "median_absolute_deviation": 2,
+            "mode": [[3.0, 5.0, 12.0], [], [1, 15, 0.5, 0]],
+            "median_absolute_deviation": 1.50,
```
taylorfturner, line 1688: changes as `suggested_bin_counts` changed for PSI calculation

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -247,8 +248,8 @@ def test_disabling_columns_during_primitive_diff(self):
                 "sum": -20.0,
                 "mean": -10.0,
                 "median": -10,
-                "mode": [[-2, -1, 1, 2], [], [5, 15]],
-                "median_absolute_deviation": -3.5,
+                "mode": [[-1.8, -1.0, 1.0, 1.8], [], [5, 15]],
+                "median_absolute_deviation": -3.595,
```
taylorfturner, line 252: changes as `suggested_bin_counts` changed for PSI calculation

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -164,16 +164,17 @@ def test_diff_primitive_compilers(self):
                 "max": 10.0,
                 "sum": 12.0,
                 "mean": 2.0,
-                "median": -0.5,
-                "mode": [[-2, 15, 1, 2], [], [5, -1]],
-                "median_absolute_deviation": -1,
+                "median": -0.60,
+                "mode": [[-1.15, 0.55, 2.25, 14.15], [], [5, -1]],
+                "median_absolute_deviation": -1.30,
```
taylorfturner, line 169: changes as `suggested_bin_counts` changed for PSI calculation

tyfarnan (dismissed):

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,49 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+                len_self_bin_counts = len(self_bin_counts)
+
+            # re-calculate `self` histogram
+            if not len_self_bin_counts == 10 and self_bin_counts.all() is not None:
+                histogram, hist_loss = self._regenerate_histogram(
+                    bin_counts=self._stored_histogram["histogram"]["bin_counts"],
+                    bin_edges=self._stored_histogram["histogram"]["bin_edges"],
+                    suggested_bin_count=9,
+                )
+                self._stored_histogram["histogram"]["bin_counts"] = histogram[
```
JGSweets, line 386: I don't think we should alter the stored histogram within diff, this could alter the results after calculating diff.

JGSweets (changes requested):
I suggest using the data we collect via the regenerate, but not altering the stored profile as 10/20 buns may not be as accurate as the best fit histogram

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: Great call out! working to resolve the down stream impact of this now

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,49 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
 
+        len_self_bin_counts = 0
+        self_bin_counts = self._stored_histogram["histogram"]["bin_counts"]
+
+        try:
+            if self_bin_counts.all() is not None and len(self_bin_counts) > 0:
+                len_self_bin_counts = len(self_bin_counts)
+
+            # re-calculate `self` histogram
+            if not len_self_bin_counts == 10 and self_bin_counts.all() is not None:
```
JGSweets, line 380: do we want to prepare this for allowing `20` to be a value as well?
making it a property pulled from options with a default of 10?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,49 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+                histogram, hist_loss = self._regenerate_histogram(
+                    bin_counts=self._stored_histogram["histogram"]["bin_counts"],
+                    bin_edges=self._stored_histogram["histogram"]["bin_edges"],
+                    suggested_bin_count=9,
```
JGSweets, line 384: `suggested_bin_count=9` but we check for `10`?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,49 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+            if (
+                other_profile._stored_histogram["histogram"]["bin_counts"].any()
+                != self._stored_histogram["histogram"]["bin_counts"].any()
+            ) and self_bin_counts.all() is not None:
+                histogram, hist_loss = self._regenerate_histogram(
+                    bin_counts=other_profile._stored_histogram["histogram"][
+                        "bin_counts"
+                    ],
+                    bin_edges=other_profile._stored_histogram["histogram"]["bin_edges"],
+                    suggested_bin_count=9,
```
JGSweets, line 403: `9` again here.

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: I'm definitely game to add! My though is to keep this PR small and focused on just adding the functionality when `self.diff()` is called. Then have a follow-up for configuring `PSI` through options. Thoughts?

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: changed back in most recent commit

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: changed back in most recent commit

dataprofiler/profilers/numerical_column_stats.py
JGSweets: I suggest to keep it clean, i would at the very least set this as a variable at the top to ensure consistency of all changes in the future.

e.g. just change `10` to `num_psi_bins`
then we can always do a `num_psi_bins = options.get(...., 10)` when we are ready

JGSweets: Otherwise, if someone accidentally changes one of these, these calcs will be off entirely

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,52 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+            # re-calculate `self` histogram
+            if not len_self_bin_counts == 10 and self_bin_counts.all() is not None:
+                histogram, hist_loss = self._regenerate_histogram(
+                    bin_counts=self._stored_histogram["histogram"]["bin_counts"],
+                    bin_edges=self._stored_histogram["histogram"]["bin_edges"],
+                    suggested_bin_count=10,
+                )
+                new_self_histogram["bin_counts"] = histogram["bin_counts"]
+                new_self_histogram["bin_edges"] = histogram["bin_edges"]
+
```
JGSweets, line 395: I think there's are only set if _stored_histogram doesn't have len `10`

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -171,9 +171,10 @@ def test_diff_primitive_compilers(self):
                 "stddev": 3.285085839971525,
                 "t-test": {
                     "t-statistic": 0.4155260166386663,
-                    "conservative": {"df": 1, "p-value": 0.749287157907667},
+                    "conservative": {"df": 1.0, "p-value": 0.749287157907667},
```
JGSweets, line 174: I'm not sure this should be a float.

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -193,6 +194,7 @@ def test_diff_primitive_compilers(self):
             profile_diff["statistics"].pop("median_absolute_deviation"),
             places=2,
         )
+        self.maxDiff = None
```
JGSweets, line 197: remove

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -247,7 +249,7 @@ def test_disabling_columns_during_primitive_diff(self):
                 "sum": -20.0,
                 "mean": -10.0,
                 "median": -10,
-                "mode": [[-2, -1, 1, 2], [], [5, 15]],
+                "mode": [[-2.0, -1.0, 1.0, 2.0], [], [5, 15]],
```
JGSweets, line 252: good call

dataprofiler/tests/profilers/test_column_profile_compilers.py
```diff
@@ -262,9 +264,10 @@ def test_disabling_columns_during_primitive_diff(self):
                 },
                 "t-test": {
                     "t-statistic": -1.9674775073518591,
-                    "conservative": {"df": 1, "p-value": 0.29936264581081673},
+                    "conservative": {"df": 1.0, "p-value": 0.29936264581081673},
```
JGSweets, line 267: df should be int, right?

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1698,9 +1698,10 @@ def test_diff(self):
             },
             "t-test": {
                 "t-statistic": 0.5393164101529813,
-                "conservative": {"df": 2, "p-value": 0.643676756587475},
+                "conservative": {"df": 2.0, "p-value": 0.643676756587475},
```
JGSweets, line 1701: `int` I believe

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1724,6 +1725,7 @@ def test_diff(self):
             profile_diff.pop("median_absolute_deviation"),
             places=2,
         )
+        self.maxDiff = None
```
JGSweets, line 1728: remove

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1733,3 +1735,7 @@ def test_diff(self):
             str(exc.exception),
             "Unsupported operand type(s) for diff: 'FloatColumn' and" " 'str'",
         )
+
+
+test_diff = TestFloatColumn()
+test_diff.test_diff()
```
JGSweets, line 1741: remove

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -1053,14 +1053,15 @@ def test_diff(self):
...
             "sum": -4.0,
             "variance": -94.0,
-            "median": -4,
+            "median": -4.0,
             "mode": [[2, 6, 4], [], [1, 15]],
             "median_absolute_deviation": -5,
             "t-test": {
                 "t-statistic": -0.5638091828819275,
-                "conservative": {"df": 1, "p-value": 0.6731699660830497},
+                "conservative": {"df": 1.0, "p-value": 0.6731699660830497},
```
JGSweets, line 1061: `int`

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -607,4 +608,5 @@ def test_diff(self):
             profile_diff.pop("median_absolute_deviation"),
             places=2,
         )
+        self.maxDiff = None
```
JGSweets, line 611: remove

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: Fair -- I'll add that

dataprofiler/tests/profilers/test_float_column_profile.py
taylorfturner: oh goodness :cringe: lol

dataprofiler/tests/profilers/test_text_column_profile.py
taylorfturner: done

dataprofiler/tests/profilers/test_float_column_profile.py
taylorfturner: fixed

dataprofiler/tests/profilers/test_float_column_profile.py
taylorfturner: removed

dataprofiler/tests/profilers/test_column_profile_compilers.py
taylorfturner: removed

dataprofiler/tests/profilers/test_column_profile_compilers.py
taylorfturner: I just checked L561 in `numerical_column_stats.py` and it shows the below... So, I think we are good with `1.0`?

```
results["conservative"]["df"] = float(conservative_df)
```

dataprofiler/tests/profilers/test_float_column_profile.py
taylorfturner: I just checked L561 in `numerical_column_stats.py` and it shows the below... So, I think we are good with `1.0`?

```
results["conservative"]["df"] = float(conservative_df)
```

dataprofiler/tests/profilers/test_column_profile_compilers.py
taylorfturner: I just checked L561 in `numerical_column_stats.py` and it shows the below... So, I think we are good with `1.0`?

```
results["conservative"]["df"] = float(conservative_df)
```

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: working this

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: added

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: good call out here -- now configured to assign already existing `_stored_histogram` to the `new_self_histogram` and/or `new_other_histogram` depending on the conditions that exist.

taylorfturner: This should almost never occur because of 
```
self.max_histogram_bin: int = 100000
self.min_histogram_bin: int = 1000
```

used in `self._get_histogram()`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,68 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
 
+        self_bin_counts = self._stored_histogram["histogram"]["bin_counts"]
+        self_bin_edges = self._stored_histogram["histogram"]["bin_edges"]
+        regen_histogram = False
+        num_psi_bins = 10
+
+        if isinstance(self_bin_counts, np.ndarray) and isinstance(
+            self_bin_edges, np.ndarray
+        ):
+            regen_histogram = True
```
JGSweets, line 380: should we check both self and other here?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,68 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+                ]
+                new_self_histogram["bin_edges"] = self._stored_histogram["histogram"][
+                    "bin_edges"
+                ]
+
+            # re-calculate `other_profile` histogram
+            both_profiles_bin_edges_equal = (
+                other_profile._stored_histogram["histogram"]["bin_edges"]
+                == self._stored_histogram["histogram"]["bin_edges"]
+            ).all() == 0
```
JGSweets, line 412: at the end of this you have `== 0` but the variable name suggests it is checking bin edges are all equal. I think you could either fix the variable name or the  `== 0` plus the usage below.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -516,6 +586,48 @@ def _perform_t_test(
...
+    def _calculate_psi(
+        mean1: float,
+        sum1: float,
+        histogram1: np.ndarray,
+        mean2: float,
+        sum2: float,
+        histogram2: np.ndarray,
+    ) -> Optional[float]:
+        self_count = sum1 / mean1
+        other_count = sum2 / mean2
```
JGSweets, line 599: If the intent is to get the count, we could just sum the bin_counts instead. otherwise we might lose floating point accuracies, no? Also:
```python
def mean(self) -> float:
        """Return mean value."""
        if self.match_count == 0:
            return 0
        return float(self.sum) / self.match_count
```
so could just use `match_count` as well.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -516,6 +586,48 @@ def _perform_t_test(
...
 
+    @staticmethod
+    def _calculate_psi(
+        mean1: float,
+        sum1: float,
+        histogram1: np.ndarray,
+        mean2: float,
+        sum2: float,
+        histogram2: np.ndarray,
+    ) -> Optional[float]:
```
JGSweets, line 597: docstring description

JGSweets: can link calculation too

taylorfturner: totally fair -- addding

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: Perf! I'll use `match_count` -- less code and reuse what already exists

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: once that is opened up to the user, for sure

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: this was how I resolved flake8 complaining... let me see what I can do to clean this up

taylorfturner: cleaned up a bit more

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,93 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+                other_profile._stored_histogram["histogram"]["bin_edges"][0],
+            )
+            max_max_edge = max(
+                self._stored_histogram["histogram"]["bin_edges"][-1],
+                other_profile._stored_histogram["histogram"]["bin_edges"][-1],
+            )
+
+        if regenerate_histogram:
+            len_self_bin_counts = 0
+            if len(self._stored_histogram["histogram"]["bin_counts"]) > 0:
```
JGSweets, line 399: Is this if statement needed?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,93 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+                    suggested_bin_count=num_psi_bins,
+                    options={
+                        "min_edge": min_min_edge,
+                        "max_edge": max_max_edge,
+                    },
+                )
+                new_self_histogram["bin_counts"] = histogram["bin_counts"]
+                new_self_histogram["bin_edges"] = histogram["bin_edges"]
+
+            if len_self_bin_counts == num_psi_bins:
```
JGSweets, line 418: Instead of this if, could set it above the first if, and it will get reassigned given the criteria

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,93 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
...
+                    options={
+                        "min_edge": min_min_edge,
+                        "max_edge": max_max_edge,
+                    },
+                )
+
+                new_other_histogram["bin_edges"] = histogram["bin_edges"]
+                new_other_histogram["bin_counts"] = histogram["bin_counts"]
+
+            if not histogram_edges_not_equal:
```
JGSweets, line 451: As above

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -369,6 +369,93 @@ def diff(self, other_profile: NumericStatsMixin, options: Dict = None) -> Dict:
                 "and '{}'".format(cls.__name__, other_profile.__class__.__name__)
             )
 
+        new_self_histogram = {"bin_counts": None, "bin_edges": None}
```
JGSweets, line 372: I think this is fine, but one thing I would consider is the rest of the api functions in diff are purely padding the data with all the checks within.

To keep things similar, unless the histograms were used elsewhere, I'd consider putting all of this in calculate_psi

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: totally fair -- updating now

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: totally fair -- updating now

dataprofiler/profilers/numerical_column_stats.py
JGSweets: This can reduce 3 ifs down to one

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: nah that's good -- lemme see what I can do right quick and push up

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -516,6 +609,66 @@ def _perform_t_test(
...
+        :return: psi_value
+        :rtype: optional[float]
+        """
+        psi_value = 0
+        invalid_stats = False
+
+        if isinstance(histogram2["bin_edges"], type(None)) or isinstance(
+            histogram1["bin_edges"], type(None)
+        ):
+            warnings.warn(
```
JGSweets, line 648: I think we use logger.warn in the repo? So that a user can shut off warnings? I cannot recall.

taylorfturner: I was basing this off `_perform_t_test`, but lemme double check

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -516,6 +609,66 @@ def _perform_t_test(
...
+                "No edges available in at least one histogram for calculating `PSI`",
+                RuntimeWarning,
+            )
+            invalid_stats = True
+
+        if invalid_stats:
```
JGSweets, line 654: Instead of an extra if could we put it inside the previous?

taylorfturner: lol true 🤦

taylorfturner: fixed in next push

dataprofiler/tests/profilers/test_column_profile_compilers.py
JGSweets: Hmm nvm then!

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: definitely

dataprofiler/profilers/numerical_column_stats.py
JGSweets: You are correct! My mistake, thanks!

JGSweets (approved):

ksneab7 (approved):

PR #700, opened 2022-10-25
Added a few missing types for data_readers

Sanketh7:
#610 
There were a few type annotations that were missing.


taylorfturner:
Good validation; although, we won't be using the below flags for validation of `mypy`, though. Will close but there is probably need for additional edits that this PR has shown. 

```
--disallow-untyped-defs
--disallow-incomplete-defs
```

PR #703, opened 2022-11-01
Updated setup.cfg mypy flags and resolved related errors.

Sanketh7:
Added the following flags to `setup.cfg`. Run `pre-commit install` then `pre-commit run -a` to test these changes locally.
```
[mypy]
warn_return_any = True
warn_unused_configs = True
ignore_missing_imports = True
```

All of the warnings that needed to be resolved were because of `warn_return_any`. Some of my solutions are more elegant than others so let me know if you find a better way.


dataprofiler/profilers/base_column_profilers.py
```diff
@@ -267,6 +267,7 @@ def __init__(self, name: Optional[str]) -> None:
         # Number of values that match the column type. eg. how many floats match
         # in the float column
         self.match_count: int = 0
+        self.sample_size: int  # inherited from BaseColumnProfiler
```
Sanketh7, line 270: `self.sample_size` wasn't being detected by mypy to be inherited from `BaseColumnProfiler`. This was causing problems downstream where mypy didn't have any type information on `self.sample_size`. I suspect the `with_metaclass` line above is causing this issue.

dataprofiler/profilers/graph_profiler.py
```diff
@@ -330,12 +330,12 @@ def _update_categorical_distribution(
     @BaseColumnProfiler._timeit(name="num_nodes")
     def _get_num_nodes(self, graph: nx.Graph) -> int:
         """Compute the number of nodes."""
-        return graph.number_of_nodes()
+        return cast(int, graph.number_of_nodes())
```
Sanketh7, line 333: As far as I could tell, the `networkx` module doesn't have type annotations for graphs.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -422,7 +422,7 @@ def median(self) -> float:
         :rtype: float
         """
         if not self._has_histogram or not self._median_is_enabled:
-            return np.nan
+            return cast(float, np.nan)
```
Sanketh7, line 425: I couldn't figure out why `np.nan` wasn't being typed as `float` instead of `Any`. I thought maybe mypy wasn't using numpy's type stubs but apparently type stubs come by default in recent versions of numpy (like the one in our `requirements.txt`). 

Maybe this is caused by my local setup or the project's setup so if someone could confirm if reverting this line causes a pre-commit error, that would be great.

taylorfturner:
@Sanketh7 `update branch` on this to make sure tests still passing

dataprofiler/data_readers/json_data.py
```diff
@@ -236,36 +237,38 @@ def _get_data_as_flattened_dataframe(self, json_lines):
...
+        data: JSONType
         try:
             data = json.loads(data_as_str)
         except json.JSONDecodeError:
-            data = data_utils.data_generator(data_as_str.splitlines())
+            data_generator = data_utils.data_generator(data_as_str.splitlines())
             data = data_utils.read_json(
-                data_generator=data,
+                data_generator=data_generator,
                 selected_columns=self.selected_keys,
```
taylorfturner, line 255: was this raised by `mypy`? Pretty sure the idea was to overwrite variable names to keep space to a minimum, but could be wrong here.

dataprofiler/data_readers/json_data.py
```diff
@@ -236,36 +237,38 @@ def _get_data_as_flattened_dataframe(self, json_lines):
...
                 read_in_string=False,
             )
         return data
 
-    def _load_data_from_file(self, input_file_path: str) -> List:
+    def _load_data_from_file(self, input_file_path: str) -> JSONType:
```
taylorfturner, line 260: `List` to `JSONType` which is Union[str, int, float, bool, None, List, Dict]? this is a big change...

dataprofiler/data_readers/structured_mixins.py
```diff
@@ -72,12 +72,12 @@ def _get_data_as_df(self, data: pd.DataFrame) -> pd.DataFrame:
...
+        data_ = [
             str(
                 "\n".join(data[i * records_per_line : (i + 1) * records_per_line])
                 .encode("UTF-8")
                 .decode()
             )
             for i in range((len(data) + records_per_line - 1) // records_per_line)
         ]
-        return data
+        return data_
```
taylorfturner, line 83: is `data` conflicting with a python keyword here?

dataprofiler/data_readers/avro_data.py
```diff
@@ -94,7 +94,7 @@ def is_match(
         if data_utils.is_stream_buffer(file_path):
             starting_location = file_path.tell()
 
-        is_valid_avro = fastavro.is_avro(file_path)
+        is_valid_avro: bool = fastavro.is_avro(file_path)
```
taylorfturner, line 97: nice

dataprofiler/profilers/histogram_utils.py
```diff
@@ -7,11 +7,8 @@
 from typing import List, Optional, Tuple, Union
 
 import numpy as np
-from numpy.lib.histograms import (  # type: ignore
-    _get_outer_edges,
-    _hist_bin_selectors,
-    _unsigned_subtract,
-)
+from numpy.lib.histograms import _get_outer_edges  # type: ignore
+from numpy.lib.histograms import _hist_bin_selectors, _unsigned_subtract
```
taylorfturner, line 11: Is this an `isort` hook change?

taylorfturner (commented):
just questions... I'll also test for if `np.nan` does get seen as a `float` instead of `Any` on my end. 

Need to review more too all the `cast()` statements

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: recreated the error on my end when taking out the `cast(float...)`

```
dataprofiler/profilers/numerical_column_stats.py:425: error: Returning Any from function declared to return "float"
Found 1 error in 1 file (checked 55 source files)
```

taylorfturner:
@Sanketh7 `update branch` FYI

dataprofiler/data_readers/json_data.py
```diff
@@ -236,36 +237,38 @@ def _get_data_as_flattened_dataframe(self, json_lines):
 
         return data
 
-    def _load_data_from_str(self, data_as_str: str) -> List:
+    def _load_data_from_str(self, data_as_str: str) -> JSONType:
```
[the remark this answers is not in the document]
Sanketh7: I figured changing it to `JSONType` would be safer because the return value could be from `json.loads` which could be any json.

dataprofiler/data_readers/json_data.py
Sanketh7: When I explicitly gave a type for `data`, mypy had a issue with this line because `data_generator` returns a generator type. We eventually return `data` so I didn't want to make its type more confusing so I used a different variable name.

dataprofiler/data_readers/structured_mixins.py
Sanketh7: The parameter name is also `data` so I had to change the name because we're returning a different type.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -285,7 +285,7 @@ def _get_float_precision(
         return subset_precision
 
     @classmethod
-    def _is_each_row_float(cls, df_series: pd.Series) -> List[bool]:
+    def _is_each_row_float(cls, df_series: pd.Series) -> pd.Series[bool]:
```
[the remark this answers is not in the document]
Sanketh7: Good catch. I'll update the return type.

dataprofiler/data_readers/json_data.py
Sanketh7: We assign `data` (which is what's returned) potentially to the output of `json.load` which I figured could be any json. Before `json.load` was being treated as `Any` but the new mypy flags don't allow us to return `Any`.

dataprofiler/profilers/histogram_utils.py
Sanketh7: yes

dataprofiler/profilers/graph_profiler.py
JGSweets: fair

dataprofiler/profilers/numerical_column_stats.py
JGSweets: Should we update the return type then instead of casting? Numpy does typing well, how do they handle it?

dataprofiler/profilers/base_column_profilers.py
JGSweets, line 270: nice

dataprofiler/data_readers/json_data.py
taylorfturner: I see

dataprofiler/profilers/float_column_profile.py
```diff
@@ -297,7 +297,7 @@ def _is_each_row_float(cls, df_series: pd.Series) -> List[bool]:
         :param df_series: series of values to evaluate
         :type df_series: pandas.core.series.Series
         :return: is_float_col
-        :rtype: list
+        :rtype: pandas.Series[bool]
```
taylorfturner, line 300: need to update this too based on type hint on output @Sanketh7

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -463,7 +463,7 @@ def _load_parameters(dirpath: str, load_options: Dict = None) -> Dict[str, Dict]
             load_options = {}
 
         with open(os.path.join(dirpath, "data_labeler_parameters.json")) as fp:
-            params = json.load(fp)
+            params: Dict[str, Dict] = json.load(fp)
```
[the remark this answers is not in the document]
taylorfturner: should this also be `JSONType`?

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Turns out the reason for this was that we needed to add numpy as a dependency to the mypy pre-commit hook. This gave mypy the necessary typing information to fix this issue. I just made a commit that fixes this and removes some extraneous casting. [0ad2a38](https://github.com/capitalone/DataProfiler/pull/703/commits/0ad2a38a34143307d51a127ccd472d387f46864d)

dataprofiler/labelers/base_data_labeler.py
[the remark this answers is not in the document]
Sanketh7: The code treats `params` as a `Dict[str, Dict]` so I figured it's better to make that explicit in the typing.

dataprofiler/profilers/float_column_profile.py
Sanketh7: Done.

dataprofiler/profilers/float_column_profile.py
```diff
@@ -285,7 +285,7 @@ def _get_float_precision(
         return subset_precision
 
     @classmethod
-    def _is_each_row_float(cls, df_series: pd.Series) -> List[bool]:
+    def _is_each_row_float(cls, df_series: pd.Series) -> Union[List, pd.Series[bool]]:
```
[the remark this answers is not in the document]
Sanketh7: Makes sense. Pushed the change.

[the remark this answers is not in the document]
taylorfturner: awesome

dataprofiler/profilers/float_column_profile.py
taylorfturner: thanks

dataprofiler/data_readers/structured_mixins.py
taylorfturner: +1 CC @Sanketh7

dataprofiler/profilers/float_column_profile.py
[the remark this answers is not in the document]
taylorfturner: not seeing the change @Sanketh7

dataprofiler/profilers/float_column_profile.py
taylorfturner: not seeing this change ... @Sanketh7

taylorfturner (commented):
comments

dataprofiler/profilers/float_column_profile.py
Sanketh7: oops i forgot to push it

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: does this remove any other need for `cast()` in the other parts of the PR?

dataprofiler/data_readers/structured_mixins.py
Sanketh7: To get around the changing types, I decided to cast `data` before returning it.

dataprofiler/profilers/base_column_profilers.py
taylorfturner: awesome

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Potentially, though I didn't see any other areas we could've removed it. A decent amount of casting is just due to the numpy's methods' return types being ambiguous when in reality we know what the type is.

dataprofiler/labelers/data_processing.py
```diff
@@ -1416,15 +1417,15 @@ def get_parameters(self, param_list: List[str] = None) -> Dict:
...
+        self, data: np.ndarray, labels: Optional[Union[List[str], npt.NDArray[np.str_]]]
     ) -> Tuple[str, Optional[List[Tuple[int, int, str]]]]:
         """
         Convert data samples list to StructCharPreprocessor required input data format.
 
         :param data: list of strings
         :type data: numpy.ndarray
         :param labels: labels for each input character
-        :type labels: list
+        :type labels: Union[list, numpy.ndarray]
```
taylorfturner, line 1428: potentially we should include the `optional` and `str` here too, @Sanketh7 ?

`Optional[Union[List[str], npt.NDArray[np.str_]]]`

dataprofiler/labelers/data_processing.py
```diff
@@ -1507,7 +1508,7 @@ def process(  # type: ignore
         # with rework, can be tuned to be batches > size 1
         for ind in range(len(data)):
             batch_data: np.ndarray = data[ind : ind + 1]
-            batch_labels: Optional[List[str]] = (
+            batch_labels: Optional[npt.NDArray[np.str_]] = (
```
taylorfturner, line 1511: why is this only `Optional[npt.NDArray[np.str_]]`, @Sanketh7 ? 

It appears that `self. convert_to_unstructured_format` takes in the `batch_labels` parameter as `list` or `npt.ndarray`? 

https://github.com/Sanketh7/DataProfiler/blob/mypy-flags/dataprofiler/labelers/data_processing.py#L1420

dataprofiler/labelers/character_level_cnn_model.py
```diff
@@ -58,7 +58,7 @@ def create_glove_char(n_dims: int, source_file: str = None) -> None:
     embd_table = build_embd_dictionary(source_file)
     embd_words: List[str]
     embd_matrix: List[np.ndarray]
-    embd_words, embd_matrix = [
+    embd_words, embd_matrix = [  # type: ignore
```
taylorfturner, line 61: don't love the `#type: ignore`. If there isn't a way to avoid this, that's okay. But IIRC, in prior PRs we are tried to avoid this, @Sanketh7?

dataprofiler/labelers/data_processing.py
```diff
@@ -1416,15 +1417,15 @@ def get_parameters(self, param_list: List[str] = None) -> Dict:
...
     ) -> Tuple[str, Optional[List[Tuple[int, int, str]]]]:
         """
         Convert data samples list to StructCharPreprocessor required input data format.
 
         :param data: list of strings
         :type data: numpy.ndarray
         :param labels: labels for each input character
-        :type labels: list
+        :type labels: Union[list, numpy.ndarray]
         :return: data in the following format
```
taylorfturner, line 1429: can we add a `:rtype: Tuple[str, Optional[List[Tuple[int, int, str]]]]` here too, @Sanketh7 ?

dataprofiler/labelers/data_processing.py
taylorfturner: updated

dataprofiler/labelers/data_processing.py
taylorfturner: updated

dataprofiler/labelers/data_processing.py
taylorfturner: updated

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: gotcha -- we should probably create a ticket around cleaning these up in the future IMO

dataprofiler/labelers/character_level_cnn_model.py
Sanketh7: The issue I was having was that there's just a lot of types involved and numpy's type annotations aren't very specific to help mypy out. Mypy sort of gives up and says that you're trying to assign an `object` to `List[str]` and `List[np.ndarray]`.

taylorfturner: i see

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -914,4 +914,4 @@ def load_with_components(
         data_labeler.set_preprocessor(preprocessor)
         data_labeler.set_model(model)
         data_labeler.set_postprocessor(postprocessor)
-        return data_labeler
+        return cast(TrainableDataLabeler, data_labeler)
```
JGSweets, line 917: Hmm, should we allow for `TrainableDataLabeler` or `CustomTrainableDataLabeler`?

JGSweets: or does that cause downstream issues?

dataprofiler/labelers/base_model.py
```diff
@@ -21,7 +21,7 @@ def __new__(
             cls, clsname, bases, attrs
         )
         new_class._register_subclass()
-        return new_class
+        return cast(AutoSubRegistrationMeta, new_class)
```
JGSweets, line 24: This might be better to do a `T` generic because really the class should be original of the input? Otherwise, would this convert something using this metaclass to that type later in code?

Is that why loading has the issue as well?

dataprofiler/labelers/labeler_utils.py
```diff
@@ -79,7 +79,7 @@ def evaluate_accuracy(
     predicted_entities_in_index: List[List[int]],
     true_entities_in_index: List[List[int]],
     num_labels: int,
-    entity_rev_dict: Dict,
+    entity_rev_dict: Dict[int, Any],
```
JGSweets, line 82: I believe the `Any` is required to be `str`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -405,7 +405,7 @@ def mean(self) -> float:
         """Return mean value."""
         if self.match_count == 0:
             return 0
-        return float(self.sum) / self.match_count
+        return float(self.sum) / float(self.match_count)
```
JGSweets, line 408: mypy compains if self.match_count isn't a float?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -445,7 +445,7 @@ def stddev(self) -> float:
         """Return stddev value."""
         if self.match_count == 0:
             return np.nan
-        return np.sqrt(self.variance)
+        return cast(float, np.sqrt(self.variance))
```
JGSweets, line 448: I think we convert later to a float bc this could be an np.float64. Probably good to say this could be np.float so that we are returning the real type.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -789,7 +789,7 @@ def _merge_biased_skewness(
         third_term = 3 * delta * (match_count1 * M2_2 - match_count2 * M2_1) / N
         M3 = first_term + second_term + third_term
 
-        biased_skewness = np.sqrt(N) * M3 / np.sqrt(M2**3)
+        biased_skewness: float = np.sqrt(N) * M3 / np.sqrt(M2**3)
```
JGSweets, line 792: is this a float or is it a np.float64?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -811,7 +811,7 @@ def _correct_bias_skewness(match_count: int, biased_skewness: float) -> float:
             )
             return np.nan
 
-        skewness = (
+        skewness: float = (
```
JGSweets, line 814: as above

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -886,7 +886,7 @@ def _merge_biased_kurtosis(
         fourth_term = 4 * delta * (match_count1 * M3_2 - match_count2 * M3_1) / N
         M4 = first_term + second_term + third_term + fourth_term
 
-        biased_kurtosis = N * M4 / M2**2 - 3
+        biased_kurtosis: float = N * M4 / M2**2 - 3
```
JGSweets, line 889: as above

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -947,15 +947,15 @@ def _estimate_mode_from_histogram(self) -> List[float]:
         mode = (
             bin_edges[highest_idxs] + bin_edges[highest_idxs + 1]  # type: ignore
         ) / 2
-        return mode.tolist()
+        return cast(List[float], mode.tolist())
```
JGSweets, line 950: is this a list of float or np.float64?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -947,15 +947,15 @@ def _estimate_mode_from_histogram(self) -> List[float]:
...
 
     def _estimate_stats_from_histogram(self) -> float:
         # test estimated mean and var
         bin_counts = self._stored_histogram["histogram"]["bin_counts"]
         bin_edges = self._stored_histogram["histogram"]["bin_edges"]
         mids = 0.5 * (bin_edges[1:] + bin_edges[:-1])
         mean = np.average(mids, weights=bin_counts)
-        var = np.average((mids - mean) ** 2, weights=bin_counts)
+        var: float = np.average((mids - mean) ** 2, weights=bin_counts)
```
JGSweets, line 958: as above

dataprofiler/profilers/utils.py
```diff
@@ -332,7 +332,7 @@ def biased_skew(df_series: Series) -> float:
     """
     n = len(df_series)
     if n < 1:
-        return np.nan
+        return float(np.nan)
```
JGSweets, line 335: np.nan is specific, I think we should maintain the np.nan type if possible. is np.nan not accepted or do we need to fix to allow?

dataprofiler/validators/base_validators.py
```diff
@@ -20,7 +20,7 @@ def is_in_range(x: Union[float, int], config: dict) -> bool:
     :returns: bool
     """
     try:
-        return config["start"] <= float(x) <= config["end"]
+        return float(config["start"]) <= float(x) <= float(config["end"])
```
JGSweets, line 23: is this conversion needed? interesting

Sanketh7: I ended up needing to do this because mypy doesn't know if `config["start"]` and `config["end"]` actually exist so it assigns them the type `Any`. Expressions involving `Any` evaluate to type `Any` which we're not allowed to return due to our new mypy flags. 

I could also use `cast` to prevent the use of a runtime cast but I figured since it's already in a try except block, any issues would be handled the same way as the cast `float(x)`.

dataprofiler/labelers/labeler_utils.py
Sanketh7: Makes sense. Made the change.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Actually, now that I look at the code again, I don't see `self.match_count` getting initialized anywhere. This would explain why it's being typed as `Any` (thus requiring some sort of cast).

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: I see. This would still require a `cast` though because numpy's type annotations for `sqrt` on a scalar has a return type of `Any`.

dataprofiler/profilers/utils.py
Sanketh7: Oh this cast isn't needed once I added numpy as a dependency in the mypy pre-commit hook. I'll remove it now.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: I'm trying to cast it to `np.float64` but the issue becomes that there's not really a compatible return type annotation because `np.nan` is of type `float`. Using something like `np.float_` causes mypy to complain about the `np.nan` case and sticking with `float` causes mypy to complain about the `np.sqrt` case.

Sanketh7: Changing it to a Union causes issues with line 385:
```
            "stddev": utils.find_diff_of_numbers(self.stddev, other_profile.stddev),
```
```
dataprofiler/profilers/numerical_column_stats.py:385: error: Value of type variable "T" of "find_diff_of_numbers" cannot be "object"
```

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: I fixed it by manually adding a type for it in the constructor. I assume `self.match_count` is populated in some dynamic way so mypy isn't able to pick up its type automatically.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: `bin_edges` doesn't necessarily need to contain numpy float values.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: This will end up being a np.float64. I changed the type annotation and method return type to reflect this.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -123,6 +123,7 @@ def __init__(self, options: NumericalOptions = None) -> None:
             "num_zeros": NumericStatsMixin._get_num_zeros,
             "num_negatives": NumericStatsMixin._get_num_negatives,
         }
+        self.match_count: int  # needed for mypy
```
taylorfturner, line 126: nice!

dataprofiler/labelers/base_model.py
Sanketh7: The issue is that `new_class` needs to be typed as `Any` initially because otherwise mypy complains that `._register_subclass()` doesn't exist. I'm thinking we could also do:
```py
new_class: AutoSubRegistrationMeta = super(AutoSubRegistrationMeta, cls).__new__(
    cls, clsname, bases, attrs
)
new_class._register_subclass() # type: ignore
return new_class
```
which does pass mypy.

dataprofiler/labelers/base_data_labeler.py
Sanketh7: issue is that the `CustomTrainableDataLabeler` type only exists locally so it wouldn't be a valid return type.

dataprofiler/labelers/base_model.py
JGSweets: I agree that`new_class: AutoSubRegistrationMeta = ...` would not work either. However, I was wondering if there's a generic approach that could fix this that allows us to do `new_class: T = ...` or some sort where `T` is the `cls` or something of the like. If it is too complex, fair. Either way, it doesn't feel right to me that `cast(AutoSubRegistrationMeta, new_class)` shouldn't be some `cast(T, new_class)` as the output is not a class `AutoSubRegistrationMeta`, right?

dataprofiler/profilers/numerical_column_stats.py
JGSweets: So `def stddev(self) -> Union[float, np.float64]:`
with the `cast(np.float64, np.sqrt(self.variance))` doesn't work?

dataprofiler/profilers/numerical_column_stats.py
JGSweets: @Sanketh7

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1492,19 +1493,29 @@ def median_abs_deviation(self) -> float:
...
+                bin_edges_neg,
+                np.cumsum(np.append([0], bin_counts_neg)),
+            ),
         )
         bin_counts_impose = bin_counts_impose_pos + bin_counts_impose_neg
 
         median_inds = np.abs(bin_counts_impose - 0.5) < 1e-10
         if np.sum(median_inds) > 1:
-            return np.mean(bin_edges_impose[median_inds])
+            return cast(float, np.mean(bin_edges_impose[median_inds]))
```
JGSweets, line 1516: I think these would also be np.float64

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Yes. It seems to cause issues with other functions.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: This does seem to be a np.float64. Because biased_skewness could be either `np.nan` or some `np.float64`, I've updated any usage of biased skewness to be a `Union[float, np.float64]`.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Updated `_correct_bias_skewness` to return `Union[float, np.float64]`

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Updated `_merged_biased_kurtosis` return return `Union[float, np.float64]`

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Casting this to `np.float64` and changing the return type to `Union[float, np.float64]` gives me a similar issue as one of the other methods:
```
"median_absolute_deviation": utils.find_diff_of_numbers(
                self.median_abs_deviation, other_profile.median_abs_deviation
            ),
```
```
dataprofiler/profilers/numerical_column_stats.py:381: error: Value of type variable "T" of "find_diff_of_numbers" cannot be "object"
```

dataprofiler/labelers/base_model.py
Sanketh7:
```py
T = TypeVar("T", bound="AutoSubRegistrationMeta")


class AutoSubRegistrationMeta(abc.ABCMeta):
    """For registering subclasses."""

    def __new__(
        cls: Type[T], clsname: str, bases: Tuple[type, ...], attrs: Dict[str, object]
    ) -> T:
        """Create auto registration object and return new class."""
        new_class: T = super(AutoSubRegistrationMeta, cls).__new__(
            cls, clsname, bases, attrs
        )
        new_class._register_subclass()  # type: ignore
        return new_class
```
Was able to get it working with this change (which I just committed). Couldn't get around the `._register_subclass()` line because that's done dynamically.

dataprofiler/profilers/numerical_column_stats.py
JGSweets: The issue is if we are prescribing it to contain something that isn't true. I'd rather say it could contains either if that is true because otherwise if someone is relying on static typing to ensure functionality of a function, this could be cause them to infer incorrectly.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1492,19 +1497,29 @@ def median_abs_deviation(self) -> float:
             np.append([True], np.diff(bin_edges_impose) > 1e-14)
         ]
 
-        bin_counts_impose_pos: float = np.interp(
-            bin_edges_impose, bin_edges_pos, np.cumsum(np.append([0], bin_counts_pos))
+        bin_counts_impose_pos: float = cast(
```
JGSweets, line 1500: I did a check on this. These are not actually float. That's part of the reason this is erroring and causing bugs elsewhere. This is actually returning an NDArray of np.float64

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1492,19 +1497,29 @@ def median_abs_deviation(self) -> float:
...
+            float,
+            np.interp(
+                bin_edges_impose,
+                bin_edges_pos,
+                np.cumsum(np.append([0], bin_counts_pos)),
+            ),
         )
-        bin_counts_impose_neg: float = np.interp(
-            bin_edges_impose, bin_edges_neg, np.cumsum(np.append([0], bin_counts_neg))
+        bin_counts_impose_neg: float = cast(
```
JGSweets, line 1508: as above

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1492,19 +1497,29 @@ def median_abs_deviation(self) -> float:
...
+                bin_edges_neg,
+                np.cumsum(np.append([0], bin_counts_neg)),
+            ),
         )
         bin_counts_impose = bin_counts_impose_pos + bin_counts_impose_neg
 
         median_inds = np.abs(bin_counts_impose - 0.5) < 1e-10
         if np.sum(median_inds) > 1:
-            return np.mean(bin_edges_impose[median_inds])
+            return cast(float, np.mean(bin_edges_impose[median_inds]))
```
JGSweets, line 1520: this would be a float64

Sanketh7: I'm getting a similar issue as before:
```
"median_absolute_deviation": utils.find_diff_of_numbers(
         self.median_abs_deviation, other_profile.median_abs_deviation
 ),
```
```
dataprofiler/profilers/numerical_column_stats.py:381: error: Value of type variable "T" of "find_diff_of_numbers" cannot be "object"
```
Looking into it more, I believe it's because we're trying to assign a `Union` to a `TypeVar` with `bound=Subtractable` and the `Union` itself does not follow the `Subtractable` protocol. So far the only reasonable solution I can think of is doing:
```
"median_absolute_deviation": utils.find_diff_of_numbers(
         cast(float, self.median_abs_deviation), cast(float, other_profile.median_abs_deviation)
 ),
```
which isn't very ideal. I'll look into mypy generics more to see if there's a workaround.

dataprofiler/labelers/base_model.py
```diff
@@ -5,22 +5,24 @@
...
-    ) -> AutoSubRegistrationMeta:
+        cls: Type[T], clsname: str, bases: Tuple[type, ...], attrs: Dict[str, object]
+    ) -> T:
         """Create auto registration object and return new class."""
-        new_class: Any = super(AutoSubRegistrationMeta, cls).__new__(
+        new_class: T = super(AutoSubRegistrationMeta, cls).__new__(
             cls, clsname, bases, attrs
         )
-        new_class._register_subclass()
+        new_class._register_subclass()  # type: ignore
```
JGSweets, line 25: I'm not 100% positive, but I think the bound above should be `BaseModel` which would fix this.

We should really try to avoid using these bandaid fixes with `#type: ignore` unless absolutely necessary

dataprofiler/profilers/numerical_column_stats.py
JGSweets: I would assume `def stddev(self) -> Union[float, np.float64]:` even w/o the cast should be functional.

but if this is erroring elsewhere, then we should fix that once we do this correctly.

dataprofiler/profilers/numerical_column_stats.py
JGSweets: Did google provide any response to this?

JGSweets: so it has something to do with the return of these numpy funcs being any maybe?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1842,7 +1857,7 @@ def is_int(x: str) -> bool:
             return a == b
 
     @staticmethod
-    def np_type_to_type(val: Any) -> Union[int, float]:
+    def np_type_to_type(val: Any) -> Union[int, float, Any]:
```
JGSweets, line 1860: if this can be any, we technically don't need union, right?

JGSweets: However, will this error then bc we are using Any?

Sanketh7: I guess the Union was redundant. Changing to Any actually doesn't cause issues with our mypy flag `warn_return_any = True` because that only complains when we return Any and the function return type is not Any.

dataprofiler/profilers/utils.py
```diff
@@ -412,7 +413,19 @@ def __sub__(self: T, other: T) -> Any:
 T = TypeVar("T", bound=Subtractable)
 
 
+@overload
+def find_diff_of_numbers(
```
Sanketh7, line 417: Ended up going with an overload here. The issue is that there are a couple cases where we pass in a `Union[float, np.float64]` to `find_diff_of_numbers` and Unions don't satisfy the typevar bound.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Done.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Done.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Should be fixed. See https://github.com/capitalone/DataProfiler/pull/703#discussion_r1015985771

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Done

dataprofiler/labelers/base_model.py
taylorfturner: @Sanketh7 can we get rid of this `#type: ignore`?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -929,7 +936,7 @@ def _correct_bias_kurtosis(match_count: int, biased_kurtosis: float) -> float:
         )
         return kurtosis
 
-    def _estimate_mode_from_histogram(self) -> List[float]:
+    def _estimate_mode_from_histogram(self) -> Union[List[float], List[np.float64]]:
```
taylorfturner, line 939: so `mypy` still complains on the return even with this `return` typing? still need to `cast`, @Sanketh7 ?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1453,7 +1460,7 @@ def _fold_histogram(
         return [bin_counts_pos, bin_edges_pos], [bin_counts_neg, bin_edges_neg]
 
     @property
-    def median_abs_deviation(self) -> float:
+    def median_abs_deviation(self) -> Union[float, np.float64]:
```
taylorfturner, line 1463: is the return type correct here given the possible `return` values in the function (i.e. `return np.nan`, `return 0` (int),  etc..)?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1506,19 +1513,25 @@ def median_abs_deviation(self) -> float:
...
         )
-        bin_counts_impose = bin_counts_impose_pos + bin_counts_impose_neg
 
         median_inds = np.abs(bin_counts_impose - 0.5) < 1e-10
         if np.sum(median_inds) > 1:
-            return np.mean(bin_edges_impose[median_inds])
+            return cast(np.float64, np.mean(bin_edges_impose[median_inds]))
 
-        return np.interp(0.5, bin_counts_impose, bin_edges_impose)
+        return cast(np.float64, np.interp(0.5, bin_counts_impose, bin_edges_impose))
```
taylorfturner, line 1534: how could we avoid these `casts`?

dataprofiler/profilers/utils.py
taylorfturner: nice -- I would recommend formatting slightly different based on the docs I was seeing [here](https://docs.python.org/3/library/typing.html#typing.overload), but nice improvement

taylorfturner: actually, disregard this. I'm seeing it both ways

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: I did some more investigating and numpy functions end up converting the values to `float64` so this should only be `List[np.float64]`

Sanketh7: Actually it would be `List[float]` because `.tolist()` converts values to the compatible python type.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Casting is needed because mypy's type annotation for `.tolist()` gives it a return type of `Any`.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: So far I haven't found a way to remove these casts. For `np.interp`, numpy doesn't have useful type annotations for the scalar case. For `np.mean`, mypy's type annotation has a lot of overloads but most of them are typed with a return type of `Any` and it seems like our case always falls under a return `Any` case.

dataprofiler/profilers/numerical_column_stats.py
JGSweets: We could just do `return 0.0` if we wanted. But I think for that, we should be fine.

dataprofiler/profilers/utils.py
JGSweets: Hmm, an object shouldn't be casted as `Union[float, np.float64]` but has the option to return as one or the other.

I wonder if this meant we did those returns wrong and should have done overloads on those funcs where it could return a float in one or np.float64 in the other.

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: true `0.0` should cover it. `np.nan` is `float64`

taylorfturner: @Sanketh7

dataprofiler/profilers/utils.py
Sanketh7: The main reason we have the `float` case is because `np.nan` is `float` and some of our methods could return `np.nan` if certain conditions aren't met. From what I can see, a lot of these conditions can't be solved by typing overloads (for example, some are based on class fields and not method parameters). 

At least for finding the diff between 2 `Union[float, np.float64]`, we shouldn't run into issues because `float` gets promoted to `float64` automatically.

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Changed it to `return 0.0`

dataprofiler/profilers/numerical_column_stats.py
Sanketh7: Fixed. See https://github.com/capitalone/DataProfiler/pull/703#discussion_r1015985771

dataprofiler/profilers/utils.py
JGSweets: Seems to be this problem:
https://stackoverflow.com/questions/73011336/different-behavior-of-mypy-with-generic-class-in-python-3-8-and-3-10

dataprofiler/labelers/base_model.py
Sanketh7: Changing the typevar to `BaseModel` didn't end up fixing it because mypy doesn't detect the relationship between `BaseModel` and `AutoSubRegistrationMeta`.

JGSweets: This has some interesting suggestions:
https://stackoverflow.com/questions/66121127/calling-new-on-an-any

However, if we can't do this, ultimately we should cast referring to the output as a derivation of BaseModel of some type. Returning it as `AutoSubRegistrationMeta` is not truly what matter since it is just a mixin.

dataprofiler/profilers/utils.py
taylorfturner: great

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: great

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: nice

dataprofiler/labelers/base_model.py
JGSweets: For now, @taylorfturner maybe we need to do an ignore, but we should validate that when using a model elsewhere in code, it is not consider a AutoReg by type

taylorfturner: Yeah, I lean toward just making sure elsewhere in the code models that inherit `BaseModel` are validated such that we ensure models have subclass of `BaseModel` @JGSweets @Sanketh7

taylorfturner: To be clear, I think that should go in a separate PR if we have to add anything to further validate. For now, I think as implemented is okay with regards to this PR getting merged....................

taylorfturner: +1

taylorfturner (approved):
LGTM -- let's create a follow-up issue for more fully resolving some of these comments in a subsequent, smaller PR

dataprofiler/labelers/base_model.py
taylorfturner: I'd recommend in order to get the rest of this merged, @Sanketh7, revert this change for the `AutoSubRegistration` and get it working to pass `mypy` checks. Then in a follow-up PR, we can resolve this specific issue more fully.

Sanketh7: Reverted these changes so we can tackle it in another PR.

taylorfturner (approved):
follow-up PRs as discussed

tyfarnan (approved):

JGSweets (approved):

PR #706, opened 2022-11-04
Fix bug with null replication metrics when row is all null

tonywu315:
When `null_replication_metrics` is enabled and any row is all null, this errors:
```Python
import dataprofiler as dp
import pandas as pd
import re

profiler_options = dp.ProfilerOptions()

NO_FLAG = 0
profiler_options.set({
    '*.null_values': {
        "": NO_FLAG,
        "nan": re.IGNORECASE,
        "none": re.IGNORECASE,
        "null": re.IGNORECASE,
        "  *": NO_FLAG,
        "--*": NO_FLAG,
        "__*": NO_FLAG,
        "9"*7: NO_FLAG,
    }, 
    "*.null_replication_metrics.is_enabled": True,
    "data_labeler.is_enabled": False,
    "multiprocess.is_enabled": False,
})
profiler = dp.Profiler(pd.DataFrame([[10, 10], [9999999, 9999999]]), options=profiler_options)
```

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2240,7 +2240,7 @@ class StructuredProfiler(BaseProfiler):
         :param clean_samples: input cleaned dataset
         :type clean_samples: dict
         """
-        data: pd.DataFrame = pd.DataFrame(clean_samples)
+        data = pd.DataFrame(clean_samples)
 
         # If the last row is all null, then add rows to the data DataFrame
         max_null_index = max(
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2240,7 +2240,20 @@ class StructuredProfiler(BaseProfiler):
         :param clean_samples: input cleaned dataset
         :type clean_samples: dict
         """
-        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
+        data: pd.DataFrame = pd.DataFrame(clean_samples)
+
+        # If the last row is all null, then add rows to the data DataFrame
+        max_null_index = max(
+            [max(i) for i in getattr(self._profile[0], "null_types_index").values()],
+            default=0,
+        )
+        if max_null_index > data.index.max():
+            data.loc[max_null_index] = {}
+
+        # Fill in missing rows with NaN and convert types to numeric
+        data = data.reindex(range(data.index.max() + 1), fill_value=np.nan).apply(
+            pd.to_numeric, errors="coerce"
+        )
 
         get_data_type = lambda profile: profile.profiles[  # NOQA: E731
             "data_type_profile"
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2240,20 +2240,7 @@ class StructuredProfiler(BaseProfiler):
         :param clean_samples: input cleaned dataset
         :type clean_samples: dict
         """
-        data = pd.DataFrame(clean_samples)
-
-        # If the last row is all null, then add rows to the data DataFrame
-        max_null_index = max(
-            [max(i) for i in getattr(self._profile[0], "null_types_index").values()],
-            default=0,
-        )
-        if max_null_index > data.index.max():
-            data.loc[max_null_index] = {}
-
-        # Fill in missing rows with NaN and convert types to numeric
-        data = data.reindex(range(data.index.max() + 1), fill_value=np.nan).apply(
-            pd.to_numeric, errors="coerce"
-        )
+        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
 
         get_data_type = lambda profile: profile.profiles[  # NOQA: E731
             "data_type_profile"
@@ -2300,7 +2287,11 @@ class StructuredProfiler(BaseProfiler):
             # Partition data based on whether target column value is null or not
             # Calculate sum, mean of each partition without including current column
             # in calculation
-            sum_null = data.iloc[null_indices, data.columns != col_id].sum().to_numpy()
+            sum_null = (
+                data.loc[data.index.intersection(null_indices), data.columns != col_id]
+                .sum()
+                .to_numpy()
+            )
 
             # Add old sum_null if exists
             if col_id in self._null_replication_metrics:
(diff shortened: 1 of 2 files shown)


dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -2081,6 +2081,21 @@ def test_null_replication_metrics_calculation(self):
...
+        profiler = dp.StructuredProfiler(data_4, options=profile_options)
+        report = profiler.report()
+
+        self.assertTrue("null_replication_metrics" in report["data_stats"][0])
+        column = report["data_stats"][0]["null_replication_metrics"]
+
+        np.testing.assert_array_almost_equal([0.5, 0.5], column["class_prior"])
+        np.testing.assert_array_almost_equal([[20], [0]], column["class_sum"])
+        np.testing.assert_array_almost_equal([[10], [0]], column["class_mean"])
+
```
tonywu315, line 2098: If the entire dataframe is null, there is still an error.

taylorfturner: are these values the actual values we expect from the calculations or are these values just what was output from the `report` on L2090?

taylorfturner (commented):
just one comment left around test values... otheriwse, LGTM

dataprofiler/profilers/profile_builder.py
```diff
@@ -2240,7 +2240,20 @@ def _update_null_replication_metrics(self, clean_samples: Dict) -> None:
...
+        # If the last row is all null, then add rows to the data DataFrame
+        max_null_index = max(
+            [max(i) for i in getattr(self._profile[0], "null_types_index").values()],
+            default=0,
+        )
+        if max_null_index > data.index.max():
+            data.loc[max_null_index] = {}
+
+        # Fill in missing rows with NaN and convert types to numeric
+        data = data.reindex(range(data.index.max() + 1), fill_value=np.nan).apply(
```
JGSweets, line 2254: this could be really expensive for large data. Is there another way around this?

JGSweets: What specific error is requiring this change / where?

taylorfturner: +1 good catch @JGSweets

dataprofiler/tests/profilers/test_profile_builder.py
tonywu315: They should be the actual values we expect

dataprofiler/profilers/profile_builder.py
tonywu315: If a row is all null, the dataframe will be missing a row, which causes an error on the line
`sum_null = data.iloc[null_indices, data.columns != col_id].sum().to_numpy()`.
`reindex` fills in these missing rows with `nan`.

taylorfturner: The `.reindex()` though when done on an entire dataset could be very costly in terms of runtime; so that is the concern around this operation that is being added to the process.

dataprofiler/profilers/profile_builder.py
```diff
@@ -2240,7 +2240,20 @@ def _update_null_replication_metrics(self, clean_samples: Dict) -> None:
         :param clean_samples: input cleaned dataset
         :type clean_samples: dict
         """
-        data = pd.DataFrame(clean_samples).apply(pd.to_numeric, errors="coerce")
+        data = pd.DataFrame(clean_samples)
+
+        # If the last row is all null, then add rows to the data DataFrame
+        max_null_index = max(
+            [max(i) for i in getattr(self._profile[0], "null_types_index").values()],
```
JGSweets, line 2247: `self._profile[0]` Right now this is getting the first profile every time is this correct? 

Why do we need to add rows to the df?

dataprofiler/tests/profilers/test_profile_builder.py
tonywu315: yes

dataprofiler/profilers/profile_builder.py
JGSweets: If a row is null, could we do a join with non-null rows? I believe we have the info on what rows are completely null actually.

JGSweets: `null_indices + non-complete null rows`?

JGSweets: or even better potentially would be `null_indices - complete null rows` (theoretically there should be less completely null rows)

tonywu315: Another way would be to reindex a single column in `clean_samples` and then combine all of them into `data`, which should also fix the bug.

tonywu315: That would work too

JGSweets: why can't we do it w/o re-indexing?

dataprofiler/profilers/profile_builder.py
tonywu315: If the last row(s) is all null, then the maximum null index of any profile should be the index of the last row.

dataprofiler/tests/profilers/test_profile_builder.py
taylorfturner: excellent

dataprofiler/profilers/profile_builder.py
taylorfturner: - in this test case, `len(self._profile)` is `3` and only the zero-th index has values.
```
getattr(self._profile[0], "null_types_index").values()
dict_values([{2, 4}])
getattr(self._profile[1], "null_types_index").values()
dict_values([])
getattr(self._profile[2], "null_types_index").values()
dict_values([])
```

@tonywu315 is this always the case though? are we sure we can always just pop pull the zero-th profile in the `self._profile` list?

dataprofiler/profilers/profile_builder.py
```diff
@@ -2287,7 +2287,11 @@ def _update_null_replication_metrics(self, clean_samples: Dict) -> None:
             # Partition data based on whether target column value is null or not
             # Calculate sum, mean of each partition without including current column
             # in calculation
-            sum_null = data.iloc[null_indices, data.columns != col_id].sum().to_numpy()
+            sum_null = (
+                data.loc[data.index.intersection(null_indices), data.columns != col_id]
+                .sum()
+                .to_numpy()
+            )
```
tonywu315, line 2294: This code does not error anymore if the entire DataFrame is null.

taylorfturner: this is much cleaner!

JGSweets: beautiful!

JGSweets (approved):

taylorfturner (approved):
LGTM -- great job!

PR #707, opened 2022-11-04
PSI `diff()` bug

taylorfturner:
Bug in `diff()` `PSI` calculation. Bug was issue when `min` edge and `max` edge on histogram are equal


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -538,6 +538,18 @@ def _preprocess_for_calculate_psi(
...
+            if (
+                len(self_histogram["bin_counts"]) <= 1
+                and len(other_histogram["bin_counts"]) <= 1
+            ):
+                warnings.warn(
+                    """Data were essentially the same and less
+                    than or equal to one bin. PSI cannot be performed.""",
+                    RuntimeWarning,
+                )
+                return new_self_histogram, new_other_histogram
```
taylorfturner, line 551: Not certain of `return new_self_histogram, new_other_histogram`. 

It does trigger the `return None` in the `_calculate_psi` function, which is a good thing because if the logic on L543-544 return `True` I don't want PSI to even be attempted to be calculated because we have all the values in one bin_edge. **This is the bug.**

Another route, instead of returning `new_self_hist` and `new_other_hist` is to return `self_hist` and `other_hist`, but that is basically kicking us out of the `preprocessing` with two return values that is essentially saying "this is processed data and usable in `_calculate_psi`. Whereas, returning `new_self_hist` and `new_other_hist` is returning something and when `_calcl_psi` sees it it knows to kick out and not calc PSI.... 

open to thoughts here or where / better way to implement this logic

dataprofiler/tests/profilers/test_integration_structured_labeler_diff.py
```diff
@@ -0,0 +1,24 @@
+import os
```
taylorfturner, line 1: integration test for the scenario that originally raised the issue in the example notebook

dataprofiler/tests/profilers/test_integration_structured_labeler_diff.py
```diff
@@ -0,0 +1,24 @@
...
+from dataprofiler import Data, Profiler
+
+test_root_path = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+
+
+class TestStructDiff(unittest.TestCase):
+    def test_diff(self):
+        # divide dataset in half
+        test_dir = os.path.join(test_root_path, "data")
+        data = Data(os.path.join(test_dir, "csv/diamonds.csv"))
```
JGSweets, line 16: if we can reduce this down when we finalize the PR, that'd be awesome. e.g. no labeler or anything expensive we don't need. even less data maybe if possible

taylorfturner: totally agree, this is super rough implementation

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: need to return `0`

dataprofiler/tests/profilers/test_integration_structured_labeler_diff.py
taylorfturner: actually included a new test in `test_text_column_profile.py` which is much cleaner, much less data, and re-creates exactly the problem we were trying to solve in the bug

dataprofiler/tests/profilers/test_integration_structured_labeler_diff.py
taylorfturner: added test for the exact bug issue `min == max` on histogram bin_edges to `test_text_column_profiler.py`

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -609,3 +609,41 @@ def test_diff(self):
             places=2,
         )
         self.assertDictEqual(expected_diff, profile_diff)
+
+        # re-create `diamond.csv` categorical column
+        df = pd.Series(["D", "I", "F", "H", "G"]).apply(str)
+
+        df2 = pd.Series(["D", "I", "F", "H", "G"]).apply(str)
```
taylorfturner, line 616: test now located here in `test_text_column_profile.py`

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -609,3 +609,41 @@ def test_diff(self):
...
+            "stddev": "unchanged",
+            "vocab": utils.find_diff_of_lists_and_sets(
+                profile1["vocab"], profile2["vocab"]
+            ),
+            "t-test": {
+                "t-statistic": None,
+                "conservative": {"df": None, "p-value": None},
+                "welch": {"df": None, "p-value": None},
+            },
+            "psi": 0,
```
taylorfturner, line 644: returning `0` because the PSI on an unchanged distribution is `0`

ksneab7 (approved):

micdavis (approved):

PR #711, opened 2022-11-10
PSI no calculation on `TextProfile`

taylorfturner:
don't return `PSI` on `TextColumn` profiling


tyfarnan (approved):

dataprofiler/tests/profilers/test_text_column_profile.py
```diff
@@ -638,7 +637,6 @@ def test_diff(self):
                 "conservative": {"df": None, "p-value": None},
                 "welch": {"df": None, "p-value": None},
             },
-            "psi": 0,
```
JGSweets, line 641: removing this, but this edge case needs testing elsewhere

JGSweets (changes requested):
missing edge case test

PR #712, opened 2022-11-10
PSI fixing unit tests

taylorfturner:


dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -996,3 +996,63 @@ def test_diff(self):
...
+        }
+
+        expected_psi_value = 0
+        psi_value = other1._calculate_psi(
+            self_match_count=other1.match_count,
+            self_histogram=other1._stored_histogram["histogram"],
+            other_match_count=other2.match_count,
+            other_histogram=other2._stored_histogram["histogram"],
+        )
+        self.assertEquals(expected_psi_value, psi_value)
```
JGSweets, line 1058: There might be 1 more edge case.

when one requires regen and the other doesn't and the transpose. Can use the same histograms, just flipped positions.

Thoughts?

taylorfturner: pushing

JGSweets: @taylorfturner is adding.

JGSweets (approved):

micdavis (approved):

PR #724, opened 2022-11-17
Add Python 3.11 to GHA

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>


gliptak:
https://github.com/apache/arrow/issues/14572

https://arrow.apache.org/docs/python/install.html#python-compatibility

taylorfturner:
Thanks for the PR, @gliptak. I see why you closed due to `pyarrow` not yet supporting python `3.11`. Once that is added, we'll be glad to take a look at adding `3.11` to the GH checks, though. 

Thanks again!

PR #735, opened 2022-12-03
loosening pin on typing-extensions

leos:
Dataprofiler is currently strictly depending on an old version of `typing-extensions`. This package is generally super safe to upgrade (and not pin tightly) since it's just backports to the `typing` module.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,4 @@ psutil>=4.0.0
 scipy>=1.4.1
 requests==2.27.1
 networkx==2.5.1
-typing-extensions>=4.4.0
+typing-extensions>=3.10.0.2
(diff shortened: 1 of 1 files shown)


taylorfturner (approved):
LGTM -- thanks for the contribution

JGSweets (approved):

leos:
Sorry, not sure what the schedule is but it would be great if there could be a point release that includes this in the near future - this change would unblock us from including `dataprofiler` in our monorepo config (we're hacking around it for now which is pretty ugly).

taylorfturner:
Yes, @JGSweets and I were discussing a release yesterday. Will let you know when it is out. Cheers!

taylorfturner:
@leos `0.8.5` is live. Release notes [here](https://github.com/capitalone/DataProfiler/releases/tag/0.8.5). Thanks for your contributions! 🎉

PR #768, opened 2023-02-25
Categoricalcolumn save

kshitijavis:


ksneab7 (approved):

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -324,3 +324,18 @@ def unalikeability(self) -> float | None:
...
+
+        :return: dict
+        """
+        categorical_dict = {
+            "type": self.type,
+            "_categories": self._categories,
+            "__calculations": self.__calculations,
+            "_top_k_categories": self._top_k_categories,
+        }
+        categorical_dict.update(super().to_dict())
```
JGSweets, line 340: As opposed to updating the dict, there could be overlap issues.
What if we do some mechanic to store it in a key referencing the base class?

JGSweets: This would allow us to also know where to route on load.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -324,3 +324,18 @@ def unalikeability(self) -> float | None:
...
+    def to_dict(self) -> dict:
+        """
+        Create serializable dict of this object.
+
+        :return: dict
+        """
+        categorical_dict = {
+            "type": self.type,
+            "_categories": self._categories,
+            "__calculations": self.__calculations,
```
JGSweets, line 337: I think this will cause issues in the future, but currently is unimportant bc there's no calc to track. We might need a reusable func to serialize and another for later loading funcs. However, we can address this when it shows in subsequent updates.

JGSweets (changes requested):
Let's figure out the serialization key structure for inheritance issues prior to merge.

PR #772, opened 2023-03-17
Encode order column

kshitijavis:
Builds on previous PR [https://github.com/capitalone/DataProfiler/pull/770](url)


JGSweets (approved):

JGSweets (changes requested):
waiting on #770

dataprofiler/tests/profilers/test_json_encoder.py
```diff
@@ -89,3 +89,51 @@ def test_encode_categorical_column_profiler_after_update(self):
         )
 
         self.assertEqual(serialized, expected)
+
```
JGSweets, line 92: Thinking long term, would it be better to put the profile specific tests with each profiler's tests instead?

ksneab7: I like having the tests match the file structure so yeah I second this for sure!

ksneab7 (approved):

JGSweets (approved):

PR #786, opened 2023-04-19
Decode categorical

kshitijavis:
This PR depends on #789 

Since this is a big PR, I list the changes I made below
1. ~~Update `ProfileEncoder` (in `json_encode`) to serialize with the class name. This makes it easier to decode. E.g:~~
  ```
  {
    "class": <str name of class that was serialized>
    "data": {
        <attr1>: <value1>
        <attr2>: <value2>
        ...
    }
}
  ```
2. Create decoder function (in `json_decode`) to handle deserialization. This first calls `json.loads` on the serialized string and then creates an object by going through the attributes in the resulting `dict`. **This currently only works for `CategoricalColumn`**
3. ~~Update all _encode_ tests to recognize the "class", "data" structure from step 1~~
4. Write `json_decode` test for CateogoricalColumn. Create `assert_profiles_equal` method in `tests/utils.py` to compare two profiles by class attributes

**Questions**
1. Is the `assert_profiles_equal` the best way to compare profiles? Should this method be tested on its own
2. The serialization/deserialization process converts `defaultdict` objects into plain `dict` objects. Will this cause any bugs?

### DIFF
--- a/dataprofiler/profilers/json_decoder.py
+++ b/dataprofiler/profilers/json_decoder.py
@@ -28,6 +28,16 @@ def load_column_profile(serialized_json: dict) -> BaseColumnProfiler:
     """
     Construct subclass of BaseColumnProfiler given a serialized JSON.
 
+    Exepcted format of serialzied_json (see json_encoder):
+        {
+            "class": <str name of class that was serialized>
+            "data": {
+                <attr1>: <value1>
+                <attr2>: <value2>
+                ...
+            }
+        }
+
     :param serialized_json: JSON represenation of column profiler that was
         serialized using the custom encoder in proilers.json_encoder
     :type serialized_json: a dict that was created by calling json.loads on
--- a/dataprofiler/profilers/json_decoder.py
+++ b/dataprofiler/profilers/json_decoder.py
@@ -22,14 +22,19 @@ def get_column_profiler_class(class_name: str) -> BaseColumnProfiler:
         raise ValueError(f"Invalid profiler class {class_name} " f"failed to load.")
 
 
-def decode_categorical_column(to_decode: dict):
+def load_column_profile(serialized_json: dict) -> BaseColumnProfiler:
     """
-    Specify how CategoricalColumn should be deserialized.
+    Construct subclass of BaseColumnProfiler given a serialzed_json.
 
-    :param to_decode: an object to be deserialized
-    :type to_serialize: a dictionary resullting from json.loads()
-    :return: CategoricalColumn object
+    :param serialized_json: JSON represenation of column profiler that was
+        serialized using the custom encoder in proilers.json_encoder
+    :type serialized_json: a dict that was created by calling json.loads on
+        a JSON representation using the custom encoder
+    :return: subclass of BaseColumnProfiler that has been deserialiszed from
+        JSON
     """
-    decoded = CategoricalColumn(to_decode["name"])
-    for attr, value in to_decode.items():
-        decoded.__setattr__(attr, value)
+    column_profiler = get_column_profiler_class(serialized_json["class"])
+    for attr, value in serialized_json["data"].items():
+        column_profiler.__setattr__(attr, value)
+
+    return column_profiler
--- a/dataprofiler/profilers/json_decoder.py
+++ b/dataprofiler/profilers/json_decoder.py
@@ -1,0 +1,16 @@
+"""Contains methods to decode components of a Profiler."""
+
+from . import categorical_column_profile
+
+
+def decode_categorical_column(to_decode: dict):
+    """
+    Specify how CategoricalColumn should be deserialized.
+
+    :param to_decode: an object to be deserialized
+    :type to_serialize: a dictionary resullting from json.loads()
+    :return: CategoricalColumn object
+    """
+    decoded = categorical_column_profile.CategoricalColumn(to_decode["name"])
+    for attr, value in to_decode.items():
+        decoded.__setattr__(attr, value)
--- a/dataprofiler/profilers/json_decoder.py
+++ b/dataprofiler/profilers/json_decoder.py
@@ -18,10 +18,13 @@ def get_column_profiler_class(class_name: str) -> BaseColumnProfiler:
     :type class_name: str representing name of class
     :return: subclass of BaseColumnProfiler object
     """
-    if class_name == CategoricalColumn.__name__:
-        return CategoricalColumn(None)
-    else:
+    profiles = {CategoricalColumn.__name__: CategoricalColumn}
+
+    profile_class = profiles.get(class_name)
+    if profile_class is None:
         raise ValueError(f"Invalid profiler class {class_name} " f"failed to load.")
+    else:
+        return profile_class(None)
 
 
 def load_column_profile(serialized_json: dict) -> BaseColumnProfiler:
--- a/dataprofiler/profilers/json_decoder.py
+++ b/dataprofiler/profilers/json_decoder.py
@@ -1,6 +1,25 @@
 """Contains methods to decode components of a Profiler."""
 
-from . import categorical_column_profile
+from .base_column_profilers import BaseColumnProfiler
+from .categorical_column_profile import CategoricalColumn
+
+
+def get_column_profiler_class(class_name: str) -> BaseColumnProfiler:
+    """
+    Use name of class to return default-constructed version of that class.
+
+    Raises ValueError if class_name is not name of a subclass of
+        BaseColumnProfiler.
+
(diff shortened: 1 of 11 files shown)


dataprofiler/profilers/json_encoder.py
```diff
@@ -26,7 +26,7 @@ def default(self, to_serialize):
                 numerical_column_stats.NumericStatsMixin,
```
JGSweets, line 26: Thinking on this: ` numerical_column_stats.NumericStatsMixin` would be included with any profile. Do we need it here specifically? does that cause issue? 

Not saying it is a problem, just considering the value of it here.

JGSweets:
I might suggest splitting this into two PRs:
1. The update to encode.
2. The new decode.

That way 1 intent per PR and easier to not need to context switch within the PR itself.

dataprofiler/profilers/json_decoder.py
```diff
@@ -0,0 +1,65 @@
+"""Contains methods to decode components of a Profiler."""
```
JGSweets, line 1: I think we might want to consider a slightly separate API that implements functionality in the profiles themselves.

e.g.
```python.

# presuming i have this file already saved as JSON.
filepath = <LOCATION_OF_DATA>
profiler = Profiler.load(filepath, type="json")

# or 
profiler = Profiler.load_json(filepath)
```

Within each profile we can handle the profiler load checks,
for categorical it could be something like:
```python
class Cat...Profiler(...):
    .
    .
    .
    @classmethod
    def load_json_file(cls, filepath):
        ...

    @classmethod
    def load_json(cls, json_data):
        if json_data.get("class", None) is not cls.__name__:
            raise ValueError(...)
        profile = cls(...)
        ...
        return profile
```
moreover, ^^ might be able to just exist in BaseColumnProfiler if it is abstract enough.
the `...` part in the load might be able to be private functions in each class the base calls e.g.

```python
   profile = cls(...)
   profile._set_properties(json_data.get("data"))
```
Not keen on `_set_properties` naming, but it would have to exist in each class loading data.

JGSweets: or if the code generalizes well enough,  we don't need `set_properties` and just use:
```python
for attr, value in serialized_json["data"].items():
        column_profiler.__setattr__(attr, value)
```

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -749,20 +753,90 @@ def test_json_encode_after_update(self):
...
+        )
+
+        deserialized = decode_column_profiler(serialized_json)
+        exepcted = CategoricalColumn("0")
+
+        test_utils.assert_profiles_equal(deserialized, exepcted)
+
+    def test_json_decode_after_update(self):
+        # Actual deserialization
+        serialized = json.dumps(
```
JGSweets, line 798: by not calc'ing serialized from `df_categorical`, i'm concerned they could become disconnected.

it might be better todo this more similar to the other saves like:
```python
serialized  = json.dumps(df_categorical)
```

JGSweets: then we shouldn't have to reset naming as well.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -749,20 +753,90 @@ def test_json_encode_after_update(self):
...
+                "b",
+                "b",
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
+            ]
+        )
+        profile = CategoricalColumn(df_categorical.name)
```
JGSweets, line 833: Instead of naming it by the df, let's make something up that we know is unique to ensure it calc's it.

```python
profile = CategoricalColumn("Fake Profile Name")
```

however, your current tests the case of `None` which is a good edge case

dataprofiler/profilers/json_decoder.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Raises ValueError if class_name is not name of a subclass of
+        BaseColumnProfiler.
+
+    :param class_name: name of BaseColumnProfiler subclass retreived by
+        calling type(instance).__name__
+    :type class_name: str representing name of class
+    :return: subclass of BaseColumnProfiler object
+    """
+    if class_name == CategoricalColumn.__name__:
```
JGSweets, line 21: I see, we are being declarative here to allow any class to load.

JGSweets: also avoids circular imports since not in the base.

JGSweets: I think to be optimum for future code we can use a dict:
```python
profiles = {
    CategoricalColumn.__name__: CategoricalColumn
}
if class_name not in profiles:
    raise V...
return profiles.get(class_name)(None)
```

dataprofiler/profilers/json_encoder.py
kshitijavis: This is needed only because I have a test case that serializes a NumericStatsMin `test_numeric_stats_mixin_profile.TestNumericStatsMixin.test_json_encode`. Otherwise, it's not really necessary

dataprofiler/profilers/json_decoder.py
kshitijavis: I was thinking the same thing. Something like a separate `StructuredColProfile` decode function that calls my `decode_column_profiler` whenever it sees a Column Profiler in its attributes.

dataprofiler/tests/profilers/test_categorical_column_profile.py
kshitijavis: Do you mean something like

```py
expected_profile = CategoricalColumn(df_categorical.name)
expected_profile.update(df_categorical)

serialized = json.dumps(expected_profile)
deserialized = decode_column_profile(serialized)

test_utils.assert_profiles_equal(deserialized, expected_profile)
```

dataprofiler/profilers/json_decoder.py
kshitijavis: Sounds good, I'll make this change

dataprofiler/tests/profilers/test_categorical_column_profile.py
taylorfturner: yes, I think this is a better methodology

taylorfturner (changes requested):

dataprofiler/tests/profilers/utils.py
```diff
@@ -161,3 +164,39 @@ def increment_counter():
...
+        value2 = profile2_dict[attr1]
+        if not (isinstance(value2, type(value1)) or isinstance(value1, type(value2))):
+            raise ValueError(f"Profile value types unmatched: {value1} != {value2}")
+
+        if isinstance(value1, (BaseProfiler, BaseColumnProfiler)):
+            assert_profiles_equal(value1, value2)
+        elif isinstance(value1, numbers.Number):
+            np.testing.assert_equal(value1, value2)
+        elif value1 != value2:
+            raise ValueError(f"Profile values unmatched: {value1} != {value2}")
```
taylorfturner, line 202: assuming other types of profiles have nesting in the profile "data", this appears to only be built to work with the expected `CategoricalColumn` unit tests that you wrote... not bad but would need to be further develop this to account for nesting potentially

dataprofiler/profilers/json_decoder.py
```diff
@@ -0,0 +1,68 @@
...
+    :type class_name: str representing name of class
+    :return: subclass of BaseColumnProfiler object
+    """
+    profiles = {CategoricalColumn.__name__: CategoricalColumn}
+
+    profile_class = profiles.get(class_name)
+    if profile_class is None:
+        raise ValueError(f"Invalid profiler class {class_name} " f"failed to load.")
+    else:
+        return profile_class(None)
```
JGSweets, line 27: technically no need for the else as it will error otherwise.
```python
if profile_class is None:
        raise ValueError(f"Invalid profiler class {class_name} " f"failed to load.")
return profile_class(None)
```

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets: yes

dataprofiler/tests/profilers/utils.py
JGSweets: I think this is fine to iterate on when we need to approach that in the future.

JGSweets (approved):

taylorfturner (approved):

PR #789, opened 2023-04-21
Encode update format

kshitijavis:
PR #786 is dependent on this PR


dataprofiler/profilers/json_encoder.py
```diff
@@ -26,7 +26,7 @@ def default(self, to_serialize):
                 numerical_column_stats.NumericStatsMixin,
             ),
         ):
-            return to_serialize.__dict__
+            return {"class": type(to_serialize).__name__, "data": to_serialize.__dict__}
```
JGSweets, line 29: I'm slightly concerned that the general encoder adds extra information. I'm curious if this is how other systems do this or if they have save specific encoders in addition to just serialization of a class.

JGSweets: Not sure the best approach here.

kshitijavis: One solution would be to create another class `class ColumnProfilerEncoder(json.JSONEncoder)` that focuses specifically on encoding subclasses of `BaseColumnProfiler`

kshitijavis: Then line 29 could be changed to `ColumnProfilerEncoder.default(to_serialize)`. But I'm not sure if this is actually fixing anything or just moving the issue to a different class.

JGSweets: I wonder if we should change `class` -> `__type__`.
Might be more clear. Then we also don't need to have the `"data"` field, but everything else can be inline.

Both work honestly. 

i was just trying to consider if we wanted to decouple the class info from pure serialization. However, maybe we handle that in another iteration if needed.

JGSweets (approved):

taylorfturner (approved):

PR #790, opened 2023-04-21
fix bug in dataset generation

ksneab7:


dataprofiler/tests/space_time_analysis/dataset_generation.py
```diff
@@ -51,11 +54,14 @@ def convert_data_to_df(
...
     :param index: whether to include index in output to csv
     :type path: bool, optional
+    :param column_names: The names of the columns of a dataset
+    :type path: List, None, optional
     :return: a pandas dataframe
     """
     # convert array into dataframe
-    dataframe = pd.DataFrame(np_data)
-
+    if not column_names:
```
JGSweets, line 62: I think we might be able to simply this by using a different load function.

https://www.geeksforgeeks.org/creating-a-pandas-dataframe-using-list-of-tuples/amp/

JGSweets (changes requested):
Small suggestion

dataprofiler/tests/space_time_analysis/dataset_generation.py
ksneab7: All of these fixes still require you to submit a column param. I just tried from_items (this one says its not function), from_records (produces same problem as original code unless column param is present), and pivoting seems just as complicated as what im doing now.

JGSweets (approved):

dataprofiler/tests/space_time_analysis/structured_space_time_analysis.py
```diff
@@ -28,15 +28,15 @@
 
 
 def dp_profile_space_analysis(
-    data: pd.DataFrame,
+    data_: pd.DataFrame,
```
tyfarnan, line 31: why the trailing underscore?

tyfarnan (approved):

JGSweets (approved):

PR #795, opened 2023-04-26
Part 1 fix for categorical mem opt issue

ksneab7:
Dataset used is a single column ordered dataset
Conditions set:
self.max_sample_size_to_check_stop_condition = 100
self.stop_condition_unique_value_ratio = .90

- 0
  - Profile
    - Pre change
        - Total space allocated: N/A
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.02222418785095215
    - Post change
        - Total space allocated: N/A
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.022262096405029297
  - Merge
    - Pre change
        - Total space allocated: N/A
        - Line specific space allocated (Line number/s):N/A
        - Merge runtime: 1.4066696166992188e-05
    - Post change
        - Total space allocated: N/A
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 1.0967254638671875e-05
- 100
  - Profile
    - Pre change
        - Total space allocated: 449.9 KiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.11128807067871094
    - Post change
        - Total space allocated: 449.9 KiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.10366582870483398
  - Merge
    - Pre change
        - Total space allocated:  160.8 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.02127695083618164
    - Post change
        - Total space allocated: 160.8 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.02112579345703125
- 1000
  - Profile
    - Pre change
        - Total space allocated: 269.5 KiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.07865023612976074
    - Post change
        - Total space allocated: 269.5 KiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.07574105262756348
  - Merge
    - Pre change
        - Total space allocated: 220.8 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.025424957275390625
    - Post change
        - Total space allocated: 220.8 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.022446870803833008
- 5000
  - Profile
    - Pre change
        - Total space allocated: 841.1 KiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.141251802444458
    - Post change
        - Total space allocated: 841.1 KiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.13668394088745117
  - Merge
    - Pre change
        - Total space allocated: 653.0 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.03322911262512207
    - Post change
        - Total space allocated: 653.0 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.03174471855163574
- 7500
  - Profile
    - Pre change
        - Total space allocated: 1.3 MiB
        - Line specific space allocated (Line number/s): 288.0 KiB
        - Profile runtime: 0.18341469764709473
    - Post change
        - Total space allocated: 1.2 MiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 0.1740732192993164
  - Merge
    - Pre change
        - Total space allocated: 1002.1 KiB
        - Line specific space allocated (Line number/s): <Space taken by soon-to-be-changed line in flame graph>
        - Merge runtime: 0.03673124313354492
    - Post change
        - Total space allocated: 1002.1 KiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.034219980239868164
- 100000
  - Profile
    - Pre change
        - Total space allocated: 22.6 MiB
        - Line specific space allocated (Line number/s): 5.0 MiB
        - Profile runtime: 1.42696213722229
    - Post change
        - Total space allocated: 17.6 MiB
        - Line specific space allocated (Line number/s): N/A
        - Profile runtime: 1.4447557926177979
  - Merge
    - Pre change
        - Total space allocated: 13.0 MiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.21359801292419434
    - Post change
        - Total space allocated: 13.0 MiB
        - Line specific space allocated (Line number/s): N/A
        - Merge runtime: 0.15265393257141113
        
 Separate measure for two different profiles to indicate merge differences:
     - Pre change
        - Total space allocated: 25.0 MiB
        - Line specific space allocated (Line number/s): N/A
    - Post change
        - Total space allocated: 17.7 MiB
        - Line specific space allocated (Line number/s): N/A

### DIFF
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -100,13 +100,13 @@ class CategoricalColumn(BaseColumnProfiler):
                 (
                     "unique_count",
                     utils.find_diff_of_numbers(
-                        len(self.categories), len(other_profile.categories)
+                        self.unique_count, other_profile.unique_count
                     ),
                 ),
                 (
                     "unique_ratio",
                     utils.find_diff_of_numbers(
-                        self._unique_ratio, other_profile.unique_ratio
+                        self.unique_ratio, other_profile.unique_ratio
                     ),
                 ),
             ]
@@ -170,8 +170,8 @@ class CategoricalColumn(BaseColumnProfiler):
             categorical=self.is_match,
             statistics=dict(
                 [
-                    ("unique_count", self._unique_count),
-                    ("unique_ratio", self._unique_ratio),
+                    ("unique_count", self.unique_count),
+                    ("unique_ratio", self.unique_ratio),
                 ]
             ),
             times=self.times,
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -71,14 +71,44 @@ class CategoricalColumn(BaseColumnProfiler):
             )
 
         merged_profile = CategoricalColumn(None)
+
+        # If both profiles have not met stop condition
         if not (self._stop_condition_is_met or other._stop_condition_is_met):
             merged_profile._categories = utils.add_nested_dictionaries(
                 self._categories, other._categories
             )
-        BaseColumnProfiler._add_helper(merged_profile, self, other)
-        self._merge_calculations(
-            merged_profile.__calculations, self.__calculations, other.__calculations
-        )
+
+            BaseColumnProfiler._add_helper(merged_profile, self, other)
+            self._merge_calculations(
+                merged_profile.__calculations, self.__calculations, other.__calculations
+            )
+
+            # Check merged profile w/ stop condition
+            if self._check_stop_condition_is_met(
+                merged_profile.sample_size, merged_profile.unique_ratio
+            ):
+                merged_profile._stopped_at_unique_ratio = merged_profile.unique_ratio
+                merged_profile._stopped_at_unique_count = merged_profile.unique_count
+                merged_profile._categories = {}
+                merged_profile._stop_condition_is_met = True
+
+        else:
+            # If either profile has hit stop condition, remove categories dict
+            merged_profile._stopped_at_unique_ratio = max(
+                self.unique_ratio, other.unique_ratio
+            )
+            merged_profile._stopped_at_unique_count = max(
+                self.unique_count, other.unique_count
+            )
+            merged_profile.times = (
+                self.times
+                if self.times["categories"] >= other.times["categories"]
+                else other.times
+            )
+            merged_profile.sample_size = self.sample_size + other.sample_size
+            merged_profile._categories = {}
+            merged_profile._stop_condition_is_met = True
+
         return merged_profile
 
     def diff(self, other_profile: CategoricalColumn, options: dict = None) -> dict:
@@ -201,12 +231,12 @@ class CategoricalColumn(BaseColumnProfiler):
     @property
     def unique_ratio(self) -> float:
         """Return ratio of unique categories to sample_size."""
-        unique_ratio = 1.0
         if self._stop_condition_is_met:
             return self._stopped_at_unique_ratio
 
         if self.sample_size:
             return len(self.categories) / self.sample_size
+        return 0
 
     @property
     def unique_count(self) -> float:
@@ -230,6 +260,25 @@ class CategoricalColumn(BaseColumnProfiler):
             is_match = True
         return is_match
 
+    def _check_stop_condition_is_met(self, sample_size: int, unqiue_ratio: float):
+        """Return boolean given stop conditions.
+
+        :param sample_size: Number of samples to check the stop condition
+        :type sample_size: int
+        :param unqiue_ratio: Ratio of unique values to full sample size to
+            check stop condition
+        :type unqiue_ratio: float
+        :return: boolean for stop conditions
+        """
+        if (
+            self.max_sample_size_to_check_stop_condition is not None
+            and self.stop_condition_unique_value_ratio is not None
+            and sample_size >= self.max_sample_size_to_check_stop_condition
+            and unqiue_ratio >= self.stop_condition_unique_value_ratio
+        ):
+            return True
+        return False
+
     def update_stop_condition(self, data: DataFrame):
         """Return value stop_condition_is_met given stop conditions.
 
@@ -238,16 +287,13 @@ class CategoricalColumn(BaseColumnProfiler):
         :return: boolean for stop conditions
         """
         merged_unique_count = len(self._categories)
-        merged_sample_size = (self.sample_size + len(data))
+        merged_sample_size = self.sample_size + len(data)
         merged_unique_ratio = merged_unique_count / merged_sample_size
 
-        if (
-                self.max_sample_size_to_check_stop_condition is not None
-                and self.stop_condition_unique_value_ratio is not None
-                and merged_sample_size >= self.max_sample_size_to_check_stop_condition
-                and merged_unique_ratio >= self.stop_condition_unique_value_ratio
-        ):
-            self._stop_condition_is_met = True
+        self._stop_condition_is_met = self._check_stop_condition_is_met(
+            merged_sample_size, merged_unique_ratio
+        )
+        if self._stop_condition_is_met:
             self._stopped_at_unique_ratio = merged_unique_ratio
             self._stopped_at_unique_count = merged_unique_count
 
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from operator import itemgetter
+from typing import cast
 
 from pandas import DataFrame, Series
 
@@ -49,8 +50,8 @@ class CategoricalColumn(BaseColumnProfiler):
         self.stop_condition_unique_value_ratio = None
         self._stop_condition_is_met = False
 
-        self._stopped_at_unique_ratio = 0.0
-        self._stopped_at_unique_count = 0.0
+        self._stopped_at_unique_ratio: float | None = None
+        self._stopped_at_unique_count: int | None = None
         if options:
             self._top_k_categories = options.top_k_categories
 
@@ -73,14 +74,13 @@ class CategoricalColumn(BaseColumnProfiler):
         merged_profile = CategoricalColumn(None)
         BaseColumnProfiler._add_helper(merged_profile, self, other)
 
+        self._merge_calculations(
+            merged_profile.__calculations, self.__calculations, other.__calculations
+        )
         # If both profiles have not met stop condition
         if not (self._stop_condition_is_met or other._stop_condition_is_met):
             merged_profile._categories = utils.add_nested_dictionaries(
                 self._categories, other._categories
-            )
-
-            self._merge_calculations(
-                merged_profile.__calculations, self.__calculations, other.__calculations
             )
 
             # Transfer stop condition variables of 1st profile object to merged profile
@@ -248,17 +248,17 @@ class CategoricalColumn(BaseColumnProfiler):
     def unique_ratio(self) -> float:
         """Return ratio of unique categories to sample_size."""
         if self._stop_condition_is_met:
-            return self._stopped_at_unique_ratio
+            return cast(float, self._stopped_at_unique_ratio)
 
         if self.sample_size:
             return len(self.categories) / self.sample_size
         return 0
 
     @property
-    def unique_count(self) -> float:
+    def unique_count(self) -> int:
         """Return ratio of unique categories to sample_size."""
         if self._stop_condition_is_met:
-            return self._stopped_at_unique_count
+            return cast(int, self._stopped_at_unique_count)
 
         return len(self.categories)
 
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -45,12 +45,12 @@ class CategoricalColumn(BaseColumnProfiler):
         self._top_k_categories: int | None = None
 
         # Conditions to stop categorical profiling
-        self.max_sample_size_to_check_stop_condition = None
-        self.stop_condition_unique_value_ratio = None
+        self.max_sample_size_to_check_stop_condition = 100
+        self.stop_condition_unique_value_ratio = .10
         self._stop_condition_is_met = False
 
-        self._unique_ratio = 0.0
-        self._unique_count = 0.0
+        self._stopped_at_unique_ratio = 0.0
+        self._stopped_at_unique_count = 0.0
         if options:
             self._top_k_categories = options.top_k_categories
 
@@ -200,24 +200,20 @@ class CategoricalColumn(BaseColumnProfiler):
     @property
     def unique_ratio(self) -> float:
         """Return ratio of unique categories to sample_size."""
-        if self._stop_condition_is_met:
-            return self._unique_ratio
-
         unique_ratio = 1.0
+        if self._stop_condition_is_met:
+            return self._stopped_at_unique_ratio
+
         if self.sample_size:
-            unique_ratio = len(self.categories) / self.sample_size
-        self._unique_ratio = unique_ratio
-        return self._unique_ratio
+            return len(self.categories) / self.sample_size
 
     @property
     def unique_count(self) -> float:
         """Return ratio of unique categories to sample_size."""
         if self._stop_condition_is_met:
-            return self._unique_count
+            return self._stopped_at_unique_count
 
-        self._unique_count = len(self.categories)
-
-        return self._unique_count
+        return len(self.categories)
 
     @property
     def is_match(self) -> bool:
@@ -240,14 +236,19 @@ class CategoricalColumn(BaseColumnProfiler):
         :type data: DataFrame
         :return: boolean for stop conditions
         """
+        merged_unique_count = len(self._categories)
+        merged_sample_size = (self.sample_size + len(data))
+        merged_unique_ratio = merged_unique_count / merged_sample_size
+
         if (
-            self.max_sample_size_to_check_stop_condition is not None
-            and len(data) >= self.max_sample_size_to_check_stop_condition
-            and self.stop_condition_unique_value_ratio is not None
-            and len(self._categories) / len(data)
-            >= self.stop_condition_unique_value_ratio
+                self.max_sample_size_to_check_stop_condition is not None
+                and self.stop_condition_unique_value_ratio is not None
+                and merged_sample_size >= self.max_sample_size_to_check_stop_condition
+                and merged_unique_ratio >= self.stop_condition_unique_value_ratio
         ):
             self._stop_condition_is_met = True
+            self._stopped_at_unique_ratio = merged_unique_ratio
+            self._stopped_at_unique_count = merged_unique_count
 
     @BaseColumnProfiler._timeit(name="categories")
     def _update_categories(
@@ -301,7 +302,7 @@ class CategoricalColumn(BaseColumnProfiler):
         :rtype: CategoricalColumn
         """
         # If condition for limiting profile calculations
-        if len(df_series) == 0:
+        if len(df_series) == 0 or self._stop_condition_is_met:
             return self
 
         profile = dict(sample_size=len(df_series))
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -71,6 +71,7 @@ class CategoricalColumn(BaseColumnProfiler):
             )
 
         merged_profile = CategoricalColumn(None)
+        BaseColumnProfiler._add_helper(merged_profile, self, other)
 
         # If both profiles have not met stop condition
         if not (self._stop_condition_is_met or other._stop_condition_is_met):
@@ -78,22 +79,18 @@ class CategoricalColumn(BaseColumnProfiler):
                 self._categories, other._categories
             )
 
-            BaseColumnProfiler._add_helper(merged_profile, self, other)
             self._merge_calculations(
                 merged_profile.__calculations, self.__calculations, other.__calculations
             )
 
             # Transfer stop condition variables of 1st profile object to merged profile
             # if they are not None else set to 2nd profile
-            if self.max_sample_size_to_check_stop_condition is not None:
+            profile1_product = self.sample_size * self.unique_ratio
+            profile2_product = other.sample_size * other.unique_ratio
+            if profile1_product > profile2_product:
                 merged_profile.max_sample_size_to_check_stop_condition = (
                     self.max_sample_size_to_check_stop_condition
                 )
-            else:
-                merged_profile.max_sample_size_to_check_stop_condition = (
-                    other.max_sample_size_to_check_stop_condition
-                )
-            if self.stop_condition_unique_value_ratio is not None:
                 merged_profile.stop_condition_unique_value_ratio = (
                     self.stop_condition_unique_value_ratio
                 )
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, name: str | None, options: CategoricalOptions = None) -> None
         self.__calculations: dict = {}
         self._filter_properties_w_options(self.__calculations, options)
         self._top_k_categories: int | None = None
+        self.max_sample_size_to_check_stop_condition = None
```
JGSweets, line 46: I suggest putting comments around this to clarify these are stop profiling conditions

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -211,6 +214,21 @@ def is_match(self) -> bool:
             is_match = True
         return is_match
 
+    def check_stop_condition_is_met(self, data: DataFrame) -> bool:
+        """Return value stop_condition_is_met provided the new data being added
+        :param data: Dataframe currently being processed by categorical profiler
+        :type data: DataFrame
+        :return: boolean for stop conditions
+        """
+        if not self._stop_condition_is_met:
```
JGSweets, line 223: if we move the check to inside `update` won't need this `if` here

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -211,6 +214,21 @@ def is_match(self) -> bool:
...
+            total_categories = len(self._categories) + len(data.value_counts())
```
JGSweets, line 225: maybe we should update the bool after we do the calc and merge. Otherwise:
1. we calc value_counts twice.
2. it will check the actual merge length which accounts for overlap

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -233,10 +251,15 @@ def _update_categories(
...
-        )
+        if not self.check_stop_condition_is_met(df_series):
+            category_count = df_series.value_counts(dropna=False).to_dict()
+            # If condition for limiting profile calculations
+
+            self._categories = utils.add_nested_dictionaries(
+                self._categories, category_count
+            )
+        else:
+            self._categories = {}
```
JGSweets, line 266: i think we can move the check to just the end and then do:
```python
if self.....:
    self._categories = {}
```
We are going to need to figure out how we handle the `unique_raito` calc as we probably still want to hold the last known values for profiling.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -211,6 +233,22 @@ def is_match(self) -> bool:
...
+
+        :param data: Dataframe currently being processed by categorical profiler
+        :type data: DataFrame
+        :return: boolean for stop conditions
+        """
+        if (
+            self.max_sample_size_to_check_stop_condition is not None
+            and len(data) >= self.max_sample_size_to_check_stop_condition
+            and self.stop_condition_unique_value_ratio is not None
+            and len(self._categories) / len(data)
```
JGSweets, line 247:
```python
merged_unique_count = len(self._categories)
merged_sample_size = (self.sample_size + len(data))
merged_unique_ratio =  merged_unique_count / merged_sample_size


if (
    self.max_sample_size_to_check_stop_condition is not None
    and  self. stop_condition_unique_value_ratio is not None
    and merged_sample_size >= self.max_sample_size_to_check_stop_condition
    and merged_unique_ratio >= stop_condition_unique_value_ratio
):
    self._stop_condition_is_met = True
    self._stopped_at_unique_ratio = merged_unique_ratio
    self._stopped_at_unique_count = merged_unique_count
```
I think this keeps it clear and then sets the values.

An option would be if we want this function to  just be a check, we can calculate in`_update_categories `:
merged_unique_count = len(self._categories)
merged_sample_size = (self.sample_size + len(data))
merged_unique_ratio =  merged_unique_count / merged_sample_size
```
and pass it into the check function:
```python
def _check_stop_condition_is_met(self, sample_count, sample_size, unqiue_ratio):
    """Return value stop_condition_is_met given stop conditions.
    
    :return: boolean for stop conditions
    """
    if (
          self.max_sample_size_to_check_stop_condition is not None
          and  self. stop_condition_unique_value_ratio is not None
          and sample_size >= self.max_sample_size_to_check_stop_condition
          and unique_ratio >= stop_condition_unique_value_ratio
    ):
        return True
    return False
```

then in `_update_categories`:
```python
if self._check_stop_condition_is_met(merged.....):
    self._stop_condition_is_met = True
    self._categories = {}
    self._stopped_at_unique_ratio = merged_unique_ratio
    self._stopped_at_unique_count = merged_unique_count
```
^^ similar logic can then be used in the `__add__` where we calc the:
```python
if not profile1._stop_condition_is_met and not profile2._stop_condition_is_met:
    # merge dicts and then check the values for the merging with the stop condition, etc.
else:
    # merged categories is empty
    # unique ratio and unique count can be take from the profile with the largest sample size
    # if we avg, will be problem for streams.
```
between the two profies

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -192,10 +200,24 @@ def categorical_counts(self) -> dict[str, int]:
...
+        self._unique_ratio = unique_ratio
+        return self._unique_ratio
+
+    @property
+    def unique_count(self) -> float:
+        """Return ratio of unique categories to sample_size."""
+        if self._stop_condition_is_met:
+            return self._unique_count
+
+        self._unique_count = len(self.categories)
```
JGSweets, line 218: let's only set `self._unique_count` if the stop condition was met. we can handle this below in the _update as described in the other comment

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -192,10 +200,24 @@ def categorical_counts(self) -> dict[str, int]:
...
     def unique_ratio(self) -> float:
         """Return ratio of unique categories to sample_size."""
+        if self._stop_condition_is_met:
+            return self._unique_ratio
+
         unique_ratio = 1.0
         if self.sample_size:
             unique_ratio = len(self.categories) / self.sample_size
-        return unique_ratio
+        self._unique_ratio = unique_ratio
```
JGSweets, line 209: let's only set self. _unique_ratio if the stop condition was met. we can handle this below in the _update as described in the other comment.

I think we need to update the hidden naming of the var as referenced below to declare the difference.
something like `self._stopped_at_unique_ratio`

dataprofiler/profilers/categorical_column_profile.py
JGSweets: I think we need to update the hidden naming of the var as referenced below to declare the difference.
something like `self._stopped_at_unique_count`

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -162,7 +170,7 @@ def profile(self) -> dict:
             categorical=self.is_match,
             statistics=dict(
                 [
-                    ("unique_count", len(self.categories)),
```
JGSweets, line 165: good

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -92,7 +100,7 @@ def diff(self, other_profile: CategoricalColumn, options: dict = None) -> dict:
                 (
                     "unique_count",
                     utils.find_diff_of_numbers(
-                        len(self.categories), len(other_profile.categories)
```
JGSweets, line 95: nice

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -43,6 +43,14 @@ def __init__(self, name: str | None, options: CategoricalOptions = None) -> None
...
         self._filter_properties_w_options(self.__calculations, options)
         self._top_k_categories: int | None = None
+
+        # Conditions to stop categorical profiling
+        self.max_sample_size_to_check_stop_condition = None
+        self.stop_condition_unique_value_ratio = None
+        self._stop_condition_is_met = False
+
+        self._unique_ratio = 0.0
+        self._unique_count = 0.0
```
JGSweets, line 53: I suggest we update naming based on thoughts below:
`self._stopped_at_unique_count`
`self._stopped_at_unique_ratio`
or of the like

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -49,6 +49,43 @@ def test_correct_categorical_model_string(self):
         }
         self.assertCountEqual(categories, profile.categories)
 
+    def test_stop_condition_is_met_initially(self):
+        dataset = self.aws_dataset["host"].dropna()
+        profile = CategoricalColumn(dataset.name)
+        profile.max_sample_size_to_check_stop_condition = 200
+        profile.stop_condition_unique_value_ratio = 0
```
JGSweets, line 56: nice

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -49,6 +49,43 @@ def test_correct_categorical_model_string(self):
         }
         self.assertCountEqual(categories, profile.categories)
 
+    def test_stop_condition_is_met_initially(self):
+        dataset = self.aws_dataset["host"].dropna()
```
JGSweets, line 53: I suggest we make this more understandable and not dependent on the external data if we can.
we can then modify the conditions below to fit our needs w/o changing the dataset to truly test our conditions

```python
# 20 samples, unique ratio = 0.10, unique_count = 4, sample_size = 40
dataset = pd.Series(["a"] * 10 + ["b"] * 10, ["c"] * 10, ["d"] * 10)

# test stop condition met
...
profile. max_sample_size_to_check_stop_condition = 0
profile. stop_condition_unique_value_ratio = 0
...asserts

# test stop condition no longer met, keeps all data
...
profile. max_sample_size_to_check_stop_condition = 41
profile. stop_condition_unique_value_ratio = 0
...asserts

File: dataprofiler/tests/profilers/test_categorical_column_profile.py [hunk 13]
```diff
@@ -49,6 +49,43 @@ def test_correct_categorical_model_string(self):
...
+        profile = CategoricalColumn(dataset.name)
+        profile.max_sample_size_to_check_stop_condition = 200
+        profile.stop_condition_unique_value_ratio = 0
+        profile.update(dataset)
+
+        self.assertTrue(profile._stop_condition_is_met)
+        self.assertEqual(profile.categories, [])
+        self.assertEqual(profile.unique_count, 0.0)
+        self.assertEqual(profile.unique_ratio, 0.0)
```
JGSweets, line 62: Since we are computing the categories and merging already, we should use that information to update these values.
Hence, since this is a new dataset, I would expect the `unique_ratio` and `unique_count` == to that of the dataset.
In the toy example above: `# 20 samples, unique ratio = 0.10, unique_count = 4, sample_size = 40`

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -49,6 +49,43 @@ def test_correct_categorical_model_string(self):
...
+
+    def test_stop_condition_is_met_after_initial_profile(self):
+        dataset = self.aws_dataset["host"].dropna()
+        profile = CategoricalColumn(dataset.name)
+        profile.max_sample_size_to_check_stop_condition = len(dataset.value_counts())
+        profile.stop_condition_unique_value_ratio = (
+            len(dataset.value_counts()) + 1
+        ) / len(dataset)
```
JGSweets, line 70: For this case, we could either use the unique_ratio condition or the max sample size to enforce.
`len(dataset.value_counts()) + 1`
and leave the `stop_condition_unique_value_ratio = 0` to ensure it calls

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -49,6 +49,43 @@ def test_correct_categorical_model_string(self):
...
+
+        dataset.loc[len(dataset.index)] = "Testing past ratio"
+        profile.update(dataset)
+
+        self.assertTrue(profile._stop_condition_is_met)
+        self.assertEqual(profile.categories, [])
+        self.assertEqual(profile.unique_count, (len(dataset.value_counts())) - 1)
+        self.assertEqual(
+            profile.unique_ratio, (len(dataset.value_counts()) - 1) / len(dataset)
+        )
```
JGSweets, line 87: will need a test for these being function for adding profiles too:
`profile1 + profile2`
on top of data checks, need to ensure the conditions transfer properly.

dataprofiler/tests/profilers/test_categorical_column_profile.py
ksneab7: I dont fully understand this comment, if we try to call any of your suggestions at this point in the code the values wouldnt make sense (profile.sample_size = 0 and profile.unique_ratio = None)

dataprofiler/tests/profilers/test_categorical_column_profile.py
ksneab7: This is not addressed yet, I am currently working out a solution

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,44 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+                self._categories, other._categories
+            )
+
+            BaseColumnProfiler._add_helper(merged_profile, self, other)
+            self._merge_calculations(
+                merged_profile.__calculations, self.__calculations, other.__calculations
+            )
+
+            # Check merged profile w/ stop condition
+            if self._check_stop_condition_is_met(
```
JGSweets, line 87: I think you need to transfer the properties of the stop condition prior

JGSweets (changes requested):
Not sure all props are transferred in merge, specifically the stop conditions

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -722,6 +796,11 @@ def test_json_encode(self):
                     "_categories": defaultdict(int),
                     "_CategoricalColumn__calculations": dict(),
                     "_top_k_categories": None,
+                    "max_sample_size_to_check_stop_condition": None,
```
JGSweets, line 799: nice

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -578,6 +614,44 @@ def test_categorical_merge(self):
         }
         self.assertCountEqual(report_count, expected_dict)
 
+        # Setting up of profile with stop condition not yet met
```
JGSweets, line 617: in the above, should we check that stop condition is not met for cases that don't meet and the stopped at stay none?

dataprofiler/profilers/categorical_column_profile.py
ksneab7, line 87: This should be `merged_profile` not `self`

ksneab7: Fixing both of these now

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,68 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+                merged_profile.stop_condition_unique_value_ratio = (
+                    self.stop_condition_unique_value_ratio
+                )
+            else:
+                merged_profile.stop_condition_unique_value_ratio = (
+                    other.stop_condition_unique_value_ratio
+                )
+
+            # 1st and 2nd profile values should be same so no need for conditonal set
+            merged_profile._stop_condition_is_met = self._stop_condition_is_met
```
JGSweets, line 106: I believe `False` is the default for `_stop_condition_is_met ` so no need to set here

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,68 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+            merged_profile._stopped_at_unique_ratio = self._stopped_at_unique_ratio
+            merged_profile._stopped_at_unique_count = self._stopped_at_unique_count
```
JGSweets, line 108: as above, default should suffice in this case.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -237,6 +347,9 @@ def _update_categories(
         self._categories = utils.add_nested_dictionaries(
             self._categories, category_count
         )
+        self.update_stop_condition(df_series)
+        if self._stop_condition_is_met:
+            self._categories = {}
```
JGSweets, line 352: if we already set:
```python
self._stopped_at_unique_ratio = merged_unique_ratio
self._stopped_at_unique_count = merged_unique_count
```
below, we can join it with:
```python
self._categories = {}
```

However, let's consider the function description and the action of the function.

When we say `update_stop_condition`, to me it would only update `_stop_condition_is_met ` boolean. Do we wnat to rename it to be more explicit about it doing more? or do we need `update_stop_condition`?

Could we just have `_update_categories` and `_check_stop_condition_is_met`?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -211,6 +284,43 @@ def is_match(self) -> bool:
...
+        if (
+            self.max_sample_size_to_check_stop_condition is not None
+            and self.stop_condition_unique_value_ratio is not None
+            and sample_size >= self.max_sample_size_to_check_stop_condition
+            and unqiue_ratio >= self.stop_condition_unique_value_ratio
+        ):
+            return True
+        return False
+
+    def update_stop_condition(self, data: DataFrame):
```
JGSweets, line 306: if we want this to continue to exist, I suggest making it private as we wouldn't want an external user to use this method.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,68 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+                merged_profile._stop_condition_is_met = True
+
+        else:
+            # If either profile has hit stop condition, remove categories dict
+            merged_profile._stopped_at_unique_ratio = max(
+                self.unique_ratio, other.unique_ratio
+            )
+            merged_profile._stopped_at_unique_count = max(
+                self.unique_count, other.unique_count
+            )
```
JGSweets, line 126: Should it be purely based on max?

Consider:
* `Profile 1` has 10        samples and a unique_ratio of 1.00 
* `Profile 2` has 10000 samples and a unique ratio of 0.50.

Do these values make sense?

Also, do we want to potentially choose different values from each, could that be problematic?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,68 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+            merged_profile.times = (
+                self.times
+                if self.times["categories"] >= other.times["categories"]
+                else other.times
+            )
```
JGSweets, line 131: We shouldn't later merged_profile times. This should already be handled via the `_add_helper`

JGSweets: we should make sure our tests validates this.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,68 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+            merged_profile.sample_size = self.sample_size + other.sample_size
```
JGSweets, line 132: I don't believe this is true if we are taking one value over the other above

dataprofiler/profilers/categorical_column_profile.py
JGSweets: The option of not choosing one profiles values vs the other's complicates the sample_size below I believe.

JGSweets: We should also have tests which ensure that if self has the values that get assigned vs swapping and ensuring that other's values can be choosen.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,68 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
-            merged_profile.__calculations, self.__calculations, other.__calculations
-        )
+
+        # If both profiles have not met stop condition
+        if not (self._stop_condition_is_met or other._stop_condition_is_met):
+            merged_profile._categories = utils.add_nested_dictionaries(
+                self._categories, other._categories
+            )
+
+            BaseColumnProfiler._add_helper(merged_profile, self, other)
```
JGSweets, line 81: I think this should still happen in both cases.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -578,6 +614,44 @@ def test_categorical_merge(self):
...
+        profile_w_stop_cond_1 = CategoricalColumn("merge_stop_condition_test")
+        profile_w_stop_cond_1.max_sample_size_to_check_stop_condition = 12
```
JGSweets, line 619: should ensure times are correct, but looks like we don't properly do that in this file rn.

The way we've done previously is mocking timeit as a generator. 
Example:
https://github.com/capitalone/DataProfiler/blob/f206af2f8218713f9372942ba7dbb33b57cf3aba/dataprofiler/tests/profilers/test_profile_builder.py#L106

JGSweets (changes requested):
Concerns about merge and suggestions for the update portion.

dataprofiler/profilers/categorical_column_profile.py
JGSweets: Want to note these are code readability / quality comments, not functional as I think the code is functional in this condition.

dataprofiler/profilers/categorical_column_profile.py
ksneab7: This is for clarity/readiaility not functionality but if it is not helpful I can remote it

dataprofiler/profilers/categorical_column_profile.py
ksneab7: same comment as above

dataprofiler/profilers/categorical_column_profile.py
ksneab7: No we should not add the times together because that is not consistent with a stop condition being hit

dataprofiler/profilers/categorical_column_profile.py
ksneab7: I dont agree with this statement (mentioned above)

dataprofiler/profilers/categorical_column_profile.py
ksneab7: why wouldnt we want that?

dataprofiler/profilers/categorical_column_profile.py
JGSweets: Essentially, we want to know the computational time cost to get to this point so we need the merge of both.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -63,13 +71,60 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+
+        # If both profiles have not met stop condition
+        if not (self._stop_condition_is_met or other._stop_condition_is_met):
+            merged_profile._categories = utils.add_nested_dictionaries(
+                self._categories, other._categories
+            )
+
+            self._merge_calculations(
+                merged_profile.__calculations, self.__calculations, other.__calculations
+            )
```
JGSweets, line 84: This still needs to be conducted above. right now we don't have it populated, but it should be done in either case.

dataprofiler/profilers/categorical_column_profile.py
JGSweets: Since this is a feature branch, we can refactor in subsequent PR.

dataprofiler/tests/profilers/test_categorical_column_profile.py
JGSweets: added:
https://github.com/capitalone/DataProfiler/issues/806

taylorfturner:
> Assuming refactors after this in subsequent PRs
need these refactors to be notated as follow-up issues @ksneab7

Issue #717, opened 2022-11-16
Investigate use of cast() in data_readers


Sanketh7:
Investigate whether these instances of `cast()` are needed. 
PRs should be small (at most a few files).

data_readers/csv_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L553
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L574
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L602
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L605

data_readers/data_utils.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/data_utils.py#L330
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/data_utils.py#L661

data_readers/data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/data.py#L66

data_readers/filepath_or_buffer.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/filepath_or_buffer.py#L91
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/filepath_or_buffer.py#L101

data_readers/graph_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/graph_data.py#L255 
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/graph_data.py#L257

data_readers/structured_mixins.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/structured_mixins.py#L83

data_readers/text_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/text_data.py#L87 
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/text_data.py#L103

junholee6a:
For most of these `cast`s, removing them causes an error when the file is type-checked with mypy. I think it is safe to assume that we don't have to investigate those further? Here are the ones that do _not_ cause mypy errors if you remove them (with updated line numbers):

data_readers/csv_data.py
* https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L572
[...]

PR #796, opened 2023-04-26
Remove unnecessary cast() in csv_data.py (1)

junholee6a:
Removing this cast() doesn't cause a mypy error.

Issue: https://github.com/capitalone/DataProfiler/issues/717

`self._header` is set to type `Optional[Union[str, int]]` in the constructor. Also, `self.guess_header_row()` has return type `Optional[int]`, so casting to `int` doesn't make sense here.


JGSweets (approved):
good identification!

junholee6a:
Thank you for the review!

PR #798, opened 2023-04-26
Remove unnecessary cast() in csv_data.py (2)

junholee6a:
Issue: https://github.com/capitalone/DataProfiler/issues/717

- Removing this `cast()` doesn't cause a mypy error
- `self.file_encoding` has type `Optional[str]` prior to casting
- The `read_csv_df` function’s `encoding` parameter also has type `Optional[str]`
- Therefore casting to `str` for the function call isn’t necessary

### DIFF
--- a/dataprofiler/data_readers/csv_data.py
+++ b/dataprofiler/data_readers/csv_data.py
@@ -597,7 +597,7 @@ class CSVData(SpreadSheetDataMixin, BaseData):
             cast(Optional[int], self.header),
             self.selected_columns,
             read_in_string=True,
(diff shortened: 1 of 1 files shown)


taylorfturner:
Will need to update branch @junholee6a

taylorfturner (approved):

JGSweets (approved):

taylorfturner:
@junholee6a maybe push an empty commit again... the `license/cla` is still in "expected" status. you can do `git commit -m "empty commit" --allow-empty`

Issue #721, opened 2022-11-16
Investigate use of # type: ignore in data_readers.


Sanketh7:
Investigate whether these instances of `# type: ignore` are needed.
PRs should be small (at most a few files).

data_readers/csv_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L221 
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L763

data_readers/json_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/json_data.py#L446

data_readers/parquet_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/parquet_data.py#L175

data_readers/text_data.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/text_data.py#L156

junholee6a:
For the first instance, [csv_data.py line 221](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/data_readers/csv_data.py#L221), it doesn't seem like the `type: ignore` can be removed. The method being called, `_guess_quote_and_delimiter()`, is from Python's built-in `csv` package. Removing the `type: ignore` causes this error in mypy: `"Sniffer" has no attribute "_guess_quote_and_delimiter"`. From testing with the `csv` and `random` packages, it seems like type checkers recognize "private methods" typically denoted by a leading underscore, and pretend they don't exist outside of the class. This happens for Python's built-in packages but not for user-written code. So mypy doesn't recognize `_guess_quote_and_delimiter()`, even though it can be called. I haven't been able to find documentation or a fix for this.

The other four instances are identical and have a fix so I'll make one PR for them

taylorfturner:
This is great, @junholee6a! Will keep and eye out and review PRs as they come in. Cheers!

junholee6a:
Thank you for reviewing them! This accounts for all of the `type: ignore`s in data_readers, so I think this issue can be closed

PR #805, opened 2023-05-01
Specify __init__ calls in data readers reload() methods

junholee6a:
Issue: https://github.com/capitalone/DataProfiler/issues/721

Previously, these lines without a `# type: ignore` caused a mypy error, because the `__init__()` called may have been overridden by a subclass. Specifying that the initializer called belongs to the current class fixes the mypy error.

Another example of this syntax: https://stackoverflow.com/questions/753640/inheritance-and-overriding-init-in-python/11055691#11055691


JGSweets (approved):

taylorfturner (approved):

taylorfturner:
Thanks, @junholee6a!

PR #807, opened 2023-05-03
Categorical Stop Condition Options

micdavis:


dataprofiler/profilers/profiler_options.py
```diff
@@ -895,6 +905,34 @@ def _validate_helper(self, variable_path: str = "CategoricalOptions") -> list[st
...
+        if self.stop_condition_unique_value_ratio is not None and (
+            not isinstance(self.stop_condition_unique_value_ratio, float)
+            or self.stop_condition_unique_value_ratio < 1
+        ):
+            errors.append(
+                "{}.stop_condition_unique_value_ratio must be either None"
+                " or a positive float".format(variable_path)
+            )
+
+        if bool(self.max_sample_size_to_check_stop_condition) != bool(
```
JGSweets, line 927: This will actually error if either of these values are 0. Which I think is valid as it force the condition to be checked early.
we can then fix this with a ` >= 0` which results in a bool.
Only issue becomes if these aren't numerics which I'm not sure is feasible.

JGSweets (changes requested):
issue in the `xor` conditiona.

PR #808, opened 2023-05-03
Categorical Stop Condition Options

micdavis:


dataprofiler/profilers/profiler_options.py
```diff
@@ -895,6 +905,34 @@ def _validate_helper(self, variable_path: str = "CategoricalOptions") -> list[st
                 "{}.top_k_categories must be either None"
                 " or a positive integer".format(variable_path)
             )
+
+        if self.max_sample_size_to_check_stop_condition is not None and (
+            not isinstance(self.max_sample_size_to_check_stop_condition, int)
+            or self.max_sample_size_to_check_stop_condition < 0
+        ):
```
taylorfturner, line 912:
```suggestion
        if (self.max_sample_size_to_check_stop_condition is not None) and (
            not isinstance(self.max_sample_size_to_check_stop_condition, int)
            or self.max_sample_size_to_check_stop_condition < 0
        ):
```

dataprofiler/profilers/profiler_options.py
```diff
@@ -895,6 +905,34 @@ def _validate_helper(self, variable_path: str = "CategoricalOptions") -> list[st
                 "{}.top_k_categories must be either None"
                 " or a positive integer".format(variable_path)
             )
+
+        if self.max_sample_size_to_check_stop_condition is not None and (
+            not isinstance(self.max_sample_size_to_check_stop_condition, int)
+            or self.max_sample_size_to_check_stop_condition < 0
```
taylorfturner, line 911: why would a user specify a less than zero for `max_sample_size_to_check_stop_condition`?

taylorfturner: wouldn't it need to be `<=0`?

dataprofiler/profilers/profiler_options.py
```diff
@@ -895,6 +905,34 @@ def _validate_helper(self, variable_path: str = "CategoricalOptions") -> list[st
...
+        ):
+            errors.append(
+                "{}.max_sample_size_to_check_stop_condition must be either None"
+                " or a non-negative integer".format(variable_path)
+            )
+
+        if self.stop_condition_unique_value_ratio is not None and (
+            not isinstance(self.stop_condition_unique_value_ratio, float)
+            or self.stop_condition_unique_value_ratio < 0
+        ):
```
taylorfturner, line 921:
```suggestion
        if (self.stop_condition_unique_value_ratio is not None) and (
            not isinstance(self.stop_condition_unique_value_ratio, float)
            or self.stop_condition_unique_value_ratio < 0
        ):
```

dataprofiler/profilers/profiler_options.py
taylorfturner: current it would hit the `errors.append` if it was set to `0` but the error would say it must be non-negative which `0` is

dataprofiler/profilers/profiler_options.py
```diff
@@ -895,6 +905,34 @@ def _validate_helper(self, variable_path: str = "CategoricalOptions") -> list[st
...
+            or self.max_sample_size_to_check_stop_condition < 0
+        ):
+            errors.append(
+                "{}.max_sample_size_to_check_stop_condition must be either None"
+                " or a non-negative integer".format(variable_path)
+            )
+
+        if self.stop_condition_unique_value_ratio is not None and (
+            not isinstance(self.stop_condition_unique_value_ratio, float)
+            or self.stop_condition_unique_value_ratio < 0
```
taylorfturner, line 920: see comment about on `< 0`

dataprofiler/profilers/profiler_options.py
```diff
@@ -866,7 +866,13 @@ def _validate_helper(self, variable_path: str = "OrderOptions") -> list[str]:
 class CategoricalOptions(BaseInspectorOptions):
     """For configuring options Categorical Column."""
 
-    def __init__(self, is_enabled: bool = True, top_k_categories: int = None) -> None:
+    def __init__(
+        self,
+        is_enabled: bool = True,
+        top_k_categories: int = None,
+        max_sample_size_to_check_stop_condition: int = None,
+        stop_condition_unique_value_ratio: float = None,
```
taylorfturner, line 874: nice

JGSweets, line 874: The typing should have the `| None`
`max_sample_size_to_check_stop_condition: int | None = None`
same for both.

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -11,15 +11,66 @@ class TestCategoricalOptions(TestBaseInspectorOptions):
     def test_init(self):
         option = self.get_options()
```
JGSweets, line 12: Another PR, but this is naming I should have caught before:
`get_options` -> `set_and_retrieve_option_values` or something. Needs to be an issue.

JGSweets: `get_options` is confusing when we set a value in it.

dataprofiler/profilers/profiler_options.py
micdavis: pushed

taylorfturner (approved):

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -32,6 +83,8 @@ def test_set(self):
         params_to_check = [
             dict(prop="is_enabled", value_list=[False, True]),
             dict(prop="top_k_categories", value_list=[None, 3]),
+            dict(prop="max_sample_size_to_check_stop_condition", value_list=[None, 20]),
+            dict(prop="stop_condition_unique_value_ratio", value_list=[None, 2]),
```
JGSweets, line 87: i suggest adding a float here

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
taylorfturner: +1

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -55,24 +108,115 @@ def _assert_set_helper(prop, value):
...
+        )
+        options = self.get_options(
+            max_sample_size_to_check_stop_condition=2.0,
+            stop_condition_unique_value_ratio=2.0,
+        )
+        self.assertEqual(
+            [expected_max_sample_size_type_error], options._validate_helper()
+        )
+
+        # These ones should throw a max_sample_size_type_error
```
JGSweets, line 153: comment should be for unique value I think

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -55,24 +108,115 @@ def _assert_set_helper(prop, value):
...
 
+        options = self.get_options(
+            max_sample_size_to_check_stop_condition=20,
+            stop_condition_unique_value_ratio=2.0,
+        )
+        self.assertEqual([], options._validate_helper())
+        options = self.get_options(
+            max_sample_size_to_check_stop_condition=0,
+            stop_condition_unique_value_ratio=0.0,
+        )
```
JGSweets, line 212: so the condition we discussed before, it might be good to make one of these a value. and one 0 to test that condition.

dataprofiler/profilers/profiler_options.py
JGSweets, line 921: Because it is a ratio we need to: `0 <= stop_condition_unique_value_ratio <= 1.0`

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -32,6 +83,8 @@ def test_set(self):
         params_to_check = [
             dict(prop="is_enabled", value_list=[False, True]),
             dict(prop="top_k_categories", value_list=[None, 3]),
+            dict(prop="max_sample_size_to_check_stop_condition", value_list=[None, 20]),
```
taylorfturner, line 86: and to @JGSweets comment too here -- shouldn't we test this with a `float` to validate its caught

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
taylorfturner: +1

taylorfturner (changes requested):

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
JGSweets: could do this as 
`stop_condition_unique_value_ratio=0.5,`

and in check above
set `stop_condition_unique_value_ratio=0.0`

so we don't need more tests to do the same checks

dataprofiler/profilers/profiler_options.py
taylorfturner: ah I knew it didn't makes sense for a hard core `< 0`

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
JGSweets: this isn't checking validation helper

taylorfturner (approved):

taylorfturner (approved):

JGSweets (approved):

PR #815, opened 2023-05-09
Fix for histogram merging

ksneab7:


dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
 )
 
 
+def rework_ptp(maximum, minimum):
```
JGSweets, line 17: once we fix naming, we should make these private as well.

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
+    ----------
+    x : array_like
+        Input data that is to be histogrammed, trimmed to range. May not
+        be empty.
+
+    Returns
+    -------
+    h : An estimate of the optimal bin width for the given data.
+    """
+    dataset_size = sum(profile._stored_histogram["histogram"]["bin_counts"])
```
JGSweets, line 45: We can instead use the `profile.sampling_size`.

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
+    minimum = profile._stored_histogram["histogram"]["bin_edges"][0]
+    maximum = profile._stored_histogram["histogram"]["bin_edges"][-1]
```
JGSweets, line 47: Should be able to use `profile.min` etc.

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
+    return (rework_ptp(maximum, minimum) /
```
JGSweets, line 81: as above

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
+        Input data that is to be histogrammed, trimmed to range. May not
+        be empty.
+
+    Returns
+    -------
+    h : An estimate of the optimal bin width for the given data.
+    """
+    dataset_size = sum(profile._stored_histogram["histogram"]["bin_counts"])
+    minimum = profile._stored_histogram["histogram"]["bin_edges"][0]
+    maximum = profile._stored_histogram["histogram"]["bin_edges"][-1]
```
JGSweets, line 106: as above

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
```
JGSweets, line 129: as above

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
+    x : array_like
+        Input data that is to be histogrammed, trimmed to range. May not
+        be empty.
+
+    Returns
+    -------
+    h : An estimate of the optimal bin width for the given data.
+    """
+    iqr = np.subtract(profile._get_percentile([75]), profile._get_percentile([25]))
+    dataset_size = sum(profile._stored_histogram["histogram"]["bin_counts"])
```
JGSweets, line 159: as above

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,218 @@
...
+    ----------
+    x : array_like
+        Input data that is to be histogrammed, trimmed to range. May not
+        be empty.
+
+    Returns
+    -------
+    h : An estimate of the optimal bin width for the given data.
+    """
+    dataset_size = sum(profile._stored_histogram["histogram"]["bin_counts"])
```
JGSweets, line 224: as above

dataprofiler/profilers/histogram_utils.py
```diff
@@ -90,3 +302,42 @@ def _get_bin_edges(
...
+    if isinstance(bin_method, str):
+        bin_name = bin_method
+    # if `bins` is a string for an automatic method,
+    # this will replace it with the number of bins calculated
+    if bin_name not in _hist_bin_selectors:
+        raise ValueError(f"{bin_name!r} is not a valid estimator for `bins`")
+
+    dataset_size = sum(profile._stored_histogram["histogram"]["bin_counts"])
+    minimum = profile._stored_histogram["histogram"]["bin_edges"][0]
+    maximum = profile._stored_histogram["histogram"]["bin_edges"][-1]
```
JGSweets, line 328: as above

dataprofiler/profilers/histogram_utils.py
```diff
@@ -90,3 +302,42 @@ def _get_bin_edges(
...
+
+    if dataset_size == 0:
+        n_equal_bins = 1
+    else:
+        # Do not call selectors on empty arrays
+        width = rework_hist_bin_selectors[bin_name](profile)
```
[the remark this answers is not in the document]
JGSweets: I'm curious if there are ways to reduce overall calcs by generating props before this loop and passing that in instead of hte profile.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -76,7 +76,6 @@ def __init__(self, options: NumericalOptions = None) -> None:
         self.histogram_selection: str | None = None
         self.user_set_histogram_bin: int | None = None
         self.bias_correction: bool = True  # By default, we correct for bias
-        self._mode_is_enabled: bool = True
```
JGSweets, line 79: Why is this removed?

dataprofiler/profilers/histogram_utils.py
ksneab7: AttributeError: 'IntColumn' object has no attribute 'sampling_size'

dataprofiler/profilers/numerical_column_stats.py
ksneab7: Weird I think this may solve some of my issue above. This should not be deleted

dataprofiler/profilers/histogram_utils.py
```diff
@@ -90,3 +366,61 @@ def _get_bin_edges(
...
+    :param profile: Input data's data profile that is to be histogrammed
+    :type profile: NumericStatsMixin
+    :param bin_method: the method used to calculate bins based on the profile given
+    :type bin_method: string
+
+    :return: ideal number of bins for a particular histograom calulcation method
+    """
+    # parse the overloaded bins argument
+
+    rework_hist_bin_selectors = {
```
JGSweets, line 384: instead of re init internal every time, maybe global in this file. Probably should privatize the variable too. Update name to suggest from profile in the name as oppose to rework.

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,293 @@
 )
 
 
+def _ptp(maximum, minimum):
```
JGSweets, line 17: do we need all of the imports above? also, do we know why we `ignore`?

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,293 @@
...
+    """
+    Doane's histogram bin estimator reworked to use profiles.
+
+    Improved version of Sturges' formula which works better for
+    non-normal data. See
+    stats.stackexchange.com/questions/55134/doanes-formula-for-histogram-binning
+    Function follows the numpy implementation within:
+    https://github.com/numpy/numpy/blob/main/numpy/lib/histograms.py
+
+    Parameters
```
JGSweets, line 39: I'm thinking we should reformat to use our format for docstrings until we want to convert to a different technique.

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,293 @@
...
+        Input data's data profile that is to be histogrammed
+
+    Returns
+    -------
+    An estimate of the optimal bin width for the given data.
+    """
+    try:
+        dataset_size = profile.match_count
+    except AttributeError:
+        dataset_size = sum(profile._stored_histogram["histogram"]["bin_counts"])
```
JGSweets, line 51: We do this a lot. Maybe abstract to a function we use at the top instead?

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,293 @@
...
+
+    minimum = (
+        profile.min
+        if profile.min is not None
+        else profile._stored_histogram["histogram"]["bin_edges"][0]
+    )
```
JGSweets, line 107: similar above for re-use. If we do a func, privatize.

dataprofiler/profilers/histogram_utils.py
```diff
@@ -14,6 +14,293 @@
...
+    "auto": _calc_auto_bin_width_from_profile,
+    "doane": _calc_doane_bin_width_from_profile,
+    "fd": _calc_fd_bin_width_from_profile,
+    "rice": _calc_rice_bin_width_from_profile,
+    "scott": _calc_scott_bin_width_from_profile,
+    "sqrt": _calc_sqrt_bin_width_from_profile,
+    "sturges": _calc_sturges_bin_width_from_profile,
+}
+
+
```
JGSweets, line 303: at this point in the code, have already calc'ed the values necessary to generate this? as if we did it from the profile? This would be __calculations order dependent.

JGSweets: could be a refactor to remove.

dataprofiler/profilers/histogram_utils.py
ksneab7: We do use all of the imports above in the `get_bin_edges` function and no not sure why there is an ignore, probably because of reference to protected methods?

dataprofiler/profilers/histogram_utils.py
JGSweets: nvm, maybe we cannot bc initial of hist not yet created for quantiles?

dataprofiler/profilers/histogram_utils.py
ksneab7: I am not sure. this may depend on Justin's feedback so Ill wait it out for now

dataprofiler/profilers/histogram_utils.py
```diff
@@ -90,3 +377,49 @@ def _get_bin_edges(
...
+        else profile._stored_histogram["histogram"]["bin_edges"][-1]
+    )
+
+    if dataset_size == 0:
+        n_equal_bins = 1
+    else:
+        # Do not call selectors on empty arrays
+        width = _hist_bin_width_selectors_for_profile[bin_method](profile)
+        if width and not np.isnan(width):
+            n_equal_bins = int(np.ceil(_unsigned_subtract(maximum, minimum) / width))
```
JGSweets, line 420: how is this `_unsigned...` diff from the `_ptp` above? Case for re-use?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -179,23 +179,87 @@ def _add_helper_merge_profile_histograms(
...
                 "current_loss": 0,
                 "histogram": {"bin_counts": None, "bin_edges": None},
             }
 
-        combined_values = np.concatenate(
-            [other1._histogram_to_array(), other2._histogram_to_array()]
+        # calculate the min of the first edge and the max of the last edge
+        # between two arrays
+        if not (other1.min is None or other2.min is None):
+            global_min_of_histogram_edges = min(other1.min, other2.min)
```
JGSweets, line 198: bc this is an or, what happens if one of them is `None`. does this error?/

dataprofiler/profilers/histogram_utils.py
ksneab7: I think the correct way to do it would be to calculate it one time and use it like how I had it in the temporary rework I indicated previously. We can turn it into a function temporarily but I think the actual fix will come later when we make this code work together with the pre-existing code such that this number will only need to be calulated one time.

dataprofiler/profilers/numerical_column_stats.py
ksneab7: if one them is None the whole expression is false so no it does not. If the expression was 
`not (other1.min is None and other2.min is None)` we may have an issue

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -179,23 +179,87 @@ def _add_helper_merge_profile_histograms(
...
+        else:
+            global_min_of_histogram_edges = min(
+                other1._stored_histogram["histogram"]["bin_edges"][0],
+                other2._stored_histogram["histogram"]["bin_edges"][0],
+            )
+
+        if not (other1.max is None or other2.max is None):
```
JGSweets, line 205: as above

dataprofiler/profilers/numerical_column_stats.py
JGSweets: at this point we have the `self.min` I think since we moved the calc below.

We could do:
```python
global_min_of_histogram_edges = self.min
if self.min is None:
    global_min_of_histogram_edges = min(
        other1._stored_histogram["histogram"]["bin_edges"][0],
        other2._stored_histogram["histogram"]["bin_edges"][0],
    )
```
it removes the `else` and utilizes existing functionality.

This is conditional on `self.min` being updated prior

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1240,11 +1305,43 @@ def _regenerate_histogram(
...
             )
+        return self._assimilate_histogram(
+            entity_count_by_bin,
+            bin_edges,
+            new_bin_counts,
+            new_bin_edges,
+            suggested_bin_count,
+        )
 
+    def _assimilate_histogram(
```
JGSweets, line 1316: can this be a class method? does it need to use `self`?

JGSweets: We would have to get around this:
```python
is_float_profile = self.__class__.__name__ == "FloatColumn"
```

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1240,11 +1305,43 @@ def _regenerate_histogram(
...
+        self,
+        entity_count_by_bin: np.ndarray,
+        bin_edges: np.ndarray,
+        new_entity_count_by_bin: np.ndarray,
+        new_bin_edges: np.ndarray,
```
JGSweets, line 1321: Trying to think how we can bring clarity to the word assimilate. instead of `new`, maybe `destination`

dataprofiler/profilers/text_column_profile.py
```diff
@@ -56,8 +56,8 @@ def __add__(self, other: TextColumn) -> TextColumn:
                 "'TextColumn' and '{}'".format(other.__class__.__name__)
             )
         merged_profile = TextColumn(None)
-        NumericStatsMixin._add_helper(merged_profile, self, other)
         BaseColumnPrimitiveTypeProfiler._add_helper(merged_profile, self, other)
+        NumericStatsMixin._add_helper(merged_profile, self, other)
```
JGSweets, line 60: I need to think about how this order could have affected other things, but tests do pass.

dataprofiler/tests/profilers/test_histogram_utils.py
```diff
@@ -0,0 +1,348 @@
...
+            "dataprofiler.profilers.NumericStatsMixin.stddev", new_callable=mock_stddev
+        ):
+            # Case 1: min, max, match_count, biased_skewness, and stddev are set
+            sg1 = np.sqrt(6.0 * (5 - 2) / ((5 + 1.0) * (5 + 3)))
+            expected = histogram_utils._ptp(5, 1) / (
+                1.0 + np.log2(5) + np.log2(1.0 + np.absolute(1.0) / sg1)
+            )
+
+            actual = histogram_utils._calc_doane_bin_width_from_profile(profile)
+            self.assertAlmostEqual(expected, actual, places=9)
```
JGSweets, line 91: do we need to almost equals after the update to the calc?

dataprofiler/tests/profilers/test_histogram_utils.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def test_calc_doane_bin_width_from_profile(self):
+        # Initial setup of profile
+        profile = TestColumn()
+
+        with mock.patch(
+            "dataprofiler.profilers.NumericStatsMixin.stddev", new_callable=mock_stddev
+        ):
+            # Case 1: min, max, match_count, biased_skewness, and stddev are set
+            sg1 = np.sqrt(6.0 * (5 - 2) / ((5 + 1.0) * (5 + 3)))
```
JGSweets, line 85: Should we replace the values themselves with the settings above?

e.g. `1-> self.min`

dataprofiler/tests/profilers/test_histogram_utils.py
```diff
@@ -0,0 +1,348 @@
...
+        # Initial setup of profile
+        profile = TestColumn()
+
+        # Case 1: bin method not in set of valid bin methods
+        with self.assertRaises(ValueError):
+            histogram_utils._calculate_bins_from_profile(profile, "test_not_in_method")
+
+        # Case 2: min, max, and match_count are set and bin_method is valid
+        expected_sqrt_answer = 4 / 2.23606797749979
+        expected_buckets = int(np.ceil(4 / expected_sqrt_answer))
```
JGSweets, line 302: for this, could probably just set the int value as opposed to the entire calc.

dataprofiler/profilers/histogram_utils.py
JGSweets: we could update the ignore to specifically capture that condition as oppose to a totality

dataprofiler/profilers/histogram_utils.py
JGSweets: Agreed, a better system. But, until then probably still ideal to nav towards dry. no extra tests needed.

dataprofiler/profilers/numerical_column_stats.py
ksneab7: else is needed. mypy fails because of  requirements from lispace parameter types

JGSweets: We shouldn't be changing code to meet mypy expectations. I'd defer to cleanliness and readability over mypy, but it typically is a good guide towards that end.

JGSweets: It failing from that change also would indicate an issue in our description / typing else where.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -179,23 +179,89 @@ def _add_helper_merge_profile_histograms(
...
                 "current_loss": 0,
                 "histogram": {"bin_counts": None, "bin_edges": None},
             }
 
-        combined_values = np.concatenate(
-            [other1._histogram_to_array(), other2._histogram_to_array()]
+        # calculate the min of the first edge and the max of the last edge
+        # between two arrays
+        global_min_of_histogram_edges = (
+            float(self.min)
```
JGSweets, line 198: if self.min is None, this will error

dataprofiler/profilers/numerical_column_stats.py
JGSweets: I figured it out, it was bc i wrote `if self.min is None` needs to be `if global_min_of_histogram_edges is None`

```python
global_min_of_histogram_edges = self.min
        if global_min_of_histogram_edges is None:
            global_min_of_histogram_edges =  min(
                other1._stored_histogram["histogram"]["bin_edges"][0],
                other2._stored_histogram["histogram"]["bin_edges"][0],
            )
```

dataprofiler/profilers/text_column_profile.py
ksneab7: So, im not sure if there was a deliberate reason to have this order not match the other profiles that use these two functions, but  the other two (float and int) do it in the updated order you see in the change. Previously order didn't matter so that might have been why it differed and no one said anything in the PR.

ksneab7: Looking farther into this, this specific line was implemented on release 3 years ago, so I wouldn't really know the reasoning behind the order differing decision as thats before my time...

dataprofiler/profilers/numerical_column_stats.py
JGSweets: probably unnecessary right now.

taylorfturner (approved):

dataprofiler/profilers/histogram_utils.py
```diff
@@ -90,3 +311,37 @@ def _get_bin_edges(
         raise ValueError("`bins` must be 1d, when an array")
 
     return bin_edges, n_equal_bins
+
+
+def _calculate_bins_from_profile(profile, bin_method):
+    """
+    Compute the bins used internally by `histogram`.
```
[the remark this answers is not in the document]
ksneab7: This doesnt have a numpy equivalent. So no I dont think so

taylorfturner (approved):

taylorfturner (approved):

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -179,23 +179,93 @@ def _add_helper_merge_profile_histograms(
...
+            calculated_loss = (
+                (histogram_loss_1 * other1.sample_size)
+                + (histogram_loss_2 * other2.sample_size)
+            ) / (other1.sample_size + other2.sample_size)
+        except AttributeError:
+            sample_size_1 = sum(other1._stored_histogram["histogram"]["bin_counts"])
+            sample_size_2 = sum(other2._stored_histogram["histogram"]["bin_counts"])
+            calculated_loss = (
+                (histogram_loss_1 * sample_size_1) + (histogram_loss_2 * sample_size_2)
+            ) / (sample_size_1 + sample_size_2)
```
JGSweets, line 256: Is `sample_size` the correct value or should it be `match_count`? Do we have tests evaluating this?

JGSweets: Also, I think loss should be the sum, right? It's the aggregate loss of both being inserted into the `_stored_histogrtam`.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -179,23 +179,93 @@ def _add_helper_merge_profile_histograms(
...
+        _, histogram_loss_2 = self._assimilate_histogram(
+            from_hist_entity_count_per_bin=other2._stored_histogram["histogram"][
+                "bin_counts"
+            ],
+            from_hist_bin_edges=other2._stored_histogram["histogram"]["bin_edges"],
+            dest_hist_entity_count_per_bin=new_entity_count_by_bin,
+            dest_hist_bin_edges=ideal_bin_edges,
+            dest_hist_num_bin=ideal_count_of_bins,
+        )
+        try:
```
JGSweets, line 246: we could separate out getting sizes from the actual calc so the calc is not within the try itself. that is if sizes are needed.

JGSweets: see below

dataprofiler/profilers/numerical_column_stats.py
ksneab7: sure match_count is probably right.
so in this calculation we are taking the the loss of each histogram and using the size of the dataset associated with that loss as a weight factor. So yes it is a sum, in this case I am doing a weighted sum.

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1280,23 +1377,30 @@ def _regenerate_histogram(
...
                 ) ** 2 * count_in_left_bin
 
                 # allocate leftovers to the right bin
-                new_bin_counts[new_bin_id + 1] += bin_count - count_in_left_bin
+                dest_hist_entity_count_per_bin[new_bin_id + 1] += (
+                    bin_count - count_in_left_bin
+                )
                 hist_loss += (
-                    ((new_bin_edge[2] - new_bin_edge[1]) - (bin_edge[1] - bin_edge[0]))
+                    ((new_bin_edge[2] + new_bin_edge[1]) - (bin_edge[1] + bin_edge[0]))
```
ksneab7, line 1391: This was the last bug to be fixed

taylorfturner (approved):

JGSweets (approved):

dataprofiler/profilers/histogram_utils.py
```diff
@@ -1,19 +1,241 @@
...
+    "fd": _calc_fd_bin_width_from_profile,
+    "rice": _calc_rice_bin_width_from_profile,
+    "scott": _calc_scott_bin_width_from_profile,
+    "sqrt": _calc_sqrt_bin_width_from_profile,
+    "sturges": _calc_sturges_bin_width_from_profile,
+}
+
+
 def _get_bin_edges(
     a: np.ndarray,
```
tyfarnan, line 240: think we might like a more descriptive variable name here?

JGSweets: A future fix, not in code we changed and is actually how `numpy` also implements.

taylorfturner: good call

tyfarnan (approved):

Issue #747, opened 2023-01-05
MyPy Configuration Issue LSP in profilers


JGSweets:
Currently, we are avoiding some MyPy errors by ignoring override errors:
https://github.com/capitalone/DataProfiler/blob/main/setup.cfg#L18

Need to address / understand how to address this issue so we can remove the flag. 

This is related to LSP. https://en.wikipedia.org/wiki/Liskov_substitution_principle

junholee6a:
When I remove the `disable_error_code = override` from setup.cfg and run `pre-commit run --all-files`, mypy says there are 12 errors, and they're all in `dataprofiler/profilers/`. But if I run `mypy dataprofiler/data_readers`, it shows 7 errors in `dataprofiler/data_readers/`. So is the pre-commit only type checking `dataprofiler/profilers/`, and is this unintentional?

JGSweets:
`data_readers` should be included. I'm not seeing a specific config that would suggest otherwise. 

Since `data_readers` is in a directory, is it possible it is bc of relative paths?

JGSweets:
@junholee6a maybe provide a sample of the `data_reader` errors from pre-commit?

taylorfturner:
Was just typing this haha... @JGSweets beat me to it

junholee6a:
Output of `pre-commit run --all-files`:
<img width="1421" alt="Screenshot 2023-05-02 at 6 10 31 PM" src="https://user-images.githubusercontent.com/53921230/235796795-6bb1cdb1-d430-4210-a92c-c4b7f7462fbb.png">

Output of `mypy dataprofiler/data_readers`
<img width="1278" alt="Screenshot 2023-05-02 at 6 11 36 PM" src="https://user-images.githubusercontent.com/53921230/235796958-fb85ab2a-4a46-430b-a863-85b5c2880693.png">

taylorfturner:
Interesting... the `pre-commit` config should only exclude what is here: https://github.com/capitalone/DataProfiler/blob/main/.pre-commit-config.yaml#L40

junholee6a:
I was able to get this output from `pre-commit run --all-files` by creating a type error in `dataprofiler/data_readers/csv_data.py`. The first error is in `dataprofiler/data_readers/`, so it seems like the pre-commit is checking all of the files as it should. It's probably checking them with a different configuration or dependency list from a basic `mypy` command though. This might be okay as it is since it seems to catch egregious type errors.
[...]

PR #819, opened 2023-05-14
Fix override errors

junholee6a:
Issue: https://github.com/capitalone/DataProfiler/issues/747

* Make `NumericStatsMixin` a subclass of `BaseColumnProfiler`
  * [Comments](https://github.com/capitalone/DataProfiler/blob/a8c3d727275b8f9e89f9848d5fc4efa0b965b68f/dataprofiler/profilers/numerical_column_stats.py#L36) indicate that `NumericStatsMixin` is supposed to be a subclass of `BaseColumnProfiler`
  * Several classes have both `NumericStatsMixin` and `BaseColumnProfiler` as superclasses. The subclasses, `NumericStatsMixin`, and `BaseColumnProfiler` all implement a `diff` method, with different method signatures. This causes the subclass to override two parent methods at once, which isn't ideal
* Make subclass methods compliant with Liskov Substitution Principle (LSP), which says you need to be able to substitute in a subclass wherever a superclass goes. This means that the parameter types for overridden methods cannot be any narrower than what the superclass has. We resolve this by making the method signatures for subclass methods the same as those of the superclass. In case someone passes in a value of the wrong type, the methods already check the types and throw errors accordingly.
* Relevant superclasses are `BaseColumnProfiler` and `BaseCompiler`
* Remove the line in `setup.cfg` which causes mypy to ignore `override` errors

### DIFF
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 from collections import defaultdict
 from operator import itemgetter
+from typing import cast
 
 from pandas import DataFrame, Series
 
@@ -72,7 +73,7 @@ class CategoricalColumn(BaseColumnProfiler):
         )
         return merged_profile
 
-    def diff(self, other_profile: CategoricalColumn, options: dict = None) -> dict:
+    def diff(self, other_profile: BaseColumnProfiler, options: dict = None) -> dict:
         """
         Find the differences for CategoricalColumns.
 
@@ -81,7 +82,9 @@ class CategoricalColumn(BaseColumnProfiler):
         :return: the CategoricalColumn differences
         :rtype: dict
         """
+        # Make sure other_profile's type matches this class
         differences: dict = super().diff(other_profile, options)
+        other_profile = cast(CategoricalColumn, other_profile)
 
         differences["categorical"] = utils.find_diff_of_strings_and_bools(
             self.is_match, other_profile.is_match
--- a/dataprofiler/profilers/data_labeler_column_profile.py
+++ b/dataprofiler/profilers/data_labeler_column_profile.py
@@ -318,7 +318,7 @@ class DataLabelerColumn(BaseColumnProfiler):
         """
         return self.profile
 
-    def diff(self, other_profile: DataLabelerColumn, options: dict = None) -> dict:
+    def diff(self, other_profile: BaseColumnProfiler, options: dict = None) -> dict:
         """
         Generate differences between the orders of two DataLabeler columns.
 
@@ -326,7 +326,9 @@ class DataLabelerColumn(BaseColumnProfiler):
         appropriate output formats
         :rtype: dict
         """
+        # Make sure other_profile's type matches this class
         differences = super().diff(other_profile, options)
+        other_profile = cast(DataLabelerColumn, other_profile)
 
         self_labels = None
         if self.sample_size:
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -31,7 +31,7 @@ class abstractstaticmethod(staticmethod):
     __isabstractmethod__ = True
 
 
-class NumericStatsMixin(metaclass=abc.ABCMeta):  # type: ignore
+class NumericStatsMixin(BaseColumnProfiler, metaclass=abc.ABCMeta):
     """
     Abstract numerical column profile subclass of BaseColumnProfiler.
 
@@ -201,8 +201,8 @@ class NumericStatsMixin(metaclass=abc.ABCMeta):  # type: ignore
 
     def _add_helper(
         self,
-        other1: NumericStatsMixin,
-        other2: NumericStatsMixin,
+        other1: BaseColumnProfiler,
+        other2: BaseColumnProfiler,
     ) -> None:
         """
         Help merge profiles.
@@ -211,6 +211,12 @@ class NumericStatsMixin(metaclass=abc.ABCMeta):  # type: ignore
         :param other2: profile2 being added to self
         :return: None
         """
+        if not (
+            isinstance(other1, NumericStatsMixin)
+            and isinstance(other2, NumericStatsMixin)
+        ):
+            raise ValueError("Parameters must be of type NumericStatsMixin.")
+
         BaseColumnProfiler._merge_calculations(
             self._NumericStatsMixin__calculations,
             other1._NumericStatsMixin__calculations,
@@ -362,7 +368,7 @@ class NumericStatsMixin(metaclass=abc.ABCMeta):  # type: ignore
 
     def diff(
         self,
-        other_profile: NumericStatsMixin,
+        other_profile: BaseColumnProfiler,
         options: dict = None,
     ) -> dict:
         """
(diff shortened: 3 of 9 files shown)


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -81,7 +82,9 @@ def diff(self, other_profile: CategoricalColumn, options: dict = None) -> dict:
         :return: the CategoricalColumn differences
         :rtype: dict
         """
+        # Make sure other_profile's type matches this class
         differences: dict = super().diff(other_profile, options)
+        other_profile = cast(CategoricalColumn, other_profile)
```
taylorfturner, line 87: Have you attempted to run this without adding the `cast()`s --> just changing the types to the base type / super type?

taylorfturner: @junholee6a

junholee6a: Yep, some of the `diff()` methods work fine without the `cast()` while some require them. I casted in all of them because it seems intuitive; at that point, the code has checked that `other_profile` is an instance of the current class, so we can treat it as such. Would you prefer removing the `cast()`s that aren't strictly needed?

taylorfturner: I'm thinking about LSP and the `cast()` part feels like a violation of LSP. If we assume LSP, my intuition is that `mypy` should pass without casting because the super type / base class type should be able to pass all the necessary things for `mypy` to say "yeah, all is well" without the `cast()`. My inference of this code is that the `cast()` is not solving the LSP issue fundamentally. 

So, I would say avoid `cast()` where possible -- especially in the case of this issue which appears to be LSP fundamentally. CC @JGSweets

junholee6a: I definitely agree that this is a workaround and violates LSP in principle, but I think the reason is that the code throws an error for certain inputs rather than because of `cast()`. I guess I'm thinking more about the Python runtime than mypy specifically. If you still want me to remove the excess `cast()`s just let me know and I'll add a commit to this branch.

An alternative solution: To strictly adhere to LSP, we could remove (or rename) the `diff()` method from the superclasses `NumericStatsMixin` and `BaseColumnProfiler`, then implement `diff()` for each subclass without overriding anything. This way, the subclasses can have the correct argument types without violating LSP. Would that possibly be a better solution? I can make another PR for that if you want to compare it to this one. Thanks!

taylorfturner: I definitely am of the opinion we should remove `cast()` where at all possible. 

I'm super game if you want to pop up a PR (outside this one) around the `alternative solution` you proposed above. To keep things a bit more clean on the PRs, I'd recommend:

- remove the casts where possible in this PR
- comment on the issue related to this PR on the remaining `cast()` that violate `LSP` 
- then in a follow-up PR propose the alternative solution to fix the fundamental LSP issue 

trying to avoid PRs getting too unwieldy. Thx, @junholee6a!

taylorfturner: also, needs an `update branch` when you have a moment

junholee6a: Thanks, I'll proceed with your recommendations!

taylorfturner: Sounds good -- thanks, @junholee6a!

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -211,6 +211,12 @@ def _add_helper(
         :param other2: profile2 being added to self
         :return: None
         """
+        if not (
+            isinstance(other1, NumericStatsMixin)
+            and isinstance(other2, NumericStatsMixin)
+        ):
+            raise ValueError("Parameters must be of type NumericStatsMixin.")
+
```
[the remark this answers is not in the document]
taylorfturner: @junholee6a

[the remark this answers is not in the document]
junholee6a: Removing this if-statement causes the mypy errors in the attached image, because the if-statement implicitly casts `other1` and `other2` to type `NumericStatsMixin`.

Like the other subclasses we discussed above, some of the methods in `NumericStatsMixin` override superclass methods despite not all input types being valid, which in principle violates LSP. I was thinking of fixing this in the same way as the other subclasses in the next PR.

Screenshot 2023-05-17 at 2 17 10 PM

taylorfturner (approved):

dataprofiler/profilers/numerical_column_stats.py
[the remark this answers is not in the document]
taylorfturner: got it -- just make sure to bullet point out the fixes in the next PR in the issue too https://github.com/capitalone/DataProfiler/issues/747

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -318,15 +318,17 @@ def report(self, remove_disabled_flag: bool = False) -> dict:
...
-    def diff(self, other_profile: DataLabelerColumn, options: dict = None) -> dict:
+    def diff(self, other_profile: BaseColumnProfiler, options: dict = None) -> dict:
         """
         Generate differences between the orders of two DataLabeler columns.
 
         :return: Dict containing the differences between orders in their
         appropriate output formats
         :rtype: dict
         """
+        # Make sure other_profile's type matches this class
```
JGSweets, line 329: as a temp solution, would it work to assert type instead of a cast?
```python
assert type(other_profile) is DataLabelerColumn
```
that way it clearly describes an issue when typing doesn't clarify.

JGSweets: sorry to prolong, just don't want to regress on clarity to the user.

taylorfturner: the mypy error will arise when the `cast()` is removed

taylorfturner: so although the `assert` raises an issue at run time, during `mypy` checks, the error will arise and pre-commit will fail

taylorfturner: > sorry to prolong, just don't want to regress on clarity to the user.

not at all -- definitely not wanting to regress, either

JGSweets: Realizing now that the diff already does this assert

junholee6a: That `assert` does an implicit cast to `DataLablerColumn` so it would work. But these `diff()` methods call `super().diff()` right before casting, which already does an equivalent of `assert` to check `other_profile`'s type and raise an appropriate `TypeError`. `assert` is definitely more readable though so I'll replace the `cast()`s with `assert` statements if you'd prefer.
https://github.com/capitalone/DataProfiler/blob/df86943d2f06610f51c07e33e8e990ed2ff54e1b/dataprofiler/profilers/base_column_profilers.py#L189

taylorfturner: yeah the `assert type(<class>) is <base_class_name>` seeminly does pass the `mypy` checks

junholee6a: > Realizing now that the diff already does this assert

My bad, didn't see this while I was typing

taylorfturner: although the `super().diff()` does that effectively like ya'll are saying too

JGSweets (approved):

PR #824, opened 2023-05-18
Add option for sampling_ratio at the dataset level

joshuart:
This PR enables users to set an optional sampling_ratio parameter that will determine the ratio of the data that will be profiled. sampling_ratio should be a float greater than 0 and less than or equal to 1. This PR only enables the option to be set. A future PR will implement the actual sampling logic.

### DIFF
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -1133,6 +1133,7 @@ class StructuredOptions(BaseOption):
         self,
         null_values: dict[str, re.RegexFlag | int] = None,
         column_null_values: dict[int, dict[str, re.RegexFlag | int]] = None,
+        sampling_ratio: float = None,
     ) -> None:
         """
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/profiler_options.py
```diff
@@ -1133,6 +1133,7 @@ def __init__(
         self,
         null_values: dict[str, re.RegexFlag | int] = None,
         column_null_values: dict[int, dict[str, re.RegexFlag | int]] = None,
+        sampling_ratio: float = None,
```
taylorfturner, line 1136: let's set the default to `0.2` actually. To match the current default setting in the core library [here](https://github.com/capitalone/DataProfiler/blob/df86943d2f06610f51c07e33e8e990ed2ff54e1b/dataprofiler/profilers/base_column_profilers.py#L28)

taylorfturner (changes requested):

PR #825, opened 2023-05-18
Sampling ratio option

joshuart:
This PR enables users to set an optional sampling_ratio parameter that will determine the ratio of the data that will be profiled. sampling_ratio should be a float greater than 0 and less than or equal to 1. This PR only enables the option to be set. A future PR will implement the actual sampling logic.

### DIFF
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -1133,6 +1133,7 @@ class StructuredOptions(BaseOption):
         self,
         null_values: dict[str, re.RegexFlag | int] = None,
         column_null_values: dict[int, dict[str, re.RegexFlag | int]] = None,
+        sampling_ratio: float = None,
     ) -> None:
         """
         Construct the StructuredOptions object with default values.
@@ -1164,6 +1165,9 @@ class StructuredOptions(BaseOption):
         :vartype null_replication_metrics: BooleanOptions
         :ivar null_values: option set for defined null values
         :vartype null_values: Union[None, dict]
+        :ivar sampling_ratio: What ratio of the input data to sample.
+            Float value > 0 and <= 1
+        :vartype sampling_ratio: Union[None, float]
         """
         # Option variables
         self.multiprocess = BooleanOption()
@@ -1180,6 +1184,7 @@ class StructuredOptions(BaseOption):
         # Non-Option variables
         self.null_values = null_values
         self.column_null_values = column_null_values
+        self.sampling_ratio = sampling_ratio
 
     @property
     def enabled_profiles(self) -> list[str]:
@@ -1189,6 +1194,7 @@ class StructuredOptions(BaseOption):
         properties = self.properties
         properties.pop("null_values")
         properties.pop("column_null_values")
+        properties.pop("sampling_ratio")
         for key, value in properties.items():
             if value.is_enabled:
                 enabled_profiles.append(key)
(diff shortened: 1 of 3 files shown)


dataprofiler/profilers/profiler_options.py
```diff
@@ -1133,6 +1133,7 @@ def __init__(
         self,
         null_values: dict[str, re.RegexFlag | int] = None,
         column_null_values: dict[int, dict[str, re.RegexFlag | int]] = None,
+        sampling_ratio: float = 0.2,
```
taylorfturner, line 1136: this is great!

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
```diff
@@ -276,6 +280,22 @@ def test_validate(self):
...
+            "{}.sampling_ratio must be greater than 0.0 and less than or equal to 1.0".format(
+                optpth
+            )
+        ]
+        # Test ratio is not a float
+        option.set({"sampling_ratio": 1})
+        self.assertEqual(expected_error_type, option._validate_helper())
+        # Test ratio is out of range
+        option.set({"sampling_ratio": 2.5})
+        self.assertEqual(expected_error_value, option._validate_helper())
```
taylorfturner, line 297: i'd recommend for a slightly more encompassing test to also test the down-side range of the bounds of the option. So you test two things, which are great:

- one right at the top end 
- another past the top end

Should add another test, to fully validate the logic, where `sampling_ratio <= 0.0`

taylorfturner (changes requested):

dataprofiler/profilers/profiler_options.py
```diff
@@ -1274,6 +1281,25 @@ def _validate_helper(self, variable_path: str = "StructuredOptions") -> list[str
                 "a re.RegexFlag".format(variable_path)
             )
 
+        if self.sampling_ratio is not None and not isinstance(
+            self.sampling_ratio, float
+        ):
+            errors.append(
+                "{}.sampling_ratio must be either None or an float".format(
```
JGSweets, line 1288: small fix: `or a float`

taylorfturner: +1

JGSweets: Would it be worth to accept an `int` bc `1` is valid.

taylorfturner: yeah I was thinking that too actually -- I think that's a fair point

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -30,6 +30,8 @@ def test_default_profiler_options(self, *mocks):
                 self.assertFalse(profile.options.properties[column].is_enabled)
             elif column == "null_values" or column == "column_null_values":
                 self.assertIsNone(profile.options.properties[column])
+            elif column == "sampling_ratio":
```
JGSweets, line 33: I think we need to also add tests below in the validate to ensure errors are done properly. Do we also need test to set?

JGSweets: never mind, this isn't the spot!

dataprofiler/tests/profilers/profiler_options/test_structured_options.py
taylorfturner: @joshuart let me know if you have any questions

taylorfturner (approved):

JGSweets (approved):

PR #828, opened 2023-05-22
added new options to docs

ksneab7:


docs/source/profiler.rst
```diff
@@ -757,6 +760,7 @@ Below is an breakdown of all the options.
         the default labeler 
       * max_sample_size - (Int) The max number of samples for the data 
         labeler
+	* **row_statistics** - (Boolean) Option to enable/disable row statistics calculations
```
JGSweets, line 763: does this tabbing need to be fixed?

taylorfturner: +1 one indentation... I would think this shouldn't go in the `data_labeler` section of options, no?

taylorfturner (changes requested):

taylorfturner:
@ksneab7 also would recommend PRing from your fork's `gh-pages` branch so it will build a version of docs so you can validate the fix. 

Also I think this would be a merge to `pages/memory-optimization`, right?

PR #840, opened 2023-05-23
Fix lsp violations

junholee6a:
Issue: https://github.com/capitalone/DataProfiler/issues/747

* Makes the superclasses `BaseColumnProfiler`, `NumericStatsMixin`, and
`BaseCompiler generic`, to avoid casting in subclass diff() methods and
violating LSP in principle.
* Removes unneeded casting in `diff()` methods

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -6,7 +6,7 @@ import abc
 import copy
 import itertools
 import warnings
-from typing import Any, Callable, Dict, List, cast
+from typing import Any, Callable, Dict, List, TypeVar, cast
 
 import numpy as np
 import numpy.typing as npt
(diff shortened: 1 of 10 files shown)


JGSweets:
This is amazing work!

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -272,12 +275,6 @@ def _add_helper(
         :param other2: profile2 being added to self
         :return: None
         """
-        if not (
-            isinstance(other1, NumericStatsMixin)
-            and isinstance(other2, NumericStatsMixin)
-        ):
-            raise ValueError("Parameters must be of type NumericStatsMixin.")
```
JGSweets, line 279: I think we still need this for the error, no? Or was this added for the LSP last time and this is handled by the classes using the mixins?

JGSweets (approved):

junholee6a:
@JGSweets thank you! The article you found was a great solution

PR #841, opened 2023-05-24
StructuredOptions: `hhl_row_hashing`

micdavis:
* added `hhl_row_hashing` option
Default is original row hashing method: `hhl_row_hashing=False`

### DIFF
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -1178,6 +1178,7 @@ class StructuredOptions(BaseOption):
         self,
         null_values: dict[str, re.RegexFlag | int] = None,
         column_null_values: dict[int, dict[str, re.RegexFlag | int]] = None,
+        hll_row_hashing: bool = False,
     ) -> None:
         """
         Construct the StructuredOptions object with default values.
@@ -1228,15 +1229,17 @@ class StructuredOptions(BaseOption):
         # Non-Option variables
         self.null_values = null_values
         self.column_null_values = column_null_values
+        self.hll_row_hashing = hll_row_hashing
 
     @property
     def enabled_profiles(self) -> list[str]:
         """Return a list of the enabled profilers for columns."""
         enabled_profiles = list()
-        # null_values and column_null_values do not have is_enabled
+        # null_values, column_null_values, and hll_row_hashing do not have is_enabled
         properties = self.properties
         properties.pop("null_values")
         properties.pop("column_null_values")
+        properties.pop("hll_row_hashing")
         for key, value in properties.items():
             if value.is_enabled:
                 enabled_profiles.append(key)
@@ -1275,6 +1278,7 @@ class StructuredOptions(BaseOption):
         properties = self.properties
         properties.pop("null_values")
         properties.pop("column_null_values")
+        properties.pop("hll_row_hashing")
         for column in properties:
             if not isinstance(self.properties[column], prop_check[column]):
                 errors.append(
@@ -1322,6 +1326,11 @@ class StructuredOptions(BaseOption):
                 "of type str and values == 0 or are instances of "
                 "a re.RegexFlag".format(variable_path)
             )
+
+        if self.hll_row_hashing is not None and not isinstance(
+            self.hll_row_hashing, bool
+        ):
+            errors.append("{}.hll_row_hashing must be either None or a bool ")
 
         if (
             isinstance(self.category, CategoricalOptions)
--- a/dataprofiler/profilers/profiler_options.py
+++ b/dataprofiler/profilers/profiler_options.py
@@ -985,7 +985,9 @@ class CorrelationOptions(BaseInspectorOptions):
 class HyperLogLogOptions(BooleanOption):
     """Options for alternative method of gathering unique row count."""
 
-    def __init__(self, is_enabled: bool = False, seed: int = 0) -> None:
+    def __init__(
+        self, is_enabled: bool = False, seed: int = 0, register_count: int = 10
+    ) -> None:
         """
         Initialize options for the hyperloglog method of gathering unique row count.
 
@@ -993,9 +995,12 @@ class HyperLogLogOptions(BooleanOption):
         :vartype is_enabled: bool
         :ivar seed: seed used to set HLL hashing function
         :vartype seed: int
+        :ivar register_count: number of registers is equal to 2^register_count
+        :vartype register_count: int
         """
         BooleanOption.__init__(self, is_enabled=is_enabled)
         self.seed = seed
+        self.register_count = register_count
 
     def _validate_helper(self, variable_path: str = "HyperLogLogOptions") -> list[str]:
         """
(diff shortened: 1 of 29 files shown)


micdavis:
interested to know if `hll_row_hashing_enabled` makes more sense, or something different that what I have

dataprofiler/profilers/profiler_options.py
```diff
@@ -1228,15 +1229,17 @@ def __init__(
         # Non-Option variables
         self.null_values = null_values
         self.column_null_values = column_null_values
+        self.hll_row_hashing = hll_row_hashing
```
JGSweets, line 1232: this needs to be under row_statistics since that's the hierarchy in the code. We will have to make a class for row statistics instead of using booleanoptions which will be similar to correlations like on https://github.com/capitalone/DataProfiler/pull/841/files#diff-00da4b3383cf04d1ea70e7d179d78433f0bfde3ac6909c791eb92363c5a34f3bR1225

dataprofiler/profilers/profiler_options.py
```diff
@@ -982,6 +982,64 @@ def _validate_helper(self, variable_path: str = "CorrelationOptions") -> list[st
         return errors
 
 
+class HyperLogLogOptions(BooleanOption):
```
micdavis, line 985: Created a nested HLL options class because seed is only pertinent when HLL is enabled.

dataprofiler/tests/profilers/profiler_options/test_row_statistics_options.py
```diff
@@ -0,0 +1,71 @@
+from dataprofiler.profilers.profiler_options import RowStatisticsOptions
```
micdavis, line 1: Unsure about these tests, not sure how to handle nested option testing.

dataprofiler/profilers/profiler_options.py
```diff
@@ -982,6 +982,64 @@ def _validate_helper(self, variable_path: str = "CorrelationOptions") -> list[st
...
+        """
+        Initialize options for the hyperloglog method of gathering unique row count.
+
+        :ivar is_enabled: boolean option to enable/disable.
+        :vartype is_enabled: bool
+        :ivar seed: seed used to set HLL hashing function
+        :vartype seed: int
+        """
+        BooleanOption.__init__(self, is_enabled=is_enabled)
+        self.seed = seed
```
micdavis, line 998: should I also include "p" which represents the 2^p registers in the HLL table

ksneab7: Yes per conversation with the team we should implement and option for this

taylorfturner (changes requested):
blocking for 0.9.0 release

dataprofiler/profilers/profiler_options.py
JGSweets: Let's rename to `UniqueCountOptions`. Here we should have a str choice to use `full` or `hll_row_hashing` in addition we can have the hll_options.

dataprofiler/profilers/profiler_options.py
```diff
@@ -982,6 +982,74 @@ def _validate_helper(self, variable_path: str = "CorrelationOptions") -> list[st
         return errors
 
 
+class HyperLogLogOptions(BooleanOption):
```
taylorfturner, line 985: agree -- let's make this a `UniqueCountOptions` with more granular toggles of `full` or `hll_row_hashing` options beneath that.

taylorfturner (changes requested):

dataprofiler/tests/profilers/profiler_options/test_hyper_log_log_options.py
```diff
@@ -0,0 +1,185 @@
...
+            {"is_enabled": True, "seed": 5, "register_count": 8}, option.properties
+        )
+
+    def test_validate_helper(self):
+        super().test_validate_helper()
+
+        optpth = self.get_options_path()
+
+        # Default configuration
+        option = self.get_options(seed=0, register_count=10)
```
JGSweets, line 42: In order to test default, I suggest not overriding. i.e. just do `option = self.get_options()`

taylorfturner: +1

dataprofiler/profilers/profiler_options.py
```diff
@@ -982,6 +982,130 @@ def _validate_helper(self, variable_path: str = "CorrelationOptions") -> list[st
...
+            errors.append(f"{variable_path}.register_count must be an integer.")
+        elif self.register_count <= 0:
+            errors.append(f"{variable_path}.register_count must be greater than 0.")
+
+        if not isinstance(self.seed, int):
+            errors.append(f"{variable_path}.seed must be an integer.")
+        return errors
+
+
+class UniqueCountOptions(BooleanOption):
```
micdavis, line 1026: curious if `RowStatisticsOptions`, `UniqueCountOptions`, or `HyperLogLogOptions` should inherit `BooleanOption` or `BaseInspectorOptions`

dataprofiler/profilers/profiler_options.py
```diff
@@ -982,6 +982,130 @@ def _validate_helper(self, variable_path: str = "CorrelationOptions") -> list[st
...
+        hll_hashing: bool = False,
+    ) -> None:
+        """
+        Initialize options for unique row counts.
+
+        :ivar is_enabled: boolean option to enable/disable.
+        :vartype is_enabled: bool
+        """
+        BooleanOption.__init__(self, is_enabled=is_enabled)
+        self.full_hashing = BooleanOption(is_enabled=full_hashing)
```
micdavis, line 1042: should `full_hashing` be a `BooleanOption`?

dataprofiler/tests/profilers/profiler_options/test_hyper_log_log_options.py
micdavis: fixed

dataprofiler/profilers/profiler_options.py
```diff
@@ -982,6 +982,74 @@ def _validate_helper(self, variable_path: str = "CorrelationOptions") -> list[st
...
+
+    def __init__(self, is_enabled: bool = False) -> None:
+        """
+        Initialize options for row statistics.
+
+        :ivar is_enabled: boolean option to enable/disable.
+        :vartype is_enabled: bool
+        """
+        BooleanOption.__init__(self, is_enabled=is_enabled)
+        self.hll_row_hashing = HyperLogLogOptions(is_enabled=False)
```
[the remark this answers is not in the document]
micdavis: changed, lmk your thoughts

dataprofiler/profilers/profiler_options.py
taylorfturner: @JGSweets

dataprofiler/profilers/profiler_options.py
taylorfturner: I think this would be a boolean since there are no properties for `full_hashing` ... its just `True` / `False` binary

dataprofiler/profilers/profiler_options.py
taylorfturner: I'm wondering it should be a `BaseInspectorOption` (e.g. `TextProfilerOptions` in `UnstructuredOptions`)

dataprofiler/profilers/profiler_options.py
JGSweets: This is a good idea, but the issue is what happens if someone specifies true for both? Instead, we might want a string input that selects one or the other. good call below in the validate. Or we need a way to toggle. Open to thoughts.

taylorfturner (approved):

PR #842, opened 2023-05-24
StructuredProfiler: HLLRowHashing

micdavis:
* Adding HLLRowHashing
* Merge after #841

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,8 +43,62 @@ repos:
       - id: mypy
         exclude: (^dataprofiler/tests/|^resources/|^examples|venv*/)
         language_version: python3
-        additional_dependencies: ['types-setuptools', 'types-python-dateutil',
-        'types-requests', 'types-chardet', 'types-six', 'numpy']
+        additional_dependencies: # Keep up-to-date with the respective requirement files
+          [
+            # requirements.txt
+            h5py>=2.10.0,
+            wheel>=0.33.1,
+            numpy>=1.22.0,
+            pandas>=1.1.2,
+            python-dateutil>=2.7.5,
+            pytz>=2020.1,
+            pyarrow>=1.0.1,
+            chardet>=3.0.4,
+            fastavro>=1.0.0.post1,
+            python-snappy>=0.5.4,
+            charset-normalizer>=1.3.6,
+            psutil>=4.0.0,
+            scipy>=1.4.1,
+            requests>=2.28.1,
+            networkx>=2.5.1,
+            typing-extensions>=3.10.0.2,
+
+            # requirements-dev.txt
+            check-manifest>=0.48,
+            black==22.3.0,
+            isort==5.12.0,
+            pre-commit==2.19.0,
+            tox==3.25.1,
+            types-setuptools==67.7.0.1,
+            types-python-dateutil==2.8.19.12,
+            types-requests==2.30.0.0,
+            types-chardet==5.0.4.5,
+            types-six==1.16.21.8,
+
+            # requirements-ml.txt
+            scikit-learn>=0.23.2,
+            keras>=2.4.3,
+            rapidfuzz>=2.6.1,
+            tensorflow>=2.6.4; sys.platform != 'darwin',
+            tensorflow>=2.6.4; sys_platform == 'darwin' and platform_machine != 'arm64',
+            tensorflow-macos>=2.6.4; sys_platform == 'darwin' and platform_machine == 'arm64',
+            tqdm>=4.0.0,
+
+            # requirements-reports.txt
+            'matplotlib>=3.3.3,<3.6.1',
+            seaborn>=0.11.1,
+
+            # requirements-test.txt
+            coverage>=5.0.1,
+            dask>=2.29.0,
+            fsspec>=0.3.3,
+            pytest>=6.0.1,
+            pytest-cov>=2.8.1,
+            pytest-xdist>=2.1.0,
+            pytest-forked>=1.3.0,
+            toolz>=0.10.0,
+            memray>=1.7.0,
+          ]
   # Check-manifest: ensures required non-Python files are included in MANIFEST.in
   # https://github.com/mgedmin/check-manifest/blob/master/.pre-commit-hooks.yaml
   - repo: https://github.com/mgedmin/check-manifest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -62,6 +62,7 @@ repos:
             requests>=2.28.1,
             networkx>=2.5.1,
             typing-extensions>=3.10.0.2,
+            HLL>=2.0.3,
 
             # requirements-dev.txt
             check-manifest>=0.48,
@@ -108,7 +109,7 @@ repos:
         additional_dependencies: ['h5py', 'wheel', 'future', 'numpy', 'pandas',
         'python-dateutil', 'pytz', 'pyarrow', 'chardet', 'fastavro',
         'python-snappy', 'charset-normalizer', 'psutil', 'scipy', 'requests',
-        'networkx','typing-extensions']
+        'networkx','typing-extensions', 'HLL']
   # Pyupgrade - standardize and modernize Python syntax for newer versions of the language
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.3.0
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1592,7 +1592,8 @@ class StructuredProfiler(BaseProfiler):
             )
         # Check hll seed
         if (
-            self.options.row_statistics.unique_count.hll.seed
+            self.options.row_statistics.unique_count.hashing_method == "hll"
+            and self.options.row_statistics.unique_count.hll.seed
             != other.options.row_statistics.unique_count.hll.seed
         ):
             raise ValueError(
@@ -1601,7 +1602,8 @@ class StructuredProfiler(BaseProfiler):
             )
         # Check hll register count
         if (
-            self.options.row_statistics.unique_count.hll.register_count
+            self.options.row_statistics.unique_count.hashing_method == "hll"
+            and self.options.row_statistics.unique_count.hll.register_count
             != other.options.row_statistics.unique_count.hll.register_count
         ):
             raise ValueError(
@@ -1951,18 +1953,17 @@ class StructuredProfiler(BaseProfiler):
     def _get_unique_row_ratio(self) -> float | None:
         """Return unique row ratio."""
         if (
-            self.options.row_statistics.is_enabled
-            and self.options.row_statistics.unique_count.is_enabled
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.unique_count.is_enabled
         ):
-            if self.total_samples:
-                if isinstance(self.hashed_row_object, dict):
-                    return len(self.hashed_row_object) / self.total_samples
-                elif isinstance(self.hashed_row_object, HyperLogLog):
-                    return (
-                        int(self.hashed_row_object.cardinality()) / self.total_samples
-                    )
-            return 0
-        return None
+            return None
+
+        if self.total_samples:
+            if isinstance(self.hashed_row_object, dict):
+                return len(self.hashed_row_object) / self.total_samples
+            elif isinstance(self.hashed_row_object, HyperLogLog):
+                return int(self.hashed_row_object.cardinality()) / self.total_samples
+        return 0
 
     def _get_row_is_null_ratio(self) -> float:
         """Return whether row is null ratio."""
@@ -1979,17 +1980,18 @@ class StructuredProfiler(BaseProfiler):
     def _get_duplicate_row_count(self) -> int | None:
         """Return dup row count."""
         if (
-            self.options.row_statistics.is_enabled
-            and self.options.row_statistics.unique_count.is_enabled
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.unique_count.is_enabled
         ):
-            if isinstance(self.hashed_row_object, dict):
-                return self.total_samples - len(self.hashed_row_object)
-            elif isinstance(self.hashed_row_object, HyperLogLog):
-                return max(
-                    0, self.total_samples - int(self.hashed_row_object.cardinality())
-                )
-            return 0
-        return None
+            return None
+
+        if isinstance(self.hashed_row_object, dict):
+            return self.total_samples - len(self.hashed_row_object)
+        elif isinstance(self.hashed_row_object, HyperLogLog):
+            return max(
+                0, self.total_samples - int(self.hashed_row_object.cardinality())
+            )
+        return 0
 
     @utils.method_timeit(name="row_stats")
     def _update_row_statistics(
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1522,7 +1522,9 @@ class StructuredProfiler(BaseProfiler):
         self.row_is_null_count = 0
         if options.row_statistics.hll_row_hashing.is_enabled:
             self.hyper_log_log_table: HyperLogLog = HyperLogLog(
-                p=10, seed=options.row_statistics.hll_row_hashing.seed, sparse=False
+                p=options.row_statistics.hll_row_hashing.register_count,
+                seed=options.row_statistics.hll_row_hashing.seed,
+                sparse=False
             )
         else:
             self.hashed_row_dict: dict = dict()
@@ -1590,9 +1592,8 @@ class StructuredProfiler(BaseProfiler):
             )
 
             if self.options.row_statistics.hll_row_hashing.is_enabled:
-                merged_profile.hyper_log_log_table = self.hyper_log_log_table.merge(
-                    other.hyper_log_log_table
-                )
+                self.hyper_log_log_table.merge(other.hyper_log_log_table)
+                merged_profile.hyper_log_log_table = self.hyper_log_log_table
             else:
                 merged_profile.hashed_row_dict.update(self.hashed_row_dict)
                 merged_profile.hashed_row_dict.update(other.hashed_row_dict)
@@ -1900,7 +1901,7 @@ class StructuredProfiler(BaseProfiler):
         """Return unique row ratio."""
         if self.total_samples:
             if self.options.row_statistics.hll_row_hashing.is_enabled:
-                return int(self.hyper_log_log_table.size()) / self.total_samples
+                return int(self.hyper_log_log_table.cardinality()) / self.total_samples
             return len(self.hashed_row_dict) / self.total_samples
         return 0
 
@@ -1919,7 +1920,7 @@ class StructuredProfiler(BaseProfiler):
     def _get_duplicate_row_count(self) -> int:
         """Retun dup row count."""
         if self.options.row_statistics.hll_row_hashing.is_enabled:
-            return self.total_samples - int(self.hyper_log_log_table.size())
+            return self.total_samples - int(self.hyper_log_log_table.cardinality())
         return self.total_samples - len(self.hashed_row_dict)
 
     @utils.method_timeit(name="row_stats")
@@ -1944,7 +1945,7 @@ class StructuredProfiler(BaseProfiler):
             )
 
         self.total_samples += len(data)
-        if self.options.row_statistics.hll_row_hashing:
+        if self.options.row_statistics.hll_row_hashing.is_enabled:
             for record in data.to_records(index=False):
                 self.hyper_log_log_table.add(record.tobytes())
         else:
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1739,7 +1739,7 @@ class StructuredProfiler(BaseProfiler):
                 ),
                 "duplicate_row_count": utils.find_diff_of_numbers(
                     self._get_duplicate_row_count(),
-                    other_profile._get_row_is_null_ratio(),
+                    other_profile._get_duplicate_row_count(),
                 ),
                 "correlation_matrix": utils.find_diff_of_matrices(
                     self.correlation_matrix, other_profile.correlation_matrix
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1520,7 +1520,7 @@ class StructuredProfiler(BaseProfiler):
         # Structured specific properties
         self.row_has_null_count = 0
         self.row_is_null_count = 0
-        self.hashed_row_object: dict | HyperLogLog = None
+        self.hashed_row_object: HyperLogLog | dict
         self._profile: list[StructuredColProfiler] = []  # type: ignore[assignment]
         self._col_name_to_idx: dict[str | int, list[int]] = defaultdict(list)
         self.correlation_matrix: np.ndarray = None  # type: ignore[assignment]
@@ -1529,13 +1529,13 @@ class StructuredProfiler(BaseProfiler):
         # capitalone/synthetic-data specific metrics
         self._null_replication_metrics: dict = None  # type: ignore[assignment]
 
-        if options.row_statistics.unique_count.hashing_method == "hll":
+        if isinstance(self.hashed_row_object, HyperLogLog):
             self.hashed_row_object = HyperLogLog(
                 p=options.row_statistics.unique_count.hll.register_count,
                 seed=options.row_statistics.unique_count.hll.seed,
                 sparse=False,
             )
-        elif options.row_statistics.unique_count.hashing_method == "full":
+        elif isinstance(self.hashed_row_object, dict):
             self.hashed_row_object = dict()
         if data is not None:
             self.update_profile(data)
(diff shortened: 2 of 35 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1519,7 +1520,10 @@ def __init__(
         # Structured specific properties
         self.row_has_null_count = 0
         self.row_is_null_count = 0
-        self.hashed_row_dict: dict = dict()
+        if options.hll_row_hashing:
+            self.hashed_row_dict: HyperLogLog = HyperLogLog(10, sparse=False)
```
micdavis, line 1524: Arbitrary register count (2 ^ 10). Curious how we want to make this scale with data input. I'm thinking `2 * log2(number of records)` might be good 🤷

dataprofiler/profilers/profile_builder.py
```diff
@@ -1519,7 +1520,10 @@ def __init__(
...
+        else:
+            self.hashed_row_dict: dict = dict()
```
micdavis, line 1526: Also this doesn't seem like the right way to init since it would be two different types, mypy doesn't like it either. Wonder if we would have two different attributes 🤔

dataprofiler/profilers/profile_builder.py
JGSweets: we can allow users to set the values via the `ProfileOptions` being made for this but for choice of value, is there a way to choose relative to an accuracy somehow? as oppose to the value itself? i'm not sure we will always know the true data size. What's the space req for the largest value? is it problematic

dataprofiler/profilers/profile_builder.py
```diff
@@ -1519,7 +1520,10 @@ def __init__(
         # Structured specific properties
         self.row_has_null_count = 0
         self.row_is_null_count = 0
-        self.hashed_row_dict: dict = dict()
+        if options.hll_row_hashing:
```
JGSweets, line 1523: to avoid the if else, set it to a dict and when the if condition hits, it will overwrite it.

dataprofiler/profilers/profile_builder.py
JGSweets: we might want to rename this variable, open to thoughts.

dataprofiler/profilers/profile_builder.py
JGSweets: might need to do an `|` for the var

dataprofiler/profilers/profile_builder.py
```diff
@@ -1582,8 +1586,13 @@ def __add__(  # type: ignore[override]
             merged_profile.row_is_null_count = (
                 self.row_is_null_count + other.row_is_null_count
             )
-            merged_profile.hashed_row_dict.update(self.hashed_row_dict)
-            merged_profile.hashed_row_dict.update(other.hashed_row_dict)
+            if self.options.hll_row_hashing:
```
JGSweets, line 1589: need to ensure options allow the merge, and seed the same

dataprofiler/profilers/profile_builder.py
JGSweets: Can we control the seed so it isn't random every time? set via options, default to something else.

dataprofiler/profilers/profile_builder.py
JGSweets: I wonder if `hll_row_hashing` is a good name. ppl might not know what it is. Anyway we could refer to unique row count and this is probabilistic. If no good options, keep.

dataprofiler/profilers/profiler_options.py
```diff
@@ -1178,6 +1178,7 @@ def __init__(
         self,
         null_values: dict[str, re.RegexFlag | int] = None,
         column_null_values: dict[int, dict[str, re.RegexFlag | int]] = None,
+        hll_row_hashing: bool = False,
```
JGSweets, line 1181: i understand this will disappear once options merges and this gets rebased.

taylorfturner (changes requested):
blocking for 0.9.0 release

requirements.txt
```diff
@@ -14,3 +14,4 @@ scipy>=1.4.1
 requests>=2.28.1
 networkx>=2.5.1
 typing-extensions>=3.10.0.2
+HLL>=2.0.3
```
taylorfturner, line 17: cool

.pre-commit-config.yaml
```diff
@@ -108,7 +109,7 @@ repos:
         additional_dependencies: ['h5py', 'wheel', 'future', 'numpy', 'pandas',
         'python-dateutil', 'pytz', 'pyarrow', 'chardet', 'fastavro',
         'python-snappy', 'charset-normalizer', 'psutil', 'scipy', 'requests',
-        'networkx','typing-extensions']
+        'networkx','typing-extensions', 'HLL']
```
taylorfturner, line 112: nice

.pre-commit-config.yaml
```diff
@@ -62,6 +62,7 @@ repos:
             requests>=2.28.1,
             networkx>=2.5.1,
             typing-extensions>=3.10.0.2,
+            HLL>=2.0.3,
```
taylorfturner, line 65: good

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -60,10 +60,21 @@ def setUpClass(cls):
...
         profiler_options = ProfilerOptions()
         profiler_options.set({"data_labeler.is_enabled": False})
+
+        profiler_options_hll = ProfilerOptions()
+        profiler_options_hll.set(
+            {
+                "data_labeler.is_enabled": False,
+                "row_statistics.unique_count.hashing_method": "hll",
+            }
+        )
```
taylorfturner, line 70: adding any tests into `test_row_statistics_options.py` or `test_profiler_options.py` for this new option?

taylorfturner: although I see in the prior PR you added `hashing_method`

taylorfturner (commented):
I don't see where the condition is for the toggle from `hll` to `full` or `full` to `hll`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1582,8 +1589,21 @@ def __add__(  # type: ignore[override]
             merged_profile.row_is_null_count = (
                 self.row_is_null_count + other.row_is_null_count
             )
-            merged_profile.hashed_row_dict.update(self.hashed_row_dict)
-            merged_profile.hashed_row_dict.update(other.hashed_row_dict)
+
+            if (
+                self.options.row_statistics.unique_count.is_enabled
+                and other.options.row_statistics.unique_count.is_enabled
+            ):
```
JGSweets, line 1596: if they disagree, should we raise an exception instead?

JGSweets: the problem becomes it silently continues and still profiles and can give a false report.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1582,8 +1589,21 @@ def __add__(  # type: ignore[override]
...
+                if isinstance(self.hashed_row_object, dict) and isinstance(
+                    other.hashed_row_object, dict
+                ):
+                    merged_profile.hashed_row_object.update(self.hashed_row_object)
+                    merged_profile.hashed_row_object.update(other.hashed_row_object)
+                elif isinstance(self.hashed_row_object, HyperLogLog) and isinstance(
+                    other.hashed_row_object, HyperLogLog
+                ):
+                    self.hashed_row_object.merge(other.hashed_row_object),
+                    merged_profile.hashed_row_object = self.hashed_row_object
```
JGSweets, line 1606: Similarly here, if they disagree, we currently continue and can silently report false information. An option is to continue, but disable and clear the results maybe with a warning. Or instead to just raise an error. Either way, test conditions are necessary for these cases.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1887,7 +1907,10 @@ def report(self, report_options: dict = None) -> dict:
     def _get_unique_row_ratio(self) -> float:
         """Return unique row ratio."""
         if self.total_samples:
```
JGSweets, line 1909: what if unique_row_ratio is disabled? this would return a result which may not be true to the dataset. Tests would be required here.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1927,18 +1954,39 @@ def _update_row_statistics(
...
-                dict.fromkeys(pd.util.hash_pandas_object(data, index=False), True)
-            )
-        except TypeError:
-            self.hashed_row_dict.update(
-                dict.fromkeys(
-                    pd.util.hash_pandas_object(data.astype(str), index=False), True
-                )
-            )
+        if not self.options.row_statistics.is_enabled:
+            return
```
JGSweets, line 1959: consistent with the other updates, should we move this to where the call happens?

dataprofiler/profilers/profile_builder.py
```diff
@@ -1927,18 +1954,39 @@ def _update_row_statistics(
...
 
+        if self.options.row_statistics.unique_count.is_enabled:
+            self.total_samples += len(data)
```
JGSweets, line 1961: ` self.total_samples += len(data)` this is actually dangerous to be in here, in fact, dangerous to be in `row_statistics`. Likely should move this outside as `self.total_samples` is critical to keep

dataprofiler/profilers/profile_builder.py
```diff
@@ -1927,18 +1954,39 @@ def _update_row_statistics(
...
+                    self.hashed_row_object.update(
+                        dict.fromkeys(
+                            pd.util.hash_pandas_object(data.astype(str), index=False),
+                            True,
+                        )
+                    )
+            elif isinstance(self.hashed_row_object, HyperLogLog):
+                batch_size = 2048
+
+                for batch_ind in range(len(data) // batch_size + 1):
```
JGSweets, line 1979: if `len(data) == 2048` does this cause issues?

dataprofiler/profilers/profile_builder.py
```diff
@@ -2725,14 +2773,20 @@ def save(self, filepath: str = None) -> None:
         :type filepath: String
         :return: None
         """
+        hashing_object_dict: dict = dict()
+        if isinstance(self.hashed_row_object, HyperLogLog):
+            hashing_object_dict = self.hashed_row_object.__dict__
```
JGSweets, line 2778: why the need to convert this to a `__dict__`

dataprofiler/profilers/profile_builder.py
```diff
@@ -2725,14 +2773,20 @@ def save(self, filepath: str = None) -> None:
...
+            hashing_object_dict = self.hashed_row_object
+
         # Create dictionary for all metadata, options, and profile
         data_dict = {
             "total_samples": self.total_samples,
             "encoding": self.encoding,
             "file_type": self.file_type,
             "row_has_null_count": self.row_has_null_count,
             "row_is_null_count": self.row_is_null_count,
-            "hashed_row_dict": self.hashed_row_dict,
```
JGSweets, line 2735: If this worked previously, why the change here when `hashing_object_dict = self.hashed_row_object` is the same?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -60,10 +60,21 @@ def setUpClass(cls):
         cls.aws_dataset = pd.read_csv(cls.input_file_path)
         profiler_options = ProfilerOptions()
         profiler_options.set({"data_labeler.is_enabled": False})
+
+        profiler_options_hll = ProfilerOptions()
```
JGSweets, line 64: this is going to be really expensive to recompute all of aws again with different options. I wonder if we can move these tests elsewhere or create tests that focus directly on the hll work.

JGSweets (changes requested):
Some concerns, but also would like to find a way to efficiently test without running the entire labeler for test efficiency purposes.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -280,15 +329,33 @@ def test_stream_profilers(self, *mocks):
...
+        self.assertEqual(5, profiler.hashed_row_object.cardinality())
+
     def test_correct_unique_row_ratio_test(self):
-        self.assertEqual(2999, len(self.trained_schema.hashed_row_dict))
+        self.assertEqual(2999, len(self.trained_schema.hashed_row_object))
         self.assertEqual(2999, self.trained_schema.total_samples)
         self.assertEqual(1.0, self.trained_schema._get_unique_row_ratio())
 
+        # not completely accurate since register count is 1024
+        self.assertEqual(3046, self.trained_schema_hll.hashed_row_object.cardinality())
```
JGSweets, line 357: Interesting, this would be concerning to me for the case that the number of duplicate rows would be a negative number.

JGSweets: Where is the register count evaluated.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -280,15 +329,33 @@ def test_stream_profilers(self, *mocks):
...
+
+        # check after update
+        with test_utils.mock_timeit():
+            profiler.update_profile(data[3:])
+
+        self.assertEqual(5, profiler.hashed_row_object.cardinality())
+
     def test_correct_unique_row_ratio_test(self):
-        self.assertEqual(2999, len(self.trained_schema.hashed_row_dict))
+        self.assertEqual(2999, len(self.trained_schema.hashed_row_object))
```
JGSweets, line 352: a good test (we can add to where we move things and not allow profiling everything) is validating the unique count is within an expected margin for varying sizes of datasets.

dataprofiler/profilers/profile_builder.py
micdavis: handled in `add_error_checks`

dataprofiler/profilers/profile_builder.py
micdavis: fixed

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: moved row stat tests to `TestStructuredProfilerRowStatistics` class below

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: new tests in `TestStructuredProfilerRowStatistics` class below

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -2584,27 +2580,6 @@ def test_min_max_id_properly_update(self, *mocks):
         self.assertEqual(0, profile._min_id)
         self.assertEqual(6, profile._max_id)
 
-    @mock.patch(
```
micdavis, line 2587: This tests was partially moved. It's functionality is tested elsewhere. (lines 3817-3823, 4076-4152)

micdavis: It doesn't make sense for this functionality to be in `TestStructuredColProfilerClass`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3790,6 +3814,365 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+
+            mock_file.seek(0)
+            with mock.patch("dataprofiler.profilers.profile_builder." "DataLabeler"):
+                load_profile = dp.StructuredProfiler.load("mock.pkl")
+        self.assertEqual(259, load_profile.hashed_row_object.cardinality())
+
+        # Check that reports are equivalent
+        save_report = test_utils.clean_report(self.trained_schema_hll.report())
+        load_report = test_utils.clean_report(load_profile.report())
+        self.assertDictEqual(save_report, load_report)
```
JGSweets, line 4170: instead of full report, maybe jsut validate the hll since that is what we care about.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3790,6 +3814,365 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+
+        # validate both are still usable after
+        self.trained_schema_hll.update_profile(pd.DataFrame([4, 5]))
+        load_profile.update_profile(pd.DataFrame([4, 5]))
```
JGSweets, line 4174: if we know the input data, after updating, validate the values make sense after updating as oppose to just ensuring update works. it's important the properties still update properly.

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: done

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: done

dataprofiler/profilers/profile_builder.py
```diff
@@ -1557,6 +1564,43 @@ def _add_error_checks(  # type: ignore[override]
...
+            self.options.row_statistics.unique_count.hll.seed
+            != other.options.row_statistics.unique_count.hll.seed
+            or self.options.row_statistics.unique_count.hll.register_count
+            != other.options.row_statistics.unique_count.hll.register_count
+        ):
+            raise ValueError(
+                "Attempting to merge profiles whose row hashing "
+                "objects are of different seed or register count."
+            )
+
```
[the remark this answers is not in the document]
micdavis: done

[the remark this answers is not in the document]
taylorfturner: dope thx

taylorfturner (approved):

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3530,10 +3504,58 @@ def test_options_ingested_correctly(self):
...
+            data.append(f"{x}")
+
+        cls.data = pd.DataFrame(data)
+
+        profiler_options_hll = ProfilerOptions()
+        profiler_options_hll.set(
+            {
+                "*.is_enabled": False,
+                "row_statistics.*.is_enabled": True,
+                "row_statistics.unique_count.hashing_method": "hll",
```
JGSweets, line 3538: Why the `*` in `"row_statistics.*.is_enabled": True,`? Does it enable both row and below then?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3693,17 +3714,7 @@ def test_null_calculation_with_differently_sampled_cols(self):
...
-    @mock.patch("dataprofiler.profilers.profile_builder." "ColumnStatsProfileCompiler")
-    @mock.patch("dataprofiler.profilers.profile_builder." "ColumnDataLabelerCompiler")
-    @mock.patch("dataprofiler.profilers.profile_builder.DataLabeler")
-    @mock.patch(
-        "dataprofiler.profilers.profile_builder."
-        "StructuredProfiler._update_correlation"
-    )
-    def test_null_row_stats_correct_after_updates(self, *mocks):
+    def test_null_row_stats_correct_after_updates(self):
         data1 = pd.DataFrame([[1, None], [1, 1], [None, None], [None, 1]])
```
JGSweets, line 3718: Won't removing this cause this test to run all the options as opposed ot just testing what is desired since this is using default `ProfilerOptions`?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3790,6 +3801,279 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
         # recognized as object, not float64
         self.assertSetEqual({8}, profile._profile[1].null_types_index["None"])
 
+        # Tests row stats disabled
+        options = StructuredOptions()
+        options.row_statistics.is_enabled = False
+        options.multiprocess.is_enabled = False
+        profile2 = StructuredProfiler(data1, options=options)
+        self.assertEqual(0, profile2.row_is_null_count)
+        self.assertEqual(0, profile2.row_has_null_count)
```
JGSweets, line 3810: For duplicate etc, we use `None` should we expect the same here?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3790,6 +3801,279 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+            spec=dataprofiler.profilers.profile_builder.HyperLogLog,
+        ) as hll_mock:
+            hll_mock.return_value.cardinality.return_value = 1000
+            profiler = StructuredProfiler(pd.DataFrame([]), options=profiler_options)
+
+        self.assertEqual(1000, profiler.hashed_row_object.cardinality())
+        self.assertEqual(0, profiler._get_duplicate_row_count())
+
+    def test_save_and_load_hll(self):
+        self.assertEqual(110, self.trained_schema_hll.hashed_row_object.cardinality())
```
JGSweets, line 4045: This feels a little strange to test as opposed to just comparing before and after to each other.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3530,10 +3504,58 @@ def test_options_ingested_correctly(self):
...
+            "france",
+            "morocco",
+            "chevy",
+            "ford",
+            "toyota",
+            "apple",
+        ] * 10
+
+        for x in range(100):
+            data.append(f"{x}")
```
JGSweets, line 3529: `data.extend(map(str, range(100)))`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1884,11 +1948,21 @@ def report(self, report_options: dict = None) -> dict:
 
         return _prepare_report(report, output_format, omit_keys)
 
-    def _get_unique_row_ratio(self) -> float:
+    def _get_unique_row_ratio(self) -> float | None:
         """Return unique row ratio."""
-        if self.total_samples:
-            return len(self.hashed_row_dict) / self.total_samples
-        return 0
+        if (
```
JGSweets, line 1953: similar to above with the refactoring.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1902,9 +1976,20 @@ def _get_row_has_null_ratio(self) -> float:
             return self.row_has_null_count / self._min_col_samples_used
         return 0
 
-    def _get_duplicate_row_count(self) -> int:
-        """Retun dup row count."""
-        return self.total_samples - len(self.hashed_row_dict)
+    def _get_duplicate_row_count(self) -> int | None:
+        """Return dup row count."""
+        if (
```
JGSweets, line 1981: similar to above with the refactoring

dataprofiler/profilers/profile_builder.py
```diff
@@ -1927,17 +2012,35 @@ def _update_row_statistics(
                 "Cannot calculate row statistics on data that is" "not a DataFrame"
             )
 
-        self.total_samples += len(data)
```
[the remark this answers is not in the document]
JGSweets: Do we have a test that ensures that this still counts if row_stats is disabled

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Yup, weird it was in the `TestStructuredColProfilerClass` class to begin with...

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3584,12 +3615,37 @@ def test_correct_null_row_counts(self):
         self.assertEqual(3, profile.row_is_null_count)
         self.assertEqual(3 / 24, profile._get_row_is_null_ratio())
 
+    def test_row_is_null_ratio_row_stats_disabled(self):
```
micdavis, line 3618: if row stats is disabled then return none, this is these are the tests for that

dataprofiler/profilers/profile_builder.py
```diff
@@ -1884,27 +1950,54 @@ def report(self, report_options: dict = None) -> dict:
...
+            if isinstance(self.hashed_row_object, dict):
+                return len(self.hashed_row_object) / self.total_samples
+            elif isinstance(self.hashed_row_object, HyperLogLog):
+                return int(self.hashed_row_object.cardinality()) / self.total_samples
         return 0
 
-    def _get_row_is_null_ratio(self) -> float:
+    def _get_row_is_null_ratio(self) -> float | None:
         """Return whether row is null ratio."""
+        if not self.options.row_statistics.is_enabled:
```
micdavis, line 1970: here is the addition that makes row stats enabled for null calcs to return non-null values

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: these will still be zero, they are attributes and not the method call so they will default to 0. One thought, should we change the init of `row_is_null_count` and `row_has_null_count` to be `None` vs 0?

JGSweets: Yes, I think we did that with unique as well.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1673,7 +1739,7 @@ def diff(  # type: ignore[override]
                 ),
                 "duplicate_row_count": utils.find_diff_of_numbers(
                     self._get_duplicate_row_count(),
-                    other_profile._get_row_is_null_ratio(),
```
micdavis, line 1676: why didn't this fail any tests?

micdavis: no tests failed cause the values were set to pass not to be correct, as seen below

JGSweets: Huge catch

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: I believe this still needs to be updated like we did for unique_rows and duplicate rows where it is none when disabled.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3790,6 +3855,326 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
...
+            load_report["global_stats"]["duplicate_row_count"],
+        )
+
+        # validate both are still usable after
+        self.trained_schema_hll.update_profile(
+            pd.DataFrame({"names": ["hello"], "numbers": [5], "tf_null": [None]})
+        )
+        load_profile.update_profile(
+            pd.DataFrame({"names": ["hello"], "numbers": [5], "tf_null": [None]})
+        )
```
JGSweets, line 4173: this is good, should be the same update to each. However, I suggest having one row be new and one be a duplicate such that we can validate it updates as expected and handles both conditions after serialization.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3530,18 +3501,73 @@ def test_options_ingested_correctly(self):
...
-class TestStructuredProfilerNullValues(unittest.TestCase):
+class TestStructuredProfilerRowStatistics(unittest.TestCase):
     def setUp(self):
         test_utils.set_seed(0)
 
+    @classmethod
+    def setUpClass(cls):
+        test_utils.set_seed(seed=0)
+
+        data = {
```
JGSweets, line 3512: none of the null tests use this, correct?

Future, I would suggest setting up a class dataset to be usable by all tests. So it would handle cases we want to test for unique and null. What I would see missing here would be the all null cases etc.

not to update in this, but if we want to throw up a good first issue that describes a refactor for that purpose here, that'd be ideal.

JGSweets (approved):

taylorfturner (approved):

PR #851, opened 2023-05-31
Update docs for 0.9.0

taylorfturner:
Docs build on fork: https://taylorfturner.github.io/DataProfiler/docs/0.9.0/html/index.html


docs/source/dataprofiler.profilers.rst
```diff
@@ -22,6 +22,7 @@ Modules
    dataprofiler.profilers.graph_profiler
    dataprofiler.profilers.histogram_utils
    dataprofiler.profilers.int_column_profile
+   dataprofiler.profilers.json_decoder
```
JGSweets, line 25: do we want the decoder stuff? It's not completely functional right now. Thoughts?

ksneab7: I think that might be next release. Just because we have a decent amount work left on it

taylorfturner: Agree

taylorfturner: will fix

taylorfturner: removing decoder

taylorfturner:
I have looked at the security check related to `tornado-6.2-cp37-abi3-` but when I install `requirements.txt` on `gh-pages` it does in fact install `6.3.2` when I look at `pip list | grep -i tornado`

PR #852, opened 2023-05-31
hot fixes for encode and decode of numeric stats mixin and intcol pro…

ksneab7:
…filer

PR #854, opened 2023-06-01
Float column profiler encode decode

ksneab7:
This PR is based on:
https://github.com/capitalone/DataProfiler/pull/852


dataprofiler/profilers/float_column_profile.py
```diff
@@ -273,12 +333,12 @@ def _get_float_precision(
         # Determine statistics precision
         precision_sum = len_per_float.sum()
         subset_precision = {
-            "min": len_per_float.min(),
-            "max": len_per_float.max(),
-            "biased_var": float(np.var(len_per_float)),
-            "sum": precision_sum,
-            "mean": precision_sum / sample_size,
-            "sample_size": sample_size,
+            "min": np.float64(len_per_float.min()),
```
JGSweets, line 336: are these needed?

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -358,9 +359,18 @@ def report(self, remove_disabled_flag: bool = False) -> dict:
...
         return profile
 
-    def _load_hist_helper(self, data):
-        # Pop off stuff specific to numerical stats
-        hist = data.pop("_stored_histogram")
+    def _load_stats_helper(self, deserialized_data: dict):
+        """Assistance function in the deserialization of profiler objects.
+
+            This function is to be used to enforce correct typing for attributes
+            associated with the NumericStatsMixinconversions when loading profiler
```
JGSweets, line 366: typo `NumericStatsMixinconversions`

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -358,9 +359,18 @@ def report(self, remove_disabled_flag: bool = False) -> dict:
...
+            objects in from their serialized saved format
+        :param deserialized_data: The dictionary read in from the serialized form
+            of the profiler object
+        :type deserialized_data: Dict
+        """
+        # Pop off attributes specific to numerical stats histogram creation
+        hist = deserialized_data.pop("_stored_histogram")
```
JGSweets, line 373: I could have sworn i made a comment that disappeared here. Do we need to even use the `deserialized_data` or can we just use self._stored_histogram that already is loaded just with the incorrect formats?

dataprofiler/tests/profilers/utils.py
```diff
@@ -188,7 +188,10 @@ def assert_profiles_equal(actual, expected):
     for actual_value, expected_value in zip(
         actual_dict.values(), expected_dict.values()
     ):
-        assert type(actual_value) == type(expected_value)
+        try:
+            assert type(actual_value) == type(expected_value)
+        except:
+            pass
```
JGSweets, line 194: This should fail if they don't match.

JGSweets (changes requested):

dataprofiler/tests/profilers/utils.py
taylorfturner: weird ... I'm almost positive Kenny had taken this out previously

taylorfturner: totally agree it should be out. I'll push a commit to resolve this at least

taylorfturner: resolved

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -393,20 +393,17 @@ def convert_histogram_key_types_to_np(histogram_info: dict):
...
+        float_int_to_np_conversion = (
+            np.int64 if self.type in ["int", "string", "text"] else np.float64
+        )
+
+        if self.min is not None:
+            self.min = float_int_to_np_conversion(self.min)
+        if self.max is not None:
+            self.max = float_int_to_np_conversion(self.max)
+        if self.sum is not None:
+            self.sum = float_int_to_np_conversion(self.sum)
```
JGSweets, line 406: since this is causing issues, let's use prev state and just do np.float64. This will be functional after merging int and float. At profile we can alter min and max for output.

```python
# remove unused float_int_to_np_conversion above

if self.min is not None:
    self.min = np.float64(self.min)
if self.max is not None:
    self.max = np.float64(self.max)
if self.sum is not None:
    self.sum = np.float64(self.sum)
```

JGSweets (commented):
Can also fix the utils assert to be more clear:
https://github.com/capitalone/DataProfiler/blob/b114e0ab595c42d895ea28b0439e759a6fc16861/dataprofiler/tests/profilers/utils.py#L185
```python
assert len(actual_dict.keys()) == len(expected_dict.keys())
assert actual_dict.keys() == expected_dict.keys()

for key in expected_dict.keys():
    actual_value = actual_dict.get(key, None)
    expected_value = expected_dict.get(key, None)
    assert type(actual_value) == type(
        expected_value
    ), f"{actual_value} and {expected_value} do not have the same type for key: {key}"
```

dataprofiler/profilers/numerical_column_stats.py
taylorfturner: bingo

JGSweets (approved):

taylorfturner (approved):

PR #858, opened 2023-06-06
Add forking / branch workflow image

taylorfturner:


JGSweets (commented):
in terms of staging, those would be individual new branches from main or dev, right? I think the continued line is a bit confusing.

taylorfturner:
> What's the extra arrow in main?

Shouldn't be there -- didn't update that when I changed something else. good call. updating on next commit

taylorfturner:
> in terms of staging, those would be individual new branches from main or dev, right? I think the continued line is a bit confusing.

yeah, removed that continued line. Effectively its just a branch to literally stage either the feature into dev or dev into main (for a release). So the diagram is drawn in such a way to notate that there should be no commits to those staging branches.  @JGSweets @micdavis @ksneab7 @tyfarnan

taylorfturner:
related to #857

docs/source/_static/images/branching_workflow_diagram.png
tyfarnan, line 1: pretty!

taylorfturner: ty

PR #863, opened 2023-06-09
Fix initial value of processor_type

junholee6a:
`BaseDataProcessor.processor_type` has type `str`, but is initialized in `BaseDataProcessor` as `None`, likely because it is meant to be overridden by subclasses. We initialize it as a string to remove the type error.


dataprofiler/labelers/data_processing.py
```diff
@@ -37,7 +37,7 @@ def __new__(
 class BaseDataProcessor(metaclass=abc.ABCMeta):
     """Abstract Data processing class."""
 
-    processor_type: str = None  # type: ignore[assignment]
+    processor_type: str = "none"  # override this
```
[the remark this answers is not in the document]
junholee6a: Makes sense, the subclasses which override `BaseDataProcessor` are `BaseDataPreprocessor` and `BaseDataPostprocessor`, which set `processor_type` to `"preprocessor"` and `"postprocessor"` respectively

dataprofiler/labelers/data_processing.py
```diff
@@ -37,7 +37,7 @@ def __new__(
 class BaseDataProcessor(metaclass=abc.ABCMeta):
     """Abstract Data processing class."""
 
-    processor_type: str = None  # type: ignore[assignment]
+    processor_type: str = "baseprocessor"  # override this
```
JGSweets, line 40: Hmm, will this causes issues in tests or validation to make sure this is set? Will a string give the false impression of naming?

what if we just do without assignment or does that fail?
`processor_type: str`

junholee6a: This works internally but causes a mypy error due to the property being accessed by a unit test (says property doesn't exist). I'll look into whether that test can be modified

JGSweets: Did a quick 2 sec check, might be related to: https://github.com/python/mypy/issues/4426

junholee6a:
Just edited a commit message, sorry for the inconvenience. I'm done editing this branch if the tests pass

ksneab7 (approved):

taylorfturner (approved):

PR #864, opened 2023-06-12
Added decoding for encoding of ordered column profiles

ksneab7:


dataprofiler/profilers/order_column_profile.py
```diff
@@ -270,6 +270,22 @@ def report(self, remove_disabled_flag: bool = False) -> dict:
...
+        :param data: dictionary with attributes and values.
+        :type data: dict[string, Any]
+
+        :return: Profiler with attributes populated.
+        :rtype: CategoricalColumn
+        """
+        # This is an ambiguous call to super classes.
+        profile = super().load_from_dict(data)
+
+        return profile
```
micdavis, line 287:
```suggestion
        return super().load_from_dict(data)
```
I see benefits to either way, just putting another option out there

dataprofiler/tests/profilers/test_order_column_profile.py
```diff
@@ -413,3 +415,41 @@ def test_json_encode_after_update(self):
...
+        expected_profile = OrderColumn(fake_profile_name)
+
+        serialized = json.dumps(expected_profile, cls=ProfileEncoder)
+        deserialized = load_column_profile(json.loads(serialized))
+
+        utils.assert_profiles_equal(deserialized, expected_profile)
+
+    def test_json_decode_after_update(self):
+        fake_profile_name = "Fake profile name"
+        # Actual deserialization
```
micdavis, line 430:
```suggestion
```
dont think this comment is needed

dataprofiler/profilers/order_column_profile.py
taylorfturner: cleaner +1

taylorfturner: @ksneab7

taylorfturner: the other implementation do operations on `profile`, but in this case we don't .... so no need I don't think to set it to a variable

micdavis (approved):

taylorfturner (approved):
LGTM

dataprofiler/tests/profilers/test_order_column_profile.py
```diff
@@ -413,3 +415,40 @@ def test_json_encode_after_update(self):
...
+        serialized = json.dumps(expected_profile, cls=ProfileEncoder)
+        deserialized = load_column_profile(json.loads(serialized))
+
+        utils.assert_profiles_equal(deserialized, expected_profile)
+
+    def test_json_decode_after_update(self):
+        fake_profile_name = "Fake profile name"
+
+        # Build expected orderColumn
+        df_order = pd.Series(["za", "z", "c", "a"])
```
JGSweets, line 432: might have be interesting to have this ordered and then the update below keep it ordered and validating the output expectation order. Maybe can quick add.

ksneab7: I see what hes saying there `order` attribute is not consistent after updating

PR #869, opened 2023-06-13
Created encoder for the datalabelercolumn

ksneab7:


dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -391,3 +395,172 @@ def test_empty_data(self, *mocks):
...
         self.assertIsNone(merge_profile.data_label)
+
+
+class TestDataLabelerColumnProfilerNoMock(unittest.TestCase):
+    def test_json_encode(self):
+        profiler = DataLabelerColumn("")
+
+        serialized = json.dumps(profiler, cls=ProfileEncoder)
+
+        expected = json.dumps(
```
JGSweets, line 406: good base condition

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -391,3 +395,172 @@ def test_empty_data(self, *mocks):
...
+                "class": "DataLabelerColumn",
+                "data": {
+                    "name": "",
+                    "col_index": float("nan"),
+                    "sample_size": 0,
+                    "metadata": {},
+                    "times": {},
+                    "thread_safe": False,
+                    "_max_sample_size": 1000,
+                    "data_labeler": "structured_model",
```
JGSweets, line 417: here's a q. do we want to use `data_labeler` or do we want to do something sneaky like:
`load_data_labeler_options` so it wouldn't get directly loaded accidentally, but can be identified and load the object to go in as the `data_labeler` when deserialized.

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -391,3 +395,172 @@ def test_empty_data(self, *mocks):
...
+
+    def test_json_encode_after_update(self):
+        data = pd.Series(["1", "2", "3"], dtype=object)
+        profiler = DataLabelerColumn(data.name)
+
+        time_array = [float(i) for i in range(4, 0, -1)]
+        with mock.patch("time.time", side_effect=lambda: time_array.pop()):
+            profiler.update(data)
+
+            serialized = json.dumps(profiler, cls=ProfileEncoder)
```
JGSweets, line 442: FYI, only the update needs to be in the mock, the validation after does not.

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -391,3 +395,172 @@ def test_empty_data(self, *mocks):
...
+                        "_min_voting_prob": 0.2,
+                        "_min_prob_differential": 0.2,
+                        "_top_k_labels": 3,
+                        "_min_top_label_prob": 0.35,
+                        "_DataLabelerColumn__calculations": {},
+                    },
+                }
+            )
+
+            self.assertEqual(serialized, expected)
```
JGSweets, line 566: I think our standards are `expected` first.

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -391,3 +395,172 @@ def test_empty_data(self, *mocks):
...
+                    "_min_voting_prob": 0.2,
+                    "_min_prob_differential": 0.2,
+                    "_top_k_labels": 3,
+                    "_min_top_label_prob": 0.35,
+                    "_DataLabelerColumn__calculations": {},
+                },
+            }
+        )
+
+        self.assertEqual(serialized, expected)
```
JGSweets, line 432: as below

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
taylorfturner: eh not really a standard... we've done `serialized, expected` in `test_base_column_profilers.py`, `test_categorical`, `test_datetime_column_profile`, `test_int_column` , and `test_numeric`, `test_order`

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
taylorfturner: as above

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
JGSweets: > eh not really a standard... we've done `serialized, expected` in `test_base_column_profilers.py`, `test_categorical`, `test_datetime_column_profile`, `test_int_column` , and `test_numeric`, `test_order`

Specifically in this test? IIRC , `self.assert...` returns a response where it assumes the first var is the expected.

taylorfturner: https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
taylorfturner: another option is to change this to a dictionary where the key is `from_library`... which will allow us to do `from_disk` down the road if someone brings their own model. 


```
"data_labeler": {"from_library": "structured_model"}
```


down the road we could do 
```
"data_labeler": {"from_disk": "my_model/path.json"}
```

JGSweets (approved):

micdavis (approved):

Issue #859, opened 2023-06-06
RowStatisticsOptions: Add null row count


micdavis:
**Is your feature request related to a problem? Please describe.**
RowStatisticsOptions has a `unique_count` option which toggles the calculation of "duplicate_row_count" and "unique row_ratio". But it doesn't have a `null_row_count` option. We need `null_row_count` to be an option to toggle to functionality for `row_has_null_ratio` ([link](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profile_builder.py#L1899)) and `row_is_null_ratio` ([link](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profile_builder.py#L1899)). This option should be toggled in `_update_row_statistics` as well to manage the calculation of these metrics ([link](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profile_builder.py#L1942)).

**Describe the outcome you'd like:**
RowStatisticsOptions needs to include a toggle to enable or disable the calculation for null row count. This option then should be used to the `profiler_builder` to toggle this functionality.

**Additional context:**
This is a follow up from current/previous work on RowStatisticsOptions: #841

taylorfturner:
Closing since the above references PRs were put into `dev` through `staging/dev/options` PR #909

PR #871, opened 2023-06-14
RowStatisticsOptions: Implementing option

drahc1R:
related to issue #859

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2088,11 +2088,13 @@ class StructuredProfiler(BaseProfiler):
 
         # If sample_ids provided, increment since that means only new data read
         if sample_ids is not None:
-            self.row_has_null_count += len(null_in_row_count)
-            self.row_is_null_count += len(null_rows)
+            if self.options.row_statistics.null_count.is_enabled:
+                self.row_has_null_count += len(null_in_row_count)
+                self.row_is_null_count += len(null_rows)
         else:
-            self.row_has_null_count = len(null_in_row_count)
-            self.row_is_null_count = len(null_rows)
+            if self.options.row_statistics.null_count.is_enabled:
+                self.row_has_null_count = len(null_in_row_count)
+                self.row_is_null_count = len(null_rows)
 
     def _get_correlation(
         self, clean_samples: dict, batch_properties: dict
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2087,12 +2087,11 @@ class StructuredProfiler(BaseProfiler):
                 null_in_row_count = null_in_row_count.union(null_row_indices)
 
         # If sample_ids provided, increment since that means only new data read
-        if sample_ids is not None:
-            if self.options.row_statistics.null_count.is_enabled:
+        if self.options.row_statistics.null_count.is_enabled:
+            if sample_ids is not None:
                 self.row_has_null_count += len(null_in_row_count)
                 self.row_is_null_count += len(null_rows)
-        else:
-            if self.options.row_statistics.null_count.is_enabled:
+            else:
                 self.row_has_null_count = len(null_in_row_count)
                 self.row_is_null_count = len(null_rows)
 
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2103,14 +2103,14 @@ class StructuredProfiler(BaseProfiler):
                     null_rows = null_rows.intersection(null_row_indices)
                     null_in_row_count = null_in_row_count.union(null_row_indices)
 
-                    # If sample_ids provided,
-                    # increment since that means only new data read
-                    if sample_ids is not None:
-                        self.row_has_null_count += len(null_in_row_count)
-                        self.row_is_null_count += len(null_rows)
-                    else:
-                        self.row_has_null_count = len(null_in_row_count)
-                        self.row_is_null_count = len(null_rows)
+            # If sample_ids provided,
+            # increment since that means only new data read
+            if sample_ids is not None:
+                self.row_has_null_count += len(null_in_row_count)
+                self.row_is_null_count += len(null_rows)
+            else:
+                self.row_has_null_count = len(null_in_row_count)
+                self.row_is_null_count = len(null_rows)
 
     def _get_correlation(
         self, clean_samples: dict, batch_properties: dict
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1584,7 +1584,7 @@ class StructuredProfiler(BaseProfiler):
             )
         # Check null_count options
         if (
-            self.options.row_statistics.null.is_enabled
+            self.options.row_statistics.null_count.is_enabled
             != other.options.row_statistics.null_count.is_enabled
         ):
             raise ValueError(
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -2060,49 +2060,50 @@ class StructuredProfiler(BaseProfiler):
                         self.hashed_row_object.add(record)
 
         # Calculate Null Column Count
-        null_rows = set()
-        null_in_row_count = set()
-        first_col_flag = True
-        for column in self._profile:
-            null_type_dict = column.null_types_index
-            null_row_indices = set()
-            if null_type_dict:
-                null_row_indices = set.union(*null_type_dict.values())
-
-            # If sample ids provided, only consider nulls in rows that
-            # were fully sampled
-            if sample_ids is not None:
-                # This is the amount (integer) indices were shifted by in the
-                # event of overlap
-                shift = column._index_shift
-                if shift is None:
-                    # Shift is None if index is str or if no overlap detected
-                    null_row_indices = null_row_indices.intersection(
-                        data.index[sample_ids[: self._min_sampled_from_batch]]
-                    )
-                else:
-                    # Only shift if index shift detected (must be ints)
-                    null_row_indices = null_row_indices.intersection(
-                        data.index[sample_ids[: self._min_sampled_from_batch]] + shift
-                    )
-
-            # Find the common null indices between the columns
-            if first_col_flag:
-                null_rows = null_row_indices
-                null_in_row_count = null_row_indices
-                first_col_flag = False
-            else:
-                null_rows = null_rows.intersection(null_row_indices)
-                null_in_row_count = null_in_row_count.union(null_row_indices)
-
-        # If sample_ids provided, increment since that means only new data read
         if self.options.row_statistics.null_count.is_enabled:
-            if sample_ids is not None:
-                self.row_has_null_count += len(null_in_row_count)
-                self.row_is_null_count += len(null_rows)
-            else:
-                self.row_has_null_count = len(null_in_row_count)
-                self.row_is_null_count = len(null_rows)
+            null_rows = set()
+            null_in_row_count = set()
+            first_col_flag = True
+            for column in self._profile:
+                null_type_dict = column.null_types_index
+                null_row_indices = set()
+                if null_type_dict:
+                    null_row_indices = set.union(*null_type_dict.values())
+
+                # If sample ids provided, only consider nulls in rows that
+                # were fully sampled
+                if sample_ids is not None:
+                    # This is the amount (integer) indices were shifted by in the
+                    # event of overlap
+                    shift = column._index_shift
+                    if shift is None:
+                        # Shift is None if index is str or if no overlap detected
+                        null_row_indices = null_row_indices.intersection(
+                            data.index[sample_ids[: self._min_sampled_from_batch]]
+                        )
+                    else:
+                        # Only shift if index shift detected (must be ints)
+                        null_row_indices = null_row_indices.intersection(
+                            data.index[sample_ids[: self._min_sampled_from_batch]]
+                            + shift
+                        )
+
+                # Find the common null indices between the columns
+                if first_col_flag:
+                    null_rows = null_row_indices
+                    null_in_row_count = null_row_indices
+                    first_col_flag = False
+                else:
+                    null_rows = null_rows.intersection(null_row_indices)
+                    null_in_row_count = null_in_row_count.union(null_row_indices)
+
+                # If sample_ids provided, increment since that means only new data read
+                if sample_ids is not None:
+                    self.row_has_null_count += len(null_in_row_count)
+                    self.row_is_null_count += len(null_rows)
+                else:
+                    self.row_has_null_count = len(null_in_row_count)
+                    self.row_is_null_count = len(null_rows)
 
     def _get_correlation(
         self, clean_samples: dict, batch_properties: dict
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1582,6 +1582,15 @@ class StructuredProfiler(BaseProfiler):
                 "Attempting to merge two profiles with unique row "
                 "count option enabled on one profile but not the other."
             )
+        # Check null_count options
+        if (
+            self.options.row_statistics.null.is_enabled
+            != other.options.row_statistics.null_count.is_enabled
+        ):
+            raise ValueError(
+                "Attempting to merge two profiles with null row "
+                "count option enabled on one profile but not the other."
+            )
         # Check hashing_method options
         if (
             self.options.row_statistics.unique_count.hashing_method
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1976,7 +1976,10 @@ class StructuredProfiler(BaseProfiler):
 
     def _get_row_is_null_ratio(self) -> float | None:
         """Return whether row is null ratio."""
-        if not self.options.row_statistics.is_enabled:
+        if (
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.is_enabled
+        ):
             return None
 
         if self._min_col_samples_used:
@@ -1985,7 +1988,10 @@ class StructuredProfiler(BaseProfiler):
 
     def _get_row_has_null_ratio(self) -> float | None:
         """Return whether row has null ratio."""
-        if not self.options.row_statistics.is_enabled:
+        if (
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.is_enabled
+        ):
             return None
 
         if self._min_col_samples_used:
@@ -2097,13 +2103,14 @@ class StructuredProfiler(BaseProfiler):
                     null_rows = null_rows.intersection(null_row_indices)
                     null_in_row_count = null_in_row_count.union(null_row_indices)
 
-                # If sample_ids provided, increment since that means only new data read
-                if sample_ids is not None:
-                    self.row_has_null_count += len(null_in_row_count)
-                    self.row_is_null_count += len(null_rows)
-                else:
-                    self.row_has_null_count = len(null_in_row_count)
-                    self.row_is_null_count = len(null_rows)
+                    # If sample_ids provided,
+                    # increment since that means only new data read
+                    if sample_ids is not None:
+                        self.row_has_null_count += len(null_in_row_count)
+                        self.row_is_null_count += len(null_rows)
+                    else:
+                        self.row_has_null_count = len(null_in_row_count)
+                        self.row_is_null_count = len(null_rows)
 
     def _get_correlation(
         self, clean_samples: dict, batch_properties: dict
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1978,7 +1978,7 @@ class StructuredProfiler(BaseProfiler):
         """Return whether row is null ratio."""
         if (
             not self.options.row_statistics.is_enabled
-            or not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.null_count.is_enabled
         ):
             return None
 
@@ -1990,7 +1990,7 @@ class StructuredProfiler(BaseProfiler):
         """Return whether row has null ratio."""
         if (
             not self.options.row_statistics.is_enabled
-            or not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.null_count.is_enabled
         ):
             return None
 
--- a/dataprofiler/tests/profilers/test_profile_builder.py
+++ b/dataprofiler/tests/profilers/test_profile_builder.py
@@ -3605,7 +3605,6 @@ class TestStructuredProfilerRowStatistics(unittest.TestCase):
                 data[2:], options=profiler_options_null_disabled
             )
 
-        # self.assertEqual(0, profiler_w_disabled_null_count.report()['data_stats'][0]['statistics']['null_count'])
         self.assertEqual(0, profiler_w_disabled_null_count.row_has_null_count)
         self.assertEqual(0, profiler_w_disabled_null_count.row_is_null_count)
 
--- a/dataprofiler/tests/profilers/test_profile_builder.py
+++ b/dataprofiler/tests/profilers/test_profile_builder.py
@@ -248,26 +248,7 @@ class TestStructuredProfiler(unittest.TestCase):
         self.assertEqual("multiple files", merged_profile.encoding)
         self.assertEqual("multiple files", merged_profile.file_type)
 
-    @mock.patch(
-        "dataprofiler.profilers.profile_builder." "ColumnPrimitiveTypeProfileCompiler"
-    )
-    @mock.patch("dataprofiler.profilers.profile_builder." "ColumnStatsProfileCompiler")
-    @mock.patch("dataprofiler.profilers.profile_builder." "ColumnDataLabelerCompiler")
-    @mock.patch(
-        "dataprofiler.profilers.profile_builder.DataLabeler", spec=StructuredDataLabeler
-    )
-    @mock.patch(
-        "dataprofiler.profilers.profile_builder."
-        "StructuredProfiler._update_correlation"
-    )
-    @mock.patch(
-        "dataprofiler.profilers.profile_builder."
-        "StructuredProfiler._merge_correlation"
-    )
-    @mock.patch(
-        "dataprofiler.profilers.profile_builder." "StructuredProfiler._update_chi2"
-    )
-    def test_add_profiles_null_count_not_enabled(self, *mocks):
+    def test_add_profiles_null_count_not_enabled(self):
         profiler1_options_null_count = ProfilerOptions()
         profiler1_options_null_count.set(
             {"row_statistics.null_count.is_enabled": False}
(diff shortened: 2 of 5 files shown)


taylorfturner:
tests failing here @drahc1R

dataprofiler/profilers/profile_builder.py
```diff
@@ -2088,11 +2088,13 @@ def _update_row_statistics(
...
         if sample_ids is not None:
-            self.row_has_null_count += len(null_in_row_count)
-            self.row_is_null_count += len(null_rows)
+            if self.options.row_statistics.null_count.is_enabled:
+                self.row_has_null_count += len(null_in_row_count)
+                self.row_is_null_count += len(null_rows)
         else:
-            self.row_has_null_count = len(null_in_row_count)
-            self.row_is_null_count = len(null_rows)
+            if self.options.row_statistics.null_count.is_enabled:
```
JGSweets, line 2095: what if we move this `if` outside the other `if`. Then we only have 1 conditional as oppose to 2

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3733,6 +3737,7 @@ def test_null_calculation_with_differently_sampled_cols(self):
         # The only null in those rows in second column in that subset are 5, 7
         # Therefore only 2 rows have null according to row_has_null_count
         self.assertEqual(0, profile.row_is_null_count)
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3740: I think the counts intuitively speak to whether this is enabled. The tests in the options should cover this being set properly as well.

We just want to make sure when enabled, it does collect the values as expected.

dataprofiler/profilers/profile_builder.py
```diff
@@ -2088,11 +2088,13 @@ def _update_row_statistics(
```
JGSweets, line 2088: if we want to do options and profiler, we need to also handle the `__add__` case. However, I suggest splitting up the PR.

@taylorfturner

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -233,6 +233,7 @@ def test_add_profilers(self, *mocks):
         self.assertEqual(
             "<class 'pandas.core.frame.DataFrame'>", merged_profile.file_type
         )
+        self.assertTrue(merged_profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 236: This is a fair check when doing merge as it `__add__` should ensure the consistency.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3602,6 +3603,7 @@ def test_correct_null_row_counts(self):
             }
         )
         profile = dp.StructuredProfiler(data, options=profiler_options)
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3606: I think the counts intuitively speak to whether this is enabled. The tests in the options should cover this being set properly as well.

We just want to make sure when enabled, it does collect the values as expected.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3610,6 +3612,7 @@ def test_correct_null_row_counts(self):
         file_path = os.path.join(test_root_path, "data", "csv/iris-with-null-rows.csv")
         data = pd.read_csv(file_path)
         profile = dp.StructuredProfiler(data, options=profiler_options)
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3615: I think the counts intuitively speak to whether this is enabled. The tests in the options should cover this being set properly as well.

We just want to make sure when enabled, it does collect the values as expected.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3773,6 +3778,7 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
         profile = dp.StructuredProfiler(
             data1, min_true_samples=2, samples_per_update=2, options=opts
         )
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3781: as above

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3782,6 +3788,7 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
         self.assertSetEqual({0, 2}, profile._profile[1].null_types_index["nan"])
 
         profile.update_profile(data2, min_true_samples=2, sample_size=2)
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3791: as above

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3803,6 +3810,7 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
             }
         )
         profile = dp.StructuredProfiler(data1, options=opts)
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3813: as above

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3826,6 +3835,7 @@ def test_null_row_stats_correct_after_updates(self, *mocks):
 
         # Test that update with emtpy data doesn't change stats
         profile.update_profile(pd.DataFrame([]))
+        self.assertTrue(profile.options.row_statistics.null_count.is_enabled)
```
JGSweets, line 3838: as above

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -233,6 +233,7 @@ def test_add_profilers(self, *mocks):
         self.assertEqual(
```
JGSweets, line 233: I think in the below, we are missing a test when we choose options that disable the count specifically. We would need to ensure the report return values are `None` (I think that's what they would return as).

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -233,12 +233,20 @@ def test_add_profilers(self, *mocks):
...
             "<class 'pandas.core.frame.DataFrame'>", merged_profile.file_type
         )
+        self.assertTrue(merged_profile.options.row_statistics.null_count.is_enabled)
         self.assertEqual(2, merged_profile.row_has_null_count)
         self.assertEqual(2, merged_profile.row_is_null_count)
         self.assertEqual(7, merged_profile.total_samples)
         self.assertEqual(5, len(merged_profile.hashed_row_object))
         self.assertDictEqual({"row_stats": 2}, merged_profile.times)
 
+        # test success when null_count option is disabled
```
drahc1R, line 243: should I create a new test function instead of bunching this into test_add_profilers?

dataprofiler/profilers/profile_builder.py
drahc1R: does line 211 check for that?

dataprofiler/tests/profilers/test_profile_builder.py
taylorfturner: yes, that's a fair call -- just be very descriptive with what the test is doing in the test name such as `def test_add_profiles_null_count_not_enabled()`

dataprofiler/profilers/profile_builder.py
taylorfturner: its checkiing the first `ValueError` in `_add_error_checks` but 211 is not actually checking that the options are different because the test is setup by only changing the data in the profile and not setting up the profiles with different options.

taylorfturner: so add a new if statement in the `_add_error_checks` specifically for `options.row_statistics.null_count.is_enabled` or whatever the specific option call is for your new option

dataprofiler/profilers/profile_builder.py
```diff
@@ -1582,6 +1582,15 @@ def _add_error_checks(  # type: ignore[override]
                 "Attempting to merge two profiles with unique row "
                 "count option enabled on one profile but not the other."
             )
+        # Check null_count options
+        if (
+            self.options.row_statistics.null.is_enabled
```
taylorfturner, line 1587:
```suggestion
            self.options.row_statistics.null_count.is_enabled
```

taylorfturner (changes requested):
typo

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,59 @@ def test_add_profilers(self, *mocks):
...
+        self.assertLess(0, profile2.row_is_null_count)
+
+        try:
+            with self.assertRaisesRegex(
+                ValueError,
+                "The two profilers were not setup with the "
+                "same options, hence they do not calculate "
+                "the same profiles and cannot be added "
+                "together.",
+            ):
```
taylorfturner, line 299: don't think this is the `ValueError` that should get raised....

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,59 @@ def test_add_profilers(self, *mocks):
...
+                profile1 + profile2
+        except:
+            pass
```
taylorfturner, line 302: try/except is sneaky... 💀

taylorfturner (changes requested):

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,59 @@ def test_add_profilers(self, *mocks):
...
+        "dataprofiler.profilers.profile_builder."
+        "StructuredProfiler._update_correlation"
+    )
+    @mock.patch(
+        "dataprofiler.profilers.profile_builder."
+        "StructuredProfiler._merge_correlation"
+    )
+    @mock.patch(
+        "dataprofiler.profilers.profile_builder." "StructuredProfiler._update_chi2"
+    )
```
taylorfturner, line 269: I'd also validate quickly that all these mocks are needed... they might be... but just double check

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,59 @@ def test_add_profilers(self, *mocks):
...
+            profile2 = dp.StructuredProfiler(
+                data[2:], options=profiler2_options_null_count
+            )
+
+        self.assertEqual(0, profile1.row_has_null_count)
+        self.assertEqual(0, profile1.row_is_null_count)
+        self.assertLess(0, profile2.row_has_null_count)
+        self.assertLess(0, profile2.row_is_null_count)
+
+        try:
```
JGSweets, line 292: Is something else erroring here and why we need the try? I would suggest not using it as then an exception could happen that we do not desire and it would silently fail.

The `with self.assertRaisesRegex` should capture the exception and not cause an error provided it matches the expected regex below.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,59 @@ def test_add_profilers(self, *mocks):
...
+    )
+    @mock.patch(
+        "dataprofiler.profilers.profile_builder."
+        "StructuredProfiler._merge_correlation"
+    )
+    @mock.patch(
+        "dataprofiler.profilers.profile_builder." "StructuredProfiler._update_chi2"
+    )
+    def test_add_profiles_null_count_not_enabled(self, *mocks):
+        profiler1_options_null_count = ProfilerOptions()
```
JGSweets, line 252: instead of `profiler1_options_null_count` or `profiler2_options_null_count`. I'd suggest naming them for clairity wrt to their differences to make the test more readable.

e.g. `profiler1_options_null_count` -> `profiler_options_null_disabled`

similarly, `profiler1` -> `profiler_w_disabled_null_count`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,59 @@ def test_add_profilers(self, *mocks):
...
+        profiler1_options_null_count.set(
+            {"row_statistics.null_count.is_enabled": False}
```
JGSweets, line 254: we started a new paradigm that I think makes tests more efficient, if we only want to test just the null_count info, we can pretty much disable everything else:
```
{"row_statistics.null_count.is_enabled": False}
```
becomes,
```
{
    "*.is_enabled": False,
    "row_statistics.*.is_enabled": True,
    "row_statistics.null_count.is_enabled": False,  # or True for the other case
}
```

JGSweets: I believe in this case we won't need all the mocks above either. If you run this test specifically, i'd expect times in the `miliseconds`

taylorfturner: +1

dataprofiler/tests/profilers/test_profile_builder.py
taylorfturner: already on it and fixed it

dataprofiler/tests/profilers/test_profile_builder.py
taylorfturner: yep told @drahc1R to test on the mocks if we need those... didn't think we would need them

dataprofiler/profilers/profile_builder.py
```diff
@@ -2087,12 +2096,13 @@ def _update_row_statistics(
                 null_in_row_count = null_in_row_count.union(null_row_indices)
 
         # If sample_ids provided, increment since that means only new data read
-        if sample_ids is not None:
-            self.row_has_null_count += len(null_in_row_count)
-            self.row_is_null_count += len(null_rows)
-        else:
-            self.row_has_null_count = len(null_in_row_count)
-            self.row_is_null_count = len(null_rows)
+        if self.options.row_statistics.null_count.is_enabled:
```
micdavis, line 2099: does this condition check belong up at line 2062? if null count is disabled, do we need to any code below 2062?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -246,6 +248,35 @@ def test_add_profilers(self, *mocks):
...
+                data[:2], options=profiler1_options_null_count
+            )
+            profile2 = dp.StructuredProfiler(
+                data[2:], options=profiler2_options_null_count
+            )
+
+        self.assertEqual(0, profile1.row_has_null_count)
+        self.assertEqual(0, profile1.row_is_null_count)
+        self.assertLess(0, profile2.row_has_null_count)
+        self.assertLess(0, profile2.row_is_null_count)
```
JGSweets, line 271: This is good, but the test itself doesn't describe the intent for this check.
`test_add_profiles_null_count_not_enabled` to me would suggest only validating adding profiles.

moreover, below, it checks for adding profiles that are both enabled and disabled.

Maybe, instead rename to `test_adding_profiles_of_mismatched_null_count_options`
and move the assertEquals above to a different test where we just check the functionality when the stat is disabled.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets:
```
"row_has_null_ratio": self._get_row_has_null_ratio(),
"row_is_null_ratio": self._get_row_is_null_ratio(),
```
these two int he report specifically.

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: also these tests should be moved down to the `TestStructuredProfilerRowStatistics`.

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Currently, 
```
def _get_row_has_null_ratio(self) -> float | None:
        """Return whether row has null ratio."""
        if not self.options.row_statistics.is_enabled:
            ...
```
only returns `None` for row_stats being disabled, but needs to be updated to include the case of null option disabled

drahc1R:
typo on name

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3555,6 +3557,69 @@ def setUpClass(cls):
...
+        data = pd.DataFrame([1, None, 3, 4, 5, None, 1])
+        with test_utils.mock_timeit():
+            profiler_w_null_count = dp.StructuredProfiler(
+                data[:2], options=profiler_options_null_count
+            )
+            profiler_w_disabled_null_count = dp.StructuredProfiler(
+                data[2:], options=profiler_options_null_disabled
+            )
+        self.assertLess(0, profiler_w_null_count.row_has_null_count)
+        self.assertLess(0, profiler_w_null_count.row_is_null_count)
```
taylorfturner, line 3619: shouldn't this fail?

dataprofiler/profilers/profile_builder.py
```diff
@@ -2051,48 +2060,50 @@ def _update_row_statistics(
...
-                    # Shift is None if index is str or if no overlap detected
-                    null_row_indices = null_row_indices.intersection(
-                        data.index[sample_ids[: self._min_sampled_from_batch]]
-                    )
-                else:
-                    # Only shift if index shift detected (must be ints)
-                    null_row_indices = null_row_indices.intersection(
-                        data.index[sample_ids[: self._min_sampled_from_batch]] + shift
-                    )
+        if self.options.row_statistics.null_count.is_enabled:
```
JGSweets, line 2063: good call, no need to run anything if it is disabled.

taylorfturner (approved):

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: I think the above still needs to be addressed.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3555,6 +3557,69 @@ def setUpClass(cls):
...
+        self.assertEqual(0, profiler_w_disabled_null_count.row_has_null_count)
+        self.assertEqual(0, profiler_w_disabled_null_count.row_is_null_count)
```
JGSweets, line 3621: I think we still need to split this out as this isn't testing mismatching of add in terms of the test function naming.

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3555,6 +3557,69 @@ def setUpClass(cls):
                 cls.data, len(cls.data), options=profiler_options_full
             )
 
+    def test_add_profiles_null_count_not_enabled(self):
```
JGSweets, line 3560: I believe this is the `test_adding_profiles_of_mismatched_null_count_options`

taylorfturner, line 3560: based on the name I'd expect both `row_statistics.null_count.is_enabled` set to False...

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: Actually, just noticing the swapping of names above.
If we change this to `test_add_profiles_null_count_not_enabled` we only need to test one profile and it's values.

JGSweets: We still need to include the tests in this comment: https://github.com/capitalone/DataProfiler/pull/871/files#r1230135673

taylorfturner (changes requested):
I think this should fail....

dataprofiler/tests/profilers/test_profile_builder.py
JGSweets: this should be: `test_adding_profiles_of_mismatched_null_count_options`

dataprofiler/profilers/profile_builder.py
```diff
@@ -2051,48 +2060,50 @@ def _update_row_statistics(
...
-        else:
-            self.row_has_null_count = len(null_in_row_count)
-            self.row_is_null_count = len(null_rows)
+                # If sample_ids provided, increment since that means only new data read
+                if sample_ids is not None:
+                    self.row_has_null_count += len(null_in_row_count)
+                    self.row_is_null_count += len(null_rows)
+                else:
+                    self.row_has_null_count = len(null_in_row_count)
+                    self.row_is_null_count = len(null_rows)
```
JGSweets, line 2106: I believe this is causing errors bc these should be a single tab and not two tabs over.

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: you can test this error message below by creating a new options config in `test_add_profilers_row_statistics_options` [here](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/tests/profilers/test_profile_builder.py#L3890)

dataprofiler/profilers/profile_builder.py
micdavis, line 1: you're going to want to update `_get_row_is_null_ratio()` [here](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profile_builder.py#L1968) and `_get_row_has_null_ratio()` [here](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/profilers/profile_builder.py#L1977) so that the base condition looks like:
```
        if not self.options.row_statistics.is_enabled
           or not self.options.row_statistic.null_count.is_enabled:
            return None
```

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: - `test_adding_profiles_of_mismatched_null_count_options` and  `test_add_profiles_null_count_not_enabled` can be removed since their functionality is being added to `test_add_profilers_row_statistics_options`. 

There needs to be tests to check the base conditions for `_get_row_is_null_ratio()` and `_get_row_has_null_ratio()`. 

- The test to check when `None` is [this](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/tests/profilers/test_profile_builder.py#L3618). It already exists, it just needs to be updated so the options include `"row_statistics.null_count.is_enabled": False,`. The test to check when `0` is returned can be similar to [this](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/tests/profilers/test_profile_builder.py#L4077). 
- There will need to be two test to check when `0` is returned, one for `_get_row_is_null_ratio()` and one for `_get_row_has_null_ratio()`. You can follow a similar format as [this](https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/tests/profilers/test_profile_builder.py#L4077) test.

dataprofiler/profilers/profile_builder.py
```diff
@@ -2051,48 +2066,51 @@ def _update_row_statistics(
...
+                    null_in_row_count = null_in_row_count.union(null_row_indices)
+
+                    # If sample_ids provided,
+                    # increment since that means only new data read
+                    if sample_ids is not None:
+                        self.row_has_null_count += len(null_in_row_count)
+                        self.row_is_null_count += len(null_rows)
+                    else:
+                        self.row_has_null_count = len(null_in_row_count)
+                        self.row_is_null_count = len(null_rows)
```
JGSweets, line 2113: I think we need to go the opposite direction with this. It should be outside the for loop. two tabs to the left.

JGSweets: Ultimately, I suggest looking at what indent level it was prior to the change with the `if` and make sure it has not changed relative to the other code tabbing.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1967,7 +1976,10 @@ def _get_unique_row_ratio(self) -> float | None:
 
     def _get_row_is_null_ratio(self) -> float | None:
         """Return whether row is null ratio."""
-        if not self.options.row_statistics.is_enabled:
```
JGSweets, line 1970: nice

JGSweets (changes requested):

dataprofiler/profilers/profile_builder.py
```diff
@@ -1976,7 +1988,10 @@ def _get_row_is_null_ratio(self) -> float | None:
 
     def _get_row_has_null_ratio(self) -> float | None:
         """Return whether row has null ratio."""
-        if not self.options.row_statistics.is_enabled:
+        if (
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.is_enabled
```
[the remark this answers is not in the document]
drahc1R: doh!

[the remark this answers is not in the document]
drahc1R: sorry about that

dataprofiler/profilers/profile_builder.py
taylorfturner: nice -- updated in your PR

dataprofiler/profilers/profile_builder.py
```diff
@@ -1967,7 +1976,10 @@ def _get_unique_row_ratio(self) -> float | None:
 
     def _get_row_is_null_ratio(self) -> float | None:
         """Return whether row is null ratio."""
-        if not self.options.row_statistics.is_enabled:
+        if (
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.null_count.is_enabled
+        ):
```
taylorfturner, line 1982: nice

dataprofiler/profilers/profile_builder.py
```diff
@@ -1976,7 +1988,10 @@ def _get_row_is_null_ratio(self) -> float | None:
 
     def _get_row_has_null_ratio(self) -> float | None:
         """Return whether row has null ratio."""
-        if not self.options.row_statistics.is_enabled:
+        if (
+            not self.options.row_statistics.is_enabled
+            or not self.options.row_statistics.null_count.is_enabled
+        ):
```
taylorfturner, line 1994: nice

dataprofiler/tests/profilers/test_profile_builder.py
taylorfturner: looks fixed now

dataprofiler/profilers/profile_builder.py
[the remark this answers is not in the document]
taylorfturner: looks fixed now

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -4085,6 +4136,18 @@ def test_unique_row_ratio_empty_profiler(self):
...
+        profiler_options = ProfilerOptions()
+        profiler_options.set(
+            {
+                "*.is_enabled": False,
+                "row_statistics.null_count.is_enabled": False,
+            }
+        )
+        profiler = StructuredProfiler(pd.DataFrame([]), options=profiler_options)
+        self.assertIsNone(profiler._get_row_is_null_ratio())
+        self.assertIsNone(profiler._get_row_has_null_ratio())
```
taylorfturner, line 4149: nice checking `None`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3555,6 +3557,57 @@ def setUpClass(cls):
...
+                "row_statistics.null_count.is_enabled": False,
+            }
+        )
+        data = pd.DataFrame([1, None, 3, 4, 5, None, 1])
+        with test_utils.mock_timeit():
+            profiler_w_disabled_null_count = dp.StructuredProfiler(
+                data[2:], options=profiler_options_null_disabled
+            )
+
+        self.assertEqual(0, profiler_w_disabled_null_count.row_has_null_count)
```
JGSweets, line 3608: this line needs to be deleted, right?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3555,6 +3557,57 @@ def setUpClass(cls):
...
+            )
+
+        with self.assertRaisesRegex(
+            ValueError,
+            "Attempting to merge two profiles with null row "
+            "count option enabled on one profile but not the other.",
+        ):
+            profiler_w_null_count + profiler_w_disabled_null_count
+
+    def test_add_profiles_null_count_not_enabled(self):
```
JGSweets, line 3593: rename, not adding profiles

JGSweets (approved):

taylorfturner (approved):

PR #845, opened 2023-05-30
Sampling ratio implement

joshuart:
Adding new feature of enabling `sampling_ratio` with `structured_options`

PR #873, opened 2023-06-14
Fixed documentation for `sampling_ratio` option

clee1152:
Fixed documentation for `sampling_ratio` option.

### DIFF
--- a/docs/source/profiler.rst
+++ b/docs/source/profiler.rst
@@ -657,7 +657,7 @@ Below is an breakdown of all the options.
 
         * is_enabled - (Boolean) Enables or disables precision
       
-      * sample_ratio - (Float) A percentage, as a decimal, ranging from 0 to 1 indicating how much data (identified as floats) to utilize as samples in determining precision
+      * sample_ratio - (Float) The ratio of 0 to 1 how much data (identified as floats) to utilize as samples in determining precision
 
       * min - Finds minimum value in a column
 
--- a/docs/source/profiler.rst
+++ b/docs/source/profiler.rst
@@ -601,7 +601,7 @@ Below is an breakdown of all the options.
 
       * is_enabled - (Boolean) Enables or disables multiprocessing
       
-    * **sampling_ratio** - A percentage, as a decimal, ranging from 0 to 1 indicating how much input data to sample. Default value set to 0.2.
+    * **sampling_ratio** - A percentage, as a decimal, ranging from greater than 0 to less than or equal to 1 indicating how much input data to sample. Default value set to 0.2.
     
     * **int** - Options for the integer columns
 
--- a/docs/source/profiler.rst
+++ b/docs/source/profiler.rst
@@ -600,6 +600,9 @@ Below is an breakdown of all the options.
     * **multiprocess** - Option to enable multiprocessing. Automatically selects the optimal number of processes to utilize based on system constraints.
 
       * is_enabled - (Boolean) Enables or disables multiprocessing
+      
+    * **sample_ratio** - The ratio from 0 to 1 of how much of the input data to sample. Default value set to 0.2.
+    
     * **int** - Options for the integer columns
 
       * is_enabled - (Boolean) Enables or disables the integer operations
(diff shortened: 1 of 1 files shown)


taylorfturner (approved):

docs/source/profiler.rst
```diff
@@ -600,6 +600,9 @@ Below is an breakdown of all the options.
     * **multiprocess** - Option to enable multiprocessing. Automatically selects the optimal number of processes to utilize based on system constraints.
 
       * is_enabled - (Boolean) Enables or disables multiprocessing
+      
+    * **sample_ratio** - The ratio from 0 to 1 of how much of the input data to sample. Default value set to 0.2.
```
taylorfturner, line 604:
```suggestion
    * **sampling_ratio** - The ratio from 0 to 1 of how much of the input data to sample. Default value set to 0.2.
```

taylorfturner (changes requested):

docs/source/profiler.rst
taylorfturner: its actually `sampling_ratio` per https://github.com/capitalone/DataProfiler/pull/845/files#diff-c869332fe6f1bea096dac03badb7365173fa62459962c2e9c020cb53341b1ba6R684

micdavis, line 604:
```suggestion
    * **sample_ratio** - A percentage ranging from 0 to 1 indicating how much input data to sample. Default value set to 0.2.
```

clee1152: Could you submit the same commit suggestion again? This commit suggestion is outdated after I resolved other commit suggestions.

docs/source/profiler.rst
```diff
@@ -600,6 +600,9 @@ Below is an breakdown of all the options.
     * **multiprocess** - Option to enable multiprocessing. Automatically selects the optimal number of processes to utilize based on system constraints.
 
       * is_enabled - (Boolean) Enables or disables multiprocessing
+      
+    * **sampling_ratio** - The ratio from 0 to 1 of how much of the input data to sample. Default value set to 0.2.
```
taylorfturner, line 604:
```suggestion
    * **sampling_ratio** - A percentage, as a decimal, ranging from 0 to 1 indicating how much input data to sample. Default value set to 0.2.
```

docs/source/profiler.rst
taylorfturner: done

taylorfturner (approved):

taylorfturner:
FYI @joshuart -- adding docs for your new option of `sampling_ratio`

docs/source/profiler.rst
```diff
@@ -653,7 +656,7 @@ Below is an breakdown of all the options.
       * precision - Finds the precision (significant figures) within the column
 
         * is_enabled - (Boolean) Enables or disables precision
-      * sample_ratio - (Float) The ratio of 0 to 1 how much data (identified as floats) to utilize as samples in determining precision
```
JGSweets, line 656: I think this is a value within `Float`. We should maintain this.

JGSweets (changes requested):

docs/source/profiler.rst
taylorfturner: +1 good catch -- keep this

taylorfturner (changes requested):
just revert the negative back to the original so there is no diff on this

taylorfturner (approved):
LGTM

docs/source/profiler.rst
```diff
@@ -600,6 +600,9 @@ Below is an breakdown of all the options.
     * **multiprocess** - Option to enable multiprocessing. Automatically selects the optimal number of processes to utilize based on system constraints.
 
       * is_enabled - (Boolean) Enables or disables multiprocessing
+      
+    * **sampling_ratio** - A percentage, as a decimal, ranging from 0 to 1 indicating how much input data to sample. Default value set to 0.2.
```
JGSweets, line 604: Just noticed this, correct me if I'm wrong, but i don't think the value can be 0. So as oppose to from 0 to 1, `more than 0 and <= 1`. Not sure best way to type that up though. Does any other doc have a similar to mirror?

taylorfturner: good catch

tyfarnan (approved):

PR #890, opened 2023-06-20
WIP cms implementation

tyfarnan:
WIP: for early feedback

### DIFF
--- a/dataprofiler/profilers/categorical_column_profile.py
+++ b/dataprofiler/profilers/categorical_column_profile.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from operator import itemgetter
 from typing import cast
 
+from datasketches import count_min_sketch
 from pandas import DataFrame, Series
 
 from . import BaseColumnProfiler, utils
@@ -60,6 +61,8 @@ class CategoricalColumn(BaseColumnProfiler["CategoricalColumn"]):
             self.max_sample_size_to_check_stop_condition = (
                 options.max_sample_size_to_check_stop_condition
             )
+            self._cms_confidence = options.cms_confidence
+            self._cms_relative_error = options.cms_relative_error
 
     def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
         """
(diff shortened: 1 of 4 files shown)


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -345,7 +348,16 @@ def _update_categories(
         :type df_series: pandas.DataFrame
         :return: None
         """
-        category_count = df_series.value_counts(dropna=False).to_dict()
+        if self._cms_confidence:
+            category_count = defaultdict(int)
+            num_hashes = count_min_sketch.suggest_num_hashes(self._cms_confidence)
+            num_buckets = count_min_sketch.suggest_num_buckets(self._cms_relative_error)
+            cm = count_min_sketch(num_hashes, num_buckets)
```
JGSweets, line 355: I'd suggest moving this to init of the class above. The `cm` should persist across updates in case we do batches or streaming datasets

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -345,7 +348,16 @@ def _update_categories(
...
+            for i in df_series:
```
JGSweets, line 356: I'm concerned about looping through the entire if df_series is completely unique. That's where I think the top_k which utilizes a look-up of values that are repetitive.

tyfarnan: hmm... top_k is the obtained by sorting the category_count dict (obtained via cms, which requires a single pass over the series?)... 

Are you suggesting we still run the stopping condition for CMS?

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: nice

dataprofiler/profilers/categorical_column_profile.py
JGSweets: Sorry, I wasn't clear. I'm concerned about storage of the entire df_series in  `category_count`

PR #891, opened 2023-06-20
Deserialization of datalabeler

ksneab7:


JGSweets:
FYI, this PR is missing the update to `base_column_profilers.py`

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -307,6 +307,48 @@ def profile(self) -> dict:
...
+            )
+        else:
+            raise ValueError(
+                "Deserialization cannot be done on labelers without "
+                "_default_model_loc set to known value."
+            )
+
+        # This is an ambiguous call to super classes.
+        # If load_from_dict is part of both super classes there may be issues
+        profile = super().load_from_dict(data, options=opt)
```
JGSweets, line 341: I suggest updating this to be:
```python
profile = super().load_from_dict(data, options={cls.__name__: opt})
```

JGSweets: In base we can do something like:
```python
if options is None:
            options = {}
        
class_options = options.get(cls.__name__)
profile = cls(data["name"], class_options)
```

JGSweets: The above for me solved all the tests.
However, mypy decided to get a little talkative. 
Unfortunately, a proper solve requires generics. Something that was merged into main, but not here.
Which means, we likely need to partially update for generics from #840 in this and will have to resolve rebase on to main conflicts when they arise in the future.

JGSweets: My suggestions to solve (don't include the random comments): 

`base_column_profilers.py`
```python
# in imports we need:
# add the extra imports to typing
from typing import Any, Callable, Generic, TypeVar

# refactor this import to be relative and add the BaseOption
from .profiler_options import BaseInspectorOptions, BaseOption

# add the Generic for any BaseColumn
BaseColumnProfilerT = TypeVar("BaseColumnProfilerT", bound="BaseColumnProfiler")

# update the class  to use generics, remove the typing ignore :)
class BaseColumnProfiler(Generic[BaseColumnProfilerT], metaclass=abc.ABCMeta):
    
    # update init, takes a BaseOption now... likely a more appropriate way to do this, but for now works
    def __init__(self, name: str | None, options: BaseOption | None = None):
        ...

    .
    .
    .
    # fix load from dict
    @classmethod
    def load_from_dict(
        cls: type[BaseColumnProfilerT],
        data: dict[str, Any],
        options: dict | None = None,
    ) -> BaseColumnProfilerT:
        """
        Parse attribute from json dictionary into self.

        :param data: dictionary with attributes and values.
        :type data: dict[string, Any]

        :rtype: BaseColumnProfiler
        """
        if options is None:
            options = {}

        class_options = options.get(cls.__name__)
        profile: BaseColumnProfilerT = cls(data["name"], class_options)
        .
        .
        .
```



`data_labeler_column_profile.py`
```python
# update the class
class DataLabelerColumn(BaseColumnProfiler["DataLabelerColumn"]):
    .
    .
    .
    # fi load_from_dict to specify proper output class
    @classmethod
    def load_from_dict(cls, data, options: dict | None = None) -> DataLabelerColumn:
         # other fixes in here as mentioned above for getting the profile
     .
     .
     .
```

`json_decoder.py`
```python
# new imports to add
from typing import TYPE_CHECKING, Dict, Type


def load_column_profile(serialized_json: dict) -> "BaseColumnProfiler":
    .
    .
    .
    # updating the expected type here, i think this is still incorrect, but unsure proper way to generic it; fine for now
    column_profiler_cls: Type[
        "BaseColumnProfiler[BaseColumnProfiler]"
    ] = get_column_profiler_class(serialized_json["class"])
    return column_profiler_cls.load_from_dict(serialized_json["data"])
.
.
.
```

JGSweets: in `json_decoder.py`,
can also fix cases of `Optional[...]` ->  `... | None`

JGSweets: Will be fun when we have to do these same fixes for the compiler maybe...

dataprofiler/profilers/data_labeler_column_profile.py
```diff
@@ -307,6 +307,48 @@ def profile(self) -> dict:
...
+
+        :param data: dictionary with attributes and values.
+        :type data: dict[string, Any]
+        :param options: options for loading column profiler params from dictionary
+        :type options: Dict | None
+
+        :return: Profiler with attributes populated.
+        :rtype: DataLabelerColumn
+        """
+        opt = DataLabelerOptions()
```
JGSweets, line 323: if we want a labeler to be applicable across all cases. I assume it also comes through options here... I'm curious how we recognize it here in this `load_from_dict`. Does it come via
```python
{
    "DataLabelerColumn": {"data_labeler": ...},
    ...
}
```

What does the `...` look like?
is it similar where it says load_from_library? is it just the object?

Does it override the `data`?

Or does the data have precedence and if it exists in `options` it pulls from that. e.g. from_library matches the options `from_library` and the name is the same and has the object.
e.g.
```python
{
    "DataLabelerColumn": {"data_labeler": {"load_from_library": {"labeler_name": LabelerObject}}},
}
```

JGSweets: @ksneab7 thoughts?

dataprofiler/tests/profilers/test_data_labeler_column_profile.py
```diff
@@ -27,6 +28,8 @@ def _setup_data_labeler_mock(mock_instance):
         mock_DataLabeler.reverse_label_mapping = {0: "a", 1: "b"}
         mock_DataLabeler.model.num_labels = 2
         mock_DataLabeler.model.requires_zero_mapping = False
+        # mock_instance.load_from_library.return_value = mock_instance
```
JGSweets, line 31: remove comment

JGSweets (approved):

micdavis (approved):

PR #892, opened 2023-06-21
Cms for categorical

tyfarnan:
Looking for feedback on which tests should be duplicated for CMS, and what new tests we should create specific to CMS.


dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -345,13 +358,29 @@ def _update_categories(
...
-            self._categories = {}
+        if self.cms is not None:
+            if self._top_k_categories is None:
+                raise ValueError("when using CMS, top_k must be an integer")
+            category_count = defaultdict(int)
+            for i in df_series:
+                self.cms.update(i)
+                # approximate heavy-hitters
+                if self.cms.get_estimate(i) >= int(
+                    self.sample_size / self._top_k_categories
```
tyfarnan, line 369: thoughts on creating a new var self._heavy_hitters_threshold instead of misusing top_k_categories here? In the CMS setting we are using top_k_categories as a minimum count for us to track a category, this doesn't imply we'll end up with k categories...

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -60,6 +63,18 @@ def __init__(self, name: str | None, options: CategoricalOptions = None) -> None
...
+
+            if options.cms:
+                self._heavy_hitters_threshold = options.heavy_hitters_threshold
+                self.num_hashes = count_min_sketch.suggest_num_hashes(
+                    options.cms_confidence
+                )
+                self.num_buckets = count_min_sketch.suggest_num_buckets(
+                    options.cms_relative_error
+                )
+                self.cms = count_min_sketch(self.num_hashes, self.num_buckets)
```
tyfarnan, line 77: @JGSweets having second thoughts that cms needs to be defined here... we only ever use it within _update_categories(), where we get an heavy-hitter approximate category count and use that to update self._categories. I think as long as self._categories persists we are safe and don't need to also persist cms

...Also having self.cms defined here causes serialization troubles. count_min_sketch has its own serialization method so we'd need some custom workaround to get json.dumps() to work with it out of the box if we choose to keep it defined here...

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -60,6 +63,18 @@ def __init__(self, name: str | None, options: CategoricalOptions = None) -> None
...
 
     def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
```
JGSweets, line 79: need to also solve this add

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -345,13 +360,31 @@ def _update_categories(
         :type df_series: pandas.DataFrame
         :return: None
         """
-        category_count = df_series.value_counts(dropna=False).to_dict()
```
JGSweets, line 348: I think something like this and it allows to reuse for `__add__` as well.

```python
def _get_categories(....):


def _merge_categories(...):


def _update_categories(...):

    # new_summary = _get_categories(new_data)
    cms = count_min_sketch(self..., self...)
    heavy_hitter_dict = {}
    for i in df_series:
        cms.update(i)
        # approximate heavy-hitters
        if cms.get_estimate(i) >= int(
            len(df_series) / self._heavy_hitters_threshold
        ):
            heavy_hitter_dict[i] = self.cms.get_estimate(i)

    
    
    # old_summary = self.....of info
    # _merge_categories(new_summary, old_summary)
    self.cms = self.cms + cms
    
    categories = utils.add_nested_dictionaries(
        self._categories, heavy_hitter_dict
    )
    
    total_samples = len(data) + self.total_samples
    for cat in categories:
        if categories["cat"] < (total_samples / self._heavy_hitters_threshold):
            categories.pop("cat")
            
    self._categories = categories
```

JGSweets: if we abstract to get and merge, `__add__` can utilize the merge if we do it via a new and prev

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -5,6 +5,7 @@
 from operator import itemgetter
 from typing import cast
 
+from datasketches import count_min_sketch
```
JGSweets, line 8: instead, let's do 
```python
import datasketches
```

elsewhere we can do:
```python
datasketches.count_min_sketch
```

Otherwise, we don't know the lineage of `count_min_sketch` for clairity as it is external.

tyfarnan: gotcha!

.github/workflows/test-python-package.yml
```diff
@@ -8,6 +8,7 @@ on:
     branches:
       - 'main'
       - 'feature/**'
+      - 'dev'
```
JGSweets, line 11: hmm, not sure why this would be here if dev already has this, are you sure you rebased on to dev

dataprofiler/profilers/categorical_column_profile.py
JGSweets: Updated above for new thoughts wrt secondary list to cover edge cases.

.github/workflows/test-python-package.yml
taylorfturner: +1

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -83,6 +104,28 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
         self._merge_calculations(
             merged_profile.__calculations, self.__calculations, other.__calculations
         )
+
+        if self.cms and other.cms:
```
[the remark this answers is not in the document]
JGSweets: e.g. I would assume to merge sketches, these need to be the same:
```python
self.num_hashes
self.num_buckets
```

anything else need to be checked too?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -323,6 +366,75 @@ def _update_stop_condition(self, data: DataFrame):
...
+        :type len1: int
+        :param len2: number of samples in batch 2
+        :type len2: int
+        :return: cms1, categories
+        """
+        cms2.merge(cms1)
+        categories = utils.add_nested_dictionaries(
+            heavy_hitter_dict2, heavy_hitter_dict1
+        )
+        categories.update(missing_heavy_hitter_dict)
```
JGSweets, line 430: we might want to put a comment here about this being a catchall.

JGSweets: or the intent of this

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -323,6 +366,75 @@ def _update_stop_condition(self, data: DataFrame):
...
+
+        total_samples = len1 + len2
+        for cat in list(categories):
+            if categories[cat] < (total_samples / self._heavy_hitters_threshold):
```
JGSweets, line 434: what are our thoughts on `_heavy_hitters_threshold` as naming since this is more like a top_k. Do we want it to be a ratio of total data?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -83,6 +104,28 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
         )
+
+        if self.cms and other.cms:
+            # decode prior to use
+            self.cms = datasketches.count_min_sketch.deserialize(
+                base64.b64decode(bytes(self.cms, "utf-8"))
+            )
+            other.cms = datasketches.count_min_sketch.deserialize(
+                base64.b64decode(bytes(other.cms, "utf-8"))
+            )
```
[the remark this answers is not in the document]
tyfarnan: the sketch object itself needs to be serialized on its own in order for the entire profile to be serialized. The pattern I used around CMS was to deserialize, modify, re-serialize.

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: that's a good idea

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -345,13 +457,42 @@ def _update_categories(
...
-            self._categories = {}
+        if self.cms is not None:
+            if self._heavy_hitters_threshold is None:
+                raise ValueError(
+                    "when using CMS, heavy_hitters_threshold must be an integer"
+                )
+            # decode prior to use
+            self.cms = datasketches.count_min_sketch.deserialize(
+                base64.b64decode(bytes(self.cms, "utf-8"))
+            )
```
JGSweets, line 468: Why decode just to reserialize?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -61,6 +68,20 @@ def __init__(self, name: str | None, options: CategoricalOptions = None) -> None
...
 
+            if options.cms:
+                self._heavy_hitters_threshold = options.heavy_hitters_threshold
+                self.num_hashes = datasketches.count_min_sketch.suggest_num_hashes(
+                    options.cms_confidence
+                )
+                self.num_buckets = datasketches.count_min_sketch.suggest_num_buckets(
+                    options.cms_relative_error
+                )
+                self.cms = base64.b64encode(
```
JGSweets, line 79: What's the i

dataprofiler/profilers/categorical_column_profile.py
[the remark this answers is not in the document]
tyfarnan: lower level cms.merge() handles the incompatibility issues.

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: if you don't encode the bytes as a string, you get "JSON can't serialize bytes" error on json.dumps()

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -822,7 +1012,13 @@ def test_categorical_stop_condition_options_set(self):
         self.assertEqual(100, profile.max_sample_size_to_check_stop_condition)
 
     def test_json_encode(self):
-        profile = CategoricalColumn("0")
```
JGSweets, line 825: shouldn't replace, should add

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -52,6 +54,11 @@ def __init__(self, name: str | None, options: CategoricalOptions = None) -> None
 
         self._stopped_at_unique_ratio: float | None = None
         self._stopped_at_unique_count: int | None = None
+
+        self._heavy_hitters_threshold: int | None = None
+        self.num_hashes: int | None = None
+        self.num_buckets: int | None = None
```
JGSweets, line 60: might want to preface with `cms`

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -83,6 +104,28 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
+                self.sample_size,
+                {},
+                other.cms,
+                other._categories,
+                other.sample_size,
+            )
+            merged_profile.cms = base64.b64encode(
+                merged_profile.cms.serialize()
+            ).decode("utf-8")
+
```
JGSweets, line 128: missing `elif when not cms for both`

dataprofiler/profilers/profiler_options.py
```diff
@@ -886,13 +890,28 @@ def __init__(
...
         BaseInspectorOptions.__init__(self, is_enabled=is_enabled)
         self.top_k_categories = top_k_categories
         self.max_sample_size_to_check_stop_condition = (
             max_sample_size_to_check_stop_condition
         )
         self.stop_condition_unique_value_ratio = stop_condition_unique_value_ratio
+        self.cms = cms
+        self.cms_confidence = cms_confidence
+        self.cms_relative_error = cms_relative_error
+        self.heavy_hitters_threshold = heavy_hitters_threshold
```
JGSweets, line 914: do we want a default value for this

dataprofiler/profilers/profiler_options.py
```diff
@@ -886,13 +890,28 @@ def __init__(
...
+            to store a distinct category.
+        :vartype heavy_hitters_threshold: [None, int]
         """
         BaseInspectorOptions.__init__(self, is_enabled=is_enabled)
         self.top_k_categories = top_k_categories
         self.max_sample_size_to_check_stop_condition = (
             max_sample_size_to_check_stop_condition
         )
         self.stop_condition_unique_value_ratio = stop_condition_unique_value_ratio
+        self.cms = cms
```
JGSweets, line 911: as opposed to cms = True or False.

I think it might be better to use a `method` param
which chooses `"full" | "cms".

open to thoughts or opposition here.

dataprofiler/profilers/profiler_options.py
```diff
@@ -869,9 +869,13 @@ class CategoricalOptions(BaseInspectorOptions):
...
         self,
         is_enabled: bool = True,
-        top_k_categories: int = None,
+        top_k_categories: int | None = None,
         max_sample_size_to_check_stop_condition: int | None = None,
         stop_condition_unique_value_ratio: float | None = None,
+        cms: bool = False,
+        cms_confidence: float | None = 0.95,
+        cms_relative_error: float | None = 0.01,
+        heavy_hitters_threshold: int | None = None,
```
JGSweets, line 878: i'm okay with a default of `5000` since that's when we start limiting our dataset.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -936,6 +1140,93 @@ def test_json_decode_after_update(self):
...
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
+                "c",
```
JGSweets, line 1166: I like in the below where you use `["a"] * ...+ ...`, makes it clear how many to expect for each.

JGSweets: looks like 5,5, 10 respectively

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -936,6 +1140,93 @@ def test_json_decode_after_update(self):
...
+        options = CategoricalOptions()
+        options.cms = True
+        options.cms_confidence = 0.95
+        options.cms_relative_error = 0.01
+        options.heavy_hitters_threshold = 2
+
+        profile = CategoricalColumn("test_name", options)
+        profile.update(df_categorical)
+
+        self.assertEqual(1, len(profile.categories))
```
JGSweets, line 1178: instead we should assert the dict itself for clarity.
I assume it would be {"c": 10}

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -936,6 +1140,93 @@ def test_json_decode_after_update(self):
...
+        dataset1 = pd.Series(["a"] * 9 + ["b"] * 11 + ["c"] * 9 + ["d"] * 1)
+
+        options = CategoricalOptions()
+        options.cms = True
+        options.cms_confidence = 0.95
+        options.cms_relative_error = 0.01
+        options.heavy_hitters_threshold = 3
+
+        profile = CategoricalColumn("test_name", options)
+        profile.update(dataset)
```
JGSweets, line 1192: we should do an assert after this first.

then a second assert after the next update.

We can then validate the update as expected. probably good to have  the case where a was not initially included, but is after.

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -936,6 +1140,93 @@ def test_json_decode_after_update(self):
...
+        self.assertTrue(profile._categories["c"] >= 23)
+
+        self.assertCountEqual(expected_categories, profile.categories)
+
+        self.assertTrue(profile._categories["b"] >= 23)
+
+    def test_cms_profile_merge_via_add(self):
+
+        dataset = pd.Series(["a"] * 9 + ["b"] * 12 + ["c"] * 9)
+        dataset1 = pd.Series(["a"] * 6 + ["b"] * 10 + ["c"] * 14)
```
JGSweets, line 1207: same as above

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -936,6 +1140,93 @@ def test_json_decode_after_update(self):
...
+        profile2 = CategoricalColumn("test_name", options)
+        profile2.update(dataset1)
+
+        # Add profiles
+        profile3 = profile1 + profile2
+        self.assertCountEqual(expected_categories, profile3.categories)
+        self.assertEqual(
+            profile3.sample_size, profile1.sample_size + profile2.sample_size
+        )
+        self.assertTrue(profile3._categories["b"] == 22)
```
JGSweets, line 1227: assert the dict instead

dataprofiler/profilers/categorical_column_profile.py
JGSweets, line 430: This actually might be problematic for the profile1 + profile2 case. We need to consider that categories of profile1 or profile2 may not have the full data bc it was `<` the limit. For cases where missing from one or the other, we might need to pull it from the `cms`

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -936,6 +1140,93 @@ def test_json_decode_after_update(self):
...
+        self.assertTrue(profile3._categories["c"] == 14)
```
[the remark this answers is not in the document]
tyfarnan: there is indeed a bug here - but our HH pruning actually removes the C category completely right now.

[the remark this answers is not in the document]
tyfarnan: the intent of this test was to try and illustrate the "gap" that occurs during add. Since we aren't treating as a stream, missing_heavy_hitters_dict doesn't help us capture the "c":9 from the first dataset that fell below the "batch HH" threshold...

I originally thought it would be 14 but realize now that by the same logic, 14 < (total_samples/HH)

JGSweets:
Easy fix on failed test, will review rest of PR tonight. But once that is in, should be good to go.

dataprofiler/tests/profilers/profiler_options/test_categorical_options.py
```diff
@@ -46,6 +58,10 @@ def test_init(self):
                 "top_k_categories": None,
                 "max_sample_size_to_check_stop_condition": 20,
                 "stop_condition_unique_value_ratio": None,
+                "cms": False,
+                "cms_confidence": 0.95,
+                "cms_relative_error": 0.01,
+                "cms_max_num_heavy_hitters": None,
```
JGSweets, line 64: I think the fixes to these will be 5000 since update to default

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -108,6 +131,173 @@ def test_timeit_profile(self):
             expected = defaultdict(float, {"categories": 2.0})
             self.assertEqual(expected, profile.profile["times"])
 
+    def test_cms_mixed_categorical_col_integer_string(self):
+        options = CategoricalOptions()
+        options.cms_confidence = 0.95
+        options.cms_relative_error = 0.1
+        options.top_k_categories = 5
+        dataset = self.aws_dataset["localeabbr"].dropna()
```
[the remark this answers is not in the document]
tyfarnan: removing this duplicate test, it isn't covering anything new

dataprofiler/tests/profilers/test_categorical_column_profile.py
tyfarnan: modified test to address this case

dataprofiler/profilers/profiler_options.py
tyfarnan: It's currently just being used as an on/off switch. Useful to have param method if there's many options to choose from. Are there plans to expand?

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -83,53 +103,80 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
-            else:
-                merged_profile._stopped_at_unique_ratio = other.unique_ratio
-                merged_profile._stopped_at_unique_count = other.unique_count
-                merged_profile.sample_size = other.sample_size
-
-            # If either profile has hit stop condition, remove categories dict
-            merged_profile._categories = {}
-            merged_profile._stop_condition_is_met = True
+            raise Exception(
+                "Unable to add two profiles is only one is using count min sketch."
```
JGSweets, line 178: just a grammar fix here:

> "Unable to add two profiles: One is using count min sketch and the other is using full."

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -323,6 +370,113 @@ def _update_stop_condition(self, data: DataFrame):
...
+        heavy_hitter_dict = defaultdict(int)
+        missing_heavy_hitter_dict = defaultdict(int)
+        for i, value in enumerate(df_series):
+            cms.update(value)
+            i_count = cms.get_estimate(value)
+            i_total_count = i_count + self.cms.get_estimate(value)
+            # approximate heavy-hitters
+            if i_count >= int(len_df / self._cms_max_num_heavy_hitters):
+                heavy_hitter_dict[value] = i_count
+                # missing_heavy_hitter_dict.pop(value, None)
```
JGSweets, line 391: I think this is problematic if there's a case where it ends in this `if` but gets populated in the second and then gets overwritten. Need to still read below to see if this is problematic.

was the a purpose of removing

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -323,6 +370,113 @@ def _update_stop_condition(self, data: DataFrame):
...
+        # frequencies compared to treated as a sequence of batches as part of
+        # the same stream.
+        categories.update(missing_heavy_hitter_dict)
+
+        if other_max_num_heavy_hitters:
+            max_num_heavy_hitters = min(
+                self._cms_max_num_heavy_hitters, other_max_num_heavy_hitters
+            )
+        else:
+            max_num_heavy_hitters = self._cms_max_num_heavy_hitters
```
JGSweets, line 472: let's do this in the add above as we don't need to do this for streaming.

JGSweets: `other_max_num_heavy_hitters` may not be the appropriate count, it might be just the value not acquired. We should use the actual profiles value.

dataprofiler/profilers/profiler_options.py
JGSweets: Was just trying to match the same technique we used for `unique_count` which makes it extensible and more clear wrt what the other option you are choosing or have set as an option when reading code or options.

dataprofiler/tests/profilers/test_categorical_column_profile.py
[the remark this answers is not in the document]
JGSweets: If we specify heavy_hitters to be `10` would it not limit the results to meet this case which would be a different result than the below?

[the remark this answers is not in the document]
JGSweets: or do we handle the limiting heavy_hitters case?

dataprofiler/profilers/categorical_column_profile.py
JGSweets: I created this test case which will fail:
```python
dataset = pd.Series(["b"] * 2 + ["c"] * 14)
dataset1 = pd.Series(["b"] * 5 + ["c"] * 9)

options = CategoricalOptions()
options.cms = True
options.cms_confidence = 0.95
options.cms_relative_error = 0.01
options.cms_max_num_heavy_hitters = 3

profile = CategoricalColumn("test_name", options)
profile.update(dataset)

expected_categories = ["c"]
expected_categories_dict = {"c": 14}

self.assertEqual(profile.sample_size, len(dataset))
self.assertEqual(profile._categories, expected_categories_dict)
self.assertCountEqual(expected_categories, profile.categories)

profile.update(dataset1)
expected_categories = ["c"]
expected_categories_dict = {"c": 23}

self.assertEqual(profile.sample_size, len(dataset) + len(dataset1))
self.assertEqual(profile._categories, expected_categories_dict)
self.assertCountEqual(expected_categories, profile.categories)
```

JGSweets: Ultimately, the `update` overwrites:
`AssertionError: defaultdict(<class 'int'>, {'c': 18.0}) != {'c': 24}`

JGSweets: uncommenting both of these will pass this test.

JGSweets: We can add this test, or find a way to have this be a condition in the current test with a third update.

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -323,6 +370,113 @@ def _update_stop_condition(self, data: DataFrame):
...
+        missing_in_1 = [
+            k for k in heavy_hitter_dict1.keys() if k not in heavy_hitter_dict2.keys()
+        ]
+        for k in missing_in_1:
+            heavy_hitter_dict2[k] = cms2.get_estimate(k)
+        missing_in_2 = [
+            k for k in heavy_hitter_dict2.keys() if k not in heavy_hitter_dict1.keys()
+        ]
+        for k in missing_in_2:
+            heavy_hitter_dict1[k] = cms1.get_estimate(k)
```
JGSweets, line 456: good call on this for the streaming case to catch.
I think we can clean this by doing:
```python
# re-collecting the estimates of non intersecting categories before
# re-applying heavy-hitters to the aggregate profile.
heavy_hitter_dict1 = heavy_hitter_dict1.copy()
heavy_hitter_dict2 = heavy_hitter_dict2.copy()
for k in (x for x in heavy_hitter_dict1 if x not in heavy_hitter_dict2):
    heavy_hitter_dict2[k] = cms2.get_estimate(k)
for k in (x for x in heavy_hitter_dict2 if x not in heavy_hitter_dict1):
    heavy_hitter_dict1[k] = cms1.get_estimate(k)
```

now down to a single loop and we don't have to call keys to loop.

Additionally, since this is now altering the dict, we need to copy any dict being passed in here so we don't alter the original.

JGSweets: anything passing in `self` or `other` for the heavy_hitter_dict would mutate the original dict otherwise.

dataprofiler/profilers/categorical_column_profile.py
JGSweets: We need a test here to validate this exception.

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: comments were an accident while debugging.

tyfarnan: adding this test

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: nice!

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: refactored this to __add__ using actual profile values

dataprofiler/profilers/categorical_column_profile.py
tyfarnan: added

JGSweets:
FYI, looks like the reqs for this need to be updated for scipy like the other dev PRs

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -83,53 +103,88 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
-        # If both profiles have not met stop condition
-        if not (self._stop_condition_is_met or other._stop_condition_is_met):
-            merged_profile._categories = utils.add_nested_dictionaries(
-                self._categories, other._categories
+
+        if self.cms and other.cms:
+
+            assert isinstance(self._cms_max_num_heavy_hitters, int)
+            assert isinstance(other._cms_max_num_heavy_hitters, int)
+            cms_max_num_heavy_hitters: int = min(
```
JGSweets, line 111: i'd say we could have assigned here and not output at merge (currently just passing through), but that can be a future refactor.

JGSweets (dismissed):

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -83,53 +103,88 @@ def __add__(self, other: CategoricalColumn) -> CategoricalColumn:
...
                 )
 
-            # Check merged profile w/ stop condition
-            if merged_profile._check_stop_condition_is_met(
-                merged_profile.sample_size, merged_profile.unique_ratio
-            ):
-                merged_profile._stopped_at_unique_ratio = merged_profile.unique_ratio
-                merged_profile._stopped_at_unique_count = merged_profile.unique_count
+                # Transfer stop condition variables of 1st profile object to
+                # merged profileif they are not None else set to 2nd profile
```
ksneab7, line 138:
```suggestion
                # merged profile if they are not None else set to 2nd profile
```

JGSweets (approved):

ksneab7 (approved):

taylorfturner (approved):

PR #894, opened 2023-06-21
Decode options

micdavis:
- Added `Generic` typing to Options classes
- Added Options class name resolver to `json_decoder.py`
- Added `load_from_dict()` to `BaseOptions`
- Added a test to `test_base_options.py` to test the decoding functionality


dataprofiler/profilers/json_decoder.py
```diff
@@ -3,12 +3,14 @@
 
 if TYPE_CHECKING:
     from .base_column_profilers import BaseColumnProfiler
-    from .profilers import column_profile_compilers as col_pro_compilers
+    from .column_profile_compilers import BaseCompiler
```
micdavis, line 6: not sure if how `.profilers` works or why we can't have `.column_profile_compilers`

dataprofiler/profilers/json_decoder.py
```diff
@@ -3,12 +3,14 @@
```
micdavis, line 3: Curious if we can consolidate `_profiles`, `_compilers`, and `_options` into one grouping, cause there is a lot of duplicate code. If we can, would it make sense to do so?

dataprofiler/profilers/profiler_options.py
```diff
@@ -6,11 +6,16 @@
 import copy
```
micdavis, line 6: Consists of: 

- `load_to_dict()`
- creating `Generic` typing for `BaseOptions`
- updating the typing for some attributes

dataprofiler/profilers/utils.py
```diff
@@ -34,7 +34,7 @@
 from .. import settings
 
 if TYPE_CHECKING:
-    from . import profilers
+    from . import profile_builder
```
micdavis, line 37: does this make more sense or is this some black magic that I don't know about

dataprofiler/tests/profilers/profiler_options/test_base_option.py
```diff
@@ -58,3 +62,12 @@ def test_eq(self):
         self.assertEqual(options, options)
         options2 = self.get_options()
         self.assertEqual(options, options2)
+
+    def test_json_decode(self):
+        expected_options = self.get_options()
+        if type(expected_options) == BaseOption:
```
micdavis, line 68: can we ignore test like this if `type()` is `BaseOption`?

dataprofiler/profilers/profiler_options.py
```diff
@@ -6,11 +6,16 @@
...
 import re
 import warnings
+from typing import Generic, TypeVar
+
+from dataprofiler.profilers.json_decoder import load_option
 
 from ..labelers.base_data_labeler import BaseDataLabeler
 
+BaseOptionT = TypeVar("BaseOptionT", bound="BaseOption")
```
tyfarnan, line 15: what's the T for? Type?

dataprofiler/profilers/json_decoder.py
JGSweets: Either via one dict or a func that takes in the dict. I'm curious of the mypy....

dataprofiler/profilers/profiler_options.py
```diff
@@ -138,6 +143,26 @@ def validate(self, raise_error: bool = True) -> list[str] | None:
...
+        :param data: dictionary with attributes and values.
+        :type data: dict[string, Any]
+
+        :return: Profiler with attributes populated.
+        :rtype: BaseColumnProfiler
+        """
+        profile = cls()
+
+        for attr, value in data.items():
+            if isinstance(value, dict) and "class" in value:
```
micdavis, line 160: is it safe to assume that sub-object would be an options object under these conditions? I think its safe to assume now, but for future reference?

dataprofiler/profilers/json_decoder.py
JGSweets: can you explain more? why was this changed?

dataprofiler/tests/profilers/profiler_options/test_base_option.py
JGSweets: whats the actual issue, this seems problematic if that causes an issue

dataprofiler/profilers/json_decoder.py
micdavis: this seems more straight forward. I also don't know where `.profilers` references to

dataprofiler/tests/profilers/profiler_options/test_base_option.py
JGSweets: if it doesn't error, then get rid of the skip.

JGSweets: Otherwise, use mock for abstractmethods like we do for BaseProfiler

taylorfturner (changes requested):

dataprofiler/profilers/float_column_profile.py
```diff
@@ -166,7 +166,7 @@ def load_from_dict(cls, data):
         :type data: dict[string, Any]
 
         :return: Profiler with attributes populated.
-        :rtype: CategoricalColumn
+        :rtype: FloatColumn
```
taylorfturner, line 169: nice

dataprofiler/profilers/int_column_profile.py
```diff
@@ -81,7 +81,7 @@ def load_from_dict(cls, data):
         :type data: dict[string, Any]
 
         :return: Profiler with attributes populated.
-        :rtype: CategoricalColumn
+        :rtype: IntColumn
```
taylorfturner, line 84: nice

dataprofiler/profilers/json_decoder.py
JGSweets: ahh, yes that's weird

JGSweets: `from . import column_profile_compilers as col_pro_compilers` would keep consistency

dataprofiler/tests/profilers/profiler_options/test_profiler_options.py
```diff
@@ -505,28 +503,6 @@ def test_eq(self, *mocks):
         options.structured_options.float.precision.sample_ratio = 0.1
         self.assertNotEqual(options, options2)
 
-    def test_json_encode(self, *mocks):
```
micdavis, line 508: this was moved to `test_profiler_class_options.py` since that test class inherits `TestBaseOptions`

dataprofiler/tests/profilers/profiler_options/test_base_option.py
```diff
@@ -1,13 +1,18 @@
+import json
 import unittest
 
+pass
```
JGSweets, line 4: remove pass

JGSweets (approved):

dataprofiler/profilers/json_decoder.py
```diff
@@ -2,13 +2,16 @@
 from typing import TYPE_CHECKING, Dict, Optional, Type
 
 if TYPE_CHECKING:
+    import column_profile_compilers as col_pro_compiler
+
```
taylorfturner, line 6: is this an isort spacing?

dataprofiler/profilers/profiler_options.py
```diff
@@ -150,7 +175,10 @@ def __eq__(self, other: object) -> bool:
         return self.__dict__ == other.__dict__
 
 
-class BooleanOption(BaseOption):
+BooleanOptionT = TypeVar("BooleanOptionT", bound="BooleanOption")
```
taylorfturner, line 178: small change of just moving this to the top of the file and keep all these TypeVars in the same spot for defining

taylorfturner (approved):

ksneab7 (approved):

Issue #722, opened 2022-11-16
Investigate use of # type: ignore in labelers.


Sanketh7:
Investigate whether these instances of `# type: ignore` are needed.
PRs should be small (at most a few files).

labelers/base_data_labeler.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L26
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L41
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L45
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L47
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L622
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L624
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_data_labeler.py#L627

labelers/base_model.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_model.py#L56
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/base_model.py#L224

labelers/char_load_tf_model.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/char_load_tf_model.py#L395

labelers/character_level_cnn_model.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/character_level_cnn_model.py#L38
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/character_level_cnn_model.py#L61
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/character_level_cnn_model.py#L62
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/character_level_cnn_model.py#L674

labelers/data_labelers.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/data_labelers.py#L58 
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/data_labelers.py#L98

labelers/data_processing.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/labelers/data_processing.py#L54
[...]

PR #895, opened 2023-06-22
Fix name mangling and typevar errors

junholee6a:

### DIFF
--- a/dataprofiler/labelers/data_processing.py
+++ b/dataprofiler/labelers/data_processing.py
@@ -49,16 +49,14 @@ class BaseDataProcessor(metaclass=abc.ABCMeta):
     def _register_subclass(cls) -> None:
         """Register a subclass for the class factory."""
         if not inspect.isabstract(cls):
-            cls._BaseDataProcessor__subclasses[  # type: ignore
-                cls.__name__.lower()
-            ] = cls
+            cls.__subclasses[cls.__name__.lower()] = cls
 
     @classmethod
-    def get_class(cls: type[Processor], class_name: str) -> type[Processor] | None:
+    def get_class(
+        cls: type[BaseDataProcessor], class_name: str
+    ) -> type[BaseDataProcessor] | None:
         """Get class of BaseDataProcessor object."""
-        return cls._BaseDataProcessor__subclasses.get(  # type: ignore
-            class_name.lower(), None
-        )
+        return cls.__subclasses.get(class_name.lower(), None)
 
     def __eq__(self, other: object) -> bool:
         """
--- a/dataprofiler/labelers/base_model.py
+++ b/dataprofiler/labelers/base_model.py
@@ -32,7 +32,7 @@ class AutoSubRegistrationMeta(abc.ABCMeta):
 class BaseModel(metaclass=abc.ABCMeta):
     """For labeling data."""
 
-    _BaseModel__subclasses: dict[str, type[BaseModel]] = {}
+    __subclasses: dict[str, type[BaseModel]] = {}
     __metaclass__ = abc.ABCMeta
 
     # boolean if the label mapping requires the mapping for index 0 reserved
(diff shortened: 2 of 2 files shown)


JGSweets:
> Issue: #722
> 
> Inside the `BaseDataProcessor` class definition, all references to `__subclasses` are automatically replaced with `_BaseDataProcessor__subclasses`. This remains the case even in static methods `_register_subclass()` and `get_class()`. Same with `BaseModel` and its `__subclasses` field. So we do not have to write out the full name mangled identifiers inside the class definitions.
> 
> Also, mypy doesn't seem to be able to handle the return type of `BaseDataProcessor.get_class()` being a typevar, so that was changed to `type[BaseDataProcessor]`. This does not affect the functionality of `get_class()` since it always returns a subclass of `BaseDataProcessor`.

IIRC, there's actually intent to use `_BaseDataProcessor__subclasses` if a different `cls` was used to call `create` might be an anti-pattern though.

junholee6a:
> IIRC, there's actually intent to use _BaseDataProcessor__subclasses if a different cls was used to call create might be an anti-pattern though.

@JGSweets Good point, I hadn't thought the case where `cls` is a subclass of `BaseDataProcessor`

Fortunately, it seems that since `get_class()` and `_register_subclass()` were defined in `BaseDataProcessor`, `__subclasses` is interpreted as `_BaseDataProcessor__subclasses` even if they are called with `cls` being a subclass of `BaseDataProcessor`. For example, this modified test still passes:

Screenshot 2023-06-22 at 12 39 46 PM

JGSweets:
> > IIRC, there's actually intent to use _BaseDataProcessor__subclasses if a different cls was used to call create might be an anti-pattern though.
> 
> @JGSweets Good point, I hadn't thought the case where `cls` is a subclass of `BaseDataProcessor`
> 
> Fortunately, it seems that since `get_class()` and `_register_subclass()` were defined in `BaseDataProcessor`, `__subclasses` is interpreted as `_BaseDataProcessor__subclasses` even if they are called with `cls` being a subclass of `BaseDataProcessor`. For example, this modified test still passes:
> 
> Screenshot 2023-06-22 at 12 39 46 PM

I wonder if that is a change in py version. Maybe 3.6 or 3.7 didn't allow this, but we no longer support it anyways.

dataprofiler/labelers/data_processing.py
```diff
@@ -49,16 +49,14 @@ def __init__(self, **parameters: Any) -> None:
...
-            cls._BaseDataProcessor__subclasses[  # type: ignore
-                cls.__name__.lower()
-            ] = cls
+            cls.__subclasses[cls.__name__.lower()] = cls
 
     @classmethod
-    def get_class(cls: type[Processor], class_name: str) -> type[Processor] | None:
+    def get_class(
+        cls: type[BaseDataProcessor], class_name: str
+    ) -> type[BaseDataProcessor] | None:
```
JGSweets, line 57: Yup, makes sense that it isn't `type[Processor]`. While generics make sense at the top of the class, this output isn't related to the cls, but to the str.

JGSweets:
@taylorfturner do we update for this to go into `dev`?

taylorfturner:
> @taylorfturner do we update for this to go into `dev`?

Definitely

taylorfturner:
Because we are changing the base to `dev` here, you may need to re-do this depending on where you branched from -- ideally you branch from `dev` for this and then merge back to dev... or you can cherrypick commits from here onto a new branch off of `dev` @junholee6a -- thanks for all your diligent work here. We've been busy with a lot of PRs lately, but we haven't forgot about you!

junholee6a:
> Because we are changing the base to dev here, you may need to re-do this depending on where you branched from -- ideally you branch from dev for this and then merge back to dev... or you can cherrypick commits from here onto a new branch off of dev @junholee6a -- thanks for all your diligent work here. We've been busy with a lot of PRs lately, but we haven't forgot about these!

I'm re-making my PRs based on `dev`, so I'll close this one. This is a fulfilling project to work on due to all the feedback I get, and I'm not in a rush to have things merged so don't worry! Thanks :)

PR #899, opened 2023-06-22
BaseProfiler Encode / Decode 

taylorfturner:


JGSweets (changes requested):
Blocking while waiting on tests

dataprofiler/tests/profilers/utils.py
```diff
@@ -206,4 +210,7 @@ def assert_profiles_equal(actual, expected):
         elif isinstance(actual_value, np.ndarray):
             np.testing.assert_array_equal(actual_value, expected_value)
         else:
-            assert actual_value == expected_value
+            try:
+                assert actual_value == expected_value
+            except:
+                pass
```
taylorfturner, line 216: failing here due to not deserialized expected_value...

ksneab7: adding to this comment: This is not expected to have an exception in final form

taylorfturner: correct and no try/except

dataprofiler/profilers/profile_builder.py
```diff
@@ -1307,6 +1362,19 @@ def report(self, report_options: dict = None) -> dict:
         report["data_stats"] = self._profile.report(remove_disabled_flag)
         return _prepare_report(report, output_format, omit_keys)
 
+    @classmethod
```
JGSweets, line 1365: if we are just calling super, no need to include, super contains the work.

However, nothing lower level for unstruct is implemented so better to raise a not implemented error.

dataprofiler/profilers/profile_builder.py
```diff
@@ -1879,6 +1947,32 @@ def report(self, report_options: dict = None) -> dict:
...
+        _profile = data.pop("_profile")
+
+        for idx, idx_profile in enumerate(_profile):
+            _profile[idx] = load_structured_col_profiler(idx_profile)
+
+        profile = super().load_from_dict(data)
+
+        setattr(profile, "chi2_matrix", np.array(chi2_matrix))
+        setattr(profile, "correlation_matrix", np.array(correlation_matrix))
+        setattr(profile, "_profile", _profile)
```
JGSweets, line 1972: If we are just setting it without alteration, do we need to pop it?

JGSweets: Or are we going to loop through here loading profiles

dataprofiler/profilers/profile_builder.py
```diff
@@ -834,6 +859,29 @@ def report(self, report_options: dict = None) -> dict:
...
+        :return: Profiler with attributes populated.
+        :rtype: BaseCompiler
+        """
+        profiler = cls(None)
+
+        for attr, value in data.items():
+            if "times" in attr:
+                setattr(profiler, "times", defaultdict(float, value))
+            if "_profiles" in attr:
+                value = load_compiler(value)
```
JGSweets, line 879: This is diff depending on struct or unstruct, right?
struct would be the `structuredcolprofiler` I believe.

maybe we can update `_profiles` before entering here and allow it to just be set.

dataprofiler/profilers/profile_builder.py
```diff
@@ -834,6 +859,29 @@ def report(self, report_options: dict = None) -> dict:
...
+        """
+        Parse attribute from json dictionary into self.
+
+        :param data: dictionary with attributes and values.
+        :type data: dict[string, Any]
+
+        :return: Profiler with attributes populated.
+        :rtype: BaseCompiler
+        """
+        profiler = cls(None)
```
JGSweets, line 873: is `None` correct? just want to validate.

JGSweets: In here might be where we need to peak into `data` and pull out labelers to set options to push down with the rest of the code when we load.

dataprofiler/profilers/profile_builder.py
```diff
@@ -834,6 +859,29 @@ def report(self, report_options: dict = None) -> dict:
         """
         raise NotImplementedError()
 
+    @classmethod
+    def load_from_dict(cls, data) -> BaseProfiler:
```
JGSweets, line 863: I think we also need the `options` (config) as an input here.

dataprofiler/profilers/profile_builder.py
```diff
@@ -378,6 +384,25 @@ def report(self, remove_disabled_flag: bool = False) -> OrderedDict:
 
         return report
 
+    @classmethod
+    def load_from_dict(cls, data) -> StructuredColProfiler:
```
JGSweets, line 388: I think we are missing the options to pass in here.

dataprofiler/profilers/profile_builder.py
```diff
@@ -378,6 +384,25 @@ def report(self, remove_disabled_flag: bool = False) -> OrderedDict:
...
+        :type data: dict[string, Any]
+
+        :return: Profiler with attributes populated.
+        :rtype: StructuredColProfiler
+        """
+        profile = cls()
+        for attr, value in data.items():
+            if attr == "profiles":
+                for profile_key, profile_value in value.items():
+                    value[profile_key] = load_compiler(profile_value)
```
JGSweets, line 402: we would need to pass options into this as well.

dataprofiler/profilers/profile_builder.py
JGSweets: `None` actually makes sense now that I think about it bc it avoids loading everything, doesn't it?

JGSweets:
My biggest concern is this PR is likely to be massive since it is touching many things. Also, breaking this into components will allow us to parallelize this work.

I might suggest breaking this out into:
1. StructuredColProfiler Encode.
2. StructuredColProfiler decode.
3. BaseProfiler Encode.
4.StructuredProfiler Encode \ decode.
5. UnStructuredProfiler Encode \ decode.
    - raise NotImplementedError()

Issue #855, opened 2023-06-04
Refactor: pkg_resources will be deprecated, instead use importlib.resources


JGSweets:
There are currently, many instances where we use `pkg_resources`, but it will be deprecated. Instead, let's use `importlib` as suggested.

Remove it also from `requirements.txt`.

This exists in many labeler files and tests. Replace all usages.

Example from `dataprofiler/labelers/base_data_labeler.py`
```python
# remove this
import pkg_resources

# instead use this (note that this is built-in
import importlib
```

```python
# replace this
default_labeler_dir = pkg_resources.resource_filename("resources", "labelers")

# with this
default_labeler_dir = importlib.resources.files("resources").joinpath("labelers")
```

drahc1R:
https://github.com/capitalone/DataProfiler/pull/900

PR #900, opened 2023-06-22
refactored packages and fixed pesky mypy errors

drahc1R:


drahc1R:
pre-commits all passed locally...

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -9,15 +10,14 @@
...
-import pkg_resources
 
 from dataprofiler._typing import DataArray
 
 from .. import data_readers
 from . import data_processing
 from .base_model import BaseModel
 
-default_labeler_dir = pkg_resources.resource_filename("resources", "labelers")
+default_labeler_dir = str(importlib.resources.files("resources")) + "/labelers"
```
JGSweets, line 20: does this need a future for 3.8 or a conditional on py version?

Similar issue: https://github.com/mu-editor/pup/issues/35

JGSweets: Instead of `str` around + "labelers",

i suggest following the method described in: https://github.com/capitalone/DataProfiler/issues/855

Was there an issue with using the `joinpath`? << should avoid any pathing join issues as well.

drahc1R: thank you for following up! 
The suggested `default_labeler_dir = importlib.resources.files("resources").joinpath("labelers")` 
causes this error to appear: `dataprofiler/labelers/data_processing.py:148: error: No overload variant of "join" matches argument types "Traversable", "str"`

When I implemented this: `default_labeler_dir = str(importlib.resources.files("resources")) + "/labelers"`
All `mypy` errors disappeared on my local machine. This is probably due to my Python version being greater than 3.8, or would it be something else?

taylorfturner:
changing base to `dev`

dataprofiler/labelers/base_data_labeler.py
```diff
@@ -17,7 +18,11 @@
 from . import data_processing
 from .base_model import BaseModel
 
-default_labeler_dir = pkg_resources.resource_filename("resources", "labelers")
+if sys.version_info >= (3, 9):
+    default_labeler_dir = str(importlib.resources.files("resources")) + "/labelers"
```
JGSweets, line 22: I saw the prev comment, but does:
`default_labeler_dir = importlib.resources.files("resources").joinpath("labelers")` 
Not work here now with the new changes?

JGSweets: I'm concerned about the brittleness of ` + "/labelers"`

JGSweets: This leads me to believe it should work in 3.9:
https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable

if not, we could req the version for `joinpath` to be 3.10 and higher

drahc1R: Using `default_labeler_dir = importlib.resources.files("resources").joinpath("labelers")`
returns this error: `No overload variant of "join" matches argument types "Traversable", "str"`

The Python version I am using locally is 3.10

drahc1R: This is a mypy error, so I used this: `str(importlib.resources.files("resources")) + "/labelers"` in order to avoid triggering the mypy error of: `No overload variant of "join" matches argument types "Traversable", "str"`.

I will try to come up with a different solution

PR #912, opened 2023-06-27
Remove process() from BaseDataProcessor

junholee6a:

`BaseDataProcessor.process()` is an abstract method that is overridden by subclasses. The subclass methods `BaseDataPreprocessor.process()` and `BaseDataPostprocessor.process()` have different signatures from the superclass method so they currently violate the Liskov substitution principle. This is fixed by removing the `process()` method from `BaseDataProcessor`.

The subclasses of `BaseDataPreprocessor` have similar method signatures for `process()` so there is no need to remove `BaseDataPreprocessor.process()` (its signature was modified to be inclusive of all subclasses). Same with `BaseDataPostprocessor.process()`.

### DIFF
--- a/dataprofiler/labelers/data_processing.py
+++ b/dataprofiler/labelers/data_processing.py
@@ -128,11 +128,6 @@ class BaseDataProcessor(metaclass=abc.ABCMeta):
         for param in kwargs:
             self._parameters[param] = kwargs[param]
 
-    @abc.abstractmethod
-    def process(self, *args: Any) -> Any:
-        """Process data."""
-        raise NotImplementedError()
-
     @classmethod
     def load_from_disk(cls: type[Processor], dirpath: str) -> Processor:
         """Load data processor from a given path on disk."""
@@ -169,13 +164,15 @@ class BaseDataPreprocessor(BaseDataProcessor):
         super().__init__(**parameters)
 
     @abc.abstractmethod
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         labels: np.ndarray | None = None,
         label_mapping: dict[str, int] | None = None,
         batch_size: int = 32,
-    ) -> Generator[tuple[np.ndarray, np.ndarray] | np.ndarray, None, None]:
+    ) -> Generator[tuple[np.ndarray, np.ndarray] | np.ndarray, None, None] | tuple[
+        np.ndarray, np.ndarray
+    ] | np.ndarray:
         """Preprocess data."""
         raise NotImplementedError()
 
@@ -191,7 +188,7 @@ class BaseDataPostprocessor(BaseDataProcessor):
         super().__init__(**parameters)
 
     @abc.abstractmethod
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         results: dict,
@@ -240,7 +237,7 @@ class DirectPassPreprocessor(BaseDataPreprocessor, metaclass=AutoSubRegistration
         )
         print(help_str)
 
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         labels: np.ndarray | None = None,
@@ -668,7 +665,7 @@ class CharPreprocessor(BaseDataPreprocessor, metaclass=AutoSubRegistrationMeta):
         if batch_data["samples"]:
             yield batch_data
 
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         labels: np.ndarray | None = None,
@@ -836,7 +833,7 @@ class CharEncodedPreprocessor(CharPreprocessor, metaclass=AutoSubRegistrationMet
         if errors:
             raise ValueError("\n".join(errors))
 
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         labels: np.ndarray | None = None,
@@ -1269,7 +1266,7 @@ class CharPostprocessor(BaseDataPostprocessor, metaclass=AutoSubRegistrationMeta
 
         return results
 
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         results: dict,
@@ -1439,7 +1436,7 @@ class StructCharPreprocessor(CharPreprocessor, metaclass=AutoSubRegistrationMeta
 
         return text, entities
 
-    def process(  # type: ignore
+    def process(
         self,
         data: np.ndarray,
         labels: np.ndarray | None = None,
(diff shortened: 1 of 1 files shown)


junholee6a:
This PR removes a method from `BaseDataProcessor`. Do I need to re-generate docs on my local computer and push its changes? Or does the documentation site generate itself automatically?

junholee6a:
Tests failed in CI, but pass locally for me. Will do some debugging

JGSweets:
Have not checked the test because I'm out, but I'm guessing it's because of a change to scipy recently. It needs to be <1.11.0 until we solve

dataprofiler/labelers/data_processing.py
```diff
@@ -128,11 +128,6 @@ def set_params(self, **kwargs: Any) -> None:
         for param in kwargs:
             self._parameters[param] = kwargs[param]
 
-    @abc.abstractmethod
-    def process(self, *args: Any) -> Any:
-        """Process data."""
-        raise NotImplementedError()
-
```
JGSweets, line 135: I see the intent here, but I'm concerned that `BaseDataProcessor` No longer requires a `process`.

taylorfturner:
> Because we are changing the base to `dev` here, you may need to re-do this depending on where you branched from -- ideally you branch from `dev` for this and then merge back to dev... or you can cherrypick commits from here onto a new branch off of `dev` @junholee6a -- thanks for all your diligent work here. We've been busy with a lot of PRs lately, but we haven't forgot about these! 

Same as other PR

dataprofiler/labelers/data_processing.py
JGSweets: @junholee6a , thoughts here?

junholee6a: Not sure about generics and overloading, but there is this alternative solution: We could have an arrangement where `BaseDataProcessor.process()` has no arguments, and all subclasses have a `process()` with only optional arguments. LSP allows subclass methods to add optional arguments but not required arguments. So we would need to make the `data` argument of `BaseDataPreprocessor.process()` and all arguments of `BaseDataPostprocessor.process()` optional.

junholee6a: I'll look further into whether this can be resolved with something like [TypeVarTuple](https://docs.python.org/3/library/typing.html#typing.TypeVarTuple)

junholee6a: Since `BaseDataPreprocessor.process()` has one required argument and `BaseDataPostprocessor.process()` has three required arguments, to use generics here we'd need to use `TypeVarTuple`, which can represent an arbitrary number of types. But `TypeVarTuple` is a new feature in Python 3.11, so I wouldn't recommend it for backwards compatibility.

Looked into overloading and I haven't found docs on how it interacts with inheritance, but I don't think it would let us work around LSP.

Summary of the solutions I could currently think of:

1. Get rid of `BaseDataProcessor.process()` (current PR):
```python
class BaseDataProcessor:
    # No process() method

class BaseDataPreprocessor(BaseDataProcessor):
    def process(
        self,
        data: np.ndarray,
        labels: np.ndarray | None = None,
        label_mapping: dict[str, int] | None = None,
        batch_size: int = 32,
    ) -> Generator[tuple[np.ndarray, np.ndarray] | np.ndarray, None, None]:
        pass

class BaseDataPostprocessor(BaseDataProcessor):
    def process(
        self,
        data: np.ndarray,
        results: dict,
        label_mapping: dict[str, int],
    ) -> dict:
        pass
``` 

2. Use only optional arguments in subclass implementations of `process()`:
```python
class BaseDataProcessor:
    def process() -> Any:
        raise NotImplementedError()

class BaseDataPreprocessor(BaseDataProcessor):
    def process(
        self,
        data: np.ndarray | None = None,
        labels: np.ndarray | None = None,
        label_mapping: dict[str, int] | None = None,
        batch_size: int = 32,
    ) -> Generator[tuple[np.ndarray, np.ndarray] | np.ndarray, None, None]:
        if data is None:
            raise ValueError('data is a required argument')

class BaseDataPostprocessor(BaseDataProcessor):
    def process(
        self,
        data: np.ndarray | None = None,
        results: dict | None = None,
        label_mapping: dict[str, int] | None = None,
    ) -> dict:
        if data is None or results is None or label_mapping is None:
            raise ValueError('data, results, and label_mapping are required arguments')
``` 

Any thoughts on what would be best? I'll remake this PR on top of `dev` once this comment thread is resolved.

JGSweets: Could it be bc we aren't allowing for the `kwargs`?
```python
class BaseDataProcessor:
    def process(*args: Any, **kwargs: Any) -> Any:
        raise NotImplementedError()

```


Actually, just found this, but haven't searched well enough yet.
https://github.com/python/mypy/issues/5876

JGSweets: This comment is also interesting: https://github.com/python/mypy/issues/5876#issuecomment-706396563

junholee6a: > Could it be bc we aren't allowing for the kwargs?

I tried adding kwargs to `BaseDataProcessor.process()` like you wrote and ran mypy, and it definitely works. Looking at the issue you mentioned, it seems like this is an intentional workaround.

https://github.com/python/mypy/pull/11203 is the PR in which they resolved the issue https://github.com/python/mypy/issues/5876. This PR made it so that a method with signature `*args: Any, **kwargs: Any` is compatible with any subclass method. [They also said that this is an LSP violation, but still went ahead with merging this because it was such a popular request](https://github.com/python/mypy/pull/11203#discussion_r716402902).

It's a violation in this case because LSP asserts that wherever a superclass method is valid, the subclass methods overriding it must also be valid. For example, a call to `my_base_data_processor.process(1, 2, 3, 4, 5, 6)` is valid due to `*args: Any`, while a call to subclass method `my_base_data_preprocessor.process(1, 2, 3, 4, 5, 6)` is not valid, since `BaseDataPreprocessor.process()` has up to four arguments.

I think you're right that this is a good solution though. The feature was requested by a lot of people in mypy because it's practical. Should we proceed with this?

JGSweets: Great analysis! For the purpose of moving forward, let's do this. It feels awkward to break LSP, but this will suffice.

taylorfturner:
@junholee6a will need to update this branch with the new `dev` too. Thx!

junholee6a:
This PR is being replaced with https://github.com/capitalone/DataProfiler/pull/954

dataprofiler/labelers/data_processing.py
junholee6a: @JGSweets Thank you for your help on this fix, this is much cleaner than what I thought was possible :)

PR #923, opened 2023-06-27
BaseProfiler save() for json

micdavis:
* Implemented `_json_save_helper()`
* implemented `save()` to support json
* wrote initial tests

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -108,7 +108,7 @@ repos:
         additional_dependencies: ['h5py', 'wheel', 'future', 'numpy', 'pandas',
         'python-dateutil', 'pytz', 'pyarrow', 'chardet', 'fastavro',
         'python-snappy', 'charset-normalizer', 'psutil', 'scipy', 'requests',
-        'networkx','typing-extensions']
+        'networkx','typing-extensions', 'HLL']
   # Pyupgrade - standardize and modernize Python syntax for newer versions of the language
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -108,7 +108,7 @@ repos:
         additional_dependencies: ['h5py', 'wheel', 'future', 'numpy', 'pandas',
         'python-dateutil', 'pytz', 'pyarrow', 'chardet', 'fastavro',
         'python-snappy', 'charset-normalizer', 'psutil', 'scipy', 'requests',
-        'networkx','typing-extensions', 'HLL']
+        'networkx','typing-extensions']
   # Pyupgrade - standardize and modernize Python syntax for newer versions of the language
   - repo: https://github.com/asottile/pyupgrade
     rev: v3.3.0
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -1548,9 +1548,7 @@ class UnstructuredProfiler(BaseProfiler):
         :type filepath: String
         :return: None
         """
-        if filepath is None or (
-            filepath.split(".")[-1] != "json" and not isinstance(filepath, str)
-        ):
+        if filepath is None or filepath.split(".")[-1] != "json":
             # Create dictionary for all metadata, options, and profile
             data_dict = {
                 "total_samples": self.total_samples,
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import copy
 import json
 import logging
+import pathlib
 import pickle
 import random
 import re
@@ -1116,7 +1117,7 @@ class BaseProfiler:
         # Restore all data labelers
         self._restore_data_labelers(data_labelers)
 
-    def _json_save_helper(self, filepath: str) -> None:
+    def _json_save_helper(self, filepath: str | None) -> None:
         """
         Save profiler to disk.
 
@@ -1125,16 +1126,21 @@ class BaseProfiler:
         :return: None
         """
         if filepath is None:
-            filepath = ""
+            filepath = "profile-{}.pkl".format(
+                datetime.now().strftime("%d-%b-%Y-%H:%M:%S.%f")
+            )
+
         with open(filepath, "w") as f:
             json.dump(self, f, cls=ProfileEncoder)
 
-    def save(self, filepath: str = None) -> None:
+    def save(self, filepath: str = None, save_method: str = None) -> None:
         """
         Save profiler to disk.
 
         :param filepath: Path of file to save to
         :type filepath: String
+        :param save_method: the desired saving method ("pickle" | "json")
+        :type save_method: String
         :return: None
         """
         raise NotImplementedError()
@@ -1540,16 +1546,19 @@ class UnstructuredProfiler(BaseProfiler):
         else:
             self._profile.update_profile(data, pool=pool)
 
-    def save(self, filepath: str = None) -> None:
+    def save(self, filepath: str = None, save_method: str = None) -> None:
         """
         Save profiler to disk.
 
         :param filepath: Path of file to save to
         :type filepath: String
+        :param save_method: the desired saving method ("pickle" | "json")
+        :type save_method: String
         :return: None
         """
-        if filepath is None or filepath.split(".")[-1] != "json":
-            # Create dictionary for all metadata, options, and profile
+        if (save_method is None or save_method != "json") and (
+            filepath is None or pathlib.Path(filepath).suffix != "json"
+        ):
             data_dict = {
                 "total_samples": self.total_samples,
                 "sample": self.sample,
@@ -2838,16 +2847,19 @@ class StructuredProfiler(BaseProfiler):
         if self.options.null_replication_metrics.is_enabled:
             self._update_null_replication_metrics(clean_sampled_dict)
 
-    def save(self, filepath: str = None) -> None:
+    def save(self, filepath: str = None, save_method: str = None) -> None:
         """
         Save profiler to disk.
 
         :param filepath: Path of file to save to
         :type filepath: String
+        :param save_method: the desired saving method ("pickle" | "json")
+        :type save_method: String
         :return: None
         """
-        # Create dictionary for all metadata, options, and profile
-        if filepath is None or filepath.split(".")[-1] != "json":
+        if (save_method is None or save_method != "json") and (
+            filepath is None or pathlib.Path(filepath).suffix != "json"
+        ):
             data_dict = {
                 "total_samples": self.total_samples,
                 "encoding": self.encoding,
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import copy
+import json
 import logging
 import pickle
 import random
@@ -32,6 +33,7 @@ from .column_profile_compilers import (
 from .graph_profiler import GraphProfiler
 from .helpers.report_helpers import _prepare_report, calculate_quantiles
 from .json_decoder import load_compiler, load_option, load_structured_col_profiler
+from .json_encoder import ProfileEncoder
 from .profiler_options import (
     BaseOption,
     ProfilerOptions,
@@ -1122,7 +1124,10 @@ class BaseProfiler:
         :type filepath: String
         :return: None
         """
-        raise NotImplementedError()
+        if filepath is None:
+            filepath = ""
+        with open(filepath, "w") as f:
+            json.dump(self, f, cls=ProfileEncoder)
 
     @classmethod
     def load(cls, filepath: str) -> BaseProfiler:
@@ -1533,21 +1538,26 @@ class UnstructuredProfiler(BaseProfiler):
         :type filepath: String
         :return: None
         """
-        # Create dictionary for all metadata, options, and profile
-        data_dict = {
-            "total_samples": self.total_samples,
-            "sample": self.sample,
-            "encoding": self.encoding,
-            "file_type": self.file_type,
-            "_samples_per_update": self._samples_per_update,
-            "_min_true_samples": self._min_true_samples,
-            "_empty_line_count": self._empty_line_count,
-            "memory_size": self.memory_size,
-            "options": self.options,
-            "_profile": self.profile,
-            "times": self.times,
-        }
-        self._save_helper(filepath, data_dict)
+        if filepath is None or (
+            filepath.split(".")[-1] != "json" and not isinstance(filepath, str)
+        ):
+            # Create dictionary for all metadata, options, and profile
+            data_dict = {
+                "total_samples": self.total_samples,
+                "sample": self.sample,
+                "encoding": self.encoding,
+                "file_type": self.file_type,
+                "_samples_per_update": self._samples_per_update,
+                "_min_true_samples": self._min_true_samples,
+                "_empty_line_count": self._empty_line_count,
+                "memory_size": self.memory_size,
+                "options": self.options,
+                "_profile": self.profile,
+                "times": self.times,
+            }
+            self._save_helper(filepath, data_dict)
+        else:
+            super().save(filepath)
 
 
 class StructuredProfiler(BaseProfiler):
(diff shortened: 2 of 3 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1531,21 +1546,24 @@ def save(self, filepath: str = None) -> None:
...
-            "_samples_per_update": self._samples_per_update,
-            "_min_true_samples": self._min_true_samples,
-            "_empty_line_count": self._empty_line_count,
-            "memory_size": self.memory_size,
-            "options": self.options,
-            "_profile": self.profile,
-            "times": self.times,
-        }
-        self._save_helper(filepath, data_dict)
+        if filepath is None or filepath.split(".")[-1] != "json":
```
micdavis, line 1549: basically if the filepath doesn't end in "json" just default to pkl

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3834,6 +3898,19 @@ def test_save_and_load_no_labeler(self):
         save_profile.update_profile(pd.DataFrame(["test", "test2"]))
         load_profile.update_profile(pd.DataFrame(["test", "test2"]))
 
+    def test_save_json_file(self):
```
micdavis, line 3901: testing `NotImplementedError()`

dataprofiler/profilers/profile_builder.py
micdavis: or if the filepath doesn't exist

.pre-commit-config.yaml
```diff
@@ -108,7 +108,7 @@ repos:
         additional_dependencies: ['h5py', 'wheel', 'future', 'numpy', 'pandas',
         'python-dateutil', 'pytz', 'pyarrow', 'chardet', 'fastavro',
         'python-snappy', 'charset-normalizer', 'psutil', 'scipy', 'requests',
-        'networkx','typing-extensions']
+        'networkx','typing-extensions', 'HLL']
```
JGSweets, line 111: Was this a needed add?

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,63 @@ def test_save_and_load_no_labeler(self):
         save_profile.update_profile(pd.DataFrame({"a": [4, 5]}))
         load_profile.update_profile(pd.DataFrame({"a": [4, 5]}))
 
+    def test_save_json_file(self):
```
micdavis, line 1548: testing save on `StructuredProfiler()` with csv file

dataprofiler/profilers/profile_builder.py
```diff
@@ -1083,7 +1085,7 @@ def _restore_data_labelers(self, data_labeler: BaseDataLabeler = None) -> None:
                 data_labeler_profile = profiler._profiles["data_labeler"]
                 data_labeler_profile.data_labeler = data_labeler
 
-    def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
+    def _pkl_save_helper(self, filepath: str | None, data_dict: dict) -> None:
```
JGSweets, line 1088: Nice

dataprofiler/profilers/profile_builder.py
taylorfturner, line 1549: wonder if this should be a bit more clear... the `filepath` connotates to me just the folder path... not necessarily the name of the file ... but the filename is vital here with its extension becuause the core logic to testing wether to go with `_pkl_save_helper` or `_json_save_helper` is keying off the extension of the file

dataprofiler/profilers/profile_builder.py
```diff
@@ -1112,6 +1114,19 @@ def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
...
+        """
+        Save profiler to disk.
+
+        :param filepath: Path of file to save to
+        :type filepath: String
+        :return: None
+        """
+        if filepath is None:
+            filepath = ""
+        with open(filepath, "w") as f:
```
taylorfturner, line 1127: don't we need some kind of file name at least for this because `with open("", "w") as f:`, even if it did succeded wouldn't be ideal for finding the thing it wrote

dataprofiler/profilers/profile_builder.py
```diff
@@ -1112,6 +1114,19 @@ def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
...
+    def _json_save_helper(self, filepath: str) -> None:
+        """
+        Save profiler to disk.
+
+        :param filepath: Path of file to save to
+        :type filepath: String
+        :return: None
+        """
+        if filepath is None:
+            filepath = ""
```
taylorfturner, line 1126:
```suggestion
            if filepath is None:
              filepath = "profile-{}.json".format(
                  datetime.now().strftime("%d-%b-%Y-%H:%M:%S.%f")
              )
```

dataprofiler/profilers/profile_builder.py
taylorfturner: see suggestion above

dataprofiler/profilers/profile_builder.py
taylorfturner: we allow for `None`, yes; but the code in `_pkl_save_helper` auto creates a `filepath` (really it is create a filename) automatically with a timestamp for uniquness

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,63 @@ def test_save_and_load_no_labeler(self):
...
+        self.maxDiff = None
```
taylorfturner, line 1549: remove

taylorfturner, line 1549:
```suggestion
```

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -3834,6 +3898,19 @@ def test_save_and_load_no_labeler(self):
...
+
+        # Create Data and StructuredProfiler objects
+        data = dp.Data(os.path.join(data_folder, test_file))
+        save_profile = UnstructuredProfiler(data)
+
+        with self.assertRaisesRegex(
+            NotImplementedError, "UnstructuredProfiler serialization not supported."
+        ):
+            save_profile.save("out.json")
+
```
taylorfturner, line 3913: nice

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,63 @@ def test_save_and_load_no_labeler(self):
...
+        options = ProfilerOptions()
+        options.set({"correlation.is_enabled": True})
+        with test_utils.mock_timeit():
+            save_profile = dp.StructuredProfiler(data)
+
+        # Save and Load profile with Mock IO
+        with mock.patch("builtins.open") as m:
+            mock_file = setup_save_mock_string_open(m)
+            save_profile.save("mock.json")
+            mock_file.seek(0)
```
taylorfturner, line 1564: two thoughts (not hard and fast here), but we should validate the return_value from `json.dumps()` here... I say that because the error in the code above would have been caught. Maybe there is a return_value from `json.dumps()` that we could validate the `filepath` variable that is being actually written... we'd have to mock I reckon

taylorfturner (changes requested):
I think there is a 🐛 (where the test shouldn't pass / we are missing assert / validation) and a couple questions / thoughts for improvement

dataprofiler/profilers/profile_builder.py
taylorfturner: this is sneaky because it probably silently runs but doesn't do what we actually want ... 💀

dataprofiler/profilers/profile_builder.py
taylorfturner: this is sneaky because it probably silently runs but doesn't do what we actually want ... 💀

dataprofiler/tests/profilers/test_profile_builder.py
taylorfturner: or just add a test validating passing `None` to the `filepath` parameter... and validate it outputs the autogenerate filepath (would have to mock the time generation)

dataprofiler/profilers/profile_builder.py
```diff
@@ -1112,6 +1114,19 @@ def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
         # Restore all data labelers
         self._restore_data_labelers(data_labelers)
 
+    def _json_save_helper(self, filepath: str) -> None:
```
JGSweets, line 1117: if we decide to keep this format and we allow `filepath` to be `None`, this typing needs to have a `| None`

dataprofiler/profilers/profile_builder.py
JGSweets: good callout

dataprofiler/profilers/profile_builder.py
JGSweets: We might want to use python's pathlib as it auto will get for us properly and handle strange cases like:
```
import pathlib


print(pathlib.Path("hello/foo.bar.tar.gz").suffix)
# output: ".gz"

 print(pathlib.Path("hello/foo.bar.tar.gz/a").suffix)
# output: ""
```
notably the last case which has `.` but not as the ext

JGSweets: I think the logic we might is more like:
1. use `type` first to determine how to save
    a. basically can force save with any extension to any type 
2. if `type=None` look to extension to save
3. if extension is not clear, choose pickle
4. if filepath not specified, choose pickle

dataprofiler/profilers/profile_builder.py
```diff
@@ -1531,21 +1546,24 @@ def save(self, filepath: str = None) -> None:
         :type filepath: String
         :return: None
         """
-        # Create dictionary for all metadata, options, and profile
```
JGSweets, line 1534: above in `save` we need to add the `type: str | None` with default `None` where we can choose '"pickle "| "json"`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1112,6 +1114,19 @@ def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
...
+        :param filepath: Path of file to save to
+        :type filepath: String
+        :return: None
+        """
+        if filepath is None:
+            filepath = ""
+        with open(filepath, "w") as f:
+            json.dump(self, f, cls=ProfileEncoder)
+
     def save(self, filepath: str = None) -> None:
```
JGSweets, line 1130: Here we need to add the type: str | None with default None where we can choose '"pickle "| "json"`

dataprofiler/profilers/profile_builder.py
JGSweets: good call, should be same as pkl technique.

dataprofiler/profilers/profile_builder.py
taylorfturner: > use type first to determine how to save

agree

dataprofiler/profilers/profile_builder.py
```diff
@@ -1112,14 +1115,32 @@ def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
...
+    def _json_save_helper(self, filepath: str | None) -> None:
         """
         Save profiler to disk.
 
         :param filepath: Path of file to save to
         :type filepath: String
         :return: None
         """
+        if filepath is None:
+            filepath = "profile-{}.pkl".format(
```
JGSweets, line 1127: Fix extension here

JGSweets (changes requested):

dataprofiler/profilers/profile_builder.py
taylorfturner: good catch

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,113 @@ def test_save_and_load_no_labeler(self):
...
+        mock_open.assert_called_with(expected_second_path, "w")
+
+    @mock.patch(
+        "dataprofiler.profilers.data_labeler_column_profile." "DataLabelerColumn.update"
+    )
+    @mock.patch(
+        "dataprofiler.profilers.profile_builder.DataLabeler",
+        spec=BaseDataLabeler,
+    )
+    def test_save_value_error(self, *mocks):
```
taylorfturner, line 1635:
```suggestion
    def test_save_structured_profiler_value_error(self, *mocks):
```

taylorfturner (changes requested):

dataprofiler/tests/profilers/test_profile_builder.py
micdavis: there is a testing class that specifies this, resolving

micdavis: and all comments below will be resolve for same reason

.pre-commit-config.yaml
taylorfturner: removed

dataprofiler/profilers/profile_builder.py
micdavis: done

dataprofiler/profilers/profile_builder.py
```diff
@@ -1114,14 +1116,32 @@ def _save_helper(self, filepath: str | None, data_dict: dict) -> None:
...
+        with open(filepath, "w") as f:
+            json.dump(self, f, cls=ProfileEncoder)
+
+    def save(self, filepath: str = None, save_method: str = "pickle") -> None:
+        """
+        Save profiler to disk.
+
+        :param filepath: Path of file to save to
+        :type filepath: String
+        :param save_method: the desired saving method (must be "pickle" or "json")
```
taylorfturner, line 1141:
```suggestion
        :param save_method: The desired saving method (must be "pickle" or "json")
```

taylorfturner (approved):

tyfarnan (approved):

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,130 @@ def test_save_and_load_no_labeler(self):
...
+        )
+
+        # Create Data and StructuredProfiler objects
+        with test_utils.mock_timeit():
+            save_profile = dp.StructuredProfiler(df_structured, options=profile_options)
+
+        # Save and Load profile with Mock IO
+        with mock.patch("builtins.open") as mock_open, mock.patch(
+            "dataprofiler.profilers.profile_builder." "datetime"
+        ) as time_out:
```
JGSweets, line 1583: really should clarify any variable that is a mock as a mock.

e.g. `mock_time_now`

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,130 @@ def test_save_and_load_no_labeler(self):
...
+            }
+        )
+
+        # Create Data and StructuredProfiler objects
+        with test_utils.mock_timeit():
+            save_profile = dp.StructuredProfiler(df_structured, options=profile_options)
+
+        # Save and Load profile with Mock IO
+        with mock.patch("builtins.open") as mock_open, mock.patch(
+            "dataprofiler.profilers.profile_builder." "datetime"
```
JGSweets, line 1582: `"dataprofiler.profilers.profile_builder.datetime"

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,130 @@ def test_save_and_load_no_labeler(self):
...
+            time_out.now().strftime.return_value = "now"
+            setup_save_mock_string_open(mock_open)
+            save_profile.save(save_method="json")
+
+        expected_second_path = "profile-now.json"
+
+        mock_open.assert_called_with(expected_second_path, "w")
+
+    @mock.patch(
+        "dataprofiler.profilers.data_labeler_column_profile." "DataLabelerColumn.update"
```
JGSweets, line 1638: "dataprofiler.profilers.data_labeler_column_profile.DataLabelerColumn.update"

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1538,6 +1545,130 @@ def test_save_and_load_no_labeler(self):
...
+        }
+
+        actual_data = json.loads(mock_file.read())
+
+        mock_open.assert_called_with(expected_first_path, "w")
+        self.assertDictEqual(expected_data, actual_data)
+
+        # do a second call without a specified file path
+        with mock.patch("builtins.open") as mock_open, mock.patch(
+            "dataprofiler.profilers.profile_builder." "datetime"
```
JGSweets, line 1627: "dataprofiler.profilers.profile_builder.datetime"

PR #925, opened 2023-06-28
Profile Builder `load()` serialization

tyfarnan:
- expands BaseProfiler load function to support both pickle and json by checking the file extension first. 
- Unstructuredprofiler raised a NotImplementedError() on it's load()

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -33,7 +33,12 @@ from .column_profile_compilers import (
 )
 from .graph_profiler import GraphProfiler
 from .helpers.report_helpers import _prepare_report, calculate_quantiles
-from .json_decoder import load_compiler, load_option, load_structured_col_profiler
+from .json_decoder import (
+    load_compiler,
+    load_option,
+    load_profiler,
+    load_structured_col_profiler,
+)
 from .json_encoder import ProfileEncoder
 from .profiler_options import (
     BaseOption,
@@ -1157,7 +1162,7 @@ class BaseProfiler:
         raise NotImplementedError()
 
     @classmethod
-    def load(cls, filepath: str) -> BaseProfiler:
+    def load(cls, filepath: str, load_method: str | None = "pickle") -> BaseProfiler:
         """
         Load profiler from disk.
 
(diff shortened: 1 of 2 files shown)


dataprofiler/profilers/profile_builder.py
```diff
@@ -1078,7 +1079,15 @@ def load(cls, filepath: str) -> BaseProfiler:
         """
         # Load profile from disk
         with open(filepath, "rb") as infile:
-            data: dict = pickle.load(infile)
+
+            if filepath.endswith(".json"):
+                data: dict = json.load(infile)
+            elif filepath.endswith((".pickle", ".pkl")):
+                data = pickle.load(infile)
```
taylorfturner, line 1086: i love the simplicity of this, but I'm not sure based on the `save()` implementation that this will account for some of the edge cases.... 🤔

micdavis: do we wanna do `load_method` to load similarly to how we did `save_method` for `save()`

micdavis: Also the json functionality should be using `json_decoder.load_profiler()` vs `json.load()` that way we can directly convert from json file to profiler object.

tyfarnan: unsure about the edge cases for load...

tyfarnan: @micdavis I don't see a `json_decoder.load_profiler()`

dataprofiler/profilers/profile_builder.py
```diff
@@ -1157,7 +1157,15 @@ def load(cls, filepath: str) -> BaseProfiler:
         """
         # Load profile from disk
         with open(filepath, "rb") as infile:
-            data: dict = pickle.load(infile)
+
+            if filepath.endswith(".json"):
```
JGSweets, line 1161: let's also allow for a user to specify the file_type like we do with save since we give the option for someone to technically use any extension given a type.

dataprofiler/profilers/profile_builder.py
JGSweets: `json_decoder.load_profiler(data, config={})` This is intentional for the config to use `{}` bc it allows for reuse of already loaded datalabelers

JGSweets: after the `load_profiler` we can return with the output.

JGSweets: it honestly might be best to separate this func into two for clarity, but could be a future refactor

dataprofiler/profilers/profile_builder.py
taylorfturner: agree -- parallel API for `save()` and `load()` in terms of accepted params

dataprofiler/tests/profilers/test_profile_builder.py
```diff
@@ -1532,8 +1581,8 @@ def test_save_and_load_no_labeler(self):
             save_profile.save()
 
             mock_file.seek(0)
-            with mock.patch("dataprofiler.profilers.profile_builder.DataLabeler"):
-                load_profile = dp.StructuredProfiler.load("mock.pkl")
+            with mock.patch("dataprofiler.profilers.profile_builder." "DataLabeler"):
```
taylorfturner, line 1584:
```suggestion
            with mock.patch("dataprofiler.profilers.profile_builder.DataLabeler"):
```

taylorfturner: just to raise this as a potential change that is needed....

taylorfturner (approved):

micdavis (approved):

PR #929, opened 2023-06-28
Fix name mangling and typevar errors (dev branch)

junholee6a:

This PR is a duplicate of https://github.com/capitalone/DataProfiler/pull/895, except it is based on the `dev` branch instead of `main`

### DIFF
--- a/dataprofiler/labelers/base_model.py
+++ b/dataprofiler/labelers/base_model.py
@@ -32,7 +32,7 @@ class AutoSubRegistrationMeta(abc.ABCMeta):
 class BaseModel(metaclass=abc.ABCMeta):
     """For labeling data."""
 
-    _BaseModel__subclasses: dict[str, type[BaseModel]] = {}
+    __subclasses: dict[str, type[BaseModel]] = {}
     __metaclass__ = abc.ABCMeta
 
     # boolean if the label mapping requires the mapping for index 0 reserved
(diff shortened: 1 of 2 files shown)


junholee6a:
`license/cla` is stuck in "expected" status, so I'll push an empty commit

dataprofiler/labelers/base_model.py
```diff
@@ -32,7 +32,7 @@ def __new__(
 class BaseModel(metaclass=abc.ABCMeta):
     """For labeling data."""
 
-    _BaseModel__subclasses: dict[str, type[BaseModel]] = {}
+    __subclasses: dict[str, type[BaseModel]] = {}
```
ksneab7, line 35: Very cool and portable

ksneab7 (approved):

taylorfturner:
Need to revert to ensure `dev` is updated to `main` before we continue additional PRs to `dev` or `feature/*` branching off `dev`. @junholee6a -- not your fault. If you can, just rebase / cherry pick your commits here on `dev` once I update things. I'll pop another comment here once its ready to go

taylorfturner:
The idea here is that we will take current `dev` and rename it to something like `dev/release/0.10.0`. But since we are going to re-create `dev` from `main` we would loose your changes in the "new" `dev` branch. Hopefully that is helpful in adding a bit more clarity. Thanks!

taylorfturner:
Okay -- good to go on a re-open @junholee6a into `dev`. Thanks again!

junholee6a:
> Okay -- good to go on a re-open @junholee6a into dev. Thanks again!

Thanks for the heads up, I just re-made the branch and PR

PR #939, opened 2023-06-29
fix: order bug

JGSweets:
If a str in the middle of numbers forced conversion to a str, this would cause failures.

This fixes it and adds tests

Also fixes bug in profiler when loading to not load model multiple times

### DIFF
--- a/dataprofiler/profilers/order_column_profile.py
+++ b/dataprofiler/profilers/order_column_profile.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 from abc import abstractmethod
-from typing import Protocol, TypeVar
+from typing import Protocol, Type, TypeVar, cast
 
 import numpy as np
 from pandas import DataFrame, Series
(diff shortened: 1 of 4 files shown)


dataprofiler/profilers/order_column_profile.py
```diff
@@ -23,6 +23,10 @@ def __lt__(self: CT, other: CT) -> bool:
 
 CT = TypeVar("CT", bound=Comparable)
 
+# bc type in class attr causing issues, need to alias
+AliasFloatType = Type[np.float64]
```
micdavis, line 27: Nice!

dataprofiler/profilers/order_column_profile.py
```diff
@@ -23,6 +23,10 @@ def __lt__(self: CT, other: CT) -> bool:
...
+AliasStrType = Type[str]
+
```
taylorfturner, line 29: adequate

dataprofiler/profilers/profile_builder.py
```diff
@@ -898,7 +898,8 @@ def load_from_dict(cls: type[BaseProfilerT], data, config) -> BaseProfilerT:
         :return: Profiler with attributes populated.
         :rtype: BaseProfiler
         """
-        profiler = cls(None)
```
JGSweets, line 901: another bug fix in this method

taylorfturner (approved):

micdavis (approved):

Issue #905, opened 2023-06-26
Scipy Verision + Graph Issues


taylorfturner:
**General Information:**
 - OS: Mac OS
 - Python version: 3.9.13
 - Library version: DP `feature/profiler-serialization` branch as of 26 Jun 2023

**Describe the bug:**
1.11.0 causes graph profilers to start failing tests.

we will need to investigate what is failing and why https://scipy.github.io/devdocs/release/1.11.0-notes.html#statistical-distributions
causes these failures.

**To Reproduce:**
Add `scipy>=1.4.1,<1.11.0`

**Expected behavior:**
Should not fail as we update package dependency versions

**Screenshots:**
N/A

**Additional context:**
N/A

taylorfturner:
slowly getting into `dev` and `feature/*` branches since we have identified issues and resolution.

taylorfturner:
#896 updates into `feature/options`

PR #948, opened 2023-07-05
scipy errors fixed

drahc1R:
Fixed breaking tests due to SciPy 1.11.0 update
Issue: https://github.com/capitalone/DataProfiler/issues/905


dataprofiler/profilers/graph_profiler.py
```diff
@@ -401,7 +401,13 @@ def _get_continuous_distribution(
 
                 for distribution in distribution_candidates:
                     # compute fit, mle, kolmogorov-smirnov test to test fit, and pdf
-                    fit = distribution.fit(df)
+
+                    # scipy 1.11.0 updated the way they handle loc parameter in fit() for lognorm
+                    if distribution == st.lognorm:
+                        fit = distribution.fit(df, superfit=True)
```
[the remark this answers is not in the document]
drahc1R: yep

dataprofiler/profilers/graph_profiler.py
```diff
@@ -401,7 +402,15 @@ def _get_continuous_distribution(
 
                 for distribution in distribution_candidates:
                     # compute fit, mle, kolmogorov-smirnov test to test fit, and pdf
-                    fit = distribution.fit(df)
+
+                    # scipy 1.11.0 updated the way they handle
+                    # the loc parameter in fit() for lognorm
+                    scipy_version = float(scipy.__version__.replace(".", ""))
```
JGSweets, line 408: to get the version, i suggest we do something like:
`pkg_resources.get_distribution('scipy').version`

I'm not 100% convinced that we should convert to float and compare as sometimes versions can contain strs.

Any other options avail?

JGSweets: Actually, according to SO, `pkg_resources` is slow to import.
```python
# at top of file
import importlib.metadata

importlib.metadata.version('scipy')
```

drahc1R: done!

dataprofiler/profilers/graph_profiler.py
```diff
@@ -401,7 +402,17 @@ def _get_continuous_distribution(
 
                 for distribution in distribution_candidates:
                     # compute fit, mle, kolmogorov-smirnov test to test fit, and pdf
-                    fit = distribution.fit(df)
+
+                    # scipy 1.11.0 updated the way they handle
+                    # the loc parameter in fit() for lognorm
+                    scipy_version = int(
+                        importlib.metadata.version("scipy").replace(".", "")
+                    )
```
JGSweets, line 410: My concern is the int \ float conversion of the version. This would error if they did a post fix etc. We might want to consider for graph installing packaging so we can do a comparison between versions unless there's an easier way without having to install another repo.

If we need to, we can choose to install it only on the graph since that's where this is being used.

drahc1R:
misclicked

PR #951, opened 2023-07-05
Scipy bug fix

drahc1R:
https://github.com/capitalone/DataProfiler/issues/905


requirements.txt
```diff
@@ -10,9 +10,10 @@ fastavro>=1.0.0.post1
...
 charset-normalizer>=1.3.6
 psutil>=4.0.0
-scipy>=1.4.1,<1.11.0
+scipy>=1.4.1
 requests>=2.28.1
 networkx>=2.5.1
 typing-extensions>=3.10.0.2
 HLL>=2.0.3
 datasketches>=4.1.0
+packaging>=23.0
```
JGSweets, line 19: let's move this to only in the graph requirements list since we don't need it elsewhere

JGSweets: Nvm, there's no graph specific req file!

dataprofiler/profilers/graph_profiler.py
```diff
@@ -399,9 +401,21 @@ def _get_continuous_distribution(
                 best_mle: float = 1000
                 best_fit_properties: tuple = None  # type: ignore[assignment]
 
+                scipy_updated = False
+                scipy_version = version.parse(importlib.metadata.version("scipy"))
+                if scipy_version >= version.parse("1.11.0"):
+                    scipy_updated = True
```
JGSweets, line 407: Let's do two things here:
1. move it outside the for loop above as well since we don't need to recalc in the loop.
    * if we really want to we could also do it towards the top of the file.
2. instead of `scipy_updated` let's clarify the variable. Ideally the code reads as natural language so `scipy_updated` is quote ambiguous. I suggest `scipy_gte_1_11_0`

dataprofiler/profilers/graph_profiler.py
```diff
@@ -391,6 +393,12 @@ def _get_continuous_distribution(
             st.lognorm,
             st.gamma,
         ]
+
+        scipy_gte_1_11_0 = False
+        scipy_version = version.parse(importlib.metadata.version("scipy"))
+        if scipy_version >= version.parse("1.11.0"):
+            scipy_gte_1_11_0 = True
```
micdavis, line 400: thoughts
```suggestion
        scipy_gte_1_11_0 = version.parse(importlib.metadata.version("scipy")) >= version.parse("1.11.0")
```

JGSweets (approved):

micdavis (approved):

PR #964, opened 2023-07-19
None-check labels dependants

junholee6a:
The mypy errors addressed here occur because variables `label_mapping` (in `CharPreprocessor.process()`), `unstructured_labels`, and `unstructured_label_set` (in `StructCharPreprocessor.process()`) have optional types when they're used. This is fixed by checking that they are not `None` prior to the operation, which mypy recognizes as removing the `None` type from them.

This should have no effect on functionality because we are already checking that `labels` is not `None`, and the variables above all depend on `labels` such that they are `None` only if `labels` is `None`.


JGSweets (approved):

taylorfturner (approved):
LGTM

PR #968, opened 2023-07-25
Make sure random_state is a list before indexed assignment

junholee6a:
Currently, a mypy error occurs because we attempt to assign to `random_state[1]` when `random_state` has type
`Union[list[Any], tuple[Any]]`. Tuples are immutable so this is a type error.

We fix this by making `random_state` into a list before doing indexed assignment on it. This tells mypy that `random_state` is definitely a `list` and cannot be a `tuple`.


taylorfturner:
update branch

tyfarnan (approved):

dataprofiler/labelers/data_processing.py
```diff
@@ -1556,7 +1556,8 @@ def __init__(
         elif isinstance(random_state, (list, tuple)) and len(random_state) == 3:
             # tuple required for random state to be set, lists do not work
             if isinstance(random_state[1], list):
-                random_state[1] = tuple(random_state[1])  # type: ignore
+                random_state = list(random_state)
```
JGSweets, line 1559: This might be a case for cast as opposed to a computation. I'm surprised `isinstance(random_state, (list, tuple))` isn't alleviating this.

taylorfturner: yeah just trick `mypy` with the cast rather than at execution time actually changing the type to `list`

junholee6a: > This might be a case for cast as opposed to a computation. I'm surprised isinstance(random_state, (list, tuple)) isn't alleviating this.

It's because `random_state` could be a tuple, in which case indexed assignment isn't possible.

Is there any possibility that `random_state` is a tuple while `random_state[1]` is a list? If this will never happen we can safely cast here, or change the conditional to `if isinstance(random_state[1], list) and isinstance(random_state, list):`

junholee6a: > Is there any possibility that random_state is a tuple while random_state[1] is a list?

Though, if this ever did happen in the past it would've thrown a runtime error.

ksneab7 (approved):

taylorfturner (dismissed):
blocking for 0.10.3 release. will unblock post-deployment

taylorfturner:
will review today. Update branch with `dev` needed @junholee6a

taylorfturner:
@junholee6a sorry something else merged to dev before this.... can you update branch once more 🙂

taylorfturner:
> Rebased

still showing as out of date with `dev` 🤔

taylorfturner:
@junholee6a update branch here lol -- @ksneab7 let's review this PR first

dataprofiler/labelers/data_processing.py
ksneab7: @taylorfturner Did this get resolved?

junholee6a: > or change the line 1558 conditional to if isinstance(random_state[1], list) and isinstance(random_state, list):

I just committed this solution since it avoids casting

taylorfturner (approved):

dataprofiler/labelers/data_processing.py
```diff
@@ -1555,8 +1555,8 @@ def __init__(
             random_state = random.Random(random_state)
         elif isinstance(random_state, (list, tuple)) and len(random_state) == 3:
             # tuple required for random state to be set, lists do not work
-            if isinstance(random_state[1], list):
-                random_state[1] = tuple(random_state[1])  # type: ignore
+            if isinstance(random_state[1], list) and isinstance(random_state, list):
```
micdavis, line 1558: should we first check if `random_state` is a list and then check if `random_state[1]` is a list? Otherwise wouldn't the first boolean condition error if `random_state` isn't a list to begin with?
```suggestion
            if isinstance(random_state, list) and isinstance(random_state[1], list):
```

junholee6a: It technically doesn't matter here since we've already checked that `random_state` is either a `list` or a `tuple` in line 1556, which means `random_state` is indexable. But I agree it's better practice to check `random_state` before `random_state[1]` in case the code changes later/for readability

dataprofiler/labelers/data_processing.py
```diff
@@ -1894,8 +1894,8 @@ def __init__(
...
```
taylorfturner, line 1897:
```suggestion
            if isinstance(random_state, list) and isinstance(random_state[1], list):
```

re-order for consistency

taylorfturner (changes requested):

taylorfturner (approved):

PR #971, opened 2023-07-26
Automate multiprocessing option

clee1152:
Based on analysis with @ksneab7 and @taylorfturner, the goal is to automate turning on `multiprocess` option in `StructuredProfiler` if the number of rows exceeds 750,000, or if the number of columns exceeds 20.


dataprofiler/profilers/profile_builder.py
```diff
@@ -2801,6 +2829,11 @@ def _update_profile_from_chunk(
         elif isinstance(data, list):
             data = pd.DataFrame(data, dtype=object)
 
+        # If options.multiprocess is enabled, auto-toggle multiprocessing
+        auto_multiprocess_toggle = None
+        if self.options.multiprocess.is_enabled:
+            auto_multiprocess_toggle = self._auto_multiprocess_toggle(data, 750000, 20)
```
JGSweets, line 2835: since the default is set, no need to pass in as well.

dataprofiler/profilers/profile_builder.py
```diff
@@ -2779,6 +2779,34 @@ def _merge_null_replication_metrics(self, other: StructuredProfiler) -> dict:
 
         return merged_properties
 
+    def _auto_multiprocess_toggle(
```
JGSweets, line 2782: Should we do a new function or instead do it in: `suggest_pool_size` which is similar.

JGSweets: Also notating this is a staticmethod wrt the class. If we kept the func, maybe should be in profiler_utils

dataprofiler/profilers/profile_builder.py
taylorfturner: +1

dataprofiler/profilers/profile_builder.py
clee1152: An issue I see with doing in `suggest_pool_size` is that it uses a different metric to measure dataset sizes for thresholding (number of rows vs. estimated data size in memory). If we were to do it in there, we would have to change the logic for this (example below). 
```
    if self.options.multiprocess.is_enabled and auto_multiprocess_toggle:
            est_data_size = data[:50000].memory_usage(index=False, deep=True).sum()
            est_data_size = (est_data_size / min(50000, len(data))) * len(data)
            pool, pool_size = profiler_utils.generate_pool(
                max_pool_size=None, data_size=est_data_size, cols=len(data.columns)
            )
```

dataprofiler/profilers/profile_builder.py
tyfarnan: +1

dataprofiler/profilers/profile_builder.py
```diff
@@ -2869,7 +2874,7 @@ def tqdm(level: set[int]) -> Generator[int, None, None]:
 
         # Generate pool and estimate datasize
         pool = None
-        if self.options.multiprocess.is_enabled:
+        if self.options.multiprocess.is_enabled and auto_multiprocess_toggle:
             est_data_size = data[:50000].memory_usage(index=False, deep=True).sum()
```
tyfarnan, line 2878: outside scope of this PR, but just noticed and wondered if this 50000 should hardcoded?

tyfarnan (dismissed):

dataprofiler/profilers/profile_builder.py
```diff
@@ -2801,6 +2801,11 @@ def _update_profile_from_chunk(
         elif isinstance(data, list):
             data = pd.DataFrame(data, dtype=object)
 
+        # If options.multiprocess is enabled, auto-toggle multiprocessing
+        auto_multiprocess_toggle = None
+        if self.options.multiprocess.is_enabled:
+            auto_multiprocess_toggle = profiler_utils.auto_multiprocess_toggle(data)
```
taylorfturner, line 2807: nice -- keying on defaults

dataprofiler/profilers/profile_builder.py
taylorfturner: fixed

dataprofiler/version.py
```diff
@@ -2,7 +2,7 @@
 
 MAJOR = 0
 MINOR = 10
-MICRO = 0
+MICRO = 1
```
taylorfturner, line 5: hmm why is this changing? makes me think you need a rebase :thinking:

taylorfturner (changes requested):

dataprofiler/version.py
taylorfturner: I would recommend just reverting this

PR #980, opened 2023-07-31
Automate multiprocess option

clee1152:


taylorfturner (approved):
LGTM

dataprofiler/tests/profilers/test_profiler_utils.py
```diff
@@ -469,3 +469,52 @@ def test_odd_merge_profile_list(self, mock_data_labeler, *mocks):
...
         self.assertEqual(60.0, single_report["data_stats"][0]["statistics"]["max"])
+
+
+class TestAutoMultiProcessToggle(unittest.TestCase):
+
+    """
+    Validate profile_utils.auto_multiprocess_toggle is properly working.
+    """
+
+    def test_auto_multiprocess_toggle(self, *mocks):
```
taylorfturner, line 480: remove `*mocks`

taylorfturner (changes requested):

taylorfturner (dismissed):

dataprofiler/profilers/profile_builder.py
```diff
@@ -2869,7 +2874,7 @@ def tqdm(level: set[int]) -> Generator[int, None, None]:
 
         # Generate pool and estimate datasize
         pool = None
-        if self.options.multiprocess.is_enabled:
+        if self.options.multiprocess.is_enabled and auto_multiprocess_toggle:
```
micdavis, line 2877: I wonder if we can just do `if auto_multiprocess_toggle:` since that is really what turns on the multiprocess. We already use `self.options.multiprocess.is_enabled` above to change `auto_multiprocess_toggle` so I don't think it's needed here and below

dataprofiler/profilers/profiler_utils.py
```diff
@@ -177,6 +177,34 @@ def partition(data: list, chunk_size: int) -> Generator[list, None, Any]:
...
+    num_rows_threshold: int = 750000,
+    num_cols_threshold: int = 20,
+) -> bool:
+    """
+    Automate multiprocessing toggle depending on dataset sizes.
+
+    :param data: a dataset
+    :type data: pandas.DataFrame
+    :param num_rows_threshold: threshold for number of rows over
+        which options.multiprocess is enabled
```
taylorfturner, line 191: not technically true... its the threshold at which if `options.multiprocess == True` we will check dataset size for multiprocess?

dataprofiler/profilers/profiler_utils.py
```diff
@@ -177,6 +177,34 @@ def partition(data: list, chunk_size: int) -> Generator[list, None, Any]:
...
+    :type num_rows_threshold: int
+    :param num_cols_threshold: threshold for number of columns
+        over which options.multiprocess is enabled
```
taylorfturner, line 194: same as above

PR #999, opened 2023-08-03
Add docs to `multiprocess` option in StructuredOptions.

clee1152:
Updated `multiprocess` to describe the auto-toggle, docs are updated accordingly to match.

PR #1000, opened 2023-08-03
Staging: into dev-gh-pages the docs for `multiprocess`.

clee1152:
#999


taylorfturner (changes requested):
wrong location the PR coming from

PR #1026, opened 2023-09-14
update install instructions for mac

atl1502:
Snappy install was not working on Apple chips. This should allow for DataProfiler's install to work on Apple chips/


taylorfturner (approved):
LGTM

PR #1027, opened 2023-09-18
added psi calculation to categorical columns

ksneab7:


dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -756,6 +757,44 @@ def test_categorical_diff(self):
...
+        df_categorical = pd.Series(["y", "y", "y", "y", "n", "n", "n", "maybe"])
+        profile = CategoricalColumn(df_categorical.name)
+        profile.update(df_categorical)
+
+        df_categorical = pd.Series(["y", "maybe", "y", "y", "n", "n", "maybe"])
+        profile2 = CategoricalColumn(df_categorical.name)
+        profile2.update(df_categorical)
+
+        # Calculate expected_psi
+        expected_psi = 0
```
taylorfturner, line 770: What about other test cases for non zero?

ksneab7: It is non-zero

ksneab7: for perc_A, perc_B in zip(bin_perc, bin_perc_2):
            expected_psi += (perc_B - perc_A) * math.log(perc_B / perc_A)

micdavis: could we make `expected_psi` in the `expected_diff` hardcoded as a value? or would that be tough? I just want the expected to be completely disjoint from any calculations. In this case were using the formula for PSI to check our formula for PSI, sort of...

ksneab7:
Follow up issue to calculate PSI using the correct functionality that exists within NumericalStatsMixin:
https://github.com/capitalone/DataProfiler/issues/1028

micdavis (dismissed):

taylorfturner (changes requested):

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -304,7 +306,21 @@ def diff(self, other_profile: CategoricalColumn, options: dict = None) -> dict:
...
             )
-
+            total_psi = 0.0
+            if cat_count1.keys() == cat_count2.keys():
+                for key in cat_count1.keys():
+                    perc_A = cat_count1[key] / self.sample_size
+                    perc_B = cat_count2[key] / other_profile.sample_size
+                    total_psi += (perc_B - perc_A) * math.log(perc_B / perc_A)
+            else:
+                warnings.warn(
```
taylorfturner, line 316: Maybe if we hit this we should just either 1) have PSI to None or 2) not include in the difference ducts

dataprofiler/profilers/categorical_column_profile.py
```diff
@@ -304,7 +305,14 @@ def diff(self, other_profile: CategoricalColumn, options: dict = None) -> dict:
                     other_profile._categories.items(), key=itemgetter(1), reverse=True
                 )
             )
-
+            if cat_count1.keys() == cat_count2.keys():
+                total_psi = 0.0
```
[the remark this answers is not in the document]
ksneab7: if the categories are equal and of equal count the psi is zero. So if there are no categories (and by extension no counts so no percentages to calculate) I have a couple questions:
1) how did the code get called anyway, if there are no categories the categorical profiler should never be initialized and cant be diffed
2) Even if we get here the `psi` of nothing compared to nothing should be zero,  `psi` is used to calculate change between two datasets, if nothing changed because there is nothing in both profiles, returning 0.0 for `psi` I think as a good thing right?

dataprofiler/profilers/categorical_column_profile.py
ksneab7: I agree with this if PSI cant be calculated it shouldnt default to zero

taylorfturner (approved):

dataprofiler/tests/profilers/test_categorical_column_profile.py
```diff
@@ -756,6 +761,38 @@ def test_categorical_diff(self):
...
+        # psi = (% of records based on Sample (A) - % of records  Sample (B)) * ln(A/ B)
+        # p-value found through using chi2 CDF
+        expected_diff = {
+            "categorical": "unchanged",
+            "statistics": {
+                "unique_count": "unchanged",
+                "unique_ratio": -0.05357142857142855,
+                "chi2-test": {
+                    "chi2-statistic": 0.6122448979591839,
+                    "df": 2,
```
tyfarnan, line 784: outside scope of this PR: Does df stand for dataframe here? Looks like it's an int.

tyfarnan (approved):
lgtm - might be worth renaming the chi2-test `"df"` key to something else if it isn't a `dataframe`

dataprofiler/tests/profilers/test_categorical_column_profile.py
ksneab7: No its supposed to an int, its stands for degrees of freedom, but I agree that it is not a great choice of name

PR #969, opened 2023-07-25
Delay transforming priority_order into ndarray

junholee6a:
Issue: https://github.com/capitalone/DataProfiler/issues/722

In the changed code, we had a mypy error because numpy `ndarray`s are not compatible with `random.Random.shuffle()` (expected argument type is `MutableSequence[Any]`)

We fix this by first instantiating `priority_order` as a list, then shuffling it, then creating an ndarray from it afterwards.

PR #1045, opened 2023-09-30
Delay transforming priority_order into ndarray

junholee6a:

In the changed code, we had a mypy error because numpy `ndarray`s are not compatible with `random.Random.shuffle()` (expected argument type is `MutableSequence[Any]`)

We fix this by first instantiating `priority_order` as a list, then shuffling it, then creating an ndarray from it afterwards.


junholee6a:
This is a remake of PR https://github.com/capitalone/DataProfiler/pull/969 after branches were rebuilt

ksneab7 (approved):

ksneab7:
Please update when able :)

taylorfturner (approved):

taylorfturner:
Thanks @junholee6a!

PR #1047, opened 2023-10-05
Fix Continuous Distribution for Graph Profiler

atl1502:
The parameters for the continuous distributions in graph profiler were being passing in incorrectly. This caused the report to list the distribution's moments as a list rather than a single value. This fix addresses those issues by properly passing in the distribution parameters.


taylorfturner (changes requested):
blocking temporarily for review since this is a big change that impacts outputs from the profiler

dataprofiler/profilers/graph_profiler.py
```diff
@@ -422,21 +422,35 @@ def _get_continuous_distribution(
...
-                        best_fit_properties = fit
-
-                mean, variance, skew, kurtosis = best_distrib.stats(
-                    best_fit_properties, moments="mvsk"
-                )
+                        best_fit_attributes = fit[:-2]
+                        best_fit_scale = fit[-1]
+                        best_fit_loc = fit[-2]
+                if best_fit_attributes:
+                    mean, variance, skew, kurtosis = (
```
ksneab7, line 429: I think it would be better to iterate through expected keys here rather than the dict itself. 
What I mean is if the best_distrib.stats for some reason differs from the expected 4 values (mean, variance, skew, kurtosis), it could lead to an error. Now whether we want an error if that happens is up for debate. @taylorfturner what is the direction we want to go with this?

ksneab7: Disregard! This makes use of an external library

ksneab7 (approved):

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -99,15 +99,11 @@ def setUpClass(cls):
...
-            ],
-            "mean": [1.7085707836543698e16, 4.241852142820433, 1.0190038591415866],
-            "variance": [8.521811094505713e64, 305.76588081569196, 0.03984103474264823],
-            "skew": [4.987683961374356e48, 82.41830452500491, 0.5951548443693909],
-            "kurtosis": [7.262126433044066e129, 117436.2896499293, 0.6363254662738349],
+            "best_fit_attributes": [8.646041719759628],
+            "mean": 3315291431455125.5,
+            "variance": 3.2085541544027255e63,
+            "skew": 4.987683961374356e48,
+            "kurtosis": 7.262126433044066e129,
```
ksneab7, line 106: Why do these values change for the unit test. If functionality is equivalent shouldnt these values stay the same?

ksneab7: So we would need another fucntion to ensure the older code within the above if else statement is still functioning the same way.

atl1502: This is because the distribution was being calculated incorrectly in the original code.

ksneab7 (dismissed):

dataprofiler/profilers/graph_profiler.py
```diff
@@ -404,7 +404,7 @@ def _get_continuous_distribution(
                 df = pd.Series(data_as_list)
                 best_fit: str = None  # type: ignore[assignment]
                 best_mle: float = 1000
-                best_fit_properties: tuple = None  # type: ignore[assignment]
+                best_fit_attributes: tuple = None  # type: ignore[assignment]
```
taylorfturner, line 407: I'd recommend pairing this PR back to only the change required for cont distribution to properly calculate distribution. There are some changes (variable renaming for example) that isn't really required. Better to change only absolutely necessary things to fix the issue

dataprofiler/tests/profilers/test_graph_profiler.py
atli3: These are the values for the distribution, it is because the size of the distribution's scale parameter is so large.

taylorfturner (commented):
something is fishy with this number

micdavis (dismissed):

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -310,18 +328,27 @@ def test_diff(self):
...
         profile_3.profile["continuous_distribution"]["value"].pop("properties")
+        profile_3.profile["continuous_distribution"]["value"].pop("mean")
+        profile_3.profile["continuous_distribution"]["value"].pop("standard_deviation")
 
         diff_1 = profile_1.diff(profile_2)
         self.assertDictEqual(diff_1, self.expected_diff_1)
 
         # Tests diffs between profiles with different # nodes and different continuous/categorical attributes
         diff_2 = profile_2.diff(profile_3)
+        print(diff_2)
```
taylorfturner, line 351: def remove prints like this

taylorfturner (changes requested):

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -220,17 +227,58 @@ def setUpClass(cls):
             },
         }
 
-    def check_continuous_properties(self, continuous_distribution_props):
+    def check_profile(self, profile, report=False):
```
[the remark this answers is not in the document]
micdavis: I think you just need to pass in a reference to the dict that is going to be used as the expected value since there are now two continuous distributions (weights and uniform). So maybe something like:
```suggestion
    def check_continuous_properties(self, continuous_distribution_props, expected_distribution_props):
```

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -248,35 +296,15 @@ def test_profile(self):
...
-        self.assertAlmostEqual(scale, -15.250985118262854)
-
-        # check that properties are almost equal
-        properties = profile.profile["continuous_distribution"]["weight"].pop(
-            "properties"
-        )
-        self.check_continuous_properties(properties)
-
-        self.assertDictEqual(self.expected_profile, profile.profile)
+        self.check_profile(profile)
```
micdavis, line 299: This should be all you need here:
```suggestion
        # check that properties are almost equal
        weight_properties = profile.profile["continuous_distribution"]["weight"].pop(
            "properties"
        )
        uniform_properties = profile.profile["continuous_distribution"]["uniform"].pop(
            "properties"
        )
        
        self.check_continuous_properties(weight_properties, self.expected_properties_weight)
        self.check_continuous_properties(uniform_properties, self.expected_properties_uniform)
        self.assertDictEqual(self.expected_profile, profile.profile)
```

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -248,35 +296,15 @@ def test_profile(self):
...
-        self.assertAlmostEqual(scale, -15.250985118262854)
-
-        # check that properties are almost equal
-        properties = profile.profile["continuous_distribution"]["weight"].pop(
-            "properties"
-        )
-        self.check_continuous_properties(properties)
-
-        self.assertDictEqual(self.expected_profile, profile.report())
+        self.check_profile(profile, report=True)
```
micdavis, line 307: Almost same as the comment above just change last line to assert report

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -285,17 +313,7 @@ def test_graph_data_object(self):
...
-        self.assertAlmostEqual(scale, -15.250985118262854)
-
-        # check that properties are almost equal
-        properties = profile.profile["continuous_distribution"]["weight"].pop(
-            "properties"
-        )
-        self.check_continuous_properties(properties)
-
-        self.assertDictEqual(self.expected_profile, profile.profile)
+        self.check_profile(profile)
```
micdavis, line 316: see comment above:

dataprofiler/tests/profilers/test_graph_profiler.py
atl1502: They are the moments from the distribution not from the actual data in the graph. They are about as large as before.

dataprofiler/tests/profilers/test_graph_profiler.py
```diff
@@ -26,17 +26,19 @@ def setup_save_mock_open(mock_open):
 
 
 class TestGraphProfiler(unittest.TestCase):
+    maxDiff = None
```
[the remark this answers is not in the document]
taylorfturner: good catch @micdavis

micdavis (approved):

taylorfturner:
pending 0.10.6 release first

taylorfturner:
updated new base to `dev` -- will need to rebase probably @atl1502

PR #1050, opened 2023-10-20
update version of urllib3

atl1502:


taylorfturner (changes requested):
might not be needed

taylorfturner:
should be resolved in 0.10.6

PR #1051, opened 2023-10-24
feat: add null ratio to column stats

suprabhatgurrala:
Null ratio is computed as a global stat, but not on a column-level. I've made a simple change to compute null ratio as a column stat by dividing `null_count / sample_size`.


taylorfturner:
Small change request is to checkout your branch off of `dev` instead of `main` and open the PR into `dev`. Thanks @suprabhatgurrala  for the contribution!

suprabhatgurrala:
My fork only had the `main` branch, will create a new PR which is branched of the `dev` branch

Issue #1030, opened 2023-09-20
Confusing name for degree of freedom in Chi2 metrics


ksneab7:
We currently use `df` to represent degree of freedom in chi^2 metrics. This is confusing as it could potentially be read as dataframe, we may want to change to something like `deg_of_free` or `dff` something less ambiguous

taylorfturner:
Closing after merged and docs and code update. Thanks @ksneab7 and @mainkhan

PR #1056, opened 2023-10-27
Rename references to degree of freedom from df to deg_of_free

mainkhan:
Fixes Issue #1030 

This PR renames references to degrees of freedom in a chi-squared calculation to `deg_of_free` to not get confused with when we use `df` to refer to a data frame. 

I have run the unit test suite and verified that the 1155 tests are running without fail. 
I have also run pre-commit on all the files without issue.

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -611,8 +611,8 @@ class NumericStatsMixin(BaseColumnProfiler[NumericStatsMixinT], metaclass=abc.AB
     ) -> dict:
         results: dict = {
             "t-statistic": None,
-            "conservative": {"df": None, "p-value": None},
-            "welch": {"df": None, "p-value": None},
+            "conservative": {"deg_of_free": None, "p-value": None},
+            "welch": {"deg_of_free": None, "p-value": None},
         }
 
         invalid_stats = False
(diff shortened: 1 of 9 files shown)


taylorfturner:
@mainkhan if you don't mind, let's open this PR into `dev` and not `main`. Thanks for the contribution!

taylorfturner (changes requested):
open PR into `dev`

mainkhan:
Hey @taylorfturner I changed the base branch. Sorry about that, I know it should be obvious but I think I got confused because in CONTRIBUTING.md the first step to create a pull request is to create a feature branch off of `main`

taylorfturner:
> Hey @taylorfturner I changed the base branch. Sorry about that, I know it should be obvious but I think I got confused because in CONTRIBUTING.md the first step to create a pull request is to create a feature branch off of `main`

Totally understandable -- we will update that for clarity. Poor direction on our part. Thanks for the contribution!

dataprofiler/tests/profilers/test_categorical_column_profile.py
taylorfturner, line 1: since this is changing behavior of the output for end users
- we will need to be very clear in our communication about what changed and when (in PR notes)
- also will need to ensure documentation is updated properly to ensure that is up to date and matching code's current state with this change

taylorfturner: @mainkhan my one thought is that we should 
- research needed documentation changes [here](https://capitalone.github.io/DataProfiler/docs/0.10.5/html/index.html)
- if any needed changes open up a PR into `dev-gh-pages` in parallel to the code change

mainkhan: I looked in to the docs and could not find any information documenting explicitly what the output will be so I'm not sure we need to change any existing documentation. Perhaps just note in release notes documenting that a field in the dict has changed

I see here: https://capitalone.github.io/DataProfiler/docs/0.10.5/html/dataprofiler.dp_logging.html?highlight=chi#dataprofiler.dp_logging.get_child_logger

```
dataprofiler.profilers.profiler_utils.perform_chi_squared_test_for_homogeneity(categories1: dict, sample_size1: int, categories2: dict, sample_size2: int) → dict[str, int | float | None]

    Perform a Chi Squared test for homogeneity between two groups.

    Parameters

            categories1 (dict) – Categories and respective counts of the first group

            sample_size1 (int) – Number of samples in first group

            categories2 (dict) – Categories and respective counts of the second group

            sample_size2 (int) – Number of samples in second group

    Returns

        Results of the chi squared test
    Return type

        dict

```

the documentation only specifies that we are returning a dict but does not define the dict's keys.

taylorfturner: more like [this](https://capitalone.github.io/DataProfiler/docs/0.10.5/html/profiler.html#:~:text=%27t%2Dtest%27,value%22%3A%20float) area of the docs

mainkhan: Created a companion documentation PR #1057

taylorfturner: awesome

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -647,17 +647,17 @@ def _perform_t_test(
...
             (var1 / n1) ** 2 / (n1 - 1) + (var2 / n2) ** 2 / (n2 - 1)
         )
         results["t-statistic"] = t
-        results["conservative"]["df"] = float(conservative_df)
-        results["welch"]["df"] = float(welch_df)
+        results["conservative"]["deg_of_free"] = float(conservative_deg_of_free)
+        results["welch"]["deg_of_free"] = float(welch_deg_of_free)
 
-        conservative_t = scipy.stats.t(conservative_df)
+        conservative_t = scipy.stats.t(conservative_deg_of_free)
```
taylorfturner, line 658: scipy refers to this as `df`; however, I agree for clarity in this repo a rename would make the code. Its more readable / cause less confusion / less ambiguity  and as a result --> easier to maintain

taylorfturner (approved):

micdavis (approved):

Issue #820, opened 2023-05-15
`_assimilate_histogram` and `_regenerate_histogram` refactor into standalone functions


ksneab7:
**Is your feature request related to a problem? Please describe.**
`_assimilate_histogram`  and  `_regenerate_histogram` functions are not using self for anything of substance and as a result can be separated into their own standalone static functions.

**Describe the outcome you'd like:**
Move these two functions outside of the class to histogram_utils.py

**Additional context:**

junholee6a:
I presume this would also involve writing unit tests for `_assimilate_histogram` and `_regenerate_histogram` in `test_histogram_utils.py`? Neither have any direct unit tests at the moment, though they might be called indirectly from other tests

taylorfturner:
Yes, definitely would want to ensure coverage with testing for sure @junholee6a

junholee6a:
Working on this. It'll take a bit to understand the functions and write test cases

junholee6a:
In [`_assimilate_histogram()`](https://github.com/capitalone/DataProfiler/blob/dev/dataprofiler/profilers/numerical_column_stats.py#L1383), it seems like each bin in the `from_hist` maps to at most two bins in the `dest_bin`. This means that when `from_hist` has much larger bins than `dest_bin`, its values aren't distributed evenly:

Test code:

```python
def test_assimilate_histogram(self):
    ret_value = histogram_utils._assimilate_histogram(
        from_hist_entity_count_per_bin=np.array([50, 0]),
        from_hist_bin_edges=np.array([0, 5, 10]),
        dest_hist_entity_count_per_bin=np.zeros(10),
        dest_hist_bin_edges=np.array((0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10)),
        dest_hist_num_bin=10,
    )
    print('=============== _assimilate_histogram result ===============')
    print(ret_value)
```

Output (reformatted by hand):

```
=============== _assimilate_histogram result ===============
(
    {
        'bin_edges': array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10]),
        'bin_counts': array([10., 40.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.])
    },
    80.0
)
```

Expected output: `bin_counts` should be distributed more evenly

```
=============== _assimilate_histogram result ===============
(
    {
        'bin_edges': array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10]),
[...]

PR #1071, opened 2023-11-19
Make _assimilate_histogram() not use self

junholee6a:
NOTE: This is an alternative of PR https://github.com/capitalone/DataProfiler/pull/1073. If this is merged, then close PR https://github.com/capitalone/DataProfiler/pull/1073

Issue: https://github.com/capitalone/DataProfiler/issues/820

This is a necessary step to resolving issue https://github.com/capitalone/DataProfiler/issues/820. Previously, `_assimilate_histogram()` called `self` to decide whether the given histogram contained integers or floats, and rounded the bins for histograms that only contained integers.

`_assimilate_histogram()` is called in exactly two places: `_regenerate_histogram()` and `_add_helper_merge_profile_histograms()`. To remove the dependency on `self`, we can move the code for rounding bin edges to `_regenerate_histograms()` and add an argument indicating whether the given histogram contains integers or floats.

While `_regenerate_histogram()` should behave exactly the same as before, the histogram loss calculation in `_add_helper_merge_profile_histograms()` may change slightly since we no longer round its histogram bins.

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -13,7 +13,7 @@ import numpy.typing as npt
 import pandas as pd
 import scipy.stats
 
-from . import histogram_utils, profiler_utils
+from . import float_column_profile, histogram_utils, profiler_utils
 from .base_column_profilers import BaseColumnProfiler
 from .profiler_options import NumericalOptions
 
@@ -710,6 +710,9 @@ class NumericStatsMixin(BaseColumnProfiler[NumericStatsMixinT], metaclass=abc.AB
                     entity_count_per_bin=self_histogram["bin_counts"],
                     bin_edges=self_histogram["bin_edges"],
                     suggested_bin_count=num_psi_bins,
+                    is_float_histogram=isinstance(
+                        self, float_column_profile.FloatColumn
+                    ),
                     options={
                         "min_edge": min_min_edge,
                         "max_edge": max_max_edge,
@@ -731,6 +734,9 @@ class NumericStatsMixin(BaseColumnProfiler[NumericStatsMixinT], metaclass=abc.AB
                     entity_count_per_bin=other_histogram["bin_counts"],
                     bin_edges=other_histogram["bin_edges"],
                     suggested_bin_count=num_psi_bins,
+                    is_float_histogram=isinstance(
+                        self, float_column_profile.FloatColumn
+                    ),
                     options={
                         "min_edge": min_min_edge,
                         "max_edge": max_max_edge,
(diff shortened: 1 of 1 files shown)


dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1372,6 +1383,11 @@ def _regenerate_histogram(
             new_bin_edges = np.linspace(
                 options["min_edge"], options["max_edge"], suggested_bin_count + 1
             )
+
+        # if it's not a float histogram, then assume it only contains integer values
+        if not is_float_histogram:
+            bin_edges = np.round(bin_edges)
```
taylorfturner, line 1389: this is a big assumption ... can you comment on the rationale here?

junholee6a: The same assumption was being made in `_assimilate_histogram` ([diff](https://github.com/capitalone/DataProfiler/pull/1071/files#diff-c46fb1aabfd8fc56a5d68b0ab6743d0c27918bc5b929538411a301472d116a43L1420)). I moved it from `_assimilate_histogram` to `_regenerate_histogram`.

We could also remove that rounding code entirely, but it causes a single test failure [here](https://github.com/junholee6a/DataProfiler/blob/assimilate_histogram_remove_dependency/dataprofiler/tests/profilers/test_text_column_profile.py#L285). I haven't been able to figure out what purpose the rounding fulfills here, as it existed since the first release. I kept it so that the behavior of `_regenerate_histogram` wouldn't change from before.

But you're right that it's a big assumption, and it's worth considering its removal. I'll open an alternative PR for that

junholee6a: PR https://github.com/capitalone/DataProfiler/pull/1073 is an alternative to this PR where we remove the rounding code

taylorfturner:
Taking a gander at these today -- thanks for the patience, @junholee6a!

dataprofiler/profilers/numerical_column_stats.py
ksneab7: Okay, after extensive discussion with @taylorfturner @micdavis we have determined that this assumption is valid and all values passed into this function at the current placement in the workflow will all be numeric (either ints and floats). I think there should be an additional test where this line 1389 is called (or if a test already exists) to validate that bin edges are able to be rounded at that point in the workflow (i.e. is it possible to get a non-roundable value at this point in the code, which I believe is not possible).

Assuming this test is created I am good with this version of code and we can delete PR 1073

taylorfturner: I closed #1073

taylorfturner:
@junholee6a just one comment on this from @ksneab7. We opted for this PR over the alternative (closed now). Let us know if you have any questions. Best!

dataprofiler/profilers/numerical_column_stats.py
junholee6a: @ksneab7 Thanks for the feedback! Line 1389 is called in the existing test `TestTextColumnProfiler.test_profile` (I've attached a screenshot of my debugger stopping at line 1389 while running the test, to demonstrate this). Also, if this rounding code is removed entirely, the output changes and this test fails, as I explained in https://github.com/capitalone/DataProfiler/pull/1073.

Screenshot 2024-01-09 at 2 31 52 PM

Does this fulfill what you meant, or did you have something else in mind?

ksneab7: Nope this is exactly what I was looking for thank you!

ksneab7 (approved):

taylorfturner (approved):

taylorfturner:
Thanks @junholee6a -- merged

PR #1073, opened 2023-11-21
Make _assimilate_histogram() not use self (alternative)

junholee6a:
NOTE: This is an alternative for PR https://github.com/capitalone/DataProfiler/pull/1071. If this is merged, then close PR https://github.com/capitalone/DataProfiler/pull/1071

Issue: https://github.com/capitalone/DataProfiler/issues/820

This is a necessary step to resolving issue https://github.com/capitalone/DataProfiler/issues/820. Previously, `_assimilate_histogram()` called `self` to decide whether the given histogram contained integers or floats, and rounded the bins for histograms that only contained integers.

However, that rounding seems unnecessary. Here, we remove that rounding code entirely and modify the one test that fails, `TestTextColumnProfiler.test_profile()`. To make sure the test is still valid, here are its values:

The data in the profile of that test is:
data

The old expected histogram is:
old_histogram

And the new expected histogram is:
new_histogram

### DIFF
--- a/dataprofiler/profilers/numerical_column_stats.py
+++ b/dataprofiler/profilers/numerical_column_stats.py
@@ -1417,11 +1417,6 @@ class NumericStatsMixin(BaseColumnProfiler[NumericStatsMixinT], metaclass=abc.AB
 
             bin_edge = from_hist_bin_edges[bin_id : bin_id + 3]
 
-            # if we know not float, we can assume values in bins are integers.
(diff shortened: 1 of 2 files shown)


taylorfturner:
Taking a gander at these today -- thanks for the patience, @junholee6a!

taylorfturner:
Sticking with 1071 for this changed. Closing 1073

Issue #1100, opened 2024-02-29
JSON Serialization Error


scottiegarcia:
**General Information:**
 - OS: MacOS Sonoma 14.3.1
 - Python version: 3.10
 - Library version: 0.10.8

**Describe the bug:**

JSON Serialization errors due to datetime object not getting serialized beforehand

**To Reproduce:**
```
from dataprofiler import Profiler
import pandas as pd

df = pd.DataFrame({"test": [1209214]})

dp = Profiler(df)
dp.save("test.json", save_method="json")       
```

**Expected behavior:**
Dataprofiler should serialize all datetime attributes to strings before saving to JSON, and deserialize on load

**Screenshots:**

Traceback
```
Traceback (most recent call last):
  File "/Projects/helpful-scripts/bug2.py", line 10, in <module>
    dp.save("test.json", save_method="json")
  File "/miniconda3/envs/monitoring/lib/python3.10/site-packages/dataprofiler/profilers/profile_builder.py", line 3070, in save
    self._json_save_helper(filepath)
  File "/miniconda3/envs/monitoring/lib/python3.10/site-packages/dataprofiler/profilers/profile_builder.py", line 1155, in _json_save_helper
    json.dump(self, f, cls=ProfileEncoder)
  File "/miniconda3/envs/monitoring/lib/python3.10/json/__init__.py", line 179, in dump
    for chunk in iterable:
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 439, in _iterencode
    yield from _iterencode(o, _current_indent_level)
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 325, in _iterencode_list
    yield from chunks
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 439, in _iterencode
    yield from _iterencode(o, _current_indent_level)
  File "/miniconda3/envs/monitoring/lib/python3.10/json/encoder.py", line 431, in _iterencode
[...]

PR #1101, opened 2024-02-29
fix: datetime profile JSON encoding bug

scottiegarcia:
This fix pertains to my recently created issue [here](https://github.com/capitalone/DataProfiler/issues/1100). The JSON encoder was only expecting/being tested for `pd.Timestamp` as values for `_dt_obj_min` and `_dt_obj_max`, but I ran into a case where those fields were `datetime.datetime` objects.

Running the newly created test on the lastest release reproduces the bug, running on this branch resolves it.

### DIFF
--- a/dataprofiler/profilers/json_encoder.py
+++ b/dataprofiler/profilers/json_encoder.py
@@ -1,6 +1,7 @@
 """Contains ProfilerEncoder class."""
 
 import json
+from datetime import datetime
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


scottiegarcia:
Mocking `pd.DataFrame` in one of the tests was causing a [new function](https://github.com/scikit-learn/scikit-learn/blob/6bb2762b0b8c12b2e2578ef528ea7ea30580f9fa/sklearn/utils/validation.py#L2169-L2175) in sklearn==1.4.1 to error. Since the erroring `test_save_conf_mat` function is just trying to ensure that the correct confusion matrix is generated and other functions validate the f1_report, I figure just mocking that `classification_report` function is the easiest path forward

PR #1103, opened 2024-03-03
Numeric updates

atl1502:


dataprofiler/profilers/float_column_profile.py
```diff
@@ -332,18 +333,22 @@ def _is_each_row_float(cls, df_series: pd.Series) -> list[bool] | pd.Series[bool
...
         """
         if len(df_series) == 0:
-            return list()
-        return df_series.map(NumericStatsMixin.is_float).astype("bool")
+            return pl.Series()
+        if sum(df_series.is_null()) == len(df_series):
+            return df_series
+        df_series = df_series.map_elements(NumericStatsMixin.is_float)
+        df_series = df_series.cast(bool)
+        return df_series
```
micdavis, line 346:
```suggestion
        return df_series.map_elements(NumericStatsMixin.is_float).cast(bool)
```

dataprofiler/profilers/float_column_profile.py
```diff
@@ -294,7 +293,7 @@ def _get_float_precision(
 
         # Lead zeros: ^[+-.0\s]+ End zeros: \.?0+(\s|$)
         # Scientific Notation: (?<=[e])(.*) Any non-digits: \D
-        r = re.compile(r"^[+-.0\s]+|\.?0+(\s|$)|(?<=[e])(.*)|\D")
+        r = r"^[+-.0\s]+|\.?0+(\s|$)|([e].*)|\D"
```
ksneab7, line 296: Why is the content changing of this of regular expression

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1749,36 +1744,30 @@ def _get_quantiles(self) -> None:
...
-            not df_series_clean.empty
-            and df_series_clean.apply(pd.to_numeric, errors="coerce").dtype == "O"
-        )
+        self._greater_than_64_bit = df_series_clean.dtype == pl.Object
         if self._greater_than_64_bit:
-            df_series_clean = df_series_clean.to_numpy()
-            df_series_clean = df_series_clean[df_series_clean != np.nan]
-            if df_series_clean.size == 0:
+            df_np_series_clean = df_series_clean.to_numpy()
+            df_np_series_clean = df_np_series_clean[df_np_series_clean != np.nan]
```
ksneab7, line 1760: is this saying that there is always a 0 or 1 for index values?

ksneab7: may not be a question for this PR

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
```diff
@@ -6,7 +6,7 @@
 from unittest import mock
```
micdavis, line 6: general question about the tests for your implementation. Do you have test cases that cover where `df_series.to_numpy()` would be called? For example:
```python
if df_series.is_empty():
    num_negatives_value = 0
elif self._greater_than_64_bit:
    num_negatives_value = int((df_series.to_numpy() < 0).sum())
else:
    num_negatives_value = int((df_series < 0).sum())
```
do you have test cases that will go in both the `elif` and the `else` here?

dataprofiler/profilers/float_column_profile.py
atl1502: it had to be changed since how regex in polars is handled differently than regex in pandas. Polars does not allow for lookahead/behind so it had to be removed. Fortunately, (if I interpreted the regex properly) I don't think the lookahead/behind was needed in the first place.

dataprofiler/tests/profilers/test_numeric_stats_mixin_profile.py
atl1502: yup these are covered by the int_column_profiler when it tests values of 64 bits.

dataprofiler/profilers/float_column_profile.py
atl1502: mypy gets update with it in that form for whatever reason

dataprofiler/profilers/float_column_profile.py
ksneab7: Alright, I am not a regex expert myself, so maybe someone else should look at this, but the reasoning makes sense to me

dataprofiler/profilers/numerical_column_stats.py
```diff
@@ -1924,11 +1909,10 @@ def _get_skewness(
...
             return
 
-        if self._greater_than_64_bit and type(df_series) is pd.Series:
-            df_series = df_series.to_numpy(dtype=float)
+        if self._greater_than_64_bit and type(df_series) is pl.Series:
+            batch_biased_skewness = profiler_utils.biased_skew(df_series.to_numpy())
         else:
-            df_series = pl.from_pandas(df_series, nan_to_null=False)
-        batch_biased_skewness = profiler_utils.biased_skew(df_series)
+            batch_biased_skewness = profiler_utils.biased_skew(df_series)
```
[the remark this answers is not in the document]
atl1502: its kind of a weird edge case due to how polars works internally it won't do math on values greater than 64 bits. What would the helper function do exactly?

dataprofiler/profilers/numerical_column_stats.py
atl1502: If i am reading it correctly it is just dropping any nans in the array

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -749,13 +747,13 @@ def test_profile_merge(self):
...
-        df = pd.Series(data).apply(str)
+        data = [2.0, 8.5, None, 6.0, -3, 0]
+        df = pl.Series(data).map_elements(str)
         profiler1 = IntColumn("Int")
         profiler1.update(df)
 
-        data2 = [0.0, 3.5, "not an int", 125.0, 0, -0.1, -88]
-        df2 = pd.Series(data2).apply(str)
+        data2 = [0.0, 3.5, None, 125.0, 0, -0.1, -88]
+        df2 = pl.Series(data2).map_elements(str)
```
taylorfturner, line 756: same Q here...

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -767,26 +765,26 @@ def test_profile_merge_for_zeros_and_negatives(self):
...
+        df = pl.Series(data).map_elements(str)
         profiler1 = IntColumn(name="Int")
         profiler1.update(df)
         profiler1.match_count = 0
 
-        data2 = [10.0, 3.5, "not a float", 15.0, "not a float"]
-        df2 = pd.Series(data2).apply(str)
+        data2 = [10.0, 3.5, None, 15.0, None]
+        df2 = pl.Series(data2).map_elements(str)
         profiler2 = IntColumn(name="Int")
```
taylorfturner, line 776: as above

dataprofiler/tests/profilers/test_int_column_profile.py
```diff
@@ -836,13 +834,13 @@ def test_custom_bin_count_merge(self):
...
-        df = pd.Series(data).apply(str)
+        data = [2, None, 6, None]
+        df = pl.Series(data).map_elements(str)
         profiler1 = IntColumn("Int", options)
         profiler1.update(df)
 
-        data2 = [10, "not an int", 15, "not an int"]
-        df2 = pd.Series(data2).apply(str)
+        data2 = [10, None, 15, None]
+        df2 = pl.Series(data2).map_elements(str)
```
taylorfturner, line 843: as above

taylorfturner: and more of the same below

dataprofiler/profilers/float_column_profile.py
atl1502: Sorry typoed it meant upset.

dataprofiler/profilers/numerical_column_stats.py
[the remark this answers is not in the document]
atl1502: gotcha, I don't there is a great way to wrap this since the called function is usually different. Would also harm readability I think.

dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -1667,20 +1660,19 @@ def test_insufficient_counts(self):
             self.assertEqual(0, len(w))
 
     def test_diff(self):
-        data = [2.5, 12.5, "not a float", 5, "not a float"]
-        df = pd.Series(data).apply(str)
+        data = [2.5, 12.5, None, 5, None]
+        df = pl.Series(data).map_elements(str)
```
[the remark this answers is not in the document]
atl1502: Yes, this was the mixed types issue that we were having. Putting a string and a int/float in same series is not allowed.

dataprofiler/profilers/float_column_profile.py
taylorfturner: mypy doesn't like sequential together... so have to leave separate for now

taylorfturner (approved):

micdavis (approved):

PR #1115, opened 2024-03-13
feat: add __str__ methods to the various part of the profiler options

carlsonp:
How about this as a starting point for adding some helpful printing of Data Profiler options?

```
import dataprofiler as dp
print(profile_options)
```


taylorfturner (commented):
Yep, exactly how I was thinking about utilizing the `__str__` magic method. I may recommend two things here, though: 
- some variable name changes for clarity in the `__str__`. We use options as a variable in throughout and you are using `options` in that for loop. Just for clarity of variable naming... maybe something along the line of `iter_option` when you are referencing in the `for` loop so its super explicit that it is related to the `for` loop 
- and adding the presets on the `ProfilerOptions.__str__` return implementation for clarity

taylorfturner:
You'll want to rebase onto `dev`... had some other PRs from @gliptak merge into `dev` this morning. 

Have you done rebases before, @carlsonp?

taylorfturner:
@carlsonp you should be good to rebase onto `dev` now

carlsonp:
Thanks, I'll get to it next week. Fighting off a bug.

carlsonp:
Rebased

taylorfturner:
@carlsonp yeah, I like the route you are going. Once you add unit tests, just tag me and I'll take another look at it. Cheers!

carlsonp:
> @carlsonp yeah, I like the route you are going. Once you add unit tests, just tag me and I'll take another look at it. Cheers!

Can you please provide a suggested starting point for which file to add the unit tests?

dataprofiler/profilers/profiler_options.py
```diff
@@ -1715,6 +1839,17 @@ def __init__(self, presets: str = None) -> None:
...
+        """
+        Return a human friendly consumable output in string form.
+
+        :return: str of the option presets and properties
+        :rtype: str
+        """
+        return f"Presets: {str(self.presets)}\n \
+            {str(self.structured_options)}\n \
+            {str(self.unstructured_options)}"
+
```
taylorfturner, line 1852: this makes sense here and you would test in `test_profiler_options.py`

taylorfturner:
> > @carlsonp yeah, I like the route you are going. Once you add unit tests, just tag me and I'll take another look at it. Cheers!
> 
> @taylorfturner Can you please provide a suggested starting point for which file to add the unit tests?

Yes, indeed! I would actually move some of the `__str__` methods you overwrite in this PR into `base_options.py`. Then I think your testing would much more simple. Ultimately this would allow you to test main implementation in `test_base_option.py` then you could build out small scenarios in the other option test files in `tests/profilers/profiler_options/...`

taylorfturner:
@carlsonp you'll want a rebase here too onto `dev`

dataprofiler/profilers/profiler_options.py
```diff
@@ -958,6 +968,25 @@ def __init__(
...
+        for iter_option in [
+            a
+            for a in dir(self)
+            if not a.startswith("__") and not callable(getattr(self, a))
+        ]:
+            dict_string["CategoricalOptions"].append(
+                {str(iter_option): str(getattr(self, iter_option))}
+            )
+        return json.dumps(dict_string, indent=4)
+
```
taylorfturner, line 989: might be good to find a way to abstract this a bit more so this ends up in `BaseOption` 90%+ of this code is repeat just with string changes: so I think there is room to make this DRY-er

Issue #1098, opened 2024-02-27
Help with understanding DataProfiler options and performance


carlsonp:
This is a bit of a question and a bit of feature request I think?

I'm trying to understand why profiling on some tables is slow.  When it calculates statistics, it seems to take a long time, even for relatively small sample sizes (less than 1,000,000 rows).  I started looking into the setup of the profiling and saw the [Profile Options](https://capitalone.github.io/DataProfiler/docs/0.10.8/html/profiler.html#profile-options).

I started to go through and turn off calculations I don't need.  For example:

```
profile_options = ProfilerOptions()
profile_options.set({
    "structured_options.data_labeler.is_enabled": False,
    "structured_options.correlation.is_enabled": False,
    "structured_options.multiprocess.is_enabled": True
})

profile = Profiler(data, options=profile_options, samples_per_update=sample_rows)
```

Is there a way to print out ALL the profile options including the defaults?  This would help me debug and understand what is being calculated.  From a feature standpoint, perhaps more of the objects should expose friendly printing of objects via `__str__` methods?

Thanks

carlsonp:
I think I made some more progress in debugging the slowness I'm seeing.  When I look at the output, some columns are being marked as categorical but they have an extremely large number of unique records.

For example, loading in a ~350MB parquet file and using profiling through [SnakeViz](https://jiffyclub.github.io/snakeviz/), I'm seeing a lot of time spent on `perform_chi_squared_test_for_homogeneity`.  This makes sense because it's running this on the categorical columns.  As can be seen below, one column is being marked as categorical and has 115207 unique values!  It seems like there should be a cutoff, categorical to me would be like less than 500 unique values.  If it's more than that, it really shouldn't be marked as categorical.

It took about 800 seconds to run the profile and write the JSON results to a file.  The example dataset has 40 columns.

```
{
            "column_name": "mynumber",
            "data_type": "int",
            "categorical": true,
            "order": "random",
            "samples": "['50125', '138752', '147769', '222048', '141623']",
            "statistics": {
                "min": 1.0,
                "max": 998974.0,
                "mode": "[500.4865]",
                "median": 145109.1429,
                "sum": 220569076580.0,
                "mean": 126742.2764,
                "variance": 5985159591.3955,
                "stddev": 77363.8132,
                "skewness": -0.2028,
                "kurtosis": 0.8711,
                "histogram": {
[...]

PR #1117, opened 2024-03-14
Multiprocessing and Performance Improvements

carlsonp:
This is related to some of the discussion in #1098

In my testing, I have a single dataset.  I am running this in a Docker container.  I'm running with the following settings:

```
import sys
import json
import time
import dataprofiler as dp

filename = "myfile.parquet"

def profile_test(filename):
    data = dp.Data(filename)
    
    profile_options.set({
        "structured_options.data_labeler.is_enabled": False,
        "unstructured_options.data_labeler.is_enabled": False,
        "structured_options.correlation.is_enabled": False,
        "structured_options.multiprocess.is_enabled": True,
        "structured_options.chi2_homogeneity.is_enabled": False,
        "structured_options.category.max_sample_size_to_check_stop_condition": 1,
        "structured_options.category.stop_condition_unique_value_ratio": 0.001,
        "structured_options.sampling_ratio": 0.3,
        "structured_options.null_replication_metrics.is_enabled": False
    })

    print(profile_options)

    profile = dp.Profiler(data, options=profile_options)

    human_readable_report = profile.report(report_options={"output_format":"pretty"})

    with open("reportfile.json", "w") as outfile:
        outfile.write(json.dumps(human_readable_report, indent=4))

start_time = time.time()
profile_test(filename)
end_time = time.time()

print("Profile runtime for "+filename, end_time-start_time, 'seconds')
```

When Data Profiler gets to the first `tqdm` loop and displays `Finding the Null values in the columns...` it's pretty quick.  It also lists 19 processes corresponding to the pool_size available in the Python multiprocessing pool.  This works fine.

Then when it gets to the second `tqdm` loop and displays `Calculating the statistics...` I noticed that it was only using 4 processes.  When I looked at what was running, I am only seeing a single core being used.  When I looked at the code, `profile_builder.py` has 4 hard-coded.  This doesn't seem right.  There's a utility function `profiler_utils.suggest_pool_size` that's not even used that returns the pool size.  So I swapped that out.  Now when I run, instead of 4 processes it shows 19 so that seems better.  At least we're not leaving potential performance on the table with hard-coding.  

However, I'm still seeing only a single core being used.  I also checked the CPU affinity after reading some [comments on Stackoverflow](https://stackoverflow.com/questions/15639779/why-does-multiprocessing-use-only-a-single-core-after-i-import-numpy).  It looks reasonable to me.

```
print(f"Affinity: {os.sched_getaffinity(os.getpid())}")
```

I'm going to try profiling a bit more and see if I can figure out where it's hanging.  It seems like it should be faster, particularly on a multi-core machine.  Calculating all the statistics and stuff *is* taxing but it seems like it should be faster.

### DIFF
--- a/dataprofiler/profilers/profile_builder.py
+++ b/dataprofiler/profilers/profile_builder.py
@@ -3003,7 +3003,7 @@ class StructuredProfiler(BaseProfiler):
         notification_str = "Calculating the statistics... "
         pool = None
         if auto_multiprocess_toggle:
-            pool, pool_size = profiler_utils.generate_pool(4, est_data_size)
+            pool, pool_size = profiler_utils.generate_pool(profiler_utils.suggest_pool_size(), est_data_size)
             if pool:
                 notification_str += " (with " + str(pool_size) + " processes)"
 


taylorfturner:
> I'm going to try profiling a bit more and see if I can figure out where it's hanging.

Testing is welcome @carlsonp  -- I'll keep on comments here and in #1098

taylorfturner:
you'll want a rebase onto `dev` here

Issue #1048, opened 2023-10-10
Categorical Column Profiling Error


scottiegarcia:
**General Information:**
 - OS: Linux
 - Python version: 3.8.5
 - Library version: 0.10.5

**Describe the bug:**

On occasion, some categorical columns produce the following error when profiling

```
TypeError                                 Traceback (most recent call last)
<ipython-input-16-5eb7460647d7> in <module>
     11 })
     12 
---> 13 profile = Profiler(data=df, options=profiler_options)
     14 
     15 report = profile.report()

/opt/conda/lib/python3.8/site-packages/dataprofiler/__init__.py in profiler_tracking_wrapper(wrapped, instance, args, kwargs)
    160         _issue_warning_and_notify()
    161 
--> 162     return wrapped(*args, **kwargs)
    163 
    164 

/opt/conda/lib/python3.8/site-packages/dataprofiler/profilers/profile_builder.py in __new__(cls, data, samples_per_update, min_true_samples, options, profiler_type)
   3124             return profile
   3125         elif profiler_type == "structured":
-> 3126             return StructuredProfiler(
   3127                 data, samples_per_update, min_true_samples, options
   3128             )

/opt/conda/lib/python3.8/site-packages/dataprofiler/__init__.py in sub_profiler_tracking_wrapper(wrapped, instance, args, kwargs)
    191             _issue_warning_and_notify()
    192 
--> 193     wrapped(*args, **kwargs)
    194 
    195 _wrapt.wrap_function_wrapper(

/opt/conda/lib/python3.8/site-packages/dataprofiler/profilers/profile_builder.py in __init__(self, data, samples_per_update, min_true_samples, options)
   1678             )
   1679         if data is not None:
-> 1680             self.update_profile(data)
   1681 
   1682     def _add_error_checks(  # type: ignore[override]

/opt/conda/lib/python3.8/site-packages/dataprofiler/profilers/profile_builder.py in update_profile(self, data, sample_size, min_true_samples)
    982             sample_size = self._get_sample_size(data)
    983 
--> 984         self._update_profile_from_chunk(data, sample_size, min_true_samples)
    985 
    986         # set file properties since data will be processed

/opt/conda/lib/python3.8/site-packages/dataprofiler/profilers/profile_builder.py in _update_profile_from_chunk(self, data, sample_size, min_true_samples)
   3004 
   3005         for prof_idx in tqdm(clean_sampled_dict.keys()):
[...]

PR #1125, opened 2024-03-25
Bug fix for float precision calculation using categorical data with trailing 0s

SchadtJ:
Bug fix for https://github.com/capitalone/DataProfiler/issues/1048#issuecomment-1973429619. 

The float precision calculation errors out for **categorical** data when one of the values has leading/ trailing zeros. This is due to the regex operation stripping these zeros and the resulting value being outside the list of possible values.

Passing tests:
dataprofiler-bug

### DIFF
--- a/dataprofiler/tests/profilers/test_float_column_profile.py
+++ b/dataprofiler/tests/profilers/test_float_column_profile.py
@@ -211,6 +211,13 @@ class TestFloatColumn(unittest.TestCase):
                 msg=f"Errored for: {sample[0]}",
             )
 
+        # Validate categorical series with trailing zeros supported
+        categorical_series = pd.Series(
+            [202209, 202210, 202211], dtype="category"
+        ).apply(str)
+        float_profiler = FloatColumn("Name")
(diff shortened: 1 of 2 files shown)


dataprofiler/tests/profilers/test_float_column_profile.py
```diff
@@ -211,6 +211,13 @@ def test_profiled_precision(self):
                 msg=f"Errored for: {sample[0]}",
             )
 
+        # Validate categorical series with trailing zeros supported
+        categorical_series = pd.Series(
+            [202209, 202210, 202211], dtype="category"
+        ).apply(str)
+        float_profiler = FloatColumn("Name")
+        float_profiler.update(categorical_series)
```
taylorfturner, line 219: testing this locally so may have some additional comments, but initial thought is that there should definitely be some `assert` statements here to validate this is actually working as intended post-change @SchadtJ @scottiegarcia

taylorfturner:
reverting this accidental merge -- @SchadtJ please reopen into `dev`. Thanks!

PR #1168, opened 2025-01-25
dev-fix/Improvements for Local Development Setup and Unit Test Structure

mhmotamedi:
# PR Description:
## Improvements for Local Development Setup and Unit Test Structure
This PR focuses on resolving issues with local environment setup and unit test execution in the dataprofiler library. The changes ensure a more reliable setup process, eliminate dependency conflicts, and enhance the performance and maintainability of unit tests.

### 1. Refactored Makefile

- Dependency Installation Order: Moved `requirements-ml.txt` earlier in the installation sequence to prevent conflicts with TensorFlow and shared dependencies like NumPy.

- Fresh Dependency Downloads: Added `--no-cache-dir` to `pip install` commands to ensure fresh downloads, avoiding issues caused by stale or corrupted cached files.

- Virtual Environment Isolation: Updated the test command to explicitly use `venv/bin/python`, ensuring tests use the correct Python interpreter and dependencies from the virtual environment.

- Python Version Check: Introduced a `check-python` target to ensure the specified Python version is `3.9, 3.10, or 3.11`. If the Python version does not match, the Makefile exits with a clear error message, improving reliability across environments.

- Help Section: Added a help target to display all available Makefile commands with descriptions. This makes the development workflow more transparent and easier to understand for new and existing contributors.

### 2. Refactored Unit Tests

- Improved Test Isolation: Replaced global variables with `setUpClass` and `tearDownClass`, ensuring proper initialization and cleanup of shared resources within each test suite.
- Faster Test Execution: Centralized resource setup to reduce redundant operations, improving test performance without relying on the `--forked` flag, which can be slow for local runs.

### 3. Updated version.py

- Incremented the version in the `version.py` file to reflect the new changes introduced in this PR.


dataprofiler/_typing.py
```diff
@@ -1,9 +1,10 @@
...
-from typing import Dict, List, NewType, Union
+
+from typing import NewType, Union
 
 import numpy as np
 import pandas as pd
 
 DataArray = Union[pd.DataFrame, pd.Series, np.ndarray]
-JSONType = Union[str, int, float, bool, None, List, Dict]
+JSONType = Union[str, int, float, bool, None, list, dict]
```
armaan-dhillon, line 9: Just curious why we are going from `typing.Dict` and `typing.List` to `dict` and `list` here and throughout the rest of the files. Was it do to mypy errors? Not doubting that it's right just want to understand!

armaan-dhillon (approved):
LGTM, just left one comment to understand the typing changes

dataprofiler/_typing.py
stephenpardy: See, for example: https://docs.python.org/3/library/typing.html#typing.Dict the builtin types are supported as of several python versions ago.

stephenpardy (approved):

PR #1169, opened 2025-02-05
test refactor for floating point failures + mypy fix

armaan-dhillon:
## PR Description
  * refactors `test_categorical_diff` and `test_diff_categorical_chi2_test` which were failing locally due to some floating point errors but not failing when run on a different machine -- this refactors them so the floating point comparisons will be machine/platform agnostic
  * updates the `flake8` `rev` in `.pre-commit-config.yaml` (was running into a local error with older version)
  * fixes one mypy error in `dataprofiler/data_readers/parquet_data.py`
  * adds inline `# type: ignore[override]` comments for 4 lines causing mypy `override` errors with `TODO` statements for context
  * above changes allow us to remove `disable_error_code = override` from `setup.cfg`

The 4 mypy override errors that this PR was supposed to address  require adding a `@property` tag to the `profile()` method of `NumericStatsMixin` and this is a breaking change as it led to many tests failing. `TODO` comments highlight this.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
   # Flake8: complexity and style checking
   # https://flake8.pycqa.org/en/latest/user/using-hooks.html
   - repo: https://github.com/pycqa/flake8
-    rev: 4.0.1
+    rev: 5.0.4
     hooks:
       - id: flake8
         additional_dependencies: [flake8-docstrings]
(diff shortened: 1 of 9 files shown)


thebrianbn (dismissed):
LGTM

stephenpardy (approved):

thebrianbn (approved):
LGTM

PR #1175, opened 2025-03-18
Remove Versions from Intro page

shania-m:
Removes versions section from intro page and leaves it in the changelog untouched, so users can still access previous versions from there.


thebrianbn (approved):
LGTM

armaan-dhillon (approved):

PR #1215, opened 2026-07-31
Fix chardet import for compatibility with chardet >= 0.7.0

jellyfishing2346:
Changed from 'from chardet.universaldetector import UniversalDetector' to 'from chardet import UniversalDetector' to support chardet 0.7.0+ which changed their import structure.

Fixes #1204

### DIFF
--- a/dataprofiler/data_readers/data_utils.py
+++ b/dataprofiler/data_readers/data_utils.py
@@ -30,7 +30,7 @@ import numpy as np
 import pandas as pd
 import pyarrow.parquet as pq
 import requests
-from chardet.universaldetector import UniversalDetector
+from chardet import UniversalDetector
 from typing_extensions import TypeGuard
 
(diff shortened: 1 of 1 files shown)


jellyfishing2346:
Hi @leos, I was wondering if you happen to have any feedback on this PR.

JGSweets:
@jellyfishing2346 does this still support backwards compatibility?

If reqs don't require the new version then this could break older versions.

jellyfishing2346:
@JGSweets it doesn’t.

JGSweets:
@jellyfishing2346 do you know if they ever resolved this issue of changing the default likelihood:
https://github.com/capitalone/DataProfiler/pull/1205#issuecomment-4056660157

Also, unsure if they resolved their license issue for 7.0.0+

jellyfishing2346:
@JGSweets I don't believe I know if they did.

Issue #720, opened 2022-11-16
Investigate use of cast() in reports.


Sanketh7:
Investigate whether these instances of cast() are needed.
PRs should be small (at most a few files).

reports/graphs.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/reports/graphs.py#L77

reports/utils.py
- https://github.com/capitalone/DataProfiler/blob/main/dataprofiler/reports/utils.py#L53

PR #1222, opened 2026-09-09
refactor: replace cast() with isinstance assertions in reports

ovodela:
## Summary

- Replaces all 4 cast() calls in dataprofiler/reports/graphs.py with assert isinstance() or assert is not None checks
- Provides both static type safety (mypy narrowing) and runtime validation instead of unchecked casts
- Removes unused Union and cast imports from typing
- The cast(F, new_f) in utils.py is intentionally retained

## Test plan

- [ ] All existing test_graphs.py tests pass (16 tests)
- [ ] mypy reports no new type errors

Addresses #720


ovodela:
Closing — same issue as #1221. PR was opened against `dev` via Spork, but `dev` is stale and all active merges target `main`. Will resubmit cleanly against `main`.

